PR #55, opened 2016-07-27
added upgrate from webpack 1 content

sokra:


bebraw:
👍 

The exact wording can be tweaked later.

PR #61, opened 2016-08-02
Sidecar, Icons and Styling

gregvenech:
This PR includes these changes:

- A sidecar to contain Gitter, Medium and Github buttons.
- Added "Github" and "Support" links to the Navigation and Footer respectively.
- Adding our own custom icons stylesheet built with icomoon (I can add anything from font-awesome, entypo, linearicons etc. if we need other icons)
- Cleaning up and removing various other styles


bebraw:
Maybe "fork at GitHub" could be removed altogether. Just include a link to GitHub project at some visible place (footer?) so people can find it. Chat could work as a link too if we want to simplify.

gregvenech:
Yea I could go either way, I like the ribbon and feel maybe it's nice to have the buttons featured prominently. However I do think it would be cleaner, at least for the ribbon, maybe to just use a link in the footer.

bebraw:
Another option - minimal sidebar with icons (think those social media bars). You could link to GitHub, Gitter, and Medium blog at least. That would solve it nicely. I don't mind writing the js bit for sidecar for this if you can do visuals.

gregvenech:
Hmm do you mean add another sidebar or just modifying the existing Sidebar component? I like the icons idea but I think we could add them to the Navigation or Footer for the same effect with less noise. 

I might be confused though, do you have an example?

bebraw:
Here's the [basic idea](http://www.cssportal.com/blog/css-social-media-sidebar/). So basically sidebar with appropriate color + icon fixed to left or right.

bebraw:
About GitHub, maybe solution like [this](http://sharingbuttons.io/) would work (see top right). Minimal, gets to the point.

gregvenech:
Ok now I see... I had thought you meant a full on sidebar. I really like both approaches but I think I like the mini sidebar most because we can hit multiple options with it in a clean way. I'm not sure we even need JavaScript based on our layout. 

I'm willing to tackle this tonight, if we're sure this is what we want.

bebraw:
Yeah, we can skip JS and just link to Gitter. I was thinking of showing sidecar on click but that would work as well. It can be tweaked later if we want.

bebraw:
Here's the corner btw, http://tholman.com/github-corners/ .

bebraw:
And matching React component, https://github.com/skratchdot/react-github-corner .

gregvenech:
Oh sorry I thought you meant use JavaScript to stick and unstick the sidebar on scroll as some pages do. I actually really like the Gitter sidecar and I think it'd be helpful to keep the docs in front of you while asking (or answering) questions.

I'll lay it out and if you could connect the Gitter icon/button to the sidecar that'd be great.

bebraw:
> I'll lay it out and if you could connect the Gitter icon/button to the sidecar that'd be great.

Sure. I can do that.

gregvenech:
Most of the way there, just need to make a custom icon set that includes both gitter and medium icons as well. Should be pushing finishing touches tomorrow or friday...

bebraw:
Cool. Take your time. 👍

gregvenech:
@bebraw this is pretty much good to go. Up to you on how to go about linking to/embedding gitter... based on what we talked about before I'm thinking the gitter button on the right opens the gitter sidecar, while the support link in the footer just navigates there (as it already does).

bebraw:
Looks good to me. Maybe it could be pushed in the middle vertically but those are details. 👍 

I'll hook up the Gitter button at a later time.

PR #64, opened 2016-08-06
Content howto hmr

jmreidy:
Add documentation for a basic How Module Replacement (HMR) setup. The example uses Babel, React, and CSS modules, and explains how to edit webpack.config, what dependencies are needed, and how to structure the source code.

### DIFF
--- a/content/how-to/hot-module-reload.md
+++ b/content/how-to/hot-module-reload.md
@@ -1,4 +1,186 @@
 ---
-title: How to Configure Hot Module Reloading?
+title: How to Configure Hot Module Replacement?
 ---
+Hot Module Replacement (HMR) exchanges, adds, or removes modules while an application is running without a page reload. HMR is particularly useful in applications using a single state tree, since components are "dumb" and will reflect the latest application state, even after their source is changed and they are replaced. 
+
+Webpack's power lies in its customizablity, and there are MANY ways of configuring HMR given the needs of a partiular project. The approach described below uses Babel and React, but these tools are not necessary for HMR to work. If you'd like to see examples of other approaches, please request them or, better yet, open up a PR with an addition! 
+
+##Project Config
+To use HMR, you'll need the following dependencies:
+
+* webpack 2 (tested with 2.1.0-beta.20)
+* webpack-dev-server 2 (tested with 2.1.0-beta.0)
+
+This guide will be demonstrating the use of HMR with Babel, React, and PostCSS (using CSS Modules). To follow along, please add the following deps to your `package.json`:
+
+```js
+"devDependencies": {
+    "babel": "^6.5.2",
+    "babel-core": "^6.11.4",
+    "babel-loader": "^6.2.4",
+    "babel-preset-es2015-webpack": "^6.4.2",
+    "babel-preset-react": "^6.11.1",
+    "babel-preset-stage-2": "^6.11.0",
+    "css-loader": "^0.23.1",
+    "postcss-loader": "^0.9.1",
+    "react-hot-loader": "^3.0.0-beta.1",
+    "style-loader": "^0.13.1",
+    "webpack": "^2.1.0-beta.20",
+    "webpack-dev-server": "^2.1.0-beta.0"
+  }
+```
+
+###Babel Config
+Your `.babelrc` file should look like the following:
+
+```js
+{
+  "presets": ["es2015-webpack", "stage-2", "react"],
+  "plugins": ["react-hot-loader/babel"]
+}
+```
+
+There's a few things to note here. First, in the presets, we're using es2015-webpack, which is a Babel preset that enables support for es2015 while DISABLING conversion to CommonJS module format. Since Webpack 2 understands import/export syntax, the es2015-webpack preset (unlike the general es2015 preset) allows Webpack to perform tree-shaking and dead code elimination.
+
+The stage-2 preset is up to your comfort with adopting new technoligies; stage-0 will keep you on the bleeding age.
+
+Finally, the React preset will enable React specific transpilation, while the react-hot-loader/babel plugin will enable your React code to work with HMR.
+
+###Webpack config
+While there's many ways of setting up your Webpack config - via API, via multiple or single config files, etc - here is the basic information you should have available:
+
+```js
+const { resolve } = require('path');
+const webpack = require('webpack');
+
+const ENTRY = ['./index.js'];
+
+module.exports = env => {
+  return {
+    entry: env.dev?
+      [
+        'react-hot-loader/patch',
+        'webpack-dev-server/client?http://localhost:8080',
+        'webpack/hot/only-dev-server',
+      ].concat(ENTRY)
+      :
+      ENTRY,
+    output: {
+      filename: 'bundle.js',
+      path: resolve(__dirname, 'dist'),
+      publicPath: '/'
+    },
+    context: resolve(__dirname, 'src'),
+    devtool: env.prod ? 'source-map' : 'inline-source-map',
+    devServer: {
+      hot: true,
+      publicPath: '/'
+    },
+    module: {
+      loaders: [
+        { test: /\.js$/, loader: 'babel',
+          exclude: /node_modules/ },
+        {
+          test: /\.css$/,
+          loader: 'style!css-loader?modules&postcss-loader',
+        },
+      ],
+    },
+    plugins: [
+      new webpack.HotModuleReplacementPlugin(),
+      new webpack.NamedModulesPlugin(),
+    ],
+  }
+};
+```
+
+There's a lot going on above, and not all of it is related to HMR. If you are new to Webpack configs, please see the documentation elsewhere on this site. The basic assumption here is that your JS entry is located at `./src/index.js`, and that you're using CSS Modules for your styling. Let's look at the HMR specific items.
+
+First, the plugins. While only the HMRPlugin itself is necessary to enable HMR, the NamedModulesPlugin will make your life easier by showing an intelligible module id in your browser console whenever a file is changed.
+
+Next, working backward, is the `devServer` key. While it's possible to configure webpack-dev-server directly from the CLI, or more specifically in a server script from the Node module API, this example approach is using the `devServer` key in the overall webpack config. `hot: true` turns on HMR. `publicPath` denotes the root from wich the server will serve up your code (e.g. `localhost:8080${PUBLIC_PATH}`, so in this case `localhost:8080/`). the `publicPath` should match the `output.publicPath`.
+
+Finally, your entry file should be preceded by the elements included here: `'react-hot-loader/patch'`, `'webpack-dev-server/client?http://localhost:8080'`, and `'webpack/hot/only-dev-server'`. The react-hot-loader entry is only necessary if you are using React. The `webpack-dev-server/client?` line should make sure to point at the dev server instance you are running (in case you have configured a different port, for example). Finally, the `only` in `hot/only-dev-server` means that updates will only be sent on clean builds/bundles; you can drop the `only-` if you would prefer seeing error output in your browser via HMR.
+
+The module loaders in this config, while not specific to HMR, are worth noting here; we're making sure to use Babel (as configued in the .babelrc) for our JS files, and setting up CSS Modules for our CSS code.
+
+###Code
+In this guide, we're using the following files:
+
+```js
+// ./src/index.js
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { AppContainer } from 'react-hot-loader'
+
+import App from './components/App';
+
+const render = () => {
+  ReactDOM.render(
+    <AppContainer>
+      <App/>
+    </AppContainer>,
+    document.getElementById('root')
+  );
+};
+
+render();
+
+// Hot Module Replacement API
+if (module.hot) {
+  module.hot.accept('./components/App', render);
+}
+
+
+// ./src/components/App.js
+import React from 'react';
+import styles from './App.css';
+
+const App = () => (
+  <div className={styles.app}>
+    <h2>Hello, </h2>
+  </div>
+);
+
+export default App;
+```
+
+```css
+// ./src/components/App.css
+.app {
+    text-size-adjust: none;
+    font-family: helvetica, arial, sans-serif;
+    line-height: 200%;
+    padding: 6px 20px 30px;
+}
+```
+
+Now, the above code is using React, but it doesn't need to be. In fact, the only thing that matters above is the code refering to `module`. First, we wrap the HMR code inside of `module.hot` check; webpack exposes `module` to the code, and if we are running with `hot: true` configured, we'll enter the inside of the conditional. While the module API offers more options than what's above, the most important element is the `module.hot.accept` call. It specific how to handle changes to specific dependencies. So in this, it will file the `render` method ONLY when `src/components/App.js` changes! Note that would also include when the dependencies of `App.js` change - so the `render` method will file not just for changes made directly to the source of `App.js`, but also changes made to `App.css`, since `App.css` is included in `App.js`.
+
+###Package.json
+Finally, we need to start up webpack dev server to bundle our code and see HMR in action. We can use the following package.json entry:
+
+```js
+  "start" : "webpack-dev-server --env.dev --content-base dist",
+```
+
+Run `npm start`, open up your browser to `localhost:8080`, and you should see the folling entries printed in your console.log:
+``` 
+dev-server.js:49[HMR] Waiting for update signal from WDS...
+only-dev-server.js:74[HMR] Waiting for update signal from WDS...
+client?c7c8:24 [WDS] Hot Module Replacement enabled.
+```
+
+Go ahead and edit and save your App.js file. You should see something like the following in your console.log:
+
+```
+[WDS] App updated. Recompiling...
+client?c7c8:91 [WDS] App hot update...
+dev-server.js:45 [HMR] Checking for updates on the server...
+log-apply-result.js:20 [HMR] Updated modules:
+log-apply-result.js:22 [HMR]  - ./components/App.js
+dev-server.js:27 [HMR] App is up to date.
+```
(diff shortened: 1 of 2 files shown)


content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
+---
+title: How to Configure Hot Module Replacement?
+---
+Hot Module Replacement (HMR) exchanges, adds, or removes modules while an application is running without a page reload. HMR is particularly useful in applications using a single state tree, since components are "dumb" and will reflect the latest application state, even after their source is changed and they are replaced. 
+
+Webpack's power lies in its customizablity, and there are MANY ways of configuring HMR given the needs of a partiular project. The approach described below uses Babel and React, but these tools are not necessary for HMR to work. If you'd like to see examples of other approaches, please request them or, better yet, open up a PR with an addition!
```
bebraw: /s/partiular/particular

bebraw: Add link to "open up a PR".

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+
+Webpack's power lies in its customizablity, and there are MANY ways of configuring HMR given the needs of a partiular project. The approach described below uses Babel and React, but these tools are not necessary for HMR to work. If you'd like to see examples of other approaches, please request them or, better yet, open up a PR with an addition! 
+
+##Project Config
+To use HMR, you'll need the following dependencies:
+
+* webpack 2 (tested with 2.1.0-beta.20)
+* webpack-dev-server 2 (tested with 2.1.0-beta.0)
+
+This guide will be demonstrating the use of HMR with Babel, React, and PostCSS (using CSS Modules). To follow along, please add the following deps to your `package.json`:
```
bebraw: Would it be better to give a `npm install` command as the versions tend to live a lot? That can lead to breakage, though, but we can fix the docs. We'll end up with the same problem either case.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+```js
+{
+  "presets": ["es2015-webpack", "stage-2", "react"],
+  "plugins": ["react-hot-loader/babel"]
+}
+```
+
+There's a few things to note here. First, in the presets, we're using es2015-webpack, which is a Babel preset that enables support for es2015 while DISABLING conversion to CommonJS module format. Since Webpack 2 understands import/export syntax, the es2015-webpack preset (unlike the general es2015 preset) allows Webpack to perform tree-shaking and dead code elimination.
+
+The stage-2 preset is up to your comfort with adopting new technoligies; stage-0 will keep you on the bleeding age.
```
bebraw: /s/technoligies/technologies

bebraw: I would add a warning (`W>`) about using Stage 0 specifically as it can potentially break your code. I prefer to enable features on feature basis over stages to communicate better which custom features I'm using. Better for maintenance.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+
+const ENTRY = ['./index.js'];
+
+module.exports = env => {
+  return {
+    entry: env.dev?
+      [
+        'react-hot-loader/patch',
+        'webpack-dev-server/client?http://localhost:8080',
+        'webpack/hot/only-dev-server',
```
bebraw: These two lines (webpack-dev-server and the one after) can be skipped if you run webpack-dev-server in inline mode. A little easier to manage. If I remember right, that might be the new default for webpack 2 even.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+      :
+      ENTRY,
+    output: {
+      filename: 'bundle.js',
+      path: resolve(__dirname, 'dist'),
+      publicPath: '/'
+    },
+    context: resolve(__dirname, 'src'),
+    devtool: env.prod ? 'source-map' : 'inline-source-map',
+    devServer: {
```
bebraw: As mentioned above, if you enable inline mode here, you can simplify entries.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+      hot: true,
+      publicPath: '/'
```
bebraw: Document why `publicPath` is set this way. A comment will do.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+    },
+    module: {
+      loaders: [
+        { test: /\.js$/, loader: 'babel',
+          exclude: /node_modules/ },
```
bebraw: `include` over `exclude`? Though setting both is actually even a better idea (needed for monorepo setups).

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+        {
+          test: /\.css$/,
+          loader: 'style!css-loader?modules&postcss-loader',
```
bebraw: Maybe prefer `loaders` array over `loader`. Reads better. webpack 2 provides a better query format too so we might want to default to that.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+        },
+      ],
+    },
+    plugins: [
+      new webpack.HotModuleReplacementPlugin(),
+      new webpack.NamedModulesPlugin(),
```
bebraw: Mention why `NamedModulesPlugin` is used.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+    ],
+  }
+};
+```
+
+There's a lot going on above, and not all of it is related to HMR. If you are new to Webpack configs, please see the documentation elsewhere on this site. The basic assumption here is that your JS entry is located at `./src/index.js`, and that you're using CSS Modules for your styling. Let's look at the HMR specific items.
```
bebraw: I would rather add clarifying comments to the configuration above. Then we could skip a sentence or two here.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+
+First, the plugins. While only the HMRPlugin itself is necessary to enable HMR, the NamedModulesPlugin will make your life easier by showing an intelligible module id in your browser console whenever a file is changed.
```
bebraw: Expand `HMRPlugin` to full name. That `NamedModulesPlugin` bit can be eliminated by moving the explanation as a code comment.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+
+Next, working backward, is the `devServer` key. While it's possible to configure webpack-dev-server directly from the CLI, or more specifically in a server script from the Node module API, this example approach is using the `devServer` key in the overall webpack config. `hot: true` turns on HMR. `publicPath` denotes the root from wich the server will serve up your code (e.g. `localhost:8080${PUBLIC_PATH}`, so in this case `localhost:8080/`). the `publicPath` should match the `output.publicPath`.
+
+Finally, your entry file should be preceded by the elements included here: `'react-hot-loader/patch'`, `'webpack-dev-server/client?http://localhost:8080'`, and `'webpack/hot/only-dev-server'`. The react-hot-loader entry is only necessary if you are using React. The `webpack-dev-server/client?` line should make sure to point at the dev server instance you are running (in case you have configured a different port, for example). Finally, the `only` in `hot/only-dev-server` means that updates will only be sent on clean builds/bundles; you can drop the `only-` if you would prefer seeing error output in your browser via HMR.
```
bebraw: I would split the paragraph at `Finally`. Separate idea.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+// ./src/components/App.css
+.app {
+    text-size-adjust: none;
+    font-family: helvetica, arial, sans-serif;
+    line-height: 200%;
+    padding: 6px 20px 30px;
+}
+```
+
+Now, the above code is using React, but it doesn't need to be. In fact, the only thing that matters above is the code refering to `module`. First, we wrap the HMR code inside of `module.hot` check; webpack exposes `module` to the code, and if we are running with `hot: true` configured, we'll enter the inside of the conditional. While the module API offers more options than what's above, the most important element is the `module.hot.accept` call. It specific how to handle changes to specific dependencies. So in this, it will file the `render` method ONLY when `src/components/App.js` changes! Note that would also include when the dependencies of `App.js` change - so the `render` method will file not just for changes made directly to the source of `App.js`, but also changes made to `App.css`, since `App.css` is included in `App.js`.
```
bebraw: Split up into two, three paragraphs.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+
+###Package.json
+Finally, we need to start up webpack dev server to bundle our code and see HMR in action. We can use the following package.json entry:
+
+```js
+  "start" : "webpack-dev-server --env.dev --content-base dist",
```
bebraw: Why to define `--content-base` here? Why not to push it to the configuration above?

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+[WDS] App updated. Recompiling...
+client?c7c8:91 [WDS] App hot update...
+dev-server.js:45 [HMR] Checking for updates on the server...
+log-apply-result.js:20 [HMR] Updated modules:
+log-apply-result.js:22 [HMR]  - ./components/App.js
+dev-server.js:27 [HMR] App is up to date.
+```
+Note that HMR specifies the paths of the updated modules. That's because we're using the NamedModules plugin!
+
+
```
bebraw: This bit could use a conclusion as per writer's guide.

bebraw: You could also link to [my chapter](http://survivejs.com/webpack/developing-with-webpack/automatic-browser-refresh/) related to the topic. The [React chapter](http://survivejs.com/webpack/advanced-techniques/configuring-react/#setting-up-hot-loading) has related info too.

content/how-to/hot-module-reload.md
TheLarkInn: Maybe mention both ways, it's okay to stay opinionated right? I usually set content-base on the CLI lol.

bebraw: Sure, you can mention both ways. `T>` is ideal for alternatives.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,186 @@
...
+const { resolve } = require('path');
+const webpack = require('webpack');
+
+const ENTRY = ['./index.js'];
+
+module.exports = env => {
+  return {
+    entry: env.dev?
+      [
+        'react-hot-loader/patch',
```
ericclemmons: I've seen conflicting reports of using `react-hot-loader/webpack`:
> https://github.com/gaearon/react-hot-boilerplate/pull/61
> https://github.com/gaearon/react-hot-loader/blob/v3.0.0-beta.1/src/webpack/index.js

From what I understand, `/patch` is for `entry` and `/webpack` is a `loader`.

I haven't tested it, but it'd be great to bring some clarification to this question, especially if it means simplifying `entry` (which becomes very convoluted in dev mode).

tleunen:
Thanks a lot @jmreidy for the guide! It helped me a lot.

I have an issue though, when trying to execute this code:
screen shot 2016-08-07 at 10 40 07 pm
Which then executes this:
screen shot 2016-08-07 at 10 41 42 pm
And then fails on:
screen shot 2016-08-07 at 10 43 00 pm
screen shot 2016-08-07 at 10 43 13 pm

content/how-to/hot-module-reload.md
```diff
@@ -1,4 +1,186 @@
...
+* webpack-dev-server 2 (tested with 2.1.0-beta.0)
+
+This guide will be demonstrating the use of HMR with Babel, React, and PostCSS (using CSS Modules). To follow along, please add the following deps to your `package.json`:
+
+```js
+"devDependencies": {
+    "babel": "^6.5.2",
+    "babel-core": "^6.11.4",
+    "babel-loader": "^6.2.4",
+    "babel-preset-es2015-webpack": "^6.4.2",
```
nodkz:
```js
    "babel": "^6.5.2",
    "babel-core": "^6.13.2",
    "babel-loader": "^6.2.4",
    "babel-preset-es2015": "^6.13.2",
```
Upgrade pakages, after that you may replace deprecated https://github.com/gajus/babel-preset-es2015-webpack on
```js
{
    "presets": [
        [
            "es2015",
            {
                "modules": false
            }
        ]
    ]
}
```

content/how-to/hot-module-reload.md
```diff
@@ -1,4 +1,186 @@
...
+    "webpack-dev-server": "^2.1.0-beta.0"
+  }
+```
+
+###Babel Config
+Your `.babelrc` file should look like the following:
+
+```js
+{
+  "presets": ["es2015-webpack", "stage-2", "react"],
```
nodkz:
```js
"presets": [["es2015", {"modules": false}], "stage-2", "react"],
```
After uprgading Babel till version v6.13.*

jmreidy:
Hey all - thanks for the feedback! I've begun integrating the changes suggested here, and will update you when they're complete.

jmreidy:
Docs are updated. If the review looks good, I'll rebase to a single commit.

@tleunen, could you try using the updated example information and see if it works?

tleunen:
I was already using the same configuration, for the presets. So I updated every dependency again and now the error is a bit different.

I both tested the normal `webpack` and the `webpack-dev-server`.

`webpack` runs fine, and the generated code works perfectly. (at least from what I saw).

The problem comes from `webpack-dev-server`. With or without the hot reload, it fails at the same place. But the error is a bit different though.
screen shot 2016-08-11 at 4 34 41 pm

There's somewhere a call to `__webpack_require__(request)` with an `undefined` request. Not sure why :/ Always when loading something in react-proxy/node_modules/lodash it seems

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,258 @@
+---
+title: How to Configure Hot Module Replacement?
```
TheLarkInn: Cc @bebraw 

I think since there are many different HMR recipes, maybe we could structure this in like How-To -> HMR Recipies -> Babel/React/Redux

I think that way we encourage other recipes in here for many state libraries and frameworks. And it reinforces a much asked 'cookbook' for certain topics.

content/how-to/hot-module-reload.md
```diff
@@ -0,0 +1,258 @@
...
+
+```js
+{
+  "presets": [
+    ["es2015", {"modules": false}],
+    //Webpack understands the native import syntax, and uses it for tree shaking
+
+    "stage-2",
+    //Specifies what level of language features to activate.
+    //State 2 is "draft", 4 is finished, 0 is strawman.
```
tleunen: typo `Stage` instead of `State`

tleunen:
@jmreidy I went throught your entire guide again and in fact I had missed one important detail...
I forgot to import the AppContainer and wrap my app with it ;)
`import { AppContainer } from 'react-hot-loader'`

TheLarkInn:
@tleunen was it something easily missed or do you think it could be stated better. It's things like this which I believe help developers from getting tripped up.

tleunen:
To be honest, the first time I read it was quite in diagonal. And the setup for rendering the component was very similar to what I already had in my application so I just missed the fact "AppContainer" was something specific to add.
But after re-reading it, I noticed it. I believe adding a quick comment could help because it's the only part of the documentation that has not any comment so it could be easy to miss the fact it's required for the hot reload.

Maybe something like:
```jsx
// Render the Application inside the Hot Reload AppContainer
ReactDOM.render(
    <AppContainer>
      <App/>
    </AppContainer>,
    document.getElementById('root')
);
```

jmreidy:
@tleunen @TheLarkInn Yeah, I think I should update the docs with both an inline comment and a note. `AppContainer` is such a generic name that it's easy to gloss over it.

@TheLarkInn @bebraw I like the idea of breaking this out into a specific HMR with React and CSS Modules thread. Do you think there'd be a benefit of structuring it with the actual source code? And switching this guide into a README.md for that code? That way, people could both read the docs AND fire it up in an editor.

bebraw:
@jmreidy Maybe we'll end up with articles like this:

* `concepts/hot-module-reloading` - Something like [this Stack Overflow answer](http://stackoverflow.com/a/24587740/228885).
* `how-to/set-up-hmr-with-react` (just an example) - There can be more specific how tos like this.
* `how-to/set-up-hmr-with-redux` (simple setup, but good to mention)
* `api/hot-module-reloading` - Literally the HMR interface.

These can cross reference to each other.

andreypopp:
I'd advice to have an example wit HMR without react-hot-loader. It's not yet finished according to its author (Dan Abramov) and Webpack HMR works fine with React on its own (you lose preserving React state and DOM but that's not really too bad).

jmreidy:
I like @bebraw's idea of making this a specific example of HMR working with React. Shold I close this PR in favor of switching it to a how-to/set-up-hmr-with-reach doc?

TheLarkInn:
we will have a general documentation page for just hmr too.

TheLarkInn:
You don't have to close the PR. Just go ahead and create the how-to folder if it doesn't already exist, and if it does drop the file in there instead.

ericclemmons:
@TheLarkInn pinged me on Twitter about this, and I think Webpack on the server + HMR is worthy of a post.  Having "recipes" (like those listed above) would be good to answer HMR for each situation so that users, such as myself, can see how it all works together.
> https://github.com/ericclemmons/webpack-hot-server-example

bebraw:
@jmreidy Ready to review again? Do you have anything else in mind for this PR?

jmreidy:
IMO it's ready to go!

bebraw:
Thanks!

I split up related issues: #85 (concept), #86 (API), #88 (Redux how to).

PR #71, opened 2016-08-20
docs(concepts): added new targets section

TheLarkInn:
Gotten started on the targets section. Work in progress.

### DIFF
--- a/content/concepts/targets.md
+++ b/content/concepts/targets.md
@@ -34,8 +34,7 @@ The following is a list of values you can pass to the target property.
 * `"node"` Compile for usage in a node.js-like environment (use `require` to load chunks)
 * `"async-node"` Compile for usage in a node.js-like environment (use `fs` and `vm` to load chunks async)
 * `"node-webkit"` Compile for usage in webkit, uses jsonp chunk loading but also supports build in node.js modules plus require("nw.gui") (experimental)
-* `"electron"` Compile for usage in [Electron](http://electron.atom.io/) – supports `require`-ing Electron-specific modules.
-* `"electron-renderer"` Compile for electron renderer process, provide a target using `JsonpTemplatePlugin`, `FunctionModulePlugin` for browser environment and `NodeTargetPlugin` and `ExternalsPlugin` for commonjs and electron bulit-in modules. *Note: need `webpack` >= 1.12.15.*
+* `"electron-renderer"` Compile for electron renderer process, provide a target using `JsonpTemplatePlugin`, `FunctionModulePlugin` for browser environment and `NodeTargetPlugin` and `ExternalsPlugin` for commonjs and electron bulit-in modules. *Note: need `webpack` >= 1.12.15.
 
 Each _target_ has a variety of deployment/environment specific additions, support to fit its needs. 
 
@@ -43,10 +42,10 @@ For example, when you use the `electron` target, *webpack* includes multiple `el
 
 ### Multiple Targets
 
-Although webpack does **not** support multiple strings being passed into the target property, you can create an isomorphic/universal library by bundling two separate configurations: 
+Although webpack does **not** support multiple strings being passed into the target property, you can create an isomorphic library by bundling two separate configurations: 
 
 **webpack.config.js**
-```javascript
+```javascriptis
 
 module.exports = [ serverConfig, clientConfig ];
 
--- a/content/concepts/targets.md
+++ b/content/concepts/targets.md
@@ -3,24 +3,25 @@ title: Targets
 ---
 ## Overview
 
-Because JavaScript can be written for both server & browser, *webpack* offer's multiple deployment _targets_.
+Because JavaScript can be written for both server and browser, webpack offers multiple deployment _targets_.
 
-*webpack* offer's a variety of deployment _targets_ that you can set in your webpack [configuration](./api/configuration).
+webpack offers a variety of deployment _targets_ that you can set in your webpack [configuration](./api/configuration).
 
-W> The webpack `target` property is not to be confused with the `output.target` property. For more information see [our guide](./concepts/output) on the `output` property.
+W> The webpack `target` property is not to be confused with the `output.libraryTarget` property. For more information see [our guide](./concepts/output) on the `output` property.
 
 
 ### Usage
 
 To set the `target` property, you simply set the target value in your webpack config: 
 
-```javascript
-// webpack.config.js
 
+**webpack.config.js**
+
+```javascript
 module.exports = config;
 
 config = {
-  target: 'node' // <===== Just pass the name of the target  
+  target: 'node'
 }
 ```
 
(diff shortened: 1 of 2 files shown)


TheLarkInn:
Need to add some resources, however is ready to have some eyes look at the content as a whole.

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
-> webworker
-> node
-> async-node
-> node-webkit
-> electron
-> electron-main
-> electron-renderer
+## Overview
+
+Because JavaScript can be written for both server & browser, *webpack* offer's multiple deployment _targets_.
```
bebraw: * Better as "server and browser".
* /s/offer's/offers
* Maybe stick to just webpack everywhere so that formatting is consistent.

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
+
+W> The webpack `target` property is not to be confused with the `output.target` property. For more information see [our guide](./concepts/output) on the `output` property.
+
+
+### Usage
+
+To set the `target` property, you simply set the target value in your webpack config: 
+
+```javascript
+// webpack.config.js
```
bebraw: Another bit we could standardize. I like to use the following format myself:

**webpack.config.js**

```javascript
module.exports = {
  target: 'node'
};
```

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
+
+module.exports = config;
+
+config = {
+  target: 'node' // <===== Just pass the name of the target
```
bebraw: You can skip "just". Often it doesn't add value and it's better to leave it out.

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
+
+* `"web"` Compile for usage in a browser-like environment (default)
+* `"webworker"` Compile as WebWorker
+* `"node"` Compile for usage in a node.js-like environment (use `require` to load chunks)
+* `"async-node"` Compile for usage in a node.js-like environment (use `fs` and `vm` to load chunks async)
+* `"node-webkit"` Compile for usage in webkit, uses jsonp chunk loading but also supports build in node.js modules plus require("nw.gui") (experimental)
+* `"electron"` Compile for usage in [Electron](http://electron.atom.io/) – supports `require`-ing Electron-specific modules.
+* `"electron-renderer"` Compile for electron renderer process, provide a target using `JsonpTemplatePlugin`, `FunctionModulePlugin` for browser environment and `NodeTargetPlugin` and `ExternalsPlugin` for commonjs and electron bulit-in modules. *Note: need `webpack` >= 1.12.15.*
+
+Each _target_ has a variety of deployment/environment specific additions, support to fit its needs.
```
bebraw: It would be good to expand on this point and actually list the differences here. Considering that, a table might work better than a list.

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
+* `"electron"` Compile for usage in [Electron](http://electron.atom.io/) – supports `require`-ing Electron-specific modules.
+* `"electron-renderer"` Compile for electron renderer process, provide a target using `JsonpTemplatePlugin`, `FunctionModulePlugin` for browser environment and `NodeTargetPlugin` and `ExternalsPlugin` for commonjs and electron bulit-in modules. *Note: need `webpack` >= 1.12.15.*
+
+Each _target_ has a variety of deployment/environment specific additions, support to fit its needs. 
+
+For example, when you use the `electron` target, *webpack* includes multiple `electron` specific variables. For more information on which templates and _externals_ are used, you can refer [directly to the webpack source code](https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsApply.js#L70-L185).
+
+### Multiple Targets
+
+Although webpack does **not** support multiple strings being passed into the target property, you can create an isomorphic/universal library by bundling two separate configurations:
```
bebraw: We should choose either isomorphic or universal and use the word consistently.

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
-> electron
-> electron-main
-> electron-renderer
+## Overview
+
+Because JavaScript can be written for both server & browser, *webpack* offer's multiple deployment _targets_.
+
+*webpack* offer's a variety of deployment _targets_ that you can set in your webpack [configuration](./api/configuration).
+
+W> The webpack `target` property is not to be confused with the `output.target` property. For more information see [our guide](./concepts/output) on the `output` property.
```
sokra: There is no `output.target`, did you mean the `output.libraryTarget` property?

content/concepts/targets.md
```diff
@@ -1,10 +1,80 @@
...
+### Options
+
+The following is a list of values you can pass to the target property. 
+
+* `"web"` Compile for usage in a browser-like environment (default)
+* `"webworker"` Compile as WebWorker
+* `"node"` Compile for usage in a node.js-like environment (use `require` to load chunks)
+* `"async-node"` Compile for usage in a node.js-like environment (use `fs` and `vm` to load chunks async)
+* `"node-webkit"` Compile for usage in webkit, uses jsonp chunk loading but also supports build in node.js modules plus require("nw.gui") (experimental)
+* `"electron"` Compile for usage in [Electron](http://electron.atom.io/) – supports `require`-ing Electron-specific modules.
```
sokra: Use `electron-main` instead of `electron`. `electron` is the legacy name.

bebraw: As a sidenote, does it give a deprecation warning if you try to use the old name?

sokra: no, we should add it.

content/concepts/targets.md
TheLarkInn: React users have used the term isomorphic and angular devs usually mention universal. I don't mind either term. Any opinions?

content/concepts/targets.md
TheLarkInn: Yup changed nice catch

content/concepts/targets.md
TheLarkInn: Should we even list the property then?

content/concepts/targets.md
TheLarkInn: I'll use isomorphic it's generally understood.

content/concepts/targets.md
bebraw: Nope. Just use `electron-main` and we are good. If you want, you can have a `T>` mentioning the deprecation so people using Electron know about it.

content/concepts/targets.md
TheLarkInn: How deep do we want to dive? I think that was my initial motivation for providing the `WebpackOptionsApply` src code.

bebraw: Yeah, linking there would work for now. If you want, you could add a `?>` for a TODO. That can be hidden from the production build as discussed earlier.

TheLarkInn:
So I added a Todo also to find some more up to date targets, to supplement I whipped up a super minimal but informative repo that people can try out using all the different webpack targets in individual dist folders.

PR #79, opened 2016-08-28
Finishing the Sidebar

gregvenech:
This is still a work in a progress...

- General styling updates (mostly positioning and padding)
- Updates to the Sidebar design and functionality

We'll probably have to wait on a few updates in Antwar before this can be merged. The things we would need are...

- A way to include scripts to add interactive pieces
- An array of `sections` added to each `page` in the `pages` array

Then we would need to:

- Add a click event to toggle visibility of the nested sections
- Make the sidebar stick on scroll (but contained)
- Order the pages and sections using the `order` or `next` field discussed in #93

### DIFF
--- a/components/Body.jsx
+++ b/components/Body.jsx
@@ -24,8 +24,10 @@ export default props => {
   return (
     <Body { ...props }>
       <Navigation home={ home } pages={ pages } />
-        <Sidecar />
-        { props.children }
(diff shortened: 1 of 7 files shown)


gregvenech:
Closing this for now and migrating these changes, which are mostly style related to another branch which I'll open a PR for momentarily.

PR #81, opened 2016-08-29
feat(coreconcepts): "add Core Concepts, Entry Points, and shape of webpack config"

TheLarkInn:
Still WIP. 

CC @developit for the latest "webpack-config.js" example.


content/concepts/core-concepts.md
```diff
@@ -2,12 +2,31 @@
 title: Core Concepts
 ---
 
-## Entry, Output, Loaders, Plugins
+## Overview
```
TheLarkInn: @bebraw I'll take some input on this section. Since we are creating separate files. I felt like the 2-3 min overview is what this page is trying to accomplish. 

With that in mind, do you think the content speaks to that? Need more explanation, or less?

content/concepts/webpack-config.md
```diff
@@ -0,0 +1,172 @@
...
+        break;
+    }
+  }
+}
+```
+
+#### Using JSX (React JS Markup) and Babel to create a JSON Configuration that webpack can understand. (Courtesy of [Jason Miller](https://twitter.com/_developit))
+
+```javascript
+/* Lib: git.io/jsxobj */
```
developit: As noted on twitter - might be best to reference `h()` via an import here:

```js
import h from 'jsxobj';

// ..etc
```

content/concepts/core-concepts.md
bebraw: I would say do an article per concept so you have `concepts/entry.md`, `concepts/output.md`, and so on. `concepts/index.md` can aggregate these and provide very short explanations and literally work as an index. There's a bit of explanation already. This would be in addition to that I think.

content/concepts/webpack-config.md
```diff
@@ -0,0 +1,172 @@
+---
+title: Webpack Config
```
bebraw: Perhaps rename this as `concepts/configuration.md`. No need to mention webpack at the title as it's given through context.

TheLarkInn: Sounds good! I agree.

content/concepts/core-concepts.md
TheLarkInn: So this is purely opinionated, and I'm open to whatever my thoughts were: 

`concepts/index.md`: Precursor and 10,000ft view of terms and things to expect
`concepts/core-concepts.md`: Declarative high level overview of what we feel are the core concepts which link to them individually.
`concepts/entry.md`: Deeper understanding of the entry and the multiple ways it can be used.  (and so on for each core concept). 

I wanted to take this approach to allow the first time user to flow from one document to the next. index => core concepts => entry => output => loaders => plugins.

content/concepts/webpack-config.md
TheLarkInn: done!

TheLarkInn:
Alright @webpack/contributor-team @bebraw @jhnns This baby is ready for review.

screen shot 2016-09-03 at 2 20 05 am

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+In the example above, through the `output.filename` and `output.path` properties we are describing to webpack the name of our bundle, and where we want it to be emitted to.
+
+T> You may see the term **emitted** or **emit** used throughout our documentation and [plugin API](../api/plugins). This is a fancy term for "produced or discharged". 
+
+The `output` property has [many more configurable features](../api/configuration), but lets spend some time understanding some of the most common use cases for the `output` property. 
+
+[**Learn more!**](./output)
+
+### Loaders
+The goal is to have all of your assets in your project to be **webpack's** conscern and not the browser. (This doesn't mean that they all have to be bundled together). webpack treats [every file (.css, .html, scss, jpg, etc.) as a module](./everything-is-a-module). However, webpack **only understands javascript**.
```
dev-zippyui: @TheLarkInn 
> However, webpack **only understands javascript**.

maybe rephrase as

> However, webpack **only understands javascript by default**.

so as to better introduce the idea that loaders make webpack understand other file types as well.

radubrehar: @TheLarkInn 
> However, webpack **only understands javascript**.

maybe rephrase as

> However, webpack **only understands javascript by default**.

so as to better introduce the idea that loaders make webpack understand other file types as well.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+
+**Loaders describes to webpack _how to treat these files as modules_ as they are added to your dependency graph.**
```
radubrehar: > **Loaders describes to webpack ...

=>

> **Loaders describe to webpack

or maybe even better,

> **Loaders tell to webpack

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+```
+  module.exports = config; 
+
+  var config = {
+    entry: './path/to/my/entry/file.js',
+    output: {
+      filename: 'my-first-webpack.bundle.js',
+      path: __dirname + 'dist/'
+    },
+    modules: {
```
radubrehar: `module` instead of `modules`

content/concepts/concepts.md
TheLarkInn: done

content/concepts/concepts.md
TheLarkInn: 👍

content/concepts/concepts.md
TheLarkInn: :shipit:

content/concepts/concepts.md
kentcdodds: hmmm... technically the original copy was correct though. The loader's job is to turn other files INTO JavaScript...

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Concepts
+---
+
+## Overview
```
bebraw: Skip `##Overview` header since intros are implicit. As a result you can push the other headers to `##`.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+webpack is [incredibly configurable](./api/configuration), however, there are **4 Core Concepts** we feel you should understand before you get started! 
+
+As part of your webpack learning journey, we wrote this document aimed to give you a **high-level** overview of these concepts, while still providing links to concept specific use-cases. 
+
+### Entry
+In the [previous article](./index), we mentioned that webpack creates a graph of all of your application's depenencies. The starting point of this graph is known as an _entry point_. The _Entry point_ tells webpack _where to start_ and follows the graph of dependendencies to know _what to bundle_. You can think of your applications _entry point_ as the **contextual root** or **the first file to kick off your app**.
```
bebraw: /s/depenencies/dependencies

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+In the example above, through the `output.filename` and `output.path` properties we are describing to webpack the name of our bundle, and where we want it to be emitted to.
+
+T> You may see the term **emitted** or **emit** used throughout our documentation and [plugin API](../api/plugins). This is a fancy term for "produced or discharged". 
+
+The `output` property has [many more configurable features](../api/configuration), but lets spend some time understanding some of the most common use cases for the `output` property. 
+
+[**Learn more!**](./output)
+
+### Loaders
+The goal is to have all of your assets in your project to be **webpack's** conscern and not the browser. (This doesn't mean that they all have to be bundled together). webpack treats [every file (.css, .html, scss, jpg, etc.) as a module](./everything-is-a-module). However, webpack **only understands javascript by default**.
```
bebraw: /s/conscern/concern (spellchecker can pick these up if you configure it right, we could also do a linting process)

bebraw: Inconsistency at `every file`. Missing dots.

bebraw: /s/javascript/JavaScript - there are probably other instances to fix too.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+      ]
+    },
+    plugins: [
+      new webpack.optimize.UglifyJsPlugin();
+      new HtmlWebpackPlugin({template: './src/index.html'})
+    ]
+  };
+```
+
+You will notice that there are many [built-in plugins already created](https://webpack.github.io/docs/list-of-plugins.html) for you to use out of the box.
```
bebraw: The link reads weird standalone (think screen readers). Better rephrase that.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+
+Using plugins in your webpack config is straight-forward, however there are many use-cases that are worth disscusing futher.
```
bebraw: /s/disscusing/discussing

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+W> It is important to remember when defining loaders in your webpack config, you are defining them under `module.loaders`, and not `loaders`.
+
+There are more specific properties to define on loaders that we haven't yet covered. 
+
+[**Learn more!**]('./loaders')
+
+### Plugins
+Since Loaders only execute transforms on a per-file basis, Plugins are most commonly used (but not limited to) performing actions and custom functionality on "compilations" or "chunks" of your bundled modules [(and so much more)](./plugins). The webpack Plugin system is [extremely and powerful and customizable](../api/plugins). 
+
+Because a plugin is an ES5 JavaScript class, using it is as trivial as `require()`ing it into your configuration and passing a **`new` instance** of that class to the `plugins` array.
```
jhnns: What is a ES5 JavaScript class? :) Will ES2015 classes work as well? ;)

webpack itself just requires an object with an `apply()` function. But since it needs to be configurable, it makes sense to use a constructor function for it. However, this is just a convention.

content/concepts/concepts.md
jhnns: /s/futher/further

:dancers:

content/concepts/configuration.md
```diff
@@ -0,0 +1,172 @@
+---
+title: Configuration
+---
+
+## Overview
```
jhnns: Same as @bebraw already mentioned. This reads more like an introduction than an overview.

content/concepts/configuration.md
```diff
@@ -0,0 +1,172 @@
...
+      minChunks: Infinity
+    }),
+    new webpack.optimize.AggressiveSplittingPlugin({
+        minSize: 5000,
+        maxSize: 10000
+    }),
+  ]
+};
+
+let targets = ['web', 'webworker', 'node', 'async-node', 'node-webkit', 'electron-main'].map((target) => {
```
jhnns: `var`, `let`, `const`?? So many options to choose :wink:

I'd just use `const` for every variable here. No matter what you choose, it should be consistent :)

jhnns: @bebraw should we have a guideline for code examples?

content/concepts/configuration.md
```diff
@@ -0,0 +1,172 @@
...
+      filename: '[name].' + target + '.js'
+    }
+  });
+  return base;
+});
+
+module.exports = targets;
+```
+
+#### Using TypeScript to create a class which the angular-cli uses to [generate configs](https://github.com/angular/angular-cli/blob/master/addon/ng2/models/webpack-config.ts)
```
jhnns: I think this paragraph and the one about jsx is far too specific to be in the core concepts chapter.

content/concepts/configuration.md
bebraw: No, but setting up one would be a fine idea. Good additions to the writer's guide. I would go with `const` by default as we don't have to support old Node.

content/concepts/concepts.md
TheLarkInn: Totally true I suppose. I wanted to give some semblance of pattern or relation between use and shape. Assuming not to be vague. 

Do you have a better suggestion.

content/concepts/configuration.md
TheLarkInn: Haha this is where opinions differ!! Yeah a guide would be good. Can you eslint code examples in mark down?

content/concepts/configuration.md
TheLarkInn: This is a separate document called `configuration`. 

Essentially it gives the learner a bit of context WRT understand the consistent 'shape' (or seemingly lack there of) of the config.

TheLarkInn: I like it because it's a linked doc separate from any learning section, yet important enough to live in concepts.

content/concepts/configuration.md
bebraw: It's simple enough to extract the code fragments using something like `marked`. Given we use ellipsis here and there the code won't be valid always by definition. That said, if you want to give it a go, I can point you to the right direction. The basic idea goes like this:

1. Glob all Markdown files.
2. Hook into marked lexer and extract code fragments. [Example](https://github.com/webpack/webpack.io/blob/develop/utilities/markdown.js).
3. Eliminate any custom markup (i.e., ellipsis).
4. Run eslint against the code fragments.

I set up #92 for this task.

content/concepts/concepts.md
jhnns: "In order to use a plugin, you just need to `require()` it and add it to the `plugins` array. Since most plugins are customizable via options, you need to create an instance of it by calling it with `new`"

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+      filename: 'my-first-webpack.bundle.js',
+      path: __dirname + 'dist/'
+    },
+    module: {
+      loaders: [
+        {test: /.jsx?$/, loader: 'babel-loader'} 
+      ]
+    },
+    plugins: [
+      new webpack.optimize.UglifyJsPlugin();
```
jhnns: `;` should be `,`

content/concepts/configuration.md
jhnns: That would be nice. We could define a set of custom markup that will be removed before linting.

content/concepts/concepts.md
TheLarkInn: :shipit:

content/concepts/concepts.md
TheLarkInn: :shipit:

content/concepts/concepts.md
TheLarkInn: :shipit: :shipit:

content/concepts/concepts.md
TheLarkInn: :shipit: I like that much better.

content/concepts/concepts.md
TheLarkInn: :shipit:

content/concepts/configuration.md
TheLarkInn: FWIW, if it is a const, the user wouldn't be able to update it later :smirk: which we've all seen before in the same file. (Also I'm only a const fan for real const and imports but thats opinionated bias).

TheLarkInn: But I updated to all const, I think consistent message is a good one!

content/concepts/concepts.md
TheLarkInn: on this file we're golden, haven't checked rest of project but yes I see some good linting stuff to be setup.

TheLarkInn:
@jhnns @bebraw second round of edits finished.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+There are multiple ways to declare your `entry` property that are specific to your applications needs. 
+
+[**Learn more!**](./entry-points)
+
+## Output
+Once you've bundled all of your assets together, we still need to tell webpack **where** to bundle our application. The webpack `output` property describes to webpack **how to treat bundled code**. 
+
+**webpack.config.js**
+```
+  module.exports = config;
```
bebraw: Why not to export the configuration directly? A little simpler that way.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+
+**webpack.config.js**
+```
+  module.exports = config; 
+
+  const config = {
+    entry: './path/to/my/entry/file.js',
+    output: {
+      filename: 'my-first-webpack.bundle.js',
+      path: __dirname + 'dist/'
```
bebraw: This isn't consistent with `entry`. I know it's about absolute/relative support, but some might find this confusing still. Maybe there's a better way.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
```
bebraw: Extra newline here.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+T> You may see the term **emitted** or **emit** used throughout our documentation and [plugin API](../api/plugins). This is a fancy term for "produced or discharged". 
+
+The `output` property has [many more configurable features](../api/configuration), but lets spend some time understanding some of the most common use cases for the `output` property. 
+
+[**Learn more!**](./output)
+
+
+## Loaders
+The goal is to have all of your assets in your project to be **webpack's** concern and not the browser. (This doesn't mean that they all have to be bundled together). webpack treats [every file (.css, .html, .scss, .jpg, etc.) as a module](./everything-is-a-module). However, webpack **only understands JavaScript**.
```
bebraw: If i remember right, `everything-is-a-module` went away and was renamed as `module` concept article. So tweak link accordingly.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+
+  const config = {
+    entry: './path/to/my/entry/file.js',
+    output: {
+      filename: 'my-first-webpack.bundle.js',
+      path: __dirname + 'dist/'
+    },
+    module: {
+      loaders: [
+        {test: /.jsx?$/, loader: 'babel-loader'}
```
bebraw: The RegExp can be written as `(js|jsx)` or so. Even though more to write, it reads better. Maybe favor that instead.

content/concepts/concepts.md
```diff
@@ -0,0 +1,129 @@
...
+      loaders: [
+        {test: /.jsx?$/, loader: 'babel-loader'} 
+      ]
+    }
+  };
+```
+
+In the configuration above we have defined our loader with its two required properties: `test`, and `loader`. It tells webpack's compiler the following: 
+
+```"Hey webpack compiler, when you come across a file with extension '.js' or '.jsx' inside of a require() statement, use the babel-loader to transform it before you bundle it together".```
```
bebraw: This could be the right section to mention about `resolve.extensions` and how webpack uses that for lookup.

content/concepts/concepts.md
jhnns: This is very subjective imho. I'd write it like `jsx?`... but we surely need to escape the dot 😁

content/concepts/concepts.md
jhnns: We should talk about the path resolution logic and that it can be customized, e.g. by modifying `resolve.extensions` or `resolve.alias`

bebraw: Yeah, great point. Maybe that should go to `concepts/modules.md` since it's related to module resolution.

content/concepts/concepts.md
TheLarkInn: I wanted to embed the idea right away, that these configs are code, exporting an object. And that all that matters is the properties inside of that object. So by using a variable, it kind of indirectly reinforces that concept. @jhnns @bebraw if you both vote to export directly I am fine with that.

content/concepts/concepts.md
TheLarkInn: so maybe just ./dist for right now? And then resolution doc/modules doc talk about that?

content/concepts/concepts.md
TheLarkInn: I pulled and merged develop but didn't see that file. Where is that getting pulled in?

content/concepts/concepts.md
TheLarkInn: Yes escape the dot at the least. I think I'll side with @bebraw and do the more verbose flavor.

content/concepts/concepts.md
TheLarkInn: :shipit:

content/concepts/concepts.md
TheLarkInn: :shipit:

TheLarkInn:
Besides comments responded, second round finished. 

maxresdefault

content/concepts/concepts.md
bebraw: Ah, yeah. Good point. Renamed.

content/concepts/configuration.md
```diff
@@ -0,0 +1,171 @@
+---
+title: Configuration
+---
+
+You may have noticed that few webpack configurations look exactly alike. This is because **webpack's configuration file is a JavaScript file that exports an object.** This object, is then parsed by webpack based upon its defined properties.
+
+## Examples
```
bebraw: Maybe drop this header. If you do that, you can push the level of following headers down by one so you get `##` for them.

content/concepts/configuration.md
```diff
@@ -0,0 +1,171 @@
...
+  return obj;
+}
+
+// example of an import'd plugin
+const CustomPlugin = config => ({
+  ...config,
+  name: 'custom-plugin'
+});
+
+const CONFIG = (
```
bebraw: Any thoughts on naming? We need to stick to some convention to keep this consistent.

TheLarkInn: Ahh well I suppose I these are examples. I like `config` for _our provided examples_ I didn't want to modify the samples for this page. (Unless you think it will be confusing).

content/concepts/configuration.md
TheLarkInn: Done

content/concepts/configuration.md
bebraw: Ah, ok. So it's about configuration styles. Got it.

TheLarkInn: Yeah do you think I should add a note about it?

bebraw: A tip would work, yup.

TheLarkInn: Note added, as well as linting errors corrected for CI

TheLarkInn:
Note added, as well as linting errors corrected for CI

bebraw:
Merged as discussed earlier.

utilities/markdown.js
```diff
@@ -53,7 +53,7 @@ module.exports = function(section) {
       code = code.replace(/ *<\/details>(\n)?/g, "\nMARKDOWNDETAILSEND\n");
       code = code.replace(/<summary>/g, "\nMARKDOWNSUMMARYSTART\n");
       code = code.replace(/ *<\/summary>/g, "\nMARKDOWNSUMMARYEND");
-      code = code.replace(/(?:  )?( *)MARKDOWNDETAILSSTART([\s\S]*?)MARKDOWNSUMMARYSTART\n/g, "MARKDOWNDETAILSSTART$2MARKDOWNSUMMARYSTART\n$1");
+      code = code.replace(/(?:{2})?( *)MARKDOWNDETAILSSTART([\s\S]*?)MARKDOWNSUMMARYSTART\n/g, "MARKDOWNDETAILSSTART$2MARKDOWNSUMMARYSTART\n$1");
```
bebraw: Where does this come from? It yields

```bash
ERROR in ./utilities/markdown.js
Module parse failed: /Users/juhovepsalainen/Projects/github/js/webpack.io/utilities/markdown.js Error parsing regular expression: Invalid regular expression: /(?:{2})?( *)MARKDOWNDETAILSSTART([\s\S]*?)MARKDOWNSUMMARYSTART\n/: Nothing to repeat (56:27)
You may need an appropriate loader to handle this file type.
SyntaxError: Error parsing regular expression: Invalid regular expression: /(?:{2})?( *)MARKDOWNDETAILSSTART([\s\S]*?)MARKDOWNSUMMARYSTART\n/: Nothing to repeat (56:27)
    at Parser.pp$4.raise (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2221:15)
    at tryCreateRegexp (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2773:46)
    at Parser.pp$7.readRegexp (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2834:7)
    at Parser.pp$7.readToken_slash (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2606:52)
    at Parser.pp$7.getTokenFromCode (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2732:19)
    at Parser.pp$7.readToken (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2477:17)
    at Parser.pp$7.nextToken (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2468:15)
    at Parser.pp$7.next (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:2413:10)
    at Parser.pp.eat (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:536:12)
    at Parser.pp$3.parseSubscripts (/Users/juhovepsalainen/Projects/github/js/webpack.io/node_modules/acorn/dist/acorn.js:1738:37)
 @ ./antwar.config.js 3:15-46
```

for me.

TheLarkInn: This was a mistake I will fix this right away.

TheLarkInn:
ah you already did it thank you very much!

PR #156, opened 2016-09-07
output options with webpack 1 definitions - wip

gambtho:
with @joshq00
used [WebpackOptionsDefaulter](https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js) as reference

Need more detail for the following options 
* [ ] `"call"`
* [ ] `"hashFunction"`
* [ ] `"hashDigest"`
* [ ] `"hashDigestLength"`

Need to review for changes in webpack 2
* [ ] `"filename"`
* [ ] `"chunkFilename"`
* [ ] `"library"`
* [ ] `"hotUpdateFunction"`
* [ ] `"jsonpFunction"`
* [ ] `"libraryTarget"`
* [ ] `"path"`
* [ ] `"sourceMapFilename"`
* [ ] `"hotUpdateChunkFilename"`
* [ ] `"hotUpdateMainFilename"`
* [ ] `"crossOriginLoading"`
* [ ] `"devtoolLineToLine"`

### DIFF
--- a/content/concepts/output.md
+++ b/content/concepts/output.md
@@ -1,12 +1,227 @@
 ---
 title: Output
 ---
-> output.path
-> output.filename
-> output.publicPath
-> output.chunkFilename
-> output.jsonpFunction
-> …
+
+## Overview
+
+Options affecting the output of the compilation. `output` options tell Webpack how to write the compiled files to disk. Note, that while there can be multiple `entry` points, only one `output` configuration is specified.
+
+If you use any hashing (`[hash]` or `[chunkhash]`), make sure to have a consistent ordering of modules. Use the `OccurrenceOrderPlugin` or `recordsPath`.
+
+
+### Usage
+
+To set the `output` property, you simply set the output value in your webpack config: 
+
+**webpack.config.js**
+
+```javascript
+module.exports = config;
+
+config = {
+  output: 'blah' 
+}
+```
+
+### Options
+
+The following is a list of values you can pass to the `output` property. 
+
+
+### `output.filename`
+
+Specifies the name of each output file on disk. You must **not** specify an absolute path here! The `output.path` option determines the location on disk the files are written. `filename` is used solely for naming the individual files.
+
+**single entry**
+``` javascript
+{
+  entry: './src/app.js',
+  output: {
+    filename: 'bundle.js',
+    path: __dirname + '/build'
+  }
+}
+
+// writes to disk: ./build/bundle.js
+```
+
+**multiple entries**
+
+If your configuration creates more than a single "chunk" (as with multiple entry points or when using plugins like CommonsChunkPlugin), you should use substitutions to ensure that each file has a unique name.
+
+`[name]` is replaced by the name of the chunk.
+
+`[hash]` is replaced by the hash of the compilation.
+
+`[chunkhash]` is replaced by the hash of the chunk.
+
+``` javascript
+{
+  entry: {
+    app: './src/app.js',
+    search: './src/search.js'
+  },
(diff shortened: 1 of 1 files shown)


gambtho:
whit this commit

grabbed all output options from the code, and copied anything that was present in the webpack 1 documentation

removed options that weren't in the code

sokra:
I think this is too details for a `concept` page. This content should be in a `api` page.

joshq00:
@sokra we had some discussion about what the "concept" pages should contain. There isn't much content, yet, so we kind of winged it.

Can you describe what you're looking for in these pages? Is it
* When/why you'd use each property (e.g. `output.filename`
* What this section is meant for (use `output` to...)
* Is this for contributors? Users?

TheLarkInn:
As a first time webpack users I would like to know the following: 

- [ ] Most common 1-2 examples of using output with different properties. 
- [ ] Use cases for some more common properties: **see below list of ones to not leave here.**
- [ ] Links to other properties ( like `entry` and `output.filename: '[name].chunk.js' )  that work together with  


## DO NOT ADD INFORMATION FOR THE FOLLOWING PROPS TO THIS DOC: 
 "hotUpdateFunction"
 "jsonpFunction"
 "hotUpdateChunkFilename"
 "hotUpdateMainFilename"
 "crossOriginLoading"
 "devtoolLineToLine"
 "call"
 "hashFunction"
 "hashDigest"
 "hashDigestLength"

@bebraw can you think of anything else? I think this appropriately covers what we're trying to do in this doc.

content/concepts/output.md
```diff
@@ -1,12 +1,227 @@
...
 title: Output
 ---
-> output.path
-> output.filename
-> output.publicPath
-> output.chunkFilename
-> output.jsonpFunction
-> …
+
+## Overview
```
bebraw: You can skip `## Overview` header and then bump `###` to `##` below.

content/concepts/output.md
```diff
@@ -1,12 +1,227 @@
...
+
+Options affecting the output of the compilation. `output` options tell Webpack how to write the compiled files to disk. Note, that while there can be multiple `entry` points, only one `output` configuration is specified.
```
bebraw: Better use webpack (lowercase) so it's consistent with the other material.

content/concepts/output.md
```diff
@@ -1,12 +1,227 @@
...
+```
+
+### Options
+
+The following is a list of values you can pass to the `output` property. 
+
+
+### `output.filename`
+
+Specifies the name of each output file on disk. You must **not** specify an absolute path here! The `output.path` option determines the location on disk the files are written. `filename` is used solely for naming the individual files.
```
bebraw: Why not absolute? In my experience absolute path works as well.

bebraw:
Yeah, we'll most likely end up with two docs:

* Concept level doc to describe the idea of **output** on a high level (entry level doc). Focus on explaining the idea and basic use cases.
* API level doc to dig into details. Focus on technical detail.

It's a good start, though.

michael-ciniawsky:
Could we merge this PR? It goes in the direction I was intending to do it, but I would like to edit a few things and add small examples to ```chunkFilename, sourceMapFilename```, how it actually looks like etc, maybe some structural things aswell. It seems unnecessary I PR my changes since most of it is already done here.

bebraw:
@michael-ciniawsky Done. See my [review notes](https://github.com/webpack/webpack.js.org/pull/156/files). There are a couple of smaller things to fix. Also CI needs some TLC.

Issue #25, opened 2016-06-29
API - Configuration


bebraw:
[Stub](https://github.com/webpack/webpack.io/blob/develop/content/api/configuration.md).

Feel free to comment here if you have ideas on what this guide should cover. Link to potential resources too.

This is to stub and rewrite the [existing page](http://webpack.github.io/docs/configuration.html)

gregvenech:
Can I take a stab at this one? I already started playing around with this when I was doing the [documentation mockup](https://github.com/webpack/docs/issues/69). Might need a some advice on certain sections but at this point I'm familiar with and have used most of the configuration options...

bebraw:
Sure, that would be great.

gregvenech:
Ok great, I'll start this week.

gregvenech:
@bebraw while working on this section, I noticed that our markdown parser doesn't accept:

 ```
... [link][1] ...

[1]: http://example.com
```

Is this intentional? I find this feature super helpful when adding long external links... it doesn't bother me as much to inline the in-site links.

bebraw:
Yeah. Footnotes aren't a good idea on the web. Hard to lay out, bad UX. It's better to use `T>` and `W>` for secondary information.

gregvenech:
I don't mean actually having the links at the bottom of the page, I just mean separating out long links to make the markdown more readable. What the user sees is unchanged by this, i.e. `[link][1]` and `[link](http://example.com/path/to/my/page)` yield the same result when used in markdown but the first makes paragraphs containing multiple links much easier to read and edit.

bebraw:
Ah, like that. If [marked](https://www.npmjs.com/package/marked) doesn't support that syntax by default, then you'll probably need a syntax extension for that purpose.

gregvenech:
Ok I'll try to look into that.

gregvenech:
Looks like a [bug](https://github.com/chjj/marked/issues/310) which isn't fully addressed yet (although that issue says it was in `0.3.2`).

bebraw:
Time to close this one given there's an entire section on configuration now.

PR #157, opened 2016-09-07
Configuration Documentation

sokra:
WIP


gregvenech:
@sokra hey I started working on this page as discussed with @bebraw in #25 but I hadn't created a pull request yet as I'm not through the first pass (I'm most of the way through resolve). You can check out what I have on the [api_configuration branch](https://github.com/webpack/webpack.js.org/tree/feature/api_configuration).

Should I just scrap what I've done or do you want me to merge it into this PR? I should be done in the with the first pass in the next few days and if I merged it into here maybe others can pick up some of the sections that I don't have as much knowledge on (I added TODOs throughout and kept the existing TODOs). Also I have a few ideas re structure if you're interested.

sokra:
hi @gregvenech yes we should definitely merge our branches. You may notice that I've only layouted this page but haven't written any content yet.

My idea was to add a code block with the configuration with a short explanation comment and a link to a more detailed section. This way you get a short overview. I also separated basic options from advanced options.

gregvenech:
Sounds good, yea I'll look at it more tonight and try to merge what I have into yours and follow your layout.

sokra:
@gregvenech merged both branches, we can continue working on that.

gregvenech:
Just added documentation for the rest of `resolve` and cleaned up a few things...

I like the idea with the linked code although I think the comments can be left out. I just came across [something similar](http://browsenpm.org/package.json) which might be another option, i.e. the hover behavior.

A few other thoughts:

- Definitely agree with the Basic/Advanced breakdown and I think there's a variety of ways we can go about it. Maybe the best would be something like a per option or per group expandable advanced section, e.g...

> ### Option | Group of Options
> 
> `string` `boolean`
> 
> Description...
> 
> ```js
> code
> ```
> 
> [Advanced Usage/Options...]()

This prevents breaking up the option groups (like `output` and `module`).

- There's definitely a lot of content to squeeze into this page, maybe we could consider breaking some out into separate pages. Even if we want to keep all the config options here, we could at least separate out **Passing a Configuration** and **External Configuration**

- Try to standardize the way we display value types (e.g. `string` `boolean`) and defaults for each option. This can probably carry over to other /api sections as well.

bebraw:
The problem with hover is that it won't work on tablets.

If you want to hide extra info for some reason (maybe better avoided), you could use [details](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d).

gregvenech:
Ahh didn't think of that. I like the details idea... but as you said maybe it isn't the best solution. I just think there's advanced advanced and basic options within the same group (for example in `module`) and it would be odd to completely break the advanced options out from the basic ones.

bebraw:
We have other pages for greater detail. I feel the main purpose of this page is to work as a quick reference.

That's actually how I use the old documentation. If I'm unsure of some property, I'll look up the field from the corresponding page and see what it's about. I guess the challenge in figuring out the right amount of detail.

gregvenech:
I guess I was kind of thinking of it the other way around e.g. this being the page with detailed descriptions of each option. I might have gone a little overboard with examples and descriptions on some. But I think a lot of the descriptions on the current page are way to vague even to get a quick picture of what the given option really does.

By other pages do you mean the entry, output etc concept pages? Or creating new pages?

bebraw:
> By other pages do you mean the entry, output etc concept pages? Or creating new pages?

The concept pages are in-depth, yeah. So if you want to understand the idea of `output` in great detail, you would navigate to the specific page from here. You can also link to specific techniques (how to) related to specific fields.

sokra:
I like the idea of using a single code block with advanced options hidden via `<details>`. We just need to add markdown support for this.

``` js
output: {
  filename: "bundle.js",

  /* Advanced output options (click to show) */
}
```

mxstbr:
I'm not a big fan of the structure as one massive markdown page with headings. It's not very scannable, and finding the option you're looking for is pretty hard imo. (especially if you have no idea what you're looking for, only what you want to do)

I'd much rather have this be a table-like list of all the options and a very very short description of what they do, that then link to the detail pages. That way I can find what I'm looking for more easily, and can still dive deep if I want/need to.

sokra:
> I'd much rather have this be a table-like list of all the options and a very very short description of what they do, that then link to the detail pages. That way I can find what I'm looking for more easily, and can still dive deep if I want/need to.

That's similar to what we want to do. Instead of a table we have a code block, a short description in a comment and a link to the detailed description.

TheLarkInn:
@mxstbr this document specifically will be for "those use to going to this page" however we will have more detailed use cases for each property

gregvenech:
> I'm not a big fan of the structure as one massive markdown page with headings. It's not very scannable, and finding the option you're looking for is pretty hard imo. (especially if you have no idea what you're looking for, only what you want to do)

I definitely agree that this page is getting very long and needs to be broken up in one way or another. As @sokra said, I think the code block example with links should work well to your point. @bebraw is going to add data (e.g. nested pages and hash links) to the `pages` array, which I'll incorporate into #79 making the `Sidebar` a full table of contents. Also planning on making it stick on scroll. Hopefully all these things, plus breaking it out into at least a few pages, should address the scannability of this part of the docs.

Right now I'm just trying to finish porting everything over and adding more detailed descriptions/examples in the process. Then it can be reviewed, edited, and parsed out in many of the ways discussed above.

gregvenech:
Almost... there... 😁

gregvenech:
Ok so at this point I've pretty much ported everything over from the existing configuration documentation. People should feel free to start reviewing.

In terms of structure and page breakdown, taking into consideration everything that's already been said, maybe something like this would work:

1. Break "Configuration" into it's own top-level Nav item
    - i.e. **Concepts - How Tos - Configuration ...**
2. Break this page up into these sections:
    - Top level description and @sokra's code-with-links-with-details example in the **index.md**
    - A detailed section for each **[Group of Options]**, e.g. `output`, `module`, `resolve`, etc.
    - An **Other Options** section with top-level properties that aren't in a **[Group]**, e.g. `cache`
        - Exception: Certain top-level, non-group, options such as `entry` might still warrant their own page
    - Non-option sections like **Passing a Configuration** and **External Configurations**
3. Finish reviewing and linking all these sections together and to relevant places

This way as soon as someone hits the **Configuration** section they can get an idea of all available options via the @sokra's code example. When in need of more detail, a user would be able to jump to the specific page#section either from the example or from the sidebar. 

**TODOs**: There are still a lot of these scattered around. Some came from the existing notes on the page when we started, some relate to going back through and linking or sectioning, but most are just the gaps for options I hadn't used much, don't have a great understanding of, or couldn't give a good use case for. I'm going to do another pass through to try to hit the ones I can but I'm hoping others will fill many of these in as well.

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+
+---
+
+### External Configurations
+
+?> TODO: possible extensions, i. e. .babel.js
+
+---
+
+Webpack is fed a configuration object. It is passed in one of two ways depending on how you are using webpack: through the Command Line (CLI) or via Node. Either method also allows an array of configuration objects, which are processed in parallel. This is more efficient than calling webpack multiple times. All the available configuration options are specified below...
```
bebraw: Command Line (CLI) -> Command Line Interface (CLI). Or go with terminal since that's well known word for it too.

I would drop the `...` from the end as it leaves the sentence hanging.

bebraw: If you pass an array to webpack does it actually execute in parallel by default? At least in webpack 1 you had to use a third party solution for that (parallel-webpack). Specify that better either way.

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
-> configuration file
-> possible extensions, i. e. .babel.js
-> exporting a function and --env
-> returning a Promise
-> exporting multiple configurations
 
-> see also [[Using the Cli]]
+?> TODO: Add a reference to [[Using the Cli]]
+
+### Passing a Configuration
```
bebraw: Instead of `###` go with `##`. That would seem more fitting here.

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+---
+
+### `output.path`
+
+`string`
+
+The output directory as an **absolute** path. Note that the `[hash]` subsitution may be used here as well.
+
+```js
+path: __dirname + '/dist/assets'
```
bebraw: Why not `path.join`? Does this work on Windows?

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+```
+
+---
+
+### `output.publicPath`
+
+`string`
+
+This option specifies the public address of the output files when referenced in a browser. For [loaders]() that embed `<script>` or `<link>` tags or reference assets like images, `publicPath` is used as the `href` or `url()` to the file when it’s different than their location on disk (as specified by `path`). This can be helpful when you want to host some or all output files on a different domain or CDN. As with `path` you can use the `[hash]` substitution for a better caching profile.
```
bebraw: I would split this paragraph in two. Seems like two ideas in one.

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+
+`include: /\/src/` - Include modules using a regex, string, or function
+
+`exclude: /node_modules/` - Exclude modules using a regex, string, or function
+
+`loader: "babel!eslint"` - A ! delimited string of loaders to use on these modules
+
+`loaders: [ "babel", "eslint" ]` - An array of loaders to use on these modules
+
+W> Note that loaders are always read from **right to left** whether passed via a delimited string or an array. In the example above, the [eslint-loader]() will lint, and possibly fix syntax in, the JavaScript modules and then hand them off to the [babel-loader]() for transpiling.
```
bebraw: It can be useful to illustrate the idea with a function call (`style(css(input))` or so).

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+### `module.preLoaders`
+
+`array`
+
+An array of [loader objects]() to be used as the first step in the loading process. In the example above, linting could be broken out into a *preLoader*:
+
+```js
+module: {
+  preLoaders: [
+    { test: /\.js/, exclude: /node_modules/, loader: 'eslint' }
```
bebraw: We probably should use `-loader` suffix for each `loader` definition for clarity. See https://github.com/webpack/webpack/issues/2986 .

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+### `module.loaders`
+
+`array`
+
+An array of [loader objects]() to be used as the second step in the loading process. Many times `module.loaders` will be the only set of loader objects needed. A basic configuration might look like this:
+
+```js
+module: {
+  loaders: [
+    { test: /\.js/, exclude: /node_modules/, loader: 'babel!eslint' },
```
bebraw: Instead of `loader: 'babel!eslint'` it might be better to use `loaders: ['babel-loader', 'eslint-loader']`. Even though more verbose, it's clearer. Same goes for the other loader definitions.

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+extensions: [ "", ".webpack.js", ".web.js", ".js" ]
+```
+
+which is what enables users to leave off the extension when importing:
+
+```js
+import File from '../path/to/file'
+```
+
+W> Setting this option will **override the default array**, meaning that webpack will no longer try to resolve modules using the default extensions. For modules that are imported with their extension, e.g. `import SomeFile from "./somefile.ext"`, to be properly resolved, an empty string must be included in the array. Similarly, for modules that are imported without extensions, e.g. `import _ from "underscore"`, to be resolved to files with `.js` extensions, you must include ".js" in your array.
```
bebraw: Feels long. Either rewrite or split into two paragraphs.

content/api/configuration.md
```diff
@@ -1,10 +1,1407 @@
...
+ cheap-source-map             | +     | o       | yes        | transformed code (lines only) 
+ cheap-module-eval-source-map | o     | ++      | no         | original source (lines only)  
+ cheap-module-source-map      | o     | -       | yes        | original source (lines only)  
+ eval-source-map              | --    | +       | no         | original source               
+ source-map                   | --    | --      | yes        | original source               
+
+
+T> See [`output.sourceMapFilename`](#output-sourcemapfilename) to customize the filenames of generated source maps.
+
+?> TODO: This section is pretty much just copied over from existing docs... imo more background is needed on the different types of source mapping, maybe via links to external sites that discuss the different types of source maps in more detail.
```
bebraw: Yeah. I have a feeling sourcemaps could use an article of their own. That would fit below concepts.

content/api/configuration.md
sokra: True. `path.resolve` is even better.
`path.resolve(__dirname, "dist/assets")` vs `path.join(__dirname, "dist", "assets")`

`__dirname + '/...` should never be used in documentation!

content/api/configuration.md
gregvenech: Done (for external configurations as well). Moved these sections to the bottom for now as I think we should probably move them to their own pages.

content/api/configuration.md
gregvenech: Making the CLI => terminal change and ellipsis fixes.

> If you pass an array to webpack does it actually execute in parallel by default?

That's what the [current docs](http://webpack.github.io/docs/configuration.html#multiple-configurations) say which is where I pulled most of this paragraph's content from. However after re-reading I decided to pull that part out as it can be discussed in **Passing a Configuration**, in fact there's already a TODO for it. I don't think passing multiple configs is something the average user is doing much, so it's probably fine to push it to a more detailed section.

content/api/configuration.md
gregvenech: Done.

content/api/configuration.md
gregvenech: Done.

content/api/configuration.md
gregvenech: Good idea, added in 6bc0547.

content/api/configuration.md
gregvenech: Sounds good, maybe we can mention/point to generating non-js source maps as well, e.g. [css source maps](https://github.com/webpack/css-loader#sourcemaps).

content/api/configuration.md
gregvenech: Yea I'll work on it...

> Similarly, for modules that are imported without extensions, e.g. `import _ from "underscore"`, to be resolved to files with `.js` extensions, you must include ".js" in your array.

I've ran into this before and it had me pulling my hair out for a little which is why I felt the need to mention here. However I feel like I read somewhere that you can't override the empty string anymore in version 2? Is that true? If so, maybe it can just be removed.

gregvenech:
What do you all think, @bebraw @sokra @TheLarkInn, about breaking template string substitutions into its own section and then referring to it from throughout the rest of the page? It seems like there's a lot of overlap between sections there so this could centralize and dedupe all that in one place.

Again, maybe this is another one that can be broken out into it's own page.

bebraw:
@gregvenech The way I see it, the function of this page is primarily to work as a *quick* reference (you want to see what this or that field does). So pushing more advanced info elsewhere could be worthwhile. Curious to hear what others think.

gregvenech:
@webpack/contributor-team I'm willing to break this up as soon as we decide how to we want to do that. I wanted people to have more time to review but maybe we should just start breaking it up first and then hit the TODOs page by page later.

It seems like the original plan for most of this documentation was to split it up into the various pages under `/concepts`. IMO, and what I originally thought, was that each concept page was going to be a more high-level discussion on the various topics. Personally I think if it is just detailed documentation with examples we should either rename the section or create a new section. 

Either way though I'm willing to break this up and pass it on to others for more review and editing.

bebraw:
What if we pushed the root level documentation for configuration to root level and even linked to it at the navigation? So you would end up with `webpack.js.org/configuration` and then `Configuration | Concepts | How to | API | GitHub` at the navigation.

It's literally one of the most important pages so making it prominent wouldn't hurt.

gregvenech:
@bebraw that's exactly what I was thinking. If everyone's good with this I'll make this change tonight.

sokra:
`Concepts | How to | Configuration | API | GitHub`

gregvenech:
Starting to break it out. Will finish this up, incorporate the new nav link, and merge tonight.

gregvenech:
Merged... we can review and edit each page in more granular pull requests now.

pastelsky:
Nice to see documentation for webpack  progressing, however I, as a user, am disappointed that some content is being picked up from the existing documentation since it's been confusing to use time and again. 

While serving the documentation on a well layed out site might help with the confusion, I really doubt it's going to be any less frustrating to understand the various terminologies used. 

For example, sample the current description given for the `devServer` config -
> ### `devServer`
 This set of options is picked up by [webpack-dev-server]() and can be used to change it's behavior in various ways. Here's a simple example that gzips and serves everything from our `/dist` directory:
  ```js
 devServer: {
   contentBase: "dist/",
   compress: true,
   port:9000
 }
 ```
When the server is started, there will be a message prior to the list of resolved modules:
  ```bash
 http://localhost:9000/
 webpack result is served from /build/
 webpack is served from dist/
 ```
 
> ### `devServer.contentBase`
 Tell the server where to serve content from. [`output.publicPath`](#output-publicpath) will also be used to determine where the bundles should be served from.

Note how various terms are used refer to similar things, without a distinction. Here's a brain dump of the questions that pop up in my mind - 

1) What is a contentBase? What do you mean by content? How is it different from what webpack builds? Is content different from assets?

2) Why is a contentBase different from a publicPath? They seem to do similar things.

3) Why do examples use both `\build` and `\dist` directories? Are they exclusive of each other, or was it because the docs were written by different authors?

4) Why have two different places to pass options to devServer? Are the options specified here merged/overwritten by ones I specify while using webpackDevServer in my express server?

TheLarkInn:
I appreciate you expressing this. So we came to two conclusions:

1. The existing content can be a very helpful 'cheatsheet' for those already familiar with webpack.
2. This is only the first step in the process. We wanted to save work and time involved while still preserving a familiar format for 'well-versed' users. What we now plan to do, is iterate over the quality of the information and provide the 'why' and 'how'. Right now we just have the 'what'.

gregvenech:
Hi @pastelsky, thanks for your feedback. As @TheLarkInn mentioned, this is only the first pass at the configuration documentation and as I said in my last comment anyone is welcome to review, edit and PR any of these sections. As you can see I also left a lot of TODOs throughout these pages on the places that I think are too ambiguous/vague and those that I just didn't have enough background using to rewrite. The reason I merged this a bit early was because we were starting to see overlap between here and **Concepts** and it was somewhat unwieldy to review as one long document. 

> What is a contentBase? What do you mean by content? How is it different from what webpack builds? Is content different from assets?

Agreed that some of these are ambiguous. Maybe something like *Decide what directory to serve the site from (usually where your `index.html` file lives).* 

> Why is a contentBase different from a publicPath? They seem to do similar things(or do they?)

They do not, `publicPath` tells webpack where your *webpack output* (i.e. anything that went through webpack, not only bundles) in relation to the directory you're serving from (i.e. `contentBase`).

> Why do examples use both \build and \dist directories? Are they exclusive of each other, or was it because the docs were written by different authors? `webpack is served from \dist` - wait what? Why are you serving me webpack?

They are exclusive of each other. The issue you describe here is more about what the actual `webpack-dev-server` logs out than how it's documented as I just copied that from an actual dev-server start up process. With that said though, I think digging through and rewriting the documentation will suggest a bunch of places where improvements can be made in the actual tools as well. Aside from the making some of the logging less ambiguous, I think some of the keys for the configuration options can be simplified or clarified as well.

Hope this helps, again please feel free to edit and PR these changes and any others you might have.

jhnns:
I agree with @pastelsky that mixing terms and examples is a problem. I really like the redux documentation that uses the TODO example throughout the documentation. This makes it easier to read different parts without switching contexts too often.

PR #162, opened 2016-09-08
feat(lint): add alex social sensitivity linter

TheLarkInn:
* Adds alex social sensitivity linter
  *  Upon using social insensitive language, one will receive lint errors and warnings. 

Although this isn't entirely dire in the context of what we are writing, it's simple and super helpful for keeping people in-check/aware just in case.

### DIFF
--- a/.alexrc
+++ b/.alexrc
@@ -1,0 +1,3 @@
+{
+  "allow": ["boogeyman-boogeywoman"]
+}
--- a/package.json
+++ b/package.json
@@ -24,12 +24,14 @@
     "start": "antwar -d",
     "build": "antwar -b && npm run sitemap && echo webpack.js.org > build/CNAME",
     "deploy": "antwar -D",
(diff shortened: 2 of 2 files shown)


TheLarkInn:
@bebraw These tests will probably fail still but I wanted to get all the lints out of the way first, and then work on one PR to ensure those tests pass.

bebraw:
Yeah, looks good. I wonder if it's possible to run the linters in a "race" (parallel). Most likely there's some Unix trick for that, but then we would leave Windows users out... I guess the current is enough.

TheLarkInn:
Yes, lol tbh, this seems like webpack loaders to me haha. But I can try to find a way to do it as future feature.

bebraw:
That's a nice idea. If we wrap the linters to loaders, then it becomes a matter of preloader configuration. That solves the cross-platform problem nicely. Parallelization might still be an issue (webpack runs in single process by default), but configuration is a better place for the problem.

TheLarkInn:
Yes def

PR #163, opened 2016-09-08
chore(lint): cleaned all linting errors added from new linting

TheLarkInn:
Also changed rules to fit with antwar (ie #header doesn't have to be top etc.)

### DIFF
--- a/.markdownlintrc
+++ b/.markdownlintrc
@@ -1,10 +1,13 @@
 {
   "default": true,
-  "MD003": { "style": "atx" },
+  "MD002": false,
   "MD007": { "indent": 4 },
-  "MD013": { "line_length": 200 },
+  "MD013": { "line_length": 500 },
+  "MD029": { "style": "ordered"},
   "MD033": false,
   "MD034": false,
+  "MD036": false,
+  "MD041": false,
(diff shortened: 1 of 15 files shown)


TheLarkInn:
Weird I don't see CI for this PR?

TheLarkInn:
Ahh there we go

.markdownlintrc
```diff
@@ -1,10 +1,13 @@
 {
   "default": true,
-  "MD003": { "style": "atx" },
+  "MD002": false,
   "MD007": { "indent": 4 },
-  "MD013": { "line_length": 200 },
+  "MD013": { "line_length": 500 },
```
TheLarkInn: This is here because of our custom tooltips, I'm not sure how to approach this one other than setting it super long.

TheLarkInn:
WOOOOT! Clean (er) docs!!!! Haven't added spelling but I think for a mix of concerns.

.markdownlintrc
bebraw: Yeah. We would need a smarter parser. I wouldn't worry about it. Keep the line length sufficiently long and it will be enough. `500` is a starting point.

PR #170, opened 2016-09-14
added contributors block to each page

sokra:
image


bebraw:
Maybe it would be better to dig the info from Git history? After that it would be automatic and would require zero maintenance.

sokra:
I think it's better manually, elsewise it would be affected by reformatting, renaming/moving, linting, metadata changes. I would exclude minor changes like typos too, but this can be discussed.

bebraw:
Ok, let's go with this and tweak later.

PR #171, opened 2016-09-15
chore(travis): update the node_js version on travis config

TheLarkInn:
I was noticing that all recent travis builds were still running against node 4 as the config. Am curious if this is related to the invariant issues. Going to try this out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 6
-
+  - "6"
 script:
   - bash ./scripts/deploy.sh


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - 6
-
+  - "6"
```
TheLarkInn: @bebraw I noticed on the official documentation for travis that they were using the version numbers as strings. Do I think this makes a difference, I am not 100% sure. But it appears that the tests are running against v6 now.

bebraw: Yeah, merged as string is better style.

PR #182, opened 2016-09-21
generate a list of repositories with stats

sokra:
image

We need to add more projects to the list... (also projects outside of the webpack organization).


TheLarkInn:
I think this is a good start.

PR #197, opened 2016-09-29
Fixed links in footer

michelgotta:
Fixed URLs in footer


bebraw:
That should work across the site. Thanks!

PR #198, opened 2016-09-30
feat(concepts/plugins): add overview for plugins

TheLarkInn:


content/concepts/plugins.md
```diff
@@ -1,5 +1,84 @@
 ---
 title: Plugins
 sort: 3
+contributors:
+  - TheLarkInn
 ---
-> plugins configuration option
+
+**Plugins** are the [backbone]('https://github.com/webpack/tapable') of webpack. webpack itself is built on the **same Plugin system** that you use in your webpack configuration!!
```
bebraw: * Extra ''s within the link.
* /s/same Plugin system/same plugin system

content/concepts/plugins.md
```diff
@@ -1,5 +1,84 @@
...
 ---
-> plugins configuration option
+
+**Plugins** are the [backbone]('https://github.com/webpack/tapable') of webpack. webpack itself is built on the **same Plugin system** that you use in your webpack configuration!! 
+
+They also serve the purpose of doing **anything else** that a [loader]('./loaders') cannot do.
+
+## Anatomy
+
+A webpack **Plugin** is a JavaScript object that has an `apply` property. This `apply` property is called by the webpack compiler, giving access to the **entire** compilation lifecycle.
```
bebraw: Plugin or plugin? Bolding might give enough emphasis alone.

content/concepts/plugins.md
```diff
@@ -1,5 +1,84 @@
...
+
+
+**ConsoleLogOnBuildWebpackPlugin.js**
+
+```javascript
+function ConsoleLogOnBuildWebpackPlugin() {
+
```
bebraw: Maybe explain here why the plugin is empty and/or what it could have.

content/concepts/plugins.md
```diff
@@ -1,5 +1,84 @@
...
+
+**ConsoleLogOnBuildWebpackPlugin.js**
+
+```javascript
+function ConsoleLogOnBuildWebpackPlugin() {
+  
+};
+
+ConsoleLogOnBuildWebpackPlugin.prototype.apply = function(compiler) {
```
bebraw: Are there other styles? Are you forced to use `prototype`? How about ES6 classes (prototypes underneath).

content/concepts/plugins.md
```diff
@@ -1,5 +1,84 @@
...
+
+  const config = {
+    entry: './path/to/my/entry/file.js',
+    output: {
+      filename: 'my-first-webpack.bundle.js',
+      path: './dist'
+    },
+    module: {
+      loaders: [
+        {test: /\.(js|jsx)$/, loader: 'babel-loader'}
```
bebraw: I might split this into three lines, but that's personal preference.

content/concepts/plugins.md
TheLarkInn: I know there are other styles, but I guess showing the other ones feels more appropriate for api/plugins. (Maybe less confusing). Thoughts?

bebraw: I guess specific info like this would fit below `content/plugins` better. It's fine to show a basic example here. Still good to cover somewhere perhaps.

content/concepts/plugins.md
TheLarkInn: done

content/concepts/plugins.md
TheLarkInn: done

content/concepts/plugins.md
TheLarkInn: done

content/concepts/plugins.md
TheLarkInn: > A webpack **plugin** is a JavaScript object that has an `apply` property.

For what its worth I never mention prototype and JS class. I think we'll have more trouble keeping the information simple if we try and show multiple types here.

bebraw: Yup. This particular doc can be simple as it's entry to the concept.

TheLarkInn:
Any other requested changes you'd like to see?

PR #199, opened 2016-09-30
Update antwar to newer alpha version

michelgotta:


bebraw:
Ah, pushed and updated already while you were doing the PR. Thanks for the effort, though. Closing. 👍

PR #200, opened 2016-09-30
Update links entry point page 

zhaoshengjun:
The links on entry page were pointing to wrong place and now they are fixed.

### DIFF
--- a/content/concepts/entry-points.md
+++ b/content/concepts/entry-points.md
@@ -3,7 +3,7 @@ title: Entry Points
 sort: 0
 ---
 
-Like we mentioned in the [introduction]('./'), there are multiple ways to define the `entry` property in your webpack configuration. We will show you the ways you **can** configure the property, in addition to explaining why it may be useful to you.
+Like we mentioned in the [introduction](./), there are multiple ways to define the `entry` property in your webpack configuration. We will show you the ways you **can** configure the property, in addition to explaining why it may be useful to you.
 
 ## Single Entry (Shorthand) Syntax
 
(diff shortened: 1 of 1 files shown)


content/concepts/entry-points.md
```diff
@@ -79,7 +79,7 @@ const config = {
 
 **What does this do?** At face value this tells webpack to create dependency graphs starting at both `app.js` and `vendors.js`. These graphs are completely separate and independant of eachother. (AKA there will be a webpack bootstrap in each bundle). This is commonly seen with single page applications which have only one entry point (excluding vendors).
 
-**Why?** This setup allows you to leverage [`CommonsChunkPlugin`](../api/plugins/commonschunkplugin) and extract any vendor references from your app bundle into your vendor bundle, replacing them with `__webpack_require__()` calls. If there is no vendor code in your application bundle, then you can achieve a common pattern in webpack known as [long-term vendor-caching.](../how-to/cache).
+**Why?** This setup allows you to leverage [`CommonsChunkPlugin`](../../api/plugins/commonschunkplugin) and extract any vendor references from your app bundle into your vendor bundle, replacing them with `__webpack_require__()` calls. If there is no vendor code in your application bundle, then you can achieve a common pattern in webpack known as [long-term vendor-caching.](../../how-to/cache).
```
bebraw: I wonder if we should go through the root here. If we move the page for some reason, then we'll break links again.

zhaoshengjun:
The other way is not use the relative path but use a absolute path instead. But it also depends on the whole website structure. If it keeps changing, none of these links will work.

bebraw:
Yeah. I think I'll merge this as is. Adding 404 checker to the build process would catch these (I have a rough idea how to integrate one).

PR #203, opened 2016-10-02
Feature/rwd

oliverturner:
Basic responsiveness added: 
- Viewport metatag
- An additional breakpoint + mixin to adapt styles to increasing resolutions
- Basic mobile menu

Remaining issues: 
- The absence of accessible state on the parent means that the menu persists on route changes
- It's unclear to me how to render the sections/pages, since the page objects don't contain a reference to section items


gregvenech:
Just pulled and started yours up locally, @TheLarkInn here's one:

screen shot 2016-10-02 at 2 21 10 pm

@oliverturner I think this looks great, thank you! There are a few other tweaks we can make but this is a big first step and I agree with @TheLarkInn that we should get this in soon. 

@oliverturner I can take a stab at the Sidebar page sections tonight. @bebraw and I had discussed this being at the antwar level before, but using the same function in the `markdown.js` utility we can parse the anchor data out from each page's content and generate those links.

oliverturner:
localhost-3000-concepts ipad 1
localhost-3000-concepts ipad
localhost-3000-concepts iphone 5 1
localhost-3000-concepts iphone 5
localhost-3000-concepts iphone 6 1
localhost-3000-concepts iphone 6

bebraw:
@oliverturner Can you open separate issues for the bits you need so I will remember to tackle those? Thanks!

oliverturner:
@gregvenech @bebraw @TheLarkInn Glad I could help!

@gregvenech 
> I can take a stab at the Sidebar page sections tonight.
That would be completely awesome 👍 

@bebraw 
> Can you open separate issues for the bits you need so I will remember to tackle those? 
Absolutely. Thanks for the earlier guidance!

PR #212, opened 2016-10-04
Rough draft for Getting started guide

pksjce:
I'm proposing a starting point for the history part. 
Shall we link it to the [survivejs article](http://survivejs.com/webpack/webpack-compared/) from here?

After this I feel we need to link to `how-to/install` and `concepts/` folder. Not sure how to do this without it appearing sparse. What do you  suggest @bebraw ?
If you are fine with the approach I'll polish this to link all the pieces mentioned.

### DIFF
--- a/content/get-started.md
+++ b/content/get-started.md
@@ -3,26 +3,21 @@ title: Get Started with Webpack
 ---
 #### Why `webpack`?
 
-It has traditionally been difficult to efficiently bundle and optimise front end applications and make them browser ready.
-The problems have generally been about not having a standard module system in client side javascript, having varied static assets such 
-as css, html, pngs, svgs, fonts which are essential to a web app. All of these static assets have to be transpiled(es2015 to es5),
-bundled(concatenate js and libraries) and optimised(uglified/minified) for the web.
+`webpack` is a module bundler.
+A bundler is a tool that builds all your project files to a single asset file that can be loaded onto a browser.
+It requires that your application files be written as [modules](/concepts/modules).
+The project structure consists of a single entry file with references to other modules as dependencies creating a tree like dependency graph.
+The dependency modules can belong to your application or can be third party libraries.
+The bundler creates this dependency graph and parses through it to include all the required modules in the final bundle.
+The final bundle file will contain all the modules in the right order of dependency and none of the modules will be duplicated.
+Other examples of module bundlers are [browserify]() and [rollupjs]()
 
-There are various tools that help us to do this(grunt, gulp, browserify etc). The modern ones can be roughly divided into two types of tools.
-##### Task runners
-Grunt and gulp fall into this category. In this eco system, transformations are applied on your code using task configuration.
-Each task can be performed using the respective plugin (grunt-concat, grunt-babel etc). A file is required to configure each of the tasks
-and to specify the order of the tasks for execution.
-##### Module bundlers
-Webpack, browserify, rollup are module bundlers.
-A module bundler is different in the sense that firstly, it mandates that your app is split into modules starting with an entry point file. Secondly, it creates a dependency
-graph by statically analysing your code for modular dependencies. Lastly, it bundles your application by following the dependency graph from the entry point.
+Unlike module bundlers, [Grunt](http://grunt),[Gulp](),[Broccoli]() are task runners.
+These tools perform a set of configured tasks in the order that the config file specifies.
+In fact, `webpack` can be run as [one of the tasks](/how-to/use-with-third-party-tools) by a task runner.
 
-Module bundlers are generally accepted to be a superior and reliable way of bundling all your static assets.
-
-Webpack can also
-* Transform the asset files to modules using loaders.(Eg-convert css files to modules, convert es2015 to commonjs modules)
-* Customise optimizations granularly on the created bundle through plugins.
-* Provide stats on the dependency graph of your application.
-* Provide code-splitting capabilities, so that you can load assets for your application as and when required by the end user.
+`webpack` can bundle javascript files written using Commonjs, AMD and ES2015 modules.
+It can also transform other assets(css, png, fonts..) to a module and bundle it together with the javascript.This is done using [loaders](/concepts/loaders).
+`webpack` can perform customised operations before, after or during bundle creation using [plugins](/concepts/plugins).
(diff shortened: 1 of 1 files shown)


bebraw:
It's maybe easier if we define context for this article as that will shape the content. I would suggest the following constraints:

* The reader doesn't know what a bundler is -> explain concept and why bundlers are needed.
* The reader wants to get webpack up and running -> show minimal setup + link to other articles for more info.
* Expect the reader has 5-15 mins of time -> basic ideas only, cut through essentials so that we "upgrade" the reader and encourage the person to check out more articles.

We can link to other resources at the end. If I remember right, the idea was to have some sort of official comparison at the site with numbers and so on.

pksjce:
Okay, Now my understanding is to have 
1. Why webpack, with more emphasis on bundlers and logic. 
2. What constitutes minimum setup for webpack? Would steps for 
- have to install webpack
- create an entry.js
- connect it to a dependency file
- run webpack from cli
be enough for this part?
3. About entry, output, existence of loaders, existence of plugins?
My approach would be to add a webpack.config.js file,
- Add entry and output file config as first step
- Add babel loader as second step
- Add uglifyjsplugin as third step

bebraw:
Yeah, that's good. When discussing some specific concept, like a loader, you should link to the concept specific article.

pksjce:
Hey @bebraw , [Rollup](http://rollupjs.org/) has this awesome interactive getting started on its homepage. Can we also do something like that?

bebraw:
@pksjce You could comment on that at #32.

pksjce:
Done. Also, I simplified module bundler definitions. What do you think?

content/get-started.md
```diff
@@ -1,7 +1,23 @@
 ---
 title: Get Started with Webpack
 ---
+#### Why `webpack`?
```
bebraw: Drop this header, intro is implicit.

content/get-started.md
```diff
@@ -1,7 +1,23 @@
...
 
-* basic concepts, get webpack up and running, technical focus
-* target - the user should be able to complete this in 10 minutes
+`webpack` is a module bundler.
+A bundler is a tool that builds all your project files to a single asset file that can be loaded onto a browser.
+It requires that your application files be written as [modules](/concepts/modules).
```
bebraw: I would merge these sentences so you get longer paragraphs. Reads a little better.

content/get-started.md
```diff
@@ -1,7 +1,23 @@
...
+The bundler creates this dependency graph and parses through it to include all the required modules in the final bundle.
+The final bundle file will contain all the modules in the right order of dependency and none of the modules will be duplicated.
+Other examples of module bundlers are [browserify]() and [rollupjs]()
+
+Unlike module bundlers, [Grunt](http://grunt),[Gulp](),[Broccoli]() are task runners.
+These tools perform a set of configured tasks in the order that the config file specifies.
+In fact, `webpack` can be run as [one of the tasks](/how-to/use-with-third-party-tools) by a task runner.
+
+`webpack` can bundle javascript files written using Commonjs, AMD and ES2015 modules.
+It can also transform other assets(css, png, fonts..) to a module and bundle it together with the javascript.This is done using [loaders](/concepts/loaders).
```
bebraw: /s/javascript/JavaScript

chrisdothtml:
Much like the intro, I also think the Getting Started page should be very simple and light on wording. As an example, when I first started using gulpjs, I was able to follow their [getting started](https://github.com/gulpjs/gulp/blob/master/docs/getting-started.md) guide and actually *get started* in a couple minutes

pksjce:
Yeah, the gulp getting started guide is really clear. This piece is mostly
intended to be something of an intro to webpack and clear the confusion
around module bundlers. I feel gulpjs comes off the back of Gruntjs, so it
doesnt need much explanation of things. If you checkout Gruntjs docs
<http://gruntjs.com/>, they have an wordy explanation of what it does.
In my experience, people are not using webpack because they dont see its
difference from a task runner. Also, I was just inspired from how the rollupjs
doc <http://rollupjs.org> clearly showed what it does for my javascript
files.

Regards,
Pavithra.K

content/get-started.md
```diff
@@ -2,6 +2,18 @@
 title: Get Started with Webpack
 ---
 
-* basic concepts, get webpack up and running, technical focus
-* target - the user should be able to complete this in 10 minutes
+`webpack` is a module bundler.A bundler is a tool that builds all your project files to a single asset file that can be loaded onto a browser.It requires that your application files be written as [modules](/concepts/modules).
```
mxstbr: What does "builds all your project files" mean? We need a better word so beginners can get an idea what webpack can do.

content/get-started.md
```diff
@@ -2,6 +2,18 @@
...
+The project structure consists of a single entry file with references to other modules as dependencies creating a tree like dependency graph.The dependency modules can belong to your application or can be third party libraries.
```
mxstbr: I'd much rather make this about webpack again, what about somethingn like "Webpack starts with your entry file, which has references to other modules, which have references to other modules, etc. This creates a tree-like _dependency graph_ that webpack can work with."

content/get-started.md
```diff
@@ -2,6 +2,18 @@
...
+
+The bundler creates this dependency graph and parses through it to include all the required modules in the final bundle.The final bundle file will contain all the modules in the right order of dependency and none of the modules will be duplicated.
+Other examples of module bundlers are [browserify]() and [rollupjs]()
```
mxstbr: I wouldn't link to competing tools here.

content/get-started.md
```diff
@@ -2,6 +2,18 @@
...
+
+Unlike module bundlers, [Grunt](http://grunt),[Gulp](),[Broccoli]() are task runners.
```
mxstbr: I wouldn't link to these, though you could mention them. Instead of talking about what the difference to these is already, I'd much rather talk a bit more about webpack is.

Everything above is very… theoretical, in a way. It doesn't show me what webpack can do to _make my life easier_.

pksjce: Yes, I'm also pretty disappointed with the description I came up with. It all sounds very theorectical as you said. I only wanted to get across the idea of a module bundler and that webpack is a great one at that. Would you help me draft this again? Can I DM you?

PR #214, opened 2016-10-05
rename 'al' to 'eval'

codetony25:
I believe it 'al' was suppose to be 'eval'


bebraw:
Please PR against `develop` in the future. Thanks.

`master` has been reserved for deployment purposes only.

PR #213, opened 2016-10-04
feat(migration-guide): added System.import

TheLarkInn:
Added code-splitting guide from "Whats new in webpack 2 old doc" into new one.

I noticed that @addyosmani was linking https://gist.github.com/sokra/27b24881210b56bbaff7 in his newest Medium publications, its probably best we just add this content to the new page which is more comprehensive and detailed for webpack 2.x. 

cc// @bebraw

PR #218, opened 2016-10-05
add: how-to/lazily-load/react

iammerrick:
I offer my humble strategy for lazily loading React. 

Heavily inspired by my friend @ryanflorence

### DIFF
--- a/content/how-to/lazy-load/react.md
+++ b/content/how-to/lazy-load/react.md
@@ -1,0 +1,148 @@
+---
+title: How to lazy load React 
+contributors:
+  - iammerrick
+---
+
+By using the [bundle-loader](https://github.com/webpack/bundle-loader) we are able to create a React component that allows us to declaratively split & load code in semantic bundles. We can also provide a higher order component function that will allow a component to specify its own dependencies. This allows consumers and authors of components to decide what is lazily loaded. A component can lazily load dependencies without it's consumer knowing, and a consumer can lazily load it's children without its children knowing, or some combination of both. By using the bundle loader we can semantically name chunks to intelligently load groups of code.
+
+## LazilyLoad Component
+
+Lets have a look at a consumer choosing to lazily load some components. The `importLazy` is simply a function that returns the `default` property, this is for Babel/ES2015 interoperability. If you don't need that you can omit the `importLazy` helper.
+
+```jsx
+<LazilyLoad modules={{
+  TodoHandler: importLazy(require('bundle?lazy&name=todo!./components/TodoHandler')),
+  TodoMenuHandler: importLazy(require('bundle?lazy&name=todo!./components/TodoMenuHandler')),
+  TodoMenu: importLazy(require('bundle?lazy&name=todo!./components/TodoMenu')),
+}}>
+{({TodoDetailsHandler, TodoMenuHandler}) => (
+  <TodoHandler>
+    <TodoMenuHandler>
+      <TodoMenu />
+    </TodoMenuHandler>
+  </TodoHandler>
+)}
+</LazilyLoad>
+```
+
+## Higher Order Component
+
+As a component, you can also make sure your own dependencies are lazily loaded. This is useful if a component relies on a really heavy library. Let's say we have a Todo component that optionally supports code highlighting...
+
+```jsx
+class Todo extends React.Component {
+  render() {
+    return (
+      <div>
+        {this.props.isCode ? <Highlight>{content}</Highlight> : content}
+      </div>
+    );
+  }
+}
+```
+
+We could then make sure the expensive library powering the Highlight component is only loaded when we actually want to highlight some code:
+
+```jsx
+// Highlight.js
+class Highlight extends React.Component {
+  render() {
+    const {Highlight} = this.props.highlight;
+    // highlight js is now on our props for use
+  }
+}
+export LazilyLoadFactory(Highlight, {
+  highlight: require('bundle?lazy!highlight'),
+});
+```
+
+Notice how the consumer of Highlight component had no idea it had a dependency that was lazily loaded? Or that if a user had todos with no code we would never need to load highlight.js? Also, we didn't need to specify a name for the highlight.js library because we don't care if webpack gives it a bundle.
+
+
+
+## The Code
+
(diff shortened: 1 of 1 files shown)


content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,148 @@
...
+export const LazilyLoadFactory = (Component, modules) => {
+  return (props) => (
+    <LazilyLoad modules={modules}>
+      {(mods) => <Component {...mods} {...props} />}
+    </LazilyLoad>
+  );
+};
+
+export const importLazy = (fetch) => (cb) => (
+   fetch((mod) => cb(mod.default))
```
TheLarkInn: Instead of fetch let's use `System.import()`, and then that way you dont _have to use_ bundle-loader.

Would you consider showing both ways. Super valuable. webpack will return a Promise the same way so not much code changes.

iammerrick: No problem @TheLarkInn! Also, I created and linked to an example repository to show it in action: https://github.com/iammerrick/how-to-lazy-load-react-webpack

content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,151 @@
...
+
+export const importLazy = (promise) => (
+  promise.then((result) => result.default)
+);
+
+export default LazilyLoad;
+```
+## Tips
+
+- By using the bundle loader we can semantically name chunks to intelligently load groups of code.
```
statianzo: Link "bundle loader" or perhaps remove this as `System.import` is what's used in the example

content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,151 @@
...
+export const LazilyLoadFactory = (Component, modules) => {
+  return (props) => (
+    <LazilyLoad modules={modules}>
+      {(mods) => <Component {...mods} {...props} />}
+    </LazilyLoad>
+  );
+};
+
+export const importLazy = (promise) => (
+  promise.then((result) => result.default)
```
statianzo: Given this behavior, it should be noted this will only work with the default export of an es6 module.

content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,151 @@
...
+    </TodoMenuHandler>
+  </TodoHandler>
+)}
+</LazilyLoad>
+```
+
+## Higher Order Component
+
+As a component, you can also make sure your own dependencies are lazily loaded. This is useful if a component relies on a really heavy library. Let's say we have a Todo component that optionally supports code highlighting...
+
```
statianzo: I like this example

content/how-to/lazy-load/react.md
TheLarkInn: Yes very good to note.

content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,151 @@
...
+export const importLazy = (promise) => (
+  promise.then((result) => result.default)
+);
+
+export default LazilyLoad;
+```
+## Tips
+
+- By using the bundle loader we can semantically name chunks to intelligently load groups of code.
+- Make sure if you are using the babel-preset-2015, to turn modules to false, this will allow webpack to handle modules.
```
statianzo: I don't know how to do this. A code snippet could be a handy cue.

Also, is it necessary other than to get the tree shaking perks?

TheLarkInn:
I feel bad asking because this is such good work, but we want to keep like short life and succinct reads. I think each example is perfect as a separate page. @bebraw thoughts? If you disagree leave as is, but I think separate markdown files may be easier to digest for users.

karlhorky:
I guess this is for webpack 2. What are your collective thoughts with providing an example for webpack 1 too? Could be a followup PR too. A webpack 1 example may also be interesting for `create-react-app` @gaearon

TheLarkInn:
So for these doc pages we want to limit example to webpack v2. However I am working on our migration guide to include mention of `require.ensure` https://github.com/webpack/webpack.js.org/pull/213

content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,151 @@
+---
+title: How to lazy load React 
+contributors:
+  - iammerrick
+---
+
+A component can lazily load dependencies without it's consumer knowing using higher order functions, or a consumer can lazily load it's children without its children knowing using a component that takes a function and collection of modules, or some combination of both.
```
TheLarkInn: As someone very limited in using React, although I understand HOC, I think it would be awesome to have a click able link to something from @gaearon or React documentation.

iammerrick:
@karlhorky Originally this was Webpack 1 (see first commit :-) 

@TheLarkInn I'm not really sure what you are advocating. The implementation couples the HoC & Function as Child components. So we can't really trim the LazilyLoad.js which is why splitting into separate pages would be a bit odd. What do you think?

content/how-to/lazy-load/react.md
iammerrick: Yeah its necessary, otherwise babel converts to CommonJS (and System.import doesn't get picked up!

content/how-to/lazy-load/react.md
iammerrick: I'll link to Higher Order Components on React patterns... http://reactpatterns.com/#Higher-order component

content/how-to/lazy-load/react.md
iammerrick: I'll note that!

iammerrick:
@TheLarkInn Feedback has been implemented, as well as a link to the react-modules project (per our discussion on Twitter™).

content/how-to/lazy-load/react.md
```diff
@@ -0,0 +1,154 @@
...
```
statianzo: > ~~it's~~ its consumer knowing
> ~~it's~~ its children knowing

TheLarkInn:
I think you are right, let's leave it as one. Good context

iammerrick:
Yay! @TheLarkInn Thanks for the gratitude and help. As always, appreciate working with you!

karlhorky:
Are these docs already deployed online somewhere?

bebraw:
@karlhorky Not online yet. We can sync over the weekend or so.

karlhorky:
Thanks!

PR #219, opened 2016-10-07
Simple Intro

chrisdothtml:
This obviously isn't complete, but I'm curious what other people think. I think it's really important to start right out of the gate with a full, simple, example to demonstrate how easy it is to setup.

Once a developer is familiar with the core concept (which this is), **then** you can start slowly introducing things like plugins, CSS bundling, etc.

### DIFF
--- a/content/index.md
+++ b/content/index.md
@@ -21,12 +21,12 @@ module.exports = {
 **webpack.config.js**
 
 ```javascript
-  module.exports = {
-    entry: './index.js',
-    output: {
-      path: 'bundle.js'
-    }
-  };
+module.exports = {
+  entry: './index.js',
+  output: {
+    path: 'bundle.js'
+  }
+}
 ```
 
 ## It's really that simple, [get started](/get-started) now!
--- a/content/index.md
+++ b/content/index.md
@@ -1,5 +1,3 @@
-## Who said front-end code can't be modular?
-
 **index.js**
 
 ```js
@@ -18,14 +16,17 @@ module.exports = {
 }
 ```
 
-## Bundle your modules with webpack.
+## Bundle your [modules](/concepts/modules) with webpack.
 
 **webpack.config.js**
 
 ```javascript
   module.exports = {
-    entry: './index.js'
+    entry: './index.js',
+    output: {
+      path: 'bundle.js'
+    }
   };
 ```
 
-## It's really that simple.
+## It's really that simple, [get started](/get-started) now!
--- a/content/index.md
+++ b/content/index.md
@@ -20,7 +20,7 @@ module.exports = {
 
 **webpack.config.js**
 
-```javascript
+```js
 module.exports = {
   entry: './index.js',
   output: {
@@ -29,4 +29,18 @@ module.exports = {
 }
 ```
 
-## It's really that simple, [get started](/get-started) now!
+**page.html**
+
+```html
+<html>
+  <head>
+    ...
+  </head>
+  <body>
+    ...
+    <script src="/bundle.js"></script>
+  </body>
+</html>
+```
+
+## It's really that simple. [Get started](/get-started) now!
--- a/content/index.md
+++ b/content/index.md
@@ -1,5 +1,31 @@
----
-title: Splash Content
----
+## Who said front-end code can't be modular?
 
-Still debating what to do with this section.
+**index.js**
+
+```js
+const foo = require('./foo.js')
+
+foo.bar()
+```
+
+**foo.js**
+
+```js
+module.exports = {
+  bar: function () {
+    //
+  }
+}
+```
+
+## Bundle your modules with webpack.
+
+**webpack.config.js**
+
+```javascript
+  module.exports = {
+    entry: './index.js'
+  };
+```
+
+## It's really that simple.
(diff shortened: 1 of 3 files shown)


bebraw:
@pksjce Can you pitch in as you are working the frontpage too?

SpaceK33z:
@chrisdothtml awesome that you followed through with this :).

I think it's a good idea to start with a simple code sample to get the developer interested and then slowly introduce the concepts.

content/index.md
```diff
@@ -1,5 +1,31 @@
...
+}
+```
+
+## Bundle your modules with webpack.
+
+**webpack.config.js**
+
+```javascript
+  module.exports = {
+    entry: './index.js'
```
SpaceK33z: webpack won't compile with this, you also need to add:

```js
output: {
  path: "bundle.js"
}
```

as a minimum.

bebraw:
@chrisdothtml Can you see #222 ?

chrisdothtml:
@SpaceK33z fixed the code sample.

Also updated the index a bit more and added a link to getting started.

@bebraw I definitely like that idea; don't think it would be good for the initial intro, or even getting-started though

chrisdothtml:
Screenshot of current progress:

Pull Request Progress

jhnns:
That is a good idea! 👍 Especially because CSS, HTML and asset bundling are only extras. The main purpose is still JS bundling.

Regarding the presentation: I think, it should be visually more clear what the steps are, like Step 1: Create an `index.js` with that content, Step 2. ... etc. But this doesn't need to be part of this PR. This can be improved incrementally.

chrisdothtml:
@jhnns Definitely agree. My thoughts were this would be the simplest intro, and the getting started page would have step by step instructions

bebraw:
@pksjce Do you want to comment here? It would be useful to see how these ideas go together. We'll end up with some sort of progression, no doubt.

pksjce:
@chrisdothtml - I agree with the need for code samples on the first page. In #222 - my aim was to show natural progression of webpack config for increasing complexity ie babel and css.
So how would we make these work together? Or we choose one form of representation. I think in the previous docs, users got a bit annoyed when we portrayed webpack as simple. So better to show that it is complex, but get the simplest version of all the moving pieces in one place.

chrisdothtml:
@pksjce I actually don't think the issue is that people think webpack is too simple; I think the docs are unapproachable. It's unclear exactly what webpack does, or what it's main purpose is.

jhnns:
Although webpack *can* be simple (only two arguments are actually required), "powerful" would probably be a better adjective.

TheLarkInn:
Call me crazy, but I almost think that displaying the output of that bundle right there on the page (if we can somehow do it), may tie together the whole "you-can-use-modules-in-your-code-and-webpack-makes-it-browser-ready-for-you" which I think most people _miss_. Thoughts?

bebraw:
Yeah, showing output could be nice indeed. Maybe we could find a way to bridge this with #222 so you could get a quick overview and then start tweaking to see different possibilities. #222 does the technical bit for that, maybe we can get the layout from here.

chrisdothtml:
@TheLarkInn Maybe we could do a "cleaner" version of the bundle just to illustrate that it's all the module code processed together

bebraw:
Could we get something simple merged this week? No need to be perfect. 👍 

Any thoughts on this @skipjack?

chrisdothtml:
Better or worse than last screenshot? This one uses something more familiar (hello world), shows the terminal command, and shows the code bundled

Possible updates

skipjack:
@bebraw yea I'm thinking I'll merge this into the splash-page PR I open later tonight. I can deal with the conflicts better over there.

@chrisdothtml I like it! I feel like is as barebones as it gets which I think is what we want. I guess the only thing that could be simplified even more is the **Bundle them with webpack** section, where we could just show `webpack entry output`. However, if we want to point people towards using the config maybe that change doesn't make sense.

chrisdothtml:
@skipjack You're right. Simplest version can be without the config file entirely. Will push these changes and update the bundle section

bebraw:
I wonder if we could borrow the idea of interactivity from #222. We can show a simple example by default but also allow the user to explore more advanced scenarios through some UI control. Maybe it could work out in the same layout.

skipjack:
@bebraw yea I actually got a webpackbin through an iframe working pretty well on my splash-page branch. We can play around with it there, I've been trying to finish up one thing with the visualization before opening the PR but no matter what I'll open that tonight and share some screenshots.

bebraw:
@skipjack Ok. Yeah, I agree getting something simple done first is the way to go. I would experiment with limited interactive (template based -> fast to load, no dependency on third party) after that before going to a heavier direction (iframes).

skipjack:
@bebraw sounds good.

cc @pkjse

chrisdothtml:
What about something like what [Sequence.js](http://www.sequencejs.com/) has where it's sort of a scripted interactive demo (user can't actually edit, but is still interactive). I'm just weary of having a full blown embedded code editor on the homepage/intro

NejcZdovc:
@chrisdothtml I really like this idea. Sequence example is very nice.

skipjack:
Merged this so I can pick up the changes on my branch. @chrisdothtml any other changes to `index.md` can go in a new PR.

PR #221, opened 2016-10-07
Suggesting `webpack-isomorphic-tools` for server-side rendering.

vpanjganj:
This dependency helped me to retrieve my chunk unique hashes while rendering my views on the server side. My view wasn't an html file so I couldn't simply use `html-webpack-plugin`. I thought this worth mentioning and save people's time.


bebraw:
Thanks!

PR #222, opened 2016-10-07
[Feature] Be able to show quick code samples for getting started

pksjce:
The examples here have the full working code in [https://github.com/pksjce/webpackdocs-getting-started](https://github.com/pksjce/webpackdocs-getting-started)
The result looks like below.
You can change the common scenarios using the dropdown and I show different code snippets for them.
screen shot 2016-10-07 at 6 04 34 pm

### DIFF
--- a/components/ShowCode.jsx
+++ b/components/ShowCode.jsx
@@ -1,0 +1,105 @@
+import React, {Component} from 'react';
+import Snippet from './Snippet.jsx';
+
+export default class ShowCode extends Component {
+    constructor(){
+        super();
+        this.options = [{
+            id: 1,
+            text: 'Simple',
+            mainjs:"import knocker from './knock-knock.js';\n\nlet html = ['<div>' + knocker.knock() + '</div>',\n\t\t'<div>whos there?</div>',\n\t\t'<div>' + knocker.whosthere() + '</div>',\n\t\t'<div> webpack who? </div>',\n\t\t'<div>' + knocker.webpackwho() + '</div>'];\n\ndocument.getElementById('knock').innerHTML = html.join('');",
+            secondaryjs:{
+                name:'knock-knock.js',
+                template: "const knocker = {\n\tknock: function() { return 'Knock Knock'; },\n\twhosthere: function() { return 'webpack'; },\n\twebpackwho: function() { return 'Dayum!' }\n}\n\nexport default knocker"
+            },
+            webpackjs: {
+                template: "module.exports = function () { \n\t return {\n\t\tentry: './main.js',\n\t\toutput: {\n\t\t\tpath: './dist',\n\t\t\tfilename: 'bundle.js'\n\t\t}\n\t}\n}"
+            },
+            iframe:'<iframe width="100%" height="300" src="//jsfiddle.net/d3gv23fg/embedded/js,result/" allowfullscreen="allowfullscreen" frameborder="0"></iframe>'
+        },{
+            id:2,
+            text: 'With babel-loader',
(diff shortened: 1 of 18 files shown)


components/ShowCode.jsx
```diff
@@ -0,0 +1,105 @@
+import React, {Component} from 'react';
+import Snippet from './Snippet.jsx';
+
+export default class ShowCode extends Component {
+    constructor(){
+        super();
+        this.options = [{
```
pksjce: @bebraw - Need to push these templates to some other place ideally

styles/components/_showcode.scss
```diff
@@ -0,0 +1,34 @@
+.showCode {
```
pksjce: @bebraw - I want to use styles from here. But I'm not sure how its connected to the jsx file.

components/Snippet.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Highlight from 'react-highlight';
+require('style!css!highlight.js/styles/github-gist.css');
```
pksjce: @bebraw - Is there a better way to do this?

styles/components/_showcode.scss
bebraw: Relying on class names should be enough. I know it's global CSS but it's not too bad. 🎱

components/Snippet.jsx
bebraw: Here, https://github.com/webpack/webpack.js.org/blob/develop/webpack.config.js#L9 . You can point from the entry.

gregvenech:
Looks like a great start to me. Maybe we should show a css or json import because I feel like webpack allowing the modularization of assets as well is one of the strongest points for using it over other tools. That said we'd still have to find a way to keep it simple. I'll try to pull your branch and play around with it a bit more today.

jhnns:
We could use [webpackbin](http://www.webpackbin.com/) by @christianalfoni for that

pksjce:
Webpackbin looks really good. Where can I see the generated webpack config file?

pksjce:
@gregvenech - Did you get any time to evaluate this idea? I have three code examples as of now
1. Simple setup [entry/output]
2. Setup with babel [entry/output/babel-loader]
3. Setup with css [entry/styles/css-loader]

components/ShowCode.jsx
pksjce: Made a separate snippets folder and passed it through highlight-loader!

components/Snippet.jsx
pksjce: This worked!

gregvenech:
@pksjce sorry the last couple days have been hectic for me. Now that I'm finished with #225 I'm going to start working on the visualization section of the splash page. I'll pull your branch and test it out as well. Mind if I PM you on gitter so we can discuss this page more and try to get something up and running?

I came across webpackbin recently as well. If they have an embed option maybe thats an easier fix than doing everything custom. Either way I think we should link to webpackbin throughout the docs as it would definitely be a great resource for learning to use webpack.

bebraw:
@christianalfoni Any thoughts on webpackbin embeds to webpack docs?

pksjce:
@gregvenech - Sure. I pinged you on gitter myself :)
I looked at webpackbin but I didnt see a way to actually see the `webpack.config.js` that is generated. Maybe I missed something.

pksjce:
Decided to close this for now. We will use webpackbin and try to enhance that to our needs.

bebraw:
Having something simple and standalone has still value. The problem is that [webpackbin hasn't seen active development in months](https://github.com/christianalfoni/webpack-bin) and the embed problem is still waiting for an answer.

pksjce:
People copy/paste webpack config all the time. It would be awesome if they could copy paste it right out of the docs.

pksjce:
@skipjack - Maybe we could revisit this PR now that #264 is closed?

skipjack:
Definitely, from my memory of testing it out a while ago I think if we just tweak the example code a bit and update the design it could work well. I think the important thing is to keep it fairly simple so it doesn't become a heavy burden to maintain, but still functional enough that it's useful for beginners to play around with.

Do you have an idea where this would go in the current version of the site (e.g. replace the current splash page example, or somewhere in getting-started maybe)?

PR #224, opened 2016-10-09
Co-locating components with Stylesheets/Assets

gregvenech:
Still doing a little bit of testing but I think this is good to go:

- Created component directories containing the jsx, scss, and any assets used in that component
- This change fixes css hot reloading on the dev-server
- Updated and simplified `webpack.config.js`

I wanted to get this done before porting over @KyleAMathews's typography/design changes and starting to dig into the splash page more heavily with @pksjce.

@bebraw @oliverturner please take a look to see if I missed or broke anything. 

**Note**: This change does mean that component stylesheets will have to individually specify their `/partials` dependencies. But this is a [good thing](https://www.bensmithett.com/smarter-css-builds-with-webpack/), right?


oliverturner:
> Note: This change does mean that component stylesheets will have to individually specify their /partials dependencies. But this is a good thing, right?

Agreed: explicit dependencies are better than implicit :)

PR #232, opened 2016-10-11
Move module.exports to syntactically valid positions

### LINKED ISSUE #228: Valid configuration syntax/ordering?
In examples throughout [the Concepts section of the docs](https://github.com/webpack/webpack.js.org/blob/master/content/concepts/index.md), this kind of pattern is used:
```js
module.exports = config;

const config = {
  entry: './path/to/my/entry/file.js'
};
```

However, this appears to be invalid because the `config` variable is referenced for the export before it is defined. The result is errors like this (under Node 6.x):
```sh
ReferenceError: config is not defined
    at Object.<anonymous> (/Users/timdorr/src/conf.js:1:80)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at repl:1:14
    at sigintHandlersWrap (vm.js:22:35)
```
A `var` works without errors, but returns `undefined`, as it doesn't have a temporal dead zone like `const`/`let` do.

It should be a simple fix to move those exports to the bottom.

tizmagik:
Fixes #228 

- Moves `module.exports` to syntactically valid positions (e.g. after `const config = {...}` is defined).
- In some cases, removes `module.exports` line entirely when it has already been written at least once in the doc (as per @timdorr 's suggestion)

### DIFF
--- a/content/concepts/entry-points.md
+++ b/content/concepts/entry-points.md
@@ -12,25 +12,21 @@ Usage: `entry: string|Array<string>`
 **webpack.config.js**
 
 ```javascript
-  module.exports = config;
+const config = {
+  entry: './path/to/my/entry/file.js'
+};
 
-  const config = {
-    entry: './path/to/my/entry/file.js'
-  };
-
+module.exports = config;
 ```
 
 The single entry syntax for the `entry` property is a short hand for:
 
 ```javascript
-  module.exports = config;
-
-  const config = {
-    entry: {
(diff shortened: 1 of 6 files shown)


bebraw:
Thanks. 👍

PR #237, opened 2016-10-11
Move stats documentation to separate page

### LINKED ISSUE #226: Configuration options "stats" missing in docs
I cannot seem to find any documentation about the stats option which seems to be valid according to the error message:

```
Warning: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration has an unknown property 'failOnError'. These properties are valid:
   object { amd?, bail?, cache?, context?, dependencies?, devServer?, devtool?, entry, externals?, loader?, module?, name?, node?, output?, plugins?, profile?, recordsInputPath?, recordsOutputPath?, recordsPath?, resolve?, resolveLoader?, stats?, target?, watch?, watchOptions? }
```

SpaceK33z:
As described in #226, this moves the documentation I wrote about `devServer.stats` to a separate page, because it also applies when using webpack itself.

Fixes #226

### DIFF
--- a/content/configuration/stats.md
+++ b/content/configuration/stats.md
@@ -1,0 +1,44 @@
+---
+title: Stats
+contributors:
+  - SpaceK33z
+---
+
+The `stats` option lets you precisely control what bundle information gets displayed. This can be a nice middle ground if you don't want to use `quiet` or `noInfo` because you want some bundle information, but not all of it.
+
+T> For webpack-dev-server, this property needs to be in the `devServer` object.
+
+### `stats`
+
+`object` `string`
+
+There are some presets: `none`, `errors-only`, `minimal` and `verbose`. Use them like this:
+
+```js
+stats: "errors-only"
+```
+
+For more granular control, it is possible to specify exactly what information you want:
+
+```js
+stats: {
+  chunks: false,
+  hash: false
+}
+```
+
+The available options are:
+
+* `hash`
+* `version`
+* `timings`
+* `assets`
+* `chunks`
+* `modules`
+* `reasons`
+* `children`
+* `source`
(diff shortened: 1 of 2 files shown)


content/configuration/stats.md
```diff
@@ -0,0 +1,44 @@
...
+title: Stats
+contributors:
+  - SpaceK33z
+---
+
+The `stats` option lets you precisely control what bundle information gets displayed. This can be a nice middle ground if you don't want to use `quiet` or `noInfo` because you want some bundle information, but not all of it.
+
+T> For webpack-dev-server, this property needs to be in the `devServer` object.
+
+### `stats`
```
bebraw: Use ## over ### here.

content/configuration/stats.md
```diff
@@ -0,0 +1,44 @@
...
+For more granular control, it is possible to specify exactly what information you want:
+
+```js
+stats: {
+  chunks: false,
+  hash: false
+}
+```
+
+The available options are:
```
SpaceK33z: Aii, I feared that you would ask this. I agree with you, but is it okay if we do that later? These options are not documented anywhere (in fact, all that I've written here comes from reading the source directly), so it costs me a lot of time to find out what everything stands for.

Also note that this wasn't even documented before, so I think for the MVP this is enough. I think it's better to focus on other tickets that need to be ready for the MVP.

content/configuration/stats.md
SpaceK33z: Done.

content/configuration/stats.md
bebraw: Sure. I setup #240 for that.

Issue #239, opened 2016-10-12
Ordering


NejcZdovc:
I would suggest to order properties alphabetically, because now it's really hard to find what you are looking for when you are looking for a specific property. When you have them ordered alphabetically you can easily spot what you are looking for

For example
https://webpack.js.org/configuration/module/#rule
https://webpack.js.org/configuration/dev-server/#devserver

Current order for dev-server options:

```
devServer.publicPath
devServer.quiet
devServer.noInfo
.....
```

I would change this to:
```
devServer.compress
devServer.contentBase
devServer.hot
```

I can work on this PR if you agree with a change. I would go trough all the pages and apply alphabetically order for the properties.

bebraw:
Sounds like a worthwhile change to me.

Any thoughts @TheLarkInn, @SpaceK33z ?

SpaceK33z:
Jep, good idea!

NejcZdovc:
So, should I start working on this?

TheLarkInn:
Love it lets do it!

PR #255, opened 2016-10-16
Ordering

NejcZdovc:
PR based on #239

### DIFF
--- a/content/api/cli.md
+++ b/content/api/cli.md
@@ -98,15 +98,15 @@ This set of options allows you to manipulate certain output parameters of your b
 
 | Parameter                    | Explanation                                                     | Input type | Default value                                         |
 |------------------------------|-----------------------------------------------------------------|------------|-------------------------------------------------------|
-| --output-path                | The output path for compilation assets                          | string     | Current directory                                     |
-| --output-filename            | The output filename of the bundle                               | string     | [name].js                                             |
 | --output-chunk-filename      | The output filename for additional chunks                       | string     | filename with [id] instead of [name] or [id] prefixed |
-| --output-source-map-filename | The output filename for the SourceMap                           | string     | [name].map or [outputFilename].map                    |
-| --output-public-path         | The public path for the assets                                  | string     | /                                                     |
+| --output-filename            | The output filename of the bundle                               | string     | [name].js                                             |
 | --output-jsonp-function      | The name of the JSONP function used for chunk loading           | string     | webpackJsonp                                          |
-| --output-pathinfo            | Include a comment with the request for every dependency         | boolean    | false                                                 |
 | --output-library             | Expose the exports of the entry point as library                | string     |                                                       |
 | --output-library-target      | The type for exposing the exports of the entry,point as library | string     | var                                                   |
+| --output-path                | The output path for compilation assets                          | string     | Current directory                                     |
+| --output-pathinfo            | Include a comment with the request for every dependency         | boolean    | false                                                 |
+| --output-public-path         | The public path for the assets                                  | string     | /                                                     |
+| --output-source-map-filename | The output filename for the SourceMap                           | string     | [name].map or [outputFilename].map                    |
 
 #### Example Usage
 
@@ -163,9 +163,9 @@ These options makes the build watch for changes in files of the dependency graph
 | Parameter                 | Explanation                                             |
 |---------------------------|---------------------------------------------------------|
 | --watch, -w               | Watch the filesystem for changes                        |
-| --watch-stdin, --stdin    | Exit the process when stdin is closed                   |
 | --watch-aggregate-timeout | Timeout for gathering changes while watching            |
 | --watch-poll              | The polling interval for watching (also enable polling) |
+| --watch-stdin, --stdin    | Exit the process when stdin is closed                   |
 
 ### Optimize Options
 
@@ -173,10 +173,10 @@ These options allow to manipulate optimisations for a production build using web
 
 | Parameter                 | Explanation                                            | Plugin used                          |
 |---------------------------|--------------------------------------------------------|--------------------------------------|
+| --optimize-dedupe         | Optimize duplicate module sources in the bundle        | DedupePlugin                         |
 | --optimize-max-chunks     | Try to keep the chunk count below a limit              | LimitChunkCountPlugin                |
 | --optimize-min-chunk-size | Try to keep the chunk size above a limit               | MinChunkSizePlugin                   |
 | --optimize-minimize       | Minimize javascript and switches loaders to minimizing | UglifyJsPlugin & LoaderOptionsPlugin |
-| --optimize-dedupe         | Optimize duplicate module sources in the bundle        | DedupePlugin                         |
 
 ### Resolve Options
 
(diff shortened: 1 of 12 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@ npm-debug.log
 build
 .antwar
 
+.idea
```
SpaceK33z: This should be in your global `.gitignore` file.

NejcZdovc: As far as I know everyone is putting this things into the projects. But if this is a problem I can add it to my global `.gitignore `not a problem

SpaceK33z: Nope, that's really bad practice. That would mean an almost infinitely growing list of `.DS_Store`, vim swap files, Sublime project files, etc. All the while the developer can add this to his personal `.gitignore` file so the project is independent :).

NejcZdovc: Nice to know that, because I was always under the impression that this is a good practice, because big projects are done like this ([Angular](https://github.com/angular/angular/blob/master/.gitignore), [React ](https://github.com/facebook/react/blob/master/.gitignore)) Will remove it.

NejcZdovc:
Can we change **api/cli** from tables to the basic description like **/configuration/dev-server**. This way we will not have any problems with tables. If not can I order them inside the table to alphabetic order?

bebraw:
I know tables will receive styling work. If we stick with tables, we can add sorting and such easier later on. I guess it doesn't hurt to order them initially, though.

NejcZdovc:
Page **/configuration/module** is quite confusing at first look. Why is header h4 also a link that can be found in the left menu? 

Why is `rule` highlighted in `Rule conditions`, `Rule results` and `Rule nested rules`?

 Where are other options for module? For now there is only `rules `and `noParse`.

NejcZdovc:
@bebraw Ok will sort them for now.

bebraw:
> Page /configuration/module is quite confusing at first look. Why is header h4 also a link that can be found in the left menu?

The navigation is generated based on h elements. You could open a separate issue related to the page. Good to discuss there what to do about it.

> Why is rule highlighted in Rule conditions, Rule results and Rule nested rules?

Not sure. Likely looks better without.

> Where are other options for module? For now there is only rules and noParse

Looks correct. `preLoaders` and `postLoaders` were factored out.

NejcZdovc:
> Looks correct. preLoaders and postLoaders were factored out.

What about this other ones?

```js
interface Module {
        exprContextCritical?: boolean;
        exprContextRecursive?: boolean;
        exprContextRegExp?: RegExp;
        exprContextRequest?: string;

        // An array of automatically applied loaders.
        loaders?: RuleSetRules;

        // Don't parse files matching. It's matched against the full resolved request.
        noParse?: RegExp[] | RegExp;

        // An array of rules applied for modules.
        rules?: RuleSetRules;
        unknownContextCritical?: boolean;
        unknownContextRecursive?: boolean;
        unknownContextRegExp?: RegExp;
        unknownContextRequest?: string;
        wrappedContextCritical?: boolean;
        wrappedContextRecursive?: boolean;
        wrappedContextRegExp?: RegExp;
    }
```

NejcZdovc:
One that is missing for me is `loaders`. The only thing that I found out that is connected to the `loaders` is  `UseEntry`. Should instead of `UseEntry` be written `module.loaders`?

NejcZdovc:
> Looks correct. preLoaders and postLoaders were factored out.

What about `loaders`?

bebraw:
> What about loaders?

Renamed as `rules` as documented [here](https://webpack.js.org/how-to/upgrade-from-webpack-1/#module-loaders-is-now-module-rules).

About the missing fields, those look like something internal, but it's true they could be worth covering here.

NejcZdovc:
Should we then remove this from [TypeScript definition](https://github.com/NejcZdovc/webpack/blob/04953b62efffdfae3cf0f2228858f1e0f0cd987b/schemas/webpack.d.ts#L32) and as well from [schema.json](https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L117)?

TheLarkInn:
'Loaders' is backwards compatible for now but we would migrate to 'rules'

NejcZdovc:
Ok, leave it for now. Good. I finished reordering of the code. All reviews are welcome.

content/api/cli.md
```diff
@@ -163,8 +163,8 @@ These options makes the build watch for changes in files of the dependency graph
 | Parameter                 | Explanation                                             |
 |---------------------------|---------------------------------------------------------|
 | --watch, -w               | Watch the filesystem for changes                        |
-| --watch-stdin, --stdin    | Exit the process when stdin is closed                   |
 | --watch-aggregate-timeout | Timeout for gathering changes while watching            |
+| --watch-stdin, --stdin    | Exit the process when stdin is closed                   |
```
bebraw: Shouldn't this go after `--watch-poll`?

content/api/cli.md
```diff
@@ -173,10 +173,10 @@ These options allow to manipulate optimisations for a production build using web
 
 | Parameter                 | Explanation                                            | Plugin used                          |
 |---------------------------|--------------------------------------------------------|--------------------------------------|
+| --optimize-dedupe         | Optimize duplicate module sources in the bundle        | DedupePlugin                         |
 | --optimize-max-chunks     | Try to keep the chunk count below a limit              | LimitChunkCountPlugin                |
-| --optimize-min-chunk-size | Try to keep the chunk size above a limit               | MinChunkSizePlugin                   |
 | --optimize-minimize       | Minimize javascript and switches loaders to minimizing | UglifyJsPlugin & LoaderOptionsPlugin |
-| --optimize-dedupe         | Optimize duplicate module sources in the bundle        | DedupePlugin                         |
+| --optimize-min-chunk-size | Try to keep the chunk size above a limit               | MinChunkSizePlugin                   |
```
bebraw: Before `optimize-minimize` I think.

content/api/cli.md
```diff
@@ -195,38 +195,38 @@ These options allow webpack to display various stats and style them differently
 | Parameter               | Explanation                                                        | Type    |
 |-------------------------|--------------------------------------------------------------------|---------|
 | --color, --colors       | Enables/Disables colors on the console [default: (supports-color)] | boolean |
```
bebraw: This flag doesn't exist anymore if I remember right (handled automatically). Check this out from the source so we get it right.

content/api/cli.md
NejcZdovc: You are correct.

content/api/cli.md
NejcZdovc: I wasn't sure what should have priority here. Will change.

content/api/cli.md
NejcZdovc: I think it's still in [use](https://github.com/webpack/webpack/blob/f18989c712894bba13ebfb6dfb3f50fc44b0d12a/bin/webpack.js#L279).

bebraw: Alright.

bebraw:
Any changes in mind? Looks good to me now.

NejcZdovc:
This is it for now. It would be good if this "rule" would be written in contribution md, so PR like this will not be necessary any more.

bebraw:
> This is it for now. It would be good if this "rule" would be written in contribution md, so PR like this will not be necessary any more.

Which one? Sorry, lost context. :)

NejcZdovc:
Rule about alphabetising options, so that are rules and lists should be order alphabetically.

bebraw:
Ah, yeah. I'll add. Thanks.

Ideally we would lint for this. There's some infrastructure in place.

NejcZdovc:
@bebraw Also add something like this for properties (it's missing in this [commit](https://github.com/webpack/webpack.js.org/commit/29058fd067135378438f6cee683e04cd5982081a)):

```
### Properties
devServer.compress
devServer.contentBase
devServer.hot

Same goes for properties.
```

Other than that it's perfect.

Issue #259, opened 2016-10-17
Backers design


montogeek:
Hi!

Currently splash page shows backers list, like this:
![screen shot 2016-10-17 at 16 50 25](https://cloud.githubusercontent.com/assets/1002461/19456998/e87c5bb6-9489-11e6-8ea3-39c745a1f20a.png)

I checked Splash design https://github.com/skipjack/webpack-design/blob/master/Splash.psd, but not found information about how should it look.

I want to help fixing it!

skipjack:
@montogeek yea it definitely needs fixing. The design is pretty outdated at this point so I wouldn't refer to it anymore and the sponsor's section was added recently.

Did you have an idea about how to fix it? I was thinking about trying to wrap them but since it's an svg it's not easy fix. Let me know your thoughts, and check out #260 for the latest splash page design.

NejcZdovc:
@skipjack I think this can be closed.

PR #260, opened 2016-10-17
Splash Page

skipjack:
Ok so this is still very much a work in progress but I think it's ready for some feedback.

- [x] Update the layout of the splash page
- [x] Show example (`index.md` currently, interactive one is still in discussion)
- [x] Update sponsors section
- [x] Update footer styling
- [x] Work on mobile viewport

For the visualization:

- [x] Display output files
- [x] Start rotating the cube (needs some tweaking)
- [x] Start work on modules/output visualization

Here's a quick demo of what it looks like now, I'll add more soon:


NejcZdovc:
Created #261 so that we can easily control all html elements and we don't have a bit svg file as discussed in #259.

skipjack:
@NejcZdovc #261 has been pulled into here.

@bebraw seeing some odd [build errors](https://travis-ci.org/webpack/webpack.js.org/builds/168461877) that I can't quite make sense of starting on line 1277. Looks like it has to do with the interactive package.

bebraw:
@skipjack There's a file naming mismatch. `splash-viz.jsx` doesn't exist.

skipjack:
Doh, fixed it (js => jsx)

NejcZdovc:
One thing that was quite off for me is the bundle size for JS. Most of the webpack users use it for web sites / apps and loading 19 MB in one chunk is something that no one wants. Maybe changing this to some smaller value would be batter, so that users will not be leaded into thinking that using webpack will lead into big bundle.

skipjack:
Haha yea you're definitely right, 19 megs is way too high. I'll make sure they all have realistic values before merging.

skipjack:
Ok cleaned this up a bit more. Once I add the links between modules in the visualization I think we should merge. The top section will take a bit more work but I can open another PR for it. It's starting to look pretty good on most viewports now. Here's a few demos:

Desktop:



Mobile:



Feedback welcome! Just keep in mind this is only round one.

SpaceK33z:
@skipjack: first impression: WOW! It looks really nice :). Some feedback:

- When hovering on one of the "dependencies" on the left of the page, maybe display lines between them so it is clear that the files dependent on each other and what their relation is? Just an idea though, feel free to ignore.
- If I leave the page open, go to another tab and then come back after about 30 seconds, the spinning cube spins *very* quickly for a while.
- The "donate" button for Open Collective is gone. The original button for OC was a bit out-of-place style-wise, but maybe we can build a custom "donate" button?

skipjack:
@SpaceK33z thanks for the feedback!

- Working on adding the links now. I think showing the connections a module has on hover is a great idea, I'll try to work that in.
- Yea the cube is still using a CSS animation, but I'm planning on moving this to more of a javascript implementation which should fix that issue.
- Definitely doable, @NejcZdovc was suggesting not hiding the one from the navigation on mobile screens, meaning there'd always be one displayed on the top right. We could always do another more prominent one in the sponsors/support section as well.

skipjack:
Ok added links between the modules and cleaned up a bit. The visualization (top) section is by no means done, I'll be opening another PR for that soon. As soon as CI passes, I'm going to merge... if anyone has more feedback or notices any problems on this page feel free to open an issue.

montogeek:
Was this deployed? I am still see the old page

bebraw:
Not yet.

PR #266, opened 2016-10-20
Changing the devServer contentBase value

fdidron:
The `contentBase` value should be relative instead of absolute in the example.


SpaceK33z:
Thanks!

Issue #256, opened 2016-10-16
Menu rendering


NejcZdovc:
Created based on #255.

> Page /configuration/module is quite confusing at first look. Why is header h4 also a link that can be found in the left menu? This happens on other pages as well.

NejcZdovc:
Some other improvements:

- When you click on menu with submenu, submenu should open
- When you switch to other menu (for example from DevServer to DevTool), previous submenu should close

NejcZdovc:
@skipjack Will #254 address this things or should I go for it and also can I add sticky menu as discussed in #246?

skipjack:
#254 is more for the mobile sidebar. Once that one's merged I'd be happy for you to take a stab at finishing off the normal Sidebar component.

I'm about to push some commits to that now so I think we'll have it merged either today or tomorrow.

skipjack:
Closing with merge of #277.

PR #277, opened 2016-10-25
Desktop Sidebar

NejcZdovc:
This PR is based on #256 . The only thing that I found out that is was not yet implemented is

> When you click on menu with sub-menu, sub-menu should open

Some more things that I found:

- [x] Highlight active link
- [x] Hide empty sub-menu (empty `ul `creates space between items)


skipjack:
Looks great thanks!

Issue #184, opened 2016-09-22
Set up search for the site


bebraw:
Algolia would work.

I'm opening this based on Twitter discussion.

dzello:
Great! Per the comment on Twitter, what thoughts did you have in mind for the design?

bebraw:
@dzello Mostly fine. The structure of the site is completely different now, though, so that might affect your side. The search control needs a good location. Maybe header? See http://webpack.js.org/ .

dzello:
Ok, good to know. If the new content structure is different that's ok, I will have to make some config changes but should be mostly painless. From clicking around on the new site it looks like some content isn't there yet. Should I hold off on trying to index it until there's more there?

The header is good for location, that's what we did on the React docs.

bebraw:
You can index it, but it's correct it won't yield a lot of results yet. But then, I assume you index it regularly so going with the new structure can't hurt.

dzello:
Right, once the config is live we re-index every 24 hours to pick up any changes. I will play with indexing the new site and see where I get. Any ideas for timeline on having all the new content up?

bebraw:
All content - months.

Minimum will take less. There's a [target](https://github.com/webpack/webpack.js.org/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Webpack+2+-+Documentation+MVP%22) for that.

dzello:
Sounds good, no problem to get new indexing configuration done by then.

In the meantime, for the sake of asking, do you think it's worth deploying search for the existing docs since the PR is ready to go? I don't want to make extra work but one upside of deploying for the existing docs is that we'll have more data to improve search for V2.

bebraw:
@dzello Yeah, getting the search on the old site could be worthwhile. Just poked about that so we get a decision made.

sokra:
google custom search can work too...

dzello:
The Google custom search UX isn't great for docs. No autocomplete, irrelevant suggested searches, clumsy modal dialog for displaying results. Also there's ads and you have to pay to remove them.

jhnns:
If a documentation site uses google custom search, I'm instantly annoyed ^^. I don't like the results.

SpaceK33z:
@dzello, are you still interested in adding search for this repo? We'd love it if you could do a PR :).

dzello:
@SpaceK33z yes, I will start getting that together :)

bebraw:
[...]

PR #282, opened 2016-10-28
Search Interface

skipjack:
cc @bebraw @dzello

Still a WIP, adding the interface bits for search:
- [x] Add a togglable search bar in the Navigation
- [x] Handle focus for usability
- [x] Trigger a search event on keying for other components to hook into
- [x] Add a SearchResults component to fetch and display results
- [x] Polyfill [CustomEvent](https://github.com/webmodules/custom-event) for crossbrowser-support
- [x] Handle other events for closing and hiding the search bar and results

Still need to add another search bar in SidebarMobile for searching on screens under 768px, but we should probably take it one step at a time and first merge this so someone else can start playing around with actually displaying search results (see #184).

I'll add some gifs tonight...

### DIFF
--- a/components/container/container.jsx
+++ b/components/container/container.jsx
@@ -1,11 +1,11 @@
 import React from 'react';
 import './container-style';
 
-export default props => {
+export default (props = {}) => {
   let { className = '' } = props;
 
   return (
-    <div className={ `container ${className}`}>
+    <div className={ `container ${className}` }>
       { props.children }
     </div>
   );
--- a/components/logo/logo-style.scss
+++ b/components/logo/logo-style.scss
@@ -16,6 +16,7 @@
     font-family:'Geomanist', 'Century Gothic', sans-serif;
     font-size:1.5em;
     font-weight:500;
+    letter-spacing:0.5px;
     text-rendering: geometricPrecision;
     margin-left:0.25em;
     color:map-get($colors, emperor);
--- a/components/navigation/navigation-style.scss
+++ b/components/navigation/navigation-style.scss
@@ -31,14 +31,10 @@
 .navigation__inner {
   display: flex;
   align-items: center;
-  padding: 0.6em 1em;
+  padding: 0.6em 0 0.6em 1em;
 
   @include break {
     padding:0 1.5em;
-  }
-
-  @include break('large') {
-    padding:0 0 0 1.5em;
   }
 }
 
@@ -73,10 +69,6 @@
     align-items: center;
     justify-content: flex-end;
   }
-
-  @include break('large') {
-    padding-right:0;
-  }
 }
 
 .navigation__link {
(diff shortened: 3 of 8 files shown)


skipjack:
Here's a preview that demos toggling, typing in, and tabbing into the search bar:

SpaceK33z:
@skipjack, do you think it would be useful if I test this out on some devices / browsers, or is this too early?

skipjack:
@SpaceK33z you'd be more than welcome to test it out, however I haven't added a mobile version yet so the only devices that you might test on would be tablets (anything >= 768px width).

Initially I was planning to use the the same search bar in the nav for mobile devices as well but after trying it out and seeing the lack of screen real estate, I'm thinking it would be better to add a separate search field at the top of SidebarMobile. 

What I'm hoping is that whoever has experience with Algolia (or whichever lib we choose) and has the indexed data will jump in here at SearchResults._handleSearch...

```js
  _handleSearch(e) {
    let { text = '' } = e.detail;

    if (text.length) {
      if (!this.state.visible) this.setState({ visible: true });
      // TODO: Get search results

    } else this.setState({ visible: false });
  }
```

to start fetching and displaying results.

SpaceK33z:
Okay, I tested it:

- After a click on the search icon, it would be nice if the text input would be focused immediately. I'd expect to be able to type immediately after clicking.
- The search icon is a `<span>` now, but for accessibility reasons you should use `<button type="button">` here. This also fixes the small issue that if you click twice quickly, the browser "selects" the icon as a text.

skipjack:
@SpaceK33z thanks for reviewing, I'll fix both those things. I'll let you know when this is ready for another pass.

skipjack:
@SpaceK33z done, let me know if there's anything else.

dzello:
Lookin' good! I've got the indexing built and the search results connected, here's a preview:

search results

I'm attaching our DocSearch event in the `componentDidMount` of `navigation.jsx` which seems to be working well for now. I didn't need to hook into the search results component because the DocSearch library already knows how to handle keystrokes and a results modal. But I left all the plumbing in so that we can chat about it and make sure the integration does exactly what we want.

My instinct is to put that in a different PR since we can deal with it separately, but let me know if you'd rather I add the commits here.

skipjack:
@dzello awesome, the screenshot looks great! I had a few reasons for separate SearchResults component:

1. Modularizing any custom styling or layout so as to not complicate the Navigation component.
2. It allows us to re-use it in multiple places, for example in the SidebarMobile component when searching on smaller screens (for which I haven't added anything yet).

With that said, I'm not too familiar with Algolia so I think we should just get something up and running (as it looks like you already have) and then continue to play around with it. Based off what you said, it seems Algolia comes with some built-in styling of results? Or is that layout custom?

> My instinct is to put that in a different PR since we can deal with it separately, but let me know if you'd rather I add the commits here.

Agree completely, I was just giving this another day or two to simmer in case anyone else wanted to test/review. Merging now... feel free to continue the conversation here or on your PR 👍

dzello:
The styling is the default for the [DocSearch project](https://community.algolia.com/) but we'll be able to [customize it](https://github.com/algolia/docsearch#customization) by adding CSS selectors and get a really nice, consistent look. PR coming soon.

sokra:
👍 The search is pretty nice.

PR #283, opened 2016-10-31
link from get started to concepts

bransbury:
**Context/problem:**
I like the content on the 'get started' page, it's a nice quick overview. However, there's a few options in the conclusion for what to do/read next. This could create some 'choice paralysis' for the user. It also means the learning journey doesn't have a linear flow which may mean users don't know what they *should* read next. 

**Solution:**
Linking to the next logical section (concepts) with a main call-to-action link at the bottom should solve this. I've added the link as part of this PR. If you think this is a good idea we can change wording/styling to suit.


bebraw:
It's good. Thanks. 👍

PR #284, opened 2016-11-01
Connect search interface to search results

dzello:
This PR wires up the new search box with the [DocSearch](https://github.com/algolia/docsearch) JS and corresponding CSS. DocSearch takes care of handling keystrokes, submitting the queries to the webpack.js.org Algolia index, and displaying the results in an overlay.

For now, I commented out a small piece of code in `search-results.jsx` responsible for showing the overlay, as DocSearch has its own. DocSearch will close the overlay if the user clicks on a result, clicks outside the modal, or hits the escape key.

As you're testing locally, you might see that clicking a search result takes you to `https://webpack.js.org` and not to `localhost:3000`. That is because the index is built from `https://webpack.js.org`. Once we get search & results functionality deployed, the URL in the search result will match the URL in the browser and we can test the on-page transitions. If we want to test transitions  before then I can create an index with records pointing to `localhost:3000`, but that seemed like extra work for now and I think the transitions should just work. The index is rebuilt from the latest site every 24 hours, so any changes will be picked up automatically and we can verify soon after a new site build is out.

For now, can someone with a keen visual eye have look at the styling and suggest any improvements? This is the list of [CSS customizations](https://github.com/algolia/docsearch#customization) we can make. Thanks!


SpaceK33z:
I'll take a critical look at the functionality later today with a designer-friend.

SpaceK33z:
I tried to test this with `npm run build && npm run build-test`, but it complains about this:

```
Uncaught TypeError: window.docsearch is not a function
```

bebraw:
Yeah. `window` doesn't exist in Node environment so you probably want `window && window.docsearch({` or so. Or we could monkeypatch `window` through webpack. That would work too.

dzello:
The last commit fixes the `window.docsearch` issue - we just needed to get DocSearch on the page earlier. It was working in dev server mode but not with the generated pages because the React stack initializes earlier.

dzello:
There is still an issue running against generated pages, React and DocSearch aren't playing nicely together because of re-renders and underlying DOM mutations. I'm surprised it works with the dev server, actually. Looking into it further and will report back.

bebraw:
Couldn't that `window.docsearch` portion live in `template.ejs`? That would avoid the problem. I've done an integration like this myself.

dzello:
There's no guarantee that the `.navigation__search-input` selector is on the page yet, so when I put it there it can't find that element to attach to.

SpaceK33z:
@dzello, I added some styling to the search results in your PR. Hope you don't mind that I directly pushed it to your PR (originally I wanted to do a PR on your PR, but then I accidentally pushed the code).

The functionality itself works fine in develop mode, so after it's fixed for production it can be merged I think.

dzello:
No worries @SpaceK33z, and looks good. You might have a look at the index page too, the background is lighter and it's hard to make out the cursor and search query.

skipjack:
Yea sorry, completely forgot to tweak the search bar on the landing page. That should be easy to do in .navigation--transparent declaration.

SpaceK33z:
@skipjack thanks for the hint, I fixed the search results for the landing page. @bebraw is quite busy, would you be able to help @dzello with the `window.docsearch` issue? I don't have much experience about server side rendering unfortunately.

skipjack:
@SpaceK33z yep, I'll take a look at this tomorrow and help finalize.

skipjack:
Ok finally had a chance to play around with this and dig into algolia and docsearch a bit more. Unfortunately the problem was not a simple issue with your `componentDidMount` and the creation of that docsearch instance, which I got running fine. The problem is after that code runs, the Navigation component is fully re-rendered by preact, or maybe something in antwar-interactive, which means all the generated html for the autocomplete is completely blown away.

I'll discuss with @bebraw and try to get at the very least a temp fix in place for the time being so we can get this merged.

skipjack:
@dzello would it be possible for us to interface directly with algolia? i.e.

```js
this.client = AlgoliaSearch('bh4d9od16a', '93950c6eda05068a6f0649e4a7f7546e');
this.index = this.client.initIndex('webpack');
this.index.search('resolve', (err, content) => {
    console.log(err, content);
});
```

I tried using the id, key, and index name you mentioned in #184 but kept hitting 403's with a message of  *key or id is incorrect*. It seems this is intentional and the docsearch functionality is a little more limited than the paid full version? Just figured I'd check before I go forward so I know not to look at the full algolia documentation and just stick to docsearch.

skipjack:
Got it working, had to navigate around preact/react state for now. Any thoughts on how I should push these changes to this PR? There's a significant amount, as I cleaned up some other stuff relating to this.

dzello:
@skipjack Not sure why you'd be getting the key or ID incorrect issue. DocSearch is full functionality, no limits or restrictions, and what is in the general Algolia docs should be relevant.

dzello:
@skipjack I have no issue if you want to push them up here. Thanks for getting that to work, I'm interested to see the route you took :)

skipjack:
@dzello Np, it's a bit hacky going outside react instead of using state but I'll probably revert some of this later once I have time to add a cleaner implementation and add a search bar for mobile screens. Just pushed a branch called *pr-284-patch*, if you could pull it and merge it into yours that would be great (you shouldn't hit conflicts as I started from your forked branch). After we get this merged, would you mind if I PM you on gitter with a few more algolia/docsearch questions?

@SpaceK33z once he updates this branch that maybe you can give this one last sanity check before we merge?

SpaceK33z:
@skipjack Sure. Note that you could also push it to his branch (since a few weeks GitHub gives maintainers write access to PR branch, even if it's a fork).

skipjack:
@SpaceK33z ah ok yea I wasn't aware of that, unfortunately I'm away from my personal comp right now but I can do that tonight if it's not in by then. 

@SpaceK33z @dzello BTW, on that patch branch I also updated the `npm run build-test` command to run a `build` and and throw up the test server so as to make things a bit easier on all of us 😄 , minor change but when you have to run it every thirty seconds it adds up lol.

dzello:
Pulled in the commits from @skipjack. Everything working as far as I can call.

@skipjack nice call to add `build` to `build-test` task! I joined the gitter room, feel free to DM me any DocSearch questions.

SpaceK33z:
I pushed a small fix because our custom search results styling was not working on the production build. This was because the docsearch css was loaded after our css. Ideally we would move the docsearch css and javascript to the bottom of `<body>` to improve first paint time.

I'm going to merge this. We can iterate over it in separate PR's.

@dzello, thank you very much!!

PR #286, opened 2016-11-01
Option to define loaders without -loader suffix

grgur:
Showing how webpack can resolve loaders without the -loader suffix

### DIFF
--- a/content/configuration/index.md
+++ b/content/configuration/index.md
@@ -137,7 +137,9 @@ T> Notice that throughout the configuration we use Node's built-in [path module]
         [enforce](/configuration/module#rule-enforce): "post",
         // apply these rule even if rules are overridden (advanced option)
 
-        [loader](/configuration/module#rule-loader): "babel-loader",
+        <details><summary>[loader](/configuration/module#rule-loader): "babel-loader",</summary>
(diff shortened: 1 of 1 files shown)


bebraw:
If I remember this feature was explicitly dropped from webpack 2. @jhnns can confirm that and you can check out the main repo too. The upgrade guide might have something too.

You can still achieve the shortcut if you reconfigure `resolveLoader.moduleTemplates`. It's just that it doesn't include `-loader` there by default.

grgur:
I must have missed that, thanks for the heads up. I kind of liked that, but I see how it made more sense in Webpack 1.x where everyone writes spaghetti loader configs

SpaceK33z:
Relevant PR: https://github.com/webpack/webpack/pull/3102

bebraw:
@grgur That said, it would be a good idea to cross reference from the configuration page to the upgrade guide so people pick up the change. Or add a tip `T>` there noting the change.

grgur:
@SpaceK33z @bebraw thanks for pointing this out. I just noticed it's already mentioned in the upgrade guide so i guess no need for living in the past. I'll close this for now

PR #292, opened 2016-11-02
added voting app

sokra:
This PR adds an page to vote for todo items of the core team. There are two currencies:

* Influence: granted to everyone depending on how long you are github member
* Golden Influence: granted by donations (importing of the open collective data WIP)

It's in a prototype stage. It would be great if someone takes care of the design (I can't do that).

Here is a screenshot:

image

You authenticate with the Github API. This means login only works on webpack.js.org host. A developer emulated version is used if not deployed. See code comment in api.js for more info.

Other ideas:

* Auction stuff
* Distribute a part of the money to ecosystem authors
* Add your own ideas

### DIFF
--- a/antwar.config.js
+++ b/antwar.config.js
@@ -96,6 +96,14 @@ module.exports = {
           /^\.\/.*\.md$/
         );
       }
+    ),
+    vote: voteList(
+      'Voting',
+      {
+        index: true,
+        feedback: true,
+        moneyDistribution: true,
+      }
     )
   }
 };
@@ -155,5 +163,47 @@ function processPage() {
     contributors: function(o) {
       return Array.isArray(o.file.contributors) && o.file.contributors.length && o.file.contributors.slice().sort();
     }
+  };
+}
+
+function voteList(title, lists) {
+  return {
+    title: title,
+    path: function() {
+      function context(request) {
+        var name = /^\.\/(.*)\.md$/.exec(request)[1];
+        return {
+          name: name,
+          __content: '' // make antwar happy
+        };
+      }
+      context.keys = function() {
+        return Object.keys(lists).map(k => './' + k + '.md');
+      };
+      return context;
+    },
+    processPage: {
+      url: function(o) {
+        return 'vote/' + o.file.name;
+      },
+      name: function(o) {
+        return o.file.name;
+      },
+      anchors: function(o) {
+        return [];
+      },
+      content: function(o) {
+        return '';
+      }
+    },
+    layouts: {
+      index: function() {
+        return require('./components/vote/list.jsx').default
+      },
+      page: function() {
+        return require('./components/vote/list.jsx').default
+      }
+    },
+    redirects: {} // <from>: <to>
   };
 }
--- a/components/vote/api.dev.js
+++ b/components/vote/api.dev.js
@@ -13,18 +13,24 @@ let lists = {
         name: "influence",
         currency: "influence",
         maximum: 30,
-        score: 1
+        score: 1,
+        step: 2,
+        color: "blue"
       },
       {
         name: "golden",
         currency: "goldenInfluence",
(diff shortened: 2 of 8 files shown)


skipjack:
> It would be great if someone takes care of the design (I can't do that).

@oliverturner or @NejcZdovc, either of you interested in helping to implement a nice design here?

NejcZdovc:
Sure not a problem. Will you prepare some kind of design or should I create one by my self?

skipjack:
Feel free to lay it out how you'd like, or maybe @sokra has some ideas. Might be some good examples of voting app designs on dribbble to use as a starting point...

sokra:
I have no specific idea.

Feel free to add more example data in the development app.

We could reduce the number of votes users get. This would reduce the number of buttons we need.

You can do a PR to my branch.

NejcZdovc:
Guys sorry for a delay, but I will tackle this issue on the weekend.

NejcZdovc:
What do you think about something like this?

ezgif-3309319921

Initial value would be 0. You could go in minus 100 up to plus 100. As you can see in gif you have a snap functionality, which means that if step in circle is set to 10, and you can only set 10*x value. All values can be reconfigured, which means that you can have different settings for different sliders.

sokra:
I think this could work. But it need to be a bit smaller. There can be up to ~100 items on the page...

NejcZdovc:
Of course, this is just a demo. In reality we can do it a lot smaller.

sokra:
It would be great to have a very basic styling soon and delay the bigger styling work for later.
I'd like to publish this soon to test this with real users. (I'm not sure how well this works out)

NejcZdovc:
Sure, I will start working on this first thing tomorrow.

NejcZdovc:
@sokra Can user set only value between 0 and x, or is it possible to set negative number as well, for example between -100 and 100?

NejcZdovc:
@sokra also can you provide me limits (max value and min value) for all three sections.

sokra:
Thanks.

The limits depends on the kind of list and kind of vote. The limits are provided by the API at runtime.

This is the current configuration:

| List | Description |
|---|---|
| Todo | The todo list for the core team. We try to do most voted items first. |
| Feedback | Give feedback to various topics. You can like or dislike them with a score from -3 to +3. |
| Money Distribution | For a part of donated money you can decide who deserve it. Note that you can't undo your distribution once it's paid to the party. |


| List | Currency | Minimum | Maximum |
|---|---|---|---|
| ToDo | Influence | 0 | 30 |
| ToDo | Golden Influence | 0 | no |
| Feedback | none | -3 | 3 |
| Money Distribution | Thanks | 0 | no |

| Currency | From |
|---|---|
| Influence | You can more over time (probably 1 per day) |
| Golden Influence | You get a bunch of them for donations |
| Thanks | You get points equal to 50% of the donated money |

But I'm thinking about lowering the values to make it more intuitive.

Note: The money distribution stuff will probably be renamed/rephrased, because it's difficult to give money to people without asking them. So they probably get points which can be used to get money or golden influence.

NejcZdovc:
Thank you for the extensive explanation. How should we handle this infinity maximum. If you have 300 gold influence, it's quite hard to be very accurate with a slider.

NejcZdovc:
Created a PR #328, based on this PR.

sokra:
> How should we handle this infinity maximum. If you have 300 gold influence, it's quite hard to be very accurate with a slider.

The more you have, the less you care about the accuracy... It's like real money... :smile: 

At least we can try to assume this and check if people complain.

NejcZdovc:
For now I set it to 100, but can be easily changed to remaining currency.

sokra:
ok the prototype is live, if you want to play with it:

https://webpack.js.org/vote/

PR #293, opened 2016-11-02
Get started section

pksjce:
- [X] Make get started into a section
  - [X] Make separate mds for
       - [X] Installation
       - [X] Why webpack
       - [X] Hello webpack


pksjce:
@bebraw - Build was failing due to hyperlink errors. Fixed those. Please have a look when you can.

bebraw:
I would merge "why webpack" article with the current comparison as there's overlap and having that table there is handy. Good otherwise

pksjce:
What do you think of pulling in /concepts/modules also into the get started section?

bebraw:
It might be good to retain a footer link still. I would leave the existing concepts section as it is right now. Some overlap doesn't hurt (no need to be DRY 👍).

pksjce:
Cool! Added back the new footer link.

content/get-started/why-webpack.md
```diff
@@ -1,6 +1,31 @@
...
+  - pksjce
+  - bebraw
+sort: 2
 ---
+
+You might be familiar with tools like Make, Grunt, Gulp, Browserify, or Brunch. Make is the oldest and the most known one as it's a general build tool and its history goes back to the 70s. More specialized *task runners*, like Grunt and Gulp, are focused on JavaScript and have healthy plugin ecosystems around them.
+
+## What Are Task Runners?
+
+Task runners literally make it easier to handle tasks, such as linting, building, or developing your project. Compared to *bundlers* like Browserify, Brunch, or webpack, they have a higher level focus. As you might guess from the term bundler, the goal of these tools is to achieve something lower level.
```
dtinth: In this paragraph I have some trouble understanding what ‘they,’ ‘these tools’ refer to.

Would rewording to something like this make the meaning more clear?

> Task runners literally make it easier to handle tasks, such as linting, building, or developing your project. Compared to *bundlers* like Browserify, Brunch, or webpack, task runners have a higher level focus. Bundlers, on the other hand, has a much more specific purpose.

content/get-started/why-webpack.md
```diff
@@ -1,6 +1,31 @@
 ---
-title: Compared to Other Systems
+title: Why webpack?
+contributors:
+  - pksjce
+  - bebraw
+sort: 2
 ---
+
+You might be familiar with tools like Make, Grunt, Gulp, Browserify, or Brunch. Make is the oldest and the most known one as it's a general build tool and its history goes back to the 70s. More specialized *task runners*, like Grunt and Gulp, are focused on JavaScript and have healthy plugin ecosystems around them.
```
dtinth: When I read this page for the first time, I don’t know why “task runners” and “bundlers” have to be introduced here. My guess is that they are usually compared (webpack vs Gulp).

Would it make sense state that we’re going to draw a distinction? e.g.

> webpack is usually compared to tools like Make, Grunt, Gulp, Browserify or Brunch. However, some of these tools (Make, Grunt, and Gulp, which are *task runners*) have a much different purpose than webpack, which is a *module bundler*. Let’s draw a distinction between these categories of tools.

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+<html>
+  <head>
+      <title>Webpack demo</title>
+      <script src='https://unpkg.com/lodash@4.16.6' type='text/javascript'/>
+      <script src='index.js' type='text/javascript'/> 
+  </head>
+  <body>
+  </body>
+</html>
+```
```
sokra: This part was a bit confusing for me. You used `import` in the JS frament but added lodash via script tag in the HTML fragment.

Maybe it's a bit clearer if you first show JS without import + HTML with script tag. And than JS with import (maybe as diff) and HTML without script tag. This way you have two complete examples, first without webpack, second with webpack.

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+Now run `webpack` on this folder with the entry file to be `index.js` and to output a `bundle.js` file which bundles all the code required for the page.
+
+```bash
+webpack --entry index.js --output bundle.js
```
sokra: maybe we should also put files into folder to encorage good organization.

`webpack app/main.js dist/bundle.js`

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+Hash: a3c861a7d42fc8944524
+Version: webpack 2.2.0
+Time: 90ms
+   Asset     Size  Chunks             Chunk Names
+index.js  1.56 kB       0  [emitted]  main
+   [0] ./app/index.js 170 bytes {0} [built]
```
sokra: aha, here it's already in a folder ;)

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+By including a single `bundle.js` gets rid of these complexities.
+
+## Using webpack with a config
+
+For more complex configuration, we can use a configuration file that webpack can reference to bundle your code.
+The above CLI command would be represented in config as follows -
+
+__webpack.config.js__
+```javascript
+module.exports = function(env){
```
sokra: use a simple object in the getting started guide to keep it minimal.

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+      filename: 'bundle.js'
+    }
+  }
+}
+```
+
+This file can be run by webpack as
+
+```bash
+webpack --config webpack.config.js
```
sokra: add note about that this is the default. You can just run `webpack` if a `webpack.config.js` file is present.

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+document.body.appendChild(component());
+```
+
+To run this piece of code, one usually has the below html
+
+```html
+<html>
+  <head>
+      <title>Webpack demo</title>
+      <script src='https://unpkg.com/lodash@4.16.6' type='text/javascript'/>
```
dtinth: I think [the script tag cannot be self-closed](http://stackoverflow.com/questions/4531772/can-the-script-tag-not-be-self-closed).

content/get-started/index.md
dtinth: You can also explain a bit more about dependencies management, such as:

> In this example, there are _implicit_ dependencies between the script tags.
>
> `index.js` depends on `lodash` being included in the page before it runs. It is implicit because `index.js` never said it needs to use `lodash`; it just assumes that a global variable `_` exists.
>
> There are problems with managing JavaScript projects this way: If a dependency is missing, or is included in the wrong order, the application won’t function. If a dependency is included but is not used, then there is an unnecessary code to download.

And in the example with webpack:

> Here, `index.js` explicitly requires `lodash` to be present, and binds it as `_` (no global scope pollution).
>
> By stating what dependencies a module needs, webpack can use this information to build a _dependency graph_, and generate an optimized bundle where scripts will be executed in the correct order, and unused dependencies will not be included in the bundle.

If you’d like to include a more extreme example, I have one from my old project, with [15 script tags](https://github.com/dtinth/IO-MIDI-INSTRUMENT/blob/ec819d4b60f3a1efb627e0da642981cdbf6fb14d/instruments/keyboard/index.html#L18-L33) that implicitly depended on each other 😛.

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+  "scripts": {
+    "build": "webpack --"
+  },
+  ...
+}
+```
+
+You can now achieve the same as above by using `npm run build` command. npm picks up the scripts through it and patches the environment temporarily so that it contains the bin commands. You will see this convention a lot of projects out there.
+
+T> I use `webpack --` so that we can pass custom parameters to webpack through npm.
```
dtinth: This doesn’t work?

content/get-started/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Creating a bundle
+
+Create a demo directory to try out webpack. [Install webpack](/get-started/install-webpack).
+
+```bash
+mkdir webpack-demo && cd webpack-demo
+npm init -y
+npm install --save-dev webpack
+webpack --help //Shows a list of valid cli commands
```
dtinth: As installing webpack globally is “not a recommended option,” should we use a local version?

```
./node_modules/.bin/webpack --help
```

content/get-started/index.md
pksjce: You're right! My bad.

skipjack: Would be nice if it could though 😄

content/get-started/index.md
sokra: why is this not recommended. Here webpack is installed locally too (line above), so it shouldn't cause any issue, but look more pretty.

dtinth: The [Install webpack](https://github.com/kalcifer/webpack.js.org/blob/34fda94a3b336ad5a3360b4b2413366a175a319b/content/get-started/install-webpack.md) link on line 15 gives user choices to choose between installing globally and locally.

The global section says:

> However, this is not a recommended practice. This locks you down to a specific version of webpack and might fail in projects that use a different version. The next section tells you how to install webpack locally in a project.

However, I agree that installing globally looks nicer and is easier for beginners and ad-hoc projects. And also given that webpack CLI will [use the local version of webpack if present](https://github.com/webpack/webpack/blob/9085e1e1882d3cae7cfc635258f81ca5d2261c0a/bin/webpack.js#L8), should the Installation docs (which moved to another section in this PR) be updated to say it’s OK to install webpack both globally and locally?

content/get-started/why-webpack.md
pksjce: Your changed text looks much better. I'll add it.

content/get-started/index.md
pksjce: Great! I made some changes as the both of you have suggested :)

content/get-started/index.md
pksjce: Done.

pksjce:
@sokra @dtinth - Would be great if you guys had a look at the changes.

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
+---
+title: Get Started with Webpack
+contributors:
+  - bebraw
+sort: 3
+---
+
+## Getting Started
+
+webpack is a tool to build javascript modules in your application. To start using `webpack` from its [cli](/api/cli) or [api](/api/node), follow the [Installation instructions](/get-started/install-webpack).
```
dtinth: Nit: javascript &rarr; JavaScript?

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+webpack simplifies your workflow by quickly constructing a dependency graph of your application and bundling them in the right order. webpack can be configured to customise optimisations to your code, to split vendor/css/js code for production, have a hot reloaded dev server and many such cool features. Learn more about [why you should use webpack](/get-started/why-webpack).
```
dtinth: Nit: Any standards on whether we should use British or American English?

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Creating a bundle
+
+Create a demo directory to try out webpack. [Install webpack](/get-started/install-webpack).
+
+```bash
+mkdir webpack-demo && cd webpack-demo
+npm init -y
+npm install --save-dev webpack
+webpack --help //Shows a list of valid cli commands
```
dtinth: Nit: Better to use `#` for shell comments?

content/get-started/index.md
dtinth: > have a hot reloaded dev server

How about:

> run a development server hot-reloads your code without page refresh

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+Now create an `index.js` file.
+
+__app/index.js__
+
+```javascript
+function component () {
+  var element = document.createElement('div');
+
+  /* dependency on lodash in the next line */
```
dtinth: Maybe we can just say this?:

> lodash is required for the next line to work

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+To run this piece of code, one usually has the below html
+
+__index.html__
+
+```html
+<html>
+  <head>
+      <title>Webpack demo</title>
```
dtinth: Nit: This should be 4 spaces instead of 6?

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+function component () {
+  var element = document.createElement('div');
+
+  /* dependency on lodash in the next line */
+  element.innerHTML = _.map(['Hello','webpack'], function(item){
+    return item + ' ';
+  });
+
+  return element;
+};
```
dtinth: Nit: There’s an extra `;` here 😛

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+## Using webpack with a config
+
+For more complex configuration, we can use a configuration file that webpack can reference to bundle your code.
+The above CLI command would be represented in config as follows -
+
+__webpack.config.js__
+```javascript
+{
```
dtinth: Missing `module.exports =` here.

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+    entry: './app/index.js'
+    output: {
+      filename: 'bundle.js',
```
dtinth: Nit: There’s some inconsistent indentation here…

content/get-started/index.md
```diff
@@ -0,0 +1,167 @@
...
+Version: webpack 2.2.0
+Time: 90ms
+   Asset     Size  Chunks             Chunk Names
+index.js  1.56 kB       0  [emitted]  main
+   [0] ./app/index.js 170 bytes {0} [built]
+
+```
+T> If a `webpack.config.js` is present, `webpack` command picks it up by default.
+
+The config file allows for all the flexibility in using webpack. We can add loaders, plugins, resolve options and many other enhancements to our bundles using this configuration file.
```
dtinth: Now that it’s called `module.rules`:

add loaders &rarr; add loader rules?

content/get-started/index.md
sokra: :+1: I like it

sokra:
I'll merge it now. More changes may be done now via separate PRs.

sokra:
Thanks, great work

content/get-started/index.md
SEAPUNK: I too am curious about this: Why `webpack --` in the scripts? From what I'm reading in the npm docs, just `webpack` for the `build` scripts and `npm run -- [args for webpack]` should suffice.

bebraw: Ah, good to know! I remember I had to do this exact trick with Karma, but if it works without, it's great to fix. The note will change a little as a result.

SEAPUNK: Note that I haven't actually tested this; I am just going off what npm's docs say.

bebraw: Ok, cool. Please PR when you have some time. 👍

SEAPUNK: #320

PR #299, opened 2016-11-05
Adds how to author libraries

pksjce:


content/how-to/author-libraries.md
```diff
@@ -1,7 +1,172 @@
 ---
 title: How to Author Libraries?
+contributors:
+    - pksjce
 ---
-> library option
-> externals option
 
-> see also [[Output]]
+webpack is a tool which can be used to bundle application code and also to bundle library code. If you are the author of JavaScript library and are looking to streamline your bundle strategy then this document will help you.
```
SpaceK33z: "the author of JavaScript library" -> "the author of a JavaScript library"

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,172 @@
...
+  - Library can be accessed inside nodejs.
+
+### Add webpack
+
+Add basic webpack configuration.
+Also add a [`.babelrc`](https://babeljs.io/docs/usage/babelrc/) file so that you can write webpack config in es2015 syntax.
+
+__webpack.config.babel.js__
+```javascript
+export default () => (
```
SpaceK33z: Why is this a function? An object would also work and is more consistent with the rest.

pksjce: Shouldn't we showcase webpack2 syntax?

pksjce:
@SpaceK33z - Made the requested changes!

content/how-to/author-libraries.md
SpaceK33z: webpack v2 doesn't require using a function instead of an object.

pksjce: Oh! I didn't know that. Is it just an alternate syntax to send `env` variables?

SpaceK33z: Jup! I didn't know that it was possible to use a function, so we both learned something haha.

pksjce: Lol! So good to go?

SpaceK33z: Using objects is a bit more consistent with the other pages, but it doesn't really matter. A few of the examples are invalid though, `export default = () => ...` doesn't work.

pksjce: The example is representative right? I wanted to skip the config i showed in the prev step for brevity. Even if I put the `...` under comments, the example wouldn't be a valid webpack config either.

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,207 @@
...
+  - Without bundling lodash but requiring it to be loaded by the consumer.
+  - Name of the library is `webpack-numbers` and the variable is `webpackNumbers`.
+  - Library can be imported as `import webpackNumbers from 'webpack-numbers'` or `require('webpack-numbers')`.
+  - Library can be accessed through global variable `webpackNumbers` when included through `script` tag.
+  - Library can be accessed inside Node.js.
+
+### Add webpack
+
+Add basic webpack configuration.
+Also add a [`.babelrc`](https://babeljs.io/docs/usage/babelrc/) file so that you can write webpack config in es2015 syntax.
```
sokra: To keep it simple, avoid this and write the config in CJS.

content/how-to/author-libraries.md
sokra: But we should also keep it simple and minimal, so I would prefer an object here too.

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,207 @@
...
+Similarly a `json-loader` is required to precompile our json fixture file.
+
+```javascript
+export default () => (
+    {
+        ...
+        module: {
+            rules: [{
+                test: /.js$/,
+                exclude: /node_modules/,
```
sokra: Every use `exclude: /node_modules/`. It's a bad pratice. Use `include` instead.

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,207 @@
...
+Now, if you run `webpack`, you will find that a largish bundle file is created. If you inspect the file, you will find that lodash has been bundled along with your code.
+It would be unnecessary for your library to bundle a library like `lodash`. Hence you would want to give up control of this external library to the consumer of your library.
+
+This can be done using the `externals` configuration as
+
+```javascript
+export.default () => ({
+    ...
+    externals: [{
+        '_' : 'lodash'
```
sokra: This is wrong. It extenalize `_` instead of `lodash`

content/how-to/author-libraries.md
SpaceK33z: Oh I didn't mean the `...`, I meant that `=` is invalid after `export default`.

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,207 @@
...
+### Final Steps
+
+[Tweak your production build using webpack](/how-to/generate-production-build).
+
+Add the path to your generated bundle as the package's main file in `package.json`
+
+```json
+{
+    ...
+    "main": "dist/webpack-numbers.js"
```
sokra: also add a `module` field, which points to the src

content/how-to/author-libraries.md
pksjce: @sokra - I did not understand what change is expected here? This is working in my example code.

content/how-to/author-libraries.md
pksjce: Sorry again! I couldn't find it in the [documentation](https://docs.npmjs.com/files/package.json). Does `package.json` have a module field?

content/how-to/author-libraries.md
pksjce: Oh my bad!

pksjce:
Thank you for the feedback! Have made some of the suggested changes.

content/how-to/author-libraries.md
SpaceK33z: He means that `_` should be `lodash`. In the example code you're using `import _ from 'lodash';`, so you should externalize `lodash`.

pksjce: Facepalm! Got it.

content/how-to/author-libraries.md
sokra: I don't think it's offically, but it's the inoffical field for `ES6 modules` main.

pksjce: Hey I found something related to that [here](https://github.com/nodejs/node/wiki/ES6-Module-Detection-in-Node).
Is this what you are talking about?

sokra: https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,177 @@
...
+
+const wordToNum = function(word) {
+        return _.reduce(numRef, (accum, ref) => {
+            return ref.word === word && word.toLowerCase() ? ref.num : accum;
+        }, -1);
+    };
+
+export {numToWord, wordToNum}
+
+```
```
sokra: Why not:

```javascript
import _ from 'lodash';
import numRef from './ref.json';

export function numToWord(num) {
    return _.reduce(numRef, (accum, ref) => {
        return ref.num === num ? ref.word : accum;
    }, '');
};

export function wordToNum(word) {
    return _.reduce(numRef, (accum, ref) => {
        return ref.word === word && word.toLowerCase() ? ref.num : accum;
    }, -1);
};
```

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,177 @@
...
+
+The usage spec for the library will be as follows.
+
+```javascript
+// ES2015 modules
+
+import webpackNumbers from 'webpack-numbers';
+
+    ...
+    webpackNumbers.wordToNum('Two') /// output is 2
```
sokra: It should either be:

``` js
import * as webpackNumbers from 'webpack-numbers';
webpackNumber.wordToNum(...)
```

or

``` js
import { wordToNum } from 'webpack-numbers';
wordToNum(...);
```

content/how-to/author-libraries.md
```diff
@@ -1,7 +1,177 @@
...
+
+Now, if you run `webpack`, you will find that a largish bundle file is created. If you inspect the file, you will find that lodash has been bundled along with your code.
+It would be unnecessary for your library to bundle a library like `lodash`. Hence you would want to give up control of this external library to the consumer of your library.
+
+This can be done using the `externals` configuration as
+
+```javascript
+module.exports = {
+    ...
+    externals:['lodash']
```
sokra: I think in this case you need something like:

``` js
externals: {
  "lodash": {
    commonjs: "lodash",
    commonjs2: "lodash",
    amd: "lodash",
    root: "_"
  }
}
```

to get UMD build working. This is required because you need a different value for the script-tag case.

pksjce: Oh got it. I updated it in the related sample app too.

sokra:
Thanks, I think this reached a level where we can merge it. It still need some minor fixes and additions, but this can be done in separate PRs.

pksjce:
Thanks!

Issue #297, opened 2016-11-04
Host webpack/media logo's on webpack.js.org


SpaceK33z:
The webpack logo's should be available somewhere on a hosted place. This way we can use the logo in README's etc.

It looks like the path to the logo at the top-left of the website [is a hash](https://webpack.js.org/7291f7007988120cc0b20372736970c2.png), so this is too tricky. Also it should be a SVG version.

We need this feature in relatively short-term for the new loader README's (webpack/webpack/issues/3195).

@skipjack, do you happen to know an easy way to do this?

skipjack:
Yea I can take this up this weekend, already had the update for the logo component in mind after seeing @jhnns comment a while ago, just hadn't gotten around to it.

For this issue, basically we just want a webpack branding page where people can just go download the logo and icon files for easy inclusion anywhere, correct? If so, assign this to me and I'll pick it up this weekend.

SpaceK33z:
Very nice :). Yeah a webpack branding page would be even better, initially I meant purely hosting the files, but a page dedicated to it is much better. https://github.com/webpack/media currently also serves as a branding page, so you might use that.

> already had the update for the logo component in mind after seeing @jhnns comment a while ago

Are you talking about updating the logo + text to be a SVG? That would be nice. Let me know, I can help you test if it looks good on some devices.

skipjack:
> Very nice :). Yeah a webpack branding page would be even better, initially I meant purely hosting the files, but a page dedicated to it is much better. https://github.com/webpack/media currently also serves as a branding page, so you might use that.

Ah good point, yea maybe I'll just link there to start. I need shift things around in the Navigation component as it's starting to get a bit cluttered with the current layout. At the very least I can add a link to the media repo there or in Footer.

> Are you talking about updating the logo + text to be a SVG? That would be nice.

Exactly, I think @jhnns mentioned he'd prefer that as opposed to using geomanist directly.

> Let me know, I can help you test if it looks good on some devices.

Sounds good 👍

bebraw:
Adding `package.json` to `webpack/media` would allow that to be consumed as a npm dependency. That might be a nice way to manage it.

skipjack:
@bebraw that's a good idea, I was thinking maybe via a submodule but maybe sticking with npm for everything is simpler. Maybe we should create another issue to start discussing how we can break parts of this repo out (as we were discussing with the components) and consume other repos like `webpack/media`. 

[...]

PR #303, opened 2016-11-06
Branding Page

skipjack:
Still a WIP

Adding a branding page for easy downloading of webpack [design assets](/webpack/media) and updating the Navigation component to better display the growing number of sections/pages. See #297 for the related discussion.

- [x] Update assets directory
- [x] Update dependent components
- [x] Add branding page
- [x] Add a branding link to the Navigation or Footer component


SpaceK33z:
Cool. Tested it, and on a "normal" screen (retina looks fine) the logo at the top left has some rough edges:

webpack

I don't know where these rough edges around the cube come from and if it's fixable?

The branding page looks good :).

Maybe run [`svgo`](https://github.com/svg/svgo) on the svgs to remove some unnecessary bytes?

skipjack:
@SpaceK33z yea I see that as well, I'll play around with it. It seems odd that a vector would render that way.

skipjack:
@SpaceK33z haven't found a solution to the rough edges yet although I did add a link to the branding page in the footer. 

I think once you review again we should just get this merged since you said it's somewhat time sensitive. I started working on the Navigation a bit but realized it's going to lead to some other more serious organizational changes which I think should be pushed to another PR. I can continue looking into what's going on with the svgs over there as well.

Funny thing for me is I only see those rough edges on my external monitor, on my macbook display it looks sharp.

SpaceK33z:
Allright merging. I'm looking with @indri-indri if we can fix the logo.

PR #304, opened 2016-11-06
Update hot-reload example to use require()

grrowl:
This was the only way I could get it to work, migrating from webpack 1 on a fairly simple project.

Also, seems logical since you either need to render the new reference to App, or re-require it on every `render` call (see https://github.com/gaearon/react-hot-loader/issues/243#issuecomment-213796910)

### DIFF
--- a/content/how-to/hot-module-reload.md
+++ b/content/how-to/hot-module-reload.md
@@ -171,12 +171,10 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { AppContainer } from 'react-hot-loader';
 
-import App from './components/App';
-
 const render = () => {
   ReactDOM.render(
     <AppContainer>
-      <App/>
+      { require('./components/App').default }
     </AppContainer>,
     document.getElementById('root')
   );


SpaceK33z:
@jmreidy, can you confirm this?

content/how-to/hot-module-reload.md
```diff
@@ -171,12 +171,10 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { AppContainer } from 'react-hot-loader';
 
-import App from './components/App';
-
 const render = () => {
   ReactDOM.render(
     <AppContainer>
-      <App/>
+      { require('./components/App').default }
```
dtinth: I think the code will be more clear if we did not inline it:

```js
  const App = require('./components/App').default
  ReactDOM.render(
    <AppContainer>
      <App />
```

Also, `require('./components/App').default` results in a React Component instead of a React Element, so I believe it doesn’t work unless you change it to this:

```js
      {React.createElement(require('./components/App').default)}
```

sokra:
In webpack 2 it should work with `import` as the `accept` updates the bindings.

If it doesn't work this way, it's a webpack bug.

Issue #305, opened 2016-11-07
Fixing backer layout breaks Sponsor listings. 


TheLarkInn:
It looks like since the latest merge the master, we have broken the sponsors showing up on the splash page. We probably need to get that prioritized if possible. 

Any thoughts on where specifically that could have happened?

skipjack:
It looks like more than that was broken, the top nav on mobile is all shifted as well and the mobile sidebar looks empty. Not sure what's going on yet, but I'll take a look.

SpaceK33z:
Tried to debug this. It looks like it's happening because the `Interactive` components set an `id` on the `div` with a path to the component in it. This works fine if the component only appears once on a page, but JavaScript really doesn't like it if you use the same value in `id` twice.

I'm not really sure if this is it btw, but it's very suspicious at least. I tried to change the `antwar-interactive` package to use `data-id` instead. However, apparently `antwar` itself also has some behavior that needs to be changed.  In `src/build/build_worker.js` I changed `$el.attr('id');` to `$el.data('id');`. After that it's not complaining anymore when building, but the html is just empty after the build. So I'm a bit stuck here...

SpaceK33z:
Andddd scratch that. Fixed the issue I was having, but this was not it :(.

SpaceK33z:
Okay I was partially wrong. The issue is in that antwar uses `document.getElementbyId` to find the interactive component, but since there are two divs with the same id the first one gets overridden.

Haven't found a solution, @bebraw, ideas?

skipjack:
Ok finally had a chance to test this, with `npm start` the site runs fine but `npm run build-test` does fail with the mentioned issues. 

@TheLarkInn still looking into the sponsors/backers issue, which is only broken after a full build. I think it has something to do with the double use of Interactive as @SpaceK33z mentioned. Working on a fix today.

The shifted nav issue I mentioned seems only to be on iOS and only on the splash page, which means it must have to do with `.navigation--transparent` class using some non iOS supported CSS or possibly something else got removed. 

Tested in Chrome w/ the screen size debugger as well as on an iOS 10 device. Should be able to open a PR with the fixes later today or tomorrow.

bebraw:
> Okay I was partially wrong. The issue is in that antwar uses document.getElementbyId to find the interactive component, but since there are two divs with the same id the first one gets overridden.

I could change the algorithm to include some kind of order index since the ids have to be unique. The fix could be something simple like that.

[...]

PR #306, opened 2016-11-08
Fixing Build Issues

skipjack:
A fix for a couple build issues, see #305 for related discussion. @SpaceK33z I gotta run to work, but if you could test and get this merged when you have a chance that'd be awesome. I guess we'll also have to do another deployment once this gets merged.

cc @bebraw @TheLarkInn

### DIFF
--- a/components/navigation/navigation-style.scss
+++ b/components/navigation/navigation-style.scss
@@ -17,6 +17,10 @@
     .navigation__inner {
       position:absolute;
       left:0;
+
+      @include break {
+        left:auto;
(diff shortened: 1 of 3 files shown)


SpaceK33z:
Clean fix :).

Adding `left:0` causes the menu to be left aligned on bigger screens. I think this was meant to only apply for small screens? See screenshot:
webpack

skipjack:
Haha yea just realized that too before leaving, see my third commit. Should've fixed it.

PR #311, opened 2016-11-09
prefer es6 import over require

sokra:


content/concepts/module-resolution.md
```diff
@@ -24,26 +24,26 @@ The dependency module can be from the application code or a third party library.
 * Absolute paths
 
 ```js
-require("/home/me/file");
-require("C:\\Home\\me\\file");
+import "/home/me/file";
+import "C:\\Users\\me\\file";
```
bebraw: Question: can you use placeholders to construct import paths? I haven't tried that but it might be good to mention in either case.

content/concepts/module-resolution.md
```diff
@@ -24,26 +24,26 @@ The dependency module can be from the application code or a third party library.
...
 ```
 
 Since we already have the absolute path to the file, no further resolution is required.
 
 * Relative paths
```
bebraw: The formatting here feels weird. These probably should be subheaders (`###` most likely).

content/concepts/modules.md
```diff
@@ -18,8 +18,8 @@ webpack builds on lessons learned from these systems and applies the concept of
 
 In contrast to Node.js modules, webpack _modules_ can express their _dependencies_ in a variety of ways. A few examples are:
 
-* A JavaScript `require()` statement
 * An ECMAScript2015 `import` statement
```
bebraw: ECMAScript2015 or ES2015? Maybe we can stick with the latter convention across the docs. Or go with ES6 even.

content/concepts/module-resolution.md
sokra: Do you mean something like `require(__dirname + "/path")`? Or what do you mean with `placeholders`?

content/concepts/modules.md
sokra: AFAIK ES6 is officially wrong and ES2015 is the correct version number. They switched from numbers to years.

bebraw: Ok, cool. Good addition to the writer's guide.

content/concepts/module-resolution.md
sokra: No this can't be used here. Only in some configuration properties.

PR #313, opened 2016-11-09
minor text changes

sokra:


content/concepts/module-resolution.md
```diff
@@ -47,7 +47,7 @@ require("module/lib/file");
 ```
 
 Modules are searched for inside directories which are specified using `resolve.modules`, which is can be an array comprising of different paths.
-Aliasing, ie setting `resolve.alias` to an existing module path, allows you to replace the module path with an alias name during `require/import`.
+Aliasing, i. e. setting `resolve.alias` to an existing module path, allows you to replace the module path with an alias name during `require/import`.
```
bebraw: This would be good to add to the writer's guide. I mean i.e. vs. i. e. . Ideally that should be linted.

content/concepts/targets.md
```diff
@@ -28,7 +28,7 @@ The following is a list of values you can pass to the `target` property.
 
 * `"async-node"` Compile for usage in a node.js-like environment (use `fs` and `vm` to load chunks async)
 * `"electron-main"` Compile for electron renderer process, provide a target using `JsonpTemplatePlugin`, `FunctionModulePlugin` for browser environment and `NodeTargetPlugin` and `ExternalsPlugin` for commonjs and electron bulit-in modules. *Note: need `webpack` >= 1.12.15.
-* `"node"` Compile for usage in a node.js-like environment (use `require` to load chunks)
+* `"node"` Compile for usage in a node.js-like environment (uses node.js `require` to load chunks)
```
bebraw: Here's another one. node vs. node.js vs. Node.js.

sokra: I would go with `Node.js` as the official site uses it this way.

sokra: I'll do a separate PR which replaces all occurences...

PR #319, opened 2016-11-09
Use latest preset, remove syntax plugin

hzoo:
Syntax plugins are not necessary if you have the equivalent transform plugin http://babeljs.io/docs/plugins/#syntax-plugins

And if you want to use es2016/es2017 then you can use the latest preset instead. 😄


TheLarkInn:
Going to rebuild this again and see if this was just a snafu. @hzoo would you mind signing the CLA (if you haven't already).

hzoo:
Ok signed - I just noticed I actually removed the transform plugin and not syntax 😛

hzoo:
@TheLarkInn Also changing to use https://github.com/babel/babel-preset-env instead of latest. env without any options is the same functionality as the latest preset.

SpaceK33z:
Tested it and verified everything still works.

Thanks!

PR #320, opened 2016-11-09
get-started: Fix and clarify usage of two dashes in npm scripts

SEAPUNK:
Ref: https://github.com/webpack/webpack.js.org/pull/293#discussion_r87222508


sokra:
Thanks, I was wondering too...

PR #321, opened 2016-11-10
Deploy

SpaceK33z:
Highlights:

- Sponsors & backers on homepage are fixed
- Lots of content changes
- New "branding" page

This deploy is relatively soon after the last one, but the sponsors & backers feature being broken doesn't look good. Also, @michael-ciniawsky is waiting for this deploy in https://github.com/webpack/webpack/issues/3195#issuecomment-259682056.


antwar.config.js
```diff
@@ -37,6 +37,16 @@ module.exports = {
         );
       }
     ),
+    'get-started': section(
+      'Get-Started',
```
bebraw: This should be `Get Started` but you can fix later.

PR #322, opened 2016-11-11
Docs about externals usage

pksjce:
Not very confident of the material.
Looked through [this readme](https://github.com/webpack/webpack/tree/54aa3cd0d6167943713491fd5e1110b777336be6/examples/externals) and code for [ExternalModules](https://github.com/webpack/webpack/blob/54aa3cd0d6167943713491fd5e1110b777336be6/lib/ExternalModule.js) and [ExternalModulesPlugin](https://github.com/webpack/webpack/blob/54aa3cd0d6167943713491fd5e1110b777336be6/lib/ExternalModuleFactoryPlugin.js)

### DIFF
--- a/content/configuration/externals.md
+++ b/content/configuration/externals.md
@@ -45,10 +45,7 @@ $('.my-element').animate(...);
 
 T> __consumer__ here is any end user application that includes the library that you have bundled using webpack.
 
-## Usage
-
 Your bundle which has external dependencies can be used in various module contexts mainly [CommonJS, AMD, global and ES2015 modules](/concepts/modules). The external library may be available in any of the above form but under different variables.
-
 
 `externals` supports the following module contexts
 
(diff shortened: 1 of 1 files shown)


content/configuration/externals.md
```diff
@@ -25,18 +29,62 @@ contributors:
...
+```javascript
 import $ from 'jquery';
 
 $('.my-element').animate(...);
 ```
 
-?> TODO: Add more details on the various ways of doing this using an `object`, `function`, etc. Also add how this is connected to [`output.libraryTarget`](/configuration/output#output-librarytarget) and maybe how it's useful when building libraries if this isn't already evident.
+T> __consumer__ here is any end user application that includes the library that you have bundled using webpack.
+
+## Usage
```
SpaceK33z: Why is this heading here? It seems out of place because `externals` is also on `##`.

content/configuration/externals.md
```diff
@@ -25,18 +29,62 @@ contributors:
...
+
+
+`externals` supports the following module contexts
+
+  * __global__ - An external library can be available as a global variable. The consumer can achieve this by including the external library in a script tag. This is the default setting for externals.
+  * __commonjs__ -  The consumer application may be using a CommonJS module system and hence the external library should be available as a CommonJS module.
+  * __amd__ - Similar to the above line but using AMD module system.
+
+```javascript
+externals : {
```
SpaceK33z: Instead of showing one big example here, I think it would be better to show a specific example under every type (string, array, ...).

content/configuration/externals.md
```diff
@@ -3,15 +3,19 @@ title: Externals
...
+  - pksjce
 ---
 
-?> Description
+`externals` configuration in webpack provides a way of not including a dependency in the bundle. Instead the created bundle relies on that dependency to be present in the consumers environment.
+This typically applies to library developers though application developers can make good use of this feature too.
 
 ## `externals`
 
 `string` `regex` `function` `array` `object`
```
SpaceK33z: Using it with `regex` is not explained later in this document. Could you either explain it or add a TODO note?

pksjce: Added it below.

SpaceK33z:
I think it's good now. It could be fine-tuned a little more, but we can do that later. Thanks!

PR #323, opened 2016-11-11
Fix stats.toJson() naming

40thieves:
Fixes `stats.toJSON()` -> `stats.toJson()`.

I'm pretty sure the correct method is `toJson()`: https://github.com/webpack/webpack/blob/ac340bce1d7239c0e6c0d0cf93dbb6c26dddde1f/lib/Stats.js#L21. However, I'm definitely not an expert on webpack core, so I could be wrong. Certainly in webpack@2.1.0-beta.25 it is.


SpaceK33z:
You're right, thanks1

PR #325, opened 2016-11-13
Navigation and Structure

skipjack:
Trying to address overcrowding of the links in the header and organize/sort the various sections a bit better.

- [x] Update the Navigation component allowing for top-level and child sections
- [x] Finish sorting all unsorted `/content` articles
- [x] Rename **How To** to **Guides** and update all how to titles
- [ ] Fix all broken links (mostly how-to) -- waiting to hear what ppl think of the rename

I realize the switch to **Guides** is kind of a big change, so I'm interested in what everyone thinks. I think it's more standard than the current title and used by a lot of other documentation sites such as Algolia and React. I do understand the idea of having `.../how-to/build-for-production` urls but I think having every article titled "How to ...?" is a bit repetitive (also makes titles really long).

Here's a quick gif of the updated structure:

There's a lot more space in the header now for things like #292 although I do think we should be careful not to throw too much in the top-level nav. Maybe continuing to utilize the child sectioning can help with that. I think if we merge this and then implement @patelsky's design for the Sidebar navigation, we'll be in pretty good shape. Between the header Navigation, search, and Sidebar nav (once it's updated), it should be pretty straightforward for users to navigate through the ever growing amount of content here.

*Credit to Algolia's docs for the child sections design idea* 👍


NejcZdovc:
With a lot of content on the page, I would suggest to add fixed side menu and header, so that user can always navigate between pages really fast.

SpaceK33z:
Nice, I tried it out and really like it. The only thing I would perhaps change is the background color of the subnav, but as you kinda said, that can be tackled with the new sidebar design.

Fully agree on the move to "Guides".

skipjack:
@NejcZdovc definitely, we already started playing around with react-sticky and react-headroom but ran into some issues with preact... however at least for react-headroom I think the issue was resolved and we can probably figure out react-sticky as well. Once we refactor the Sidebar design it would probably be good to try to tackle this as well.

@SpaceK33z great! Yea I think when we update the color scheme we can tweak the navigation design to provide more contrast and make it feel a bit less wireframy. I talked to @bebraw and he said he's good with the changes as well so I'll fix the build errors and rebase tonight so we can get this merged and move onto the next thing.

PR #328, opened 2016-11-14
Voting slider

NejcZdovc:
As discussed in #292 I implemented slider for the voting. For now I didn't add any styling to the page only slider functionality, so that you can check it out if this is ok first and after that I will add styling.

Note to @sokra:
- added step in the vote settings
- added color in the vote settings
- instead of infinity I set max to 100 with step amount to 10


sokra:
Thanks

PR #332, opened 2016-11-15
fixed link

donaldpipowitch:
I guess this is the first time `entry` is mentioned.

I don't know why `T> As a rule of thumb: for one HTML use exactly one entry point.` is highlighted as change? I didn't change something.


bebraw:
There could be a newline or some such at the end. No need to worry about that.

Issue #308, opened 2016-11-09
Issue with `currentPage` variable in Sidebar


bebraw:
Example: https://webpack.js.org/how-to/code-splitting/splitting-vendor/ . Notice the sidebar. This has likely something to do with the way it resolves the status.

NejcZdovc:
I am testing this on the latest version from develop branch and it's working correctly.

SpaceK33z:
Not sure, but looking at webpack.js.org (which was just deployed) it seems like this issue was fixed?

skipjack:
Weird. It does look fixed, however yesterday when I was testing other stuff with `build-test` this was still broken.

SpaceK33z:
Yeah I noticed that too, but I think @sokra fixed it in Antwar yesterday and updated the Antwar version this morning.

skipjack:
awesome 👍

PR #335, opened 2016-11-16
Navigation and Structure

skipjack:
Re-opening after accidentally closing #325.


skipjack:
@bebraw @SpaceK33z should be good to go, feel free to review again.

SpaceK33z:
There is one thing that is bothering me a bit: if you click on Documentation -> Output, the submenu doesn't expand (showing the options in Output). I have to click on the small arrow icon next to it to expand it.

Note that I tested this with `npm run build-test`.

skipjack:
Weird, that behavior doesn't work on `npm run build-test` but does work on the dev-server. I didn't change anything in the Sidebar, this PR was just for the top-level Navigation component and sorting/renaming of articles. I'll take a look though and see if I can figure out what's going on.

skipjack:
It's a problem with the `currentPage` prop in Sidebar which is somehow not being computed properly on build.

NejcZdovc:
Maybe this issue is related https://github.com/webpack/webpack.js.org/issues/308. I don't know why @bebraw closed it.

bebraw:
Ah, yeah. It's the same issue. Reopened.

skipjack:
Yeah I tried to debug it a bit, with very little luck, the odd thing is the Navigation component is essentially passed the same thing from Site but that url works fine. We could always go back to using window but I think @bebraw was trying to avoid that (which I agree makes sense).

However seeing as this issue is somewhat separate from this PR maybe I can fix the one conflict later today and then merge this? Any Sidebar issues could be tackled all at once when we implement the new design...

bebraw:
@skipjack Yeah, we can resolve it separately. Probably good to do a little session over it when we are both online at the same time.

skipjack:
Sounds good, or if you and @NejcZdovc want to try to resolve it before the redesign PR that's cool too. Either way 👍

SpaceK33z:
So @skipjack, this is good to merge I think? After fixing the merge conflict you can merge it :).

PR #338, opened 2016-11-16
fix homepage example

kentcdodds:
The example had a small issue :)


kentcdodds:
CLA signed :+1:

sokra:
Thanks

PR #341, opened 2016-11-17
Getting Started Clarifications

varunjayaraman:
The script tag in `get-started/index.md` is incorrectly placed. If someone tries to execute the code to demonstrate how browsers load global variables, they're going to get an error that the browser can't read `appendChild` of undefined because `document.body` won't exist yet. 

The script tag should be placed at the bottom of the body so that they don't get confused (although this is a pretty inconsequential thing and the whole point of this part of the tutorial is to tell people not to code this way lol).


SpaceK33z:
There is a second time the wrong <script> src is mentioned, could you also fix that? It's in the diff example.

varunjayaraman:
@SpaceK33z should be good to go

SpaceK33z:
Thanks!

PR #343, opened 2016-11-17
Fixed mobile menu alignment

felixsanz:
The mobile menu button appears in a place it shoulnd't on Chrome (at least using the Desktop version with a small viewport).

That margin is the cause. Also, to let the logo be center aligned the best way is to absolute positioning the menu button and take it out of flow.

### DIFF
--- a/components/navigation/navigation-style.scss
+++ b/components/navigation/navigation-style.scss
@@ -35,6 +35,7 @@
 .navigation__inner {
   display: flex;
   align-items: center;
+  position: relative;
   padding: 0.6em 1em;
 
   @include break {
@@ -45,7 +46,7 @@
 .navigation__mobile {
   display: flex;
   font-size: 1.5em;
-  margin-left: -1em;
+  position: absolute;
   align-items: center;
   color: map-get($colors, white);
   cursor: pointer;


themre:
yesterday I also noticed this, you got ahead of me :) I tested it and looks god, but it's up to the others to also look.

skipjack:
This is actually pretty similar to how we had it before, but a bug on iOS with the absolute positioning caused us to go back to the negative margin (which I agree is a bit hacky).

@felixsanz would you be able to share a quick screenshot of what this fixes? 

@SpaceK33z if you have time could you test this out, particularly on an iOS device? If not, I should have time this weekend.

felixsanz:
@skipjack Take a look at the menu, it's almost off-screen. The absolute positioning fix it.

Menu

skipjack:
@SpaceK33z testing this now, if it looks ok on my iOS device I'm fine with merging.

skipjack:
Ok it looks a bit off-center vertically, especially on iOS, which was the same problem I was having before with vertical alignment. It seems maybe changing the line-height of the icon will fix it.

skipjack:
As you can see below it's a little off, but maybe this isn't such a big deal. I spent a little time trying to find a fix with no luck. I'm fine with merging and dealing with this later I guess, what do you think @SpaceK33z?

img_2232

SpaceK33z:
Sure, this is an improvement. @felixsanz, thanks!

felixsanz:
Since it's absolute positioned i think `line-height` will not fix it. Just add a `top` property with a few pixels, like `top: 3px;`

PR #351, opened 2016-11-17
Hot-fix to section: Automatic -loader module name extension removed

Jontem:
Hot-fix to section: Automatic -loader module name extension removed

Added example to opt-in to the old behavior with the resolveLoader.moduleExtensions configuration option.


sokra:
Thanks

PR #358, opened 2016-11-18
Define the ESLint config as root

silvenon:
This tells ESLint not to extend any config up the directory tree, which (for example) makes it easier to be used as a submodule.

Let's say that I'm building a Dash docset from new Webpack docs, _hipothetically_, 😜 and I'm including this repo as a submodule, this repo's ESLint config will inherit my docset's config.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,4 +1,6 @@
 {
+    "root": true,
+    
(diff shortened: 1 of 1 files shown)


skipjack:
Good idea, thanks!

PR #372, opened 2016-11-20
Use SVG logo with text

SpaceK33z:
We have talked before about replacing the current logo + text with an SVG.

Using a SVG for logo and text fixes ensures the logo is displayed exactly the same everywhere. Also, it allows us to easily use an other font since the logo text doesn't depend on the Geomanist font being loaded anymore.

Example of what happened with my crappy monitor:

**Old:**
logo-old

**New:**
logo-new

Unfortunately I couldn't use the logo and text in `assets/logo-on-dark-bg.svg`, because the viewbox was too wide.


skipjack:
LGTM 👍

Issue #337, opened 2016-11-16
Design is broken on Safari


alanhussey:
The redesign looks great! …but it's a bit broken on Safari :(

- the dependency graph is offset, and is missing the files
- the navigation in the top-right overflows off of the screen

![screen shot 2016-11-16 at 1 17 24 pm](https://cloud.githubusercontent.com/assets/383725/20366288/30781180-abff-11e6-9b84-dd801c858fc4.png)

In addition, the menu button is a bit off-screen when the viewport is small enough.

![screen shot 2016-11-16 at 1 24 05 pm](https://cloud.githubusercontent.com/assets/383725/20366442/f9ecb2f0-abff-11e6-8fb8-3835d7af635c.png)

I did a search for `"is:issue is:open safari"` and nothing relevant turned up. Please let me know if there's anything I can do to help repro this.

skipjack:
@alanhussey thanks for pointing this out, we'll try to address it soon and let you know if we can't reproduce. We have some other updates planned for this section of the splash page so this can fit right in to those 👍

skipjack:
Once 809739a1168 is cherry picked to develop this should be fixed.

SpaceK33z:
Should be fixed now.

Issue #309, opened 2016-11-09
Landing page gradient gets banding on lower quality displays (think TN panel)


bebraw:
I'm not sure if there's much we can do about that but I would thought to raise an issue. Maybe tweaking the gradient curve a notch would make it less noticeable?

skipjack:
I think this will be moot if we move to the darker background as suggested by @patelsky. Once we update the color scheme I'd be happy to help make these changes.

skipjack:
Leaving this open until the landing page updates get done.

pastelsky:
The banding is because grey is one of the hardest colors for monitors to reproduce correctly. Switching to a different color might help.

skipjack:
Closing as I believe this was fixed by the new background added in #374.

PR #374, opened 2016-11-20
Refactoring and Fixes

### LINKED ISSUE #310: There's no way to exit/cancel search after opening the widget
Activating search works just great. The problem is that there's no way to exit it once you activate it on the landing page (might apply to other pages too). I would expect "return" on an empty field or "esc" to work. Likely blur should close it too so you can click outside of the field and make it closed.

### LINKED ISSUE #345: Search unicode doesn't render on windows
Search button unicode symbol on main page isn't rendered correctly on Windows 7, chrome 54 (not on default `Western (Windows1252)`, nor `Unicode (UTF-8)` encoding).

image

### LINKED ISSUE #357: Homepage file images not showing
screen shot 2016-11-18 at 11 37 24

No errors in console, disabled ad blocker. Firefox 47.0 on MacOS Sierra.

### LINKED ISSUE #359: Hide Navigation's children section on mobile screens
All navigation should be done through the mobile sidebar and that child section should be hidden.

image

skipjack:
This branch contains a bunch of changes and cleanup to make development a bit easier for other contribs. It also contains fixes for #359, #357, #345, and #310. 

@patelsky this should make some of your upcoming PRs a bit easier -- you can now add icons to icon font yourself (for example for use in the special warning, tip, and todo blockquotes) and instead of needing `map-get($colors, ...)`, you can just use the new `getColor(...)` function. Also cleaned up the splash page a bit -- switched to dark background and removed old visualization. 

- [x] Replace icomoon font with fontgen-loader and local icons
- [x] Remove Navigation's transparent mod and clean up splash page
- [x] Replace all `map-get($colors, ...)` with a simpler `getColor` function
- [x] Cleanup tip styling and remove current icons in preparation for new design
- [x] Fix #359 by hiding `navigation__bottom` until the first breakpoint
- [x] Fix #357 -- cleanup of the splash page makes this a non-issue
- [x] Fix #345 by replacing the unicode with an icon
- [x] Fix #310 by adding a blur behavior and and changing icon to a cross when active

Sorry this did become a bit monolithic, but a lot of these things kind of fit together. Also I think #337 and #309 might be fixed by these changes as well.


SpaceK33z:
About the splash page, the idea here is to only clean it up for now so it's easier for others to make a PR, right? That's okay I think, but if we would deploy after this PR, the landing page can potentially confuse users since it only displays the webpack logo with some files next to it.

As for the rest, I checked that the issues were fixed with the production build. Looks good! Is it correct that the fix for #310 is WIP?

skipjack:
@SpaceK33z yep thanks for the quick response! Just pushed the #310 fix right now.

Good point on the splash page, I can bring the old visualization back to life for now or maybe I should just add a quick screenshot of @pastelsky's modules as a placeholder.

SpaceK33z:
The fix for #310 is only working in dev mode, with production mode it doesn't seem to be working.

Yeah a quick screenshot will do!

skipjack:
Weird I tested that fix in prod with `build-test` and it seemed to be working fine, I'll check again though and make the splash fix.

skipjack:
Looks like the blur behavior and close button are working to me, what exactly looked broken?

SpaceK33z:
Oh it works for me too now, perhaps cache was acting weird or something.

skipjack:
Ok cool, would you mind cherry picking that last commit I pushed -- it includes better temporary styling for the splash page (I can't push directly to develop).

SpaceK33z:
I have the same permissions as you have ;) - but it's fixed now!

skipjack:
Haha, oh I didn't realize! No worries, thanks for fixing.

pastelsky:
Thanks @skipjack . Although I would've appreciated it if you'd asked me for the original resources ( for eg. the modules image). The one you committed seems to be a crop from the low-res mocks I'd proposed.

skipjack:
That was just a temporary fix, you're welcome to open a new PR whenever and completely refactor the splash page.

Issue #68, opened 2016-08-16
Whats New In Webpack 2 - Resolvers


TheLarkInn:
This will be a sub issue, under "Whats New in Webpack 2"

TheLarkInn:
We can stub this out. In a "Whats New in Webpack2" folder? Any opinions on this? @bebraw

bebraw:
Maybe start with a section at #20 given that's a priority item. If it looks like the ideas can be split into separate articles, then we'll split.

TheLarkInn:
Perfect I like. Absolutely.

sebastiandeutsch:
# Webpack 2.0 Resolving

Webpack 2.0 has an updated enhanced resolving mechanism:

```
resolve: {
  extensions: ['', '.js', '.jsx'],
  modules: ['src', 'node_modules'],
  plugins: [new DirectoryNamedWebpackPlugin()]
  ...
},
```

This is a complete list of all possible fields within `resolve`:

| Field                    | Default                     | Description                                                                        |
| ------------------------ | --------------------------- | ---------------------------------------------------------------------------------- |
| modules                  | ["node_modules"]            | A list of directories to resolve modules from, can be absolute path or folder name |
| descriptionFiles         | ["package.json"]            | A list of description files to read from |
| plugins                  | []                          | A list of additional resolve plugins which should be applied |
| mainFields               | ["main"]                    | A list of main fields in description files |
| aliasFields              | []                          | A list of alias fields in description files |
| mainFiles                | ["index"]                   | A list of main files in directories |
| extensions               | [".js", ".json", ".node"]   | A list of extensions which should be tried for files |
| enforceExtension         | false                       | Enforce that a extension from extensions must be used |
| moduleExtensions         | []                          | A list of module extsions which should be tried for modules |
| enforceModuleExtension   | false                       | Enforce that a extension from moduleExtensions must be used |
| alias                    | []                          | A list of module alias configurations or an object which maps key to value |
| resolveToContext         | false                       | Resolve to a context instead of a file |
| unsafeCache              | false                       | Use this cache object to unsafely cache the successful requests |
| cachePredicate           | function() { return true }; | A function which decides wheter a request should be cached or not. An object is passed to the function with `path` and `request` properties. |
| fileSystem               |                             | The file system which should be used |
| resolver                 | undefined                   | A prepared Resolver to which the plugins are attached |

## Upgrade from Webpack 1.x to Webpack 2.x

[...]

PR #388, opened 2016-11-21
Improve resolve docs

SpaceK33z:
The resolve docs were somewhat inconsistent and sometimes wrong. I fixed most of the issues I think.

I also compared the documented options to the table in https://github.com/webpack/webpack.js.org/issues/68#issuecomment-242033413 and added some descriptions.

@TheLarkInn, I'm not sure if #68 can be closed after this?

### DIFF
--- a/content/configuration/resolve.md
+++ b/content/configuration/resolve.md
@@ -288,7 +288,11 @@ plugins: [new DirectoryNamedWebpackPlugin()]
 
 `boolean`
 
-Whether to resolve symlinks to their symlinked location. Defaults to `true`.
+Whether to resolve symlinks to their symlinked location. Default:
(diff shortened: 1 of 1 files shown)


content/configuration/resolve.md
```diff
@@ -244,32 +259,42 @@ Default:
...
+Whether to resolve symlinks to their symlinked location. Defaults to `true`.
 
-?> Description
 
 ## `resolve.cachePredicate`
 
-?> Description
+`function`
+
+A function which decides wheter a request should be cached or not. An object is passed to the function with `path` and `request` properties. Default:
```
SpaceK33z: Good catch, fixed.

PR #413, opened 2016-11-23
fix mobile sidebar from getting stuck on iOS safari.

### LINKED ISSUE #412: Sidebar mobile gets stuck on iOS safari

prashantpalikhe:
`element.style = '';`

does _NOT_ clear up inline styles on safari.

Need to target specific property we are trying to remove.

`element.style.transform = '';`

in our case.

Closes #412


SpaceK33z:
Thanks!

PR #428, opened 2016-11-24
Update index.md

GeorgeAnagnostopoulos:
Updated index.html changing script tag position from inside head tag to inside body tag and removing a not used div tag


SpaceK33z:
Thanks!

Issue #291, opened 2016-11-01
Fresh Look: Colors & Fonts


pastelsky:
As discussed in #285 — discussing this first since it is a perquisite for other changes.
## 1. Colors
<img src="https://cloud.githubusercontent.com/assets/8946207/19899432/21c043bc-a086-11e6-90eb-b0e25d534cea.png" width="350" />

- The blues are picked up from the logo. 
- Cools greys have a slight blue tint and are used in the sidebar, level 2 text etc.
- Neutrals are used as body copy (dark) or body copy background (light)
- Accent color — green  is used in code highlighting to make the code section a stand out a bit — since we use blue shades everywhere else in the rest of the website.

I'm not married to the idea of having green as an accent color, but it seemed to complement the primary colors well (suggestions welcome)

### Code

<img src="https://cloud.githubusercontent.com/assets/8946207/19899871/9f17e81e-a087-11e6-94f3-af6f094b3972.png" width="650" />

## 2. Fonts
The existing design uses a combination of 8 font families / weights / styles including Cabin, Ubuntu Mono, Geomanist and the icon font.

#### Proposed font stack
1) Roboto or Source Sans Pro for body copy and headings (regular/bold/regular italic).
2) System specific monospace font — Monaco (Mac), Menlo (Mac/iOS), Deja Vu Sans (Linux), Consolas (Windows) Droid Mono (Android), fallback to Fira Code. 
The code fonts supplied with OS'es look pretty good and there's little reason to download a new font.
3) Icon font using a subset of material design icons.

This comes out to be a total of 4 font combinations that a user has to download, hence better avoiding FOUT (which can last a good 3-4 seconds for each font style on slower networks)

#### Readability
<img width="1235" alt="screen shot 2016-11-01 at 11 43 39 pm" src="https://cloud.githubusercontent.com/assets/8946207/19901406/105d5dd8-a08d-11e6-97f2-0a8284b4da3e.png">

_Click to enlarge image_ or see [codepen example](http://codepen.io/pastelsky/pen/pEMOdj)

bebraw:
> I'm not married to the idea having green as an accent color, but it seemed to complement the primary colors well.

Green is complement. Have you considered some orange or would that be pushing it?

pastelsky:
@bebraw I did try orange ( which is infact a literal complement of blue), but the warm and cold color combination didn't seem to go to well.

skipjack:
[...]

PR #430, opened 2016-11-25
Typography Improvements

pastelsky:
This is a proposal to change our font stack for better readability. 
Had been thinking of this myself and a comment from the recent [hacker news discussion](https://news.ycombinator.com/item?id=13023787) made me go forward with this PR.
>	dismantlethesun 1 day ago [-]
Wow. This documentation is so much better than what was previously available. Thank you from the bottom of my heart.
The one thing I will point out is that from a UX perspective the font you are using "Cabin", is naturally smaller and more compact than more common sans fonts (e.g. Open Sans, Century Gothic).
Because of this, even though your site is font-size 16, everything appears smaller than it would normally be.

By smaller, I think the author meant letter spacing, which is a problem with Cabin.

# Changes
- Simplified font stack to use `Source Sans` for body and `Source Code Pro`for code.
- Requests for fonts are now non-blocking. 
- Reduced size of Geomanist font by 50% (19KB vs. 42KB) using font subsetting.
- ~Don't unnecessarily bundle both WOFF and WOFF2 as base64.~ On hold until #443 is taken care of.
- Reduction in time to paint and load time, especially on slower networks.

According to me this has better readability especially when there are large paragraphs of text.


SpaceK33z:
Thanks, this font as body font is way easier to read. I'm not sure on using it for the headings. @jhnns said that he would still like to use Geomanist as heading font (https://github.com/webpack/webpack.js.org/issues/291#issuecomment-262286202).

pastelsky:
Will update headings to use Geomanist.

jhnns:
That looks gorgeous. Thx @pastelsky!

I'm not completely happy with the font colors (especially the contrast) and vertical rhythm, but it's definitely an improvement.

I'd love to pull this in, but what's wrong with tests?

pastelsky:
@jhnns glad you liked it. We can discuss the contrast and rhythm issues if you'd like to. As for the tests, it seems that the lint check for the hyperlink to the font files are failing. Although it seems yo be working just fine in development. How do I simulate a production build locally?

SpaceK33z:
@pastelsky, the font files seem to be really failing, I'm getting 404's on my local production build. You can make a production build with `npm run build-test`.

skipjack:
I had hacked this before with the `url-loader` to inline the fonts. For some reason, antwar is able to copy certain `import`ed assets over while excluding others -- thus causing the errors. If you want to use the `file-loader`, which I agree would be the better solution, I would talk to @bebraw. (I know he still has some plans for updating antwar's asset management although I'm sure for when)

SpaceK33z:
@pastelsky, perhaps an idea to revert the change in `webpack.config.js` for now, and make an issue to look into it later?

pastelsky:
Reverted those changes for now and have created a separate issue ( #443 ) since it requires some more exploration.

SpaceK33z:
All right, nice work again @pastelsky :).

bebraw:
@skipjack I'm currently redoing the asset pipeline. The problem is that it emits the assets below `./antwar/build` on the initial pass. I'll change it to emit to `/build` directly. That will fix the problem. Still a little bit to go but I hope to have the fix out there in a few days.

I also sorted out the CSS situation (single CSS file instead of multiple) already but I need to fix asset handling to make this nice.

skipjack:
@bebraw sounds good! As we finish up the major parts of the design/interface it would be nice to clean up some of the temporary hacks we added so implementing new features and fixing bugs is easier. Speaking of which, I'm working on some cleanup and fixes that I should be opening PRs for soon. 

Unfortunately I'm super busy this month but I'll still be helping out whenever I can. It's awesome that @pastelsky and @NejcZdovc have been picking up a lot. By the new year I should be able to dive into things again.

bebraw:
@skipjack Ok, cool. Any thoughts on pushing bits like navigation to a component of its own? Those are useful bits beyond this project.

PR #431, opened 2016-11-25
Contributor heads

pastelsky:
Higher res contributor images with nick name

## Before
screen shot 2016-11-25 at 2 21 40 pm

## After
screen shot 2016-11-25 at 2 55 43 pm


skipjack:
Looks awesome!

PR #440, opened 2016-11-26
Updated styles for table, removed nowrap

pastelsky:
# Changes 
- Updated table styles to go with site colors
- Removed nowrap from table, which was causing long horizontal scrolling, especially on mobile.

screen shot 2016-11-26 at 6 53 23 pm

### DIFF
--- a/styles/markdown.scss
+++ b/styles/markdown.scss
@@ -110,31 +110,53 @@
   }
 
   table {
-    display:block;
     width:100%;
-    overflow:auto;
+    background: #fff;
+    overflow: auto;
+    border-radius: 4px;
+    border: 1px solid #E9E9E9;
+    margin: 1em 0;
+    border-collapse: separate;
+    border-spacing: 0;
+    font-size: 15px;
+
+    thead {
+      tr:first-child {
+        & > th:first-child {
+          border-radius: 4px 0 0;
+        }
+        & > th:last-child {
+          border-radius: 0 4px 0 0;
+        }
+        > th:first-child {
+        }
+      }
+    }
 
     tr {
-      border-top: 1px solid #cccccc;
-      background-color: white;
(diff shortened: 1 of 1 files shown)


pastelsky:
@SpaceK33z should be working fine now.

SpaceK33z:
@pastelsky nope, still getting this. Did you want to push something?

pastelsky:
I dropped one of my commits that was causing it. Could you do a force pull?

SpaceK33z:
Ah okay, yes it's fixed now. There's another issue though: if a table is too wide (e.g. on `/configuration/resolve/`), the whole content area will get a horizontal scroll bar now. Previously only the table would get a horizontal scrollbar.

pastelsky:
Yeah, there is a tradeoff we'll have to make:-

1) **Existing**: Either use `white-space:nowrap`  on all cells and scroll the table when its too wide — See  https://webpack.js.org/configuration/target. This get's uncomfortable to use on mobile and desktop when content gets too much.

2) **Proposed Change**: Wrap table contents when it exceeds width. This has a con that if the contents of the cell itself have  `white-space: nowrap` content (like the resolve page you mentioned), then the wrapping doesn't work and the entire section scrolls.
Maybe we can avoid writing such long code samples in tables.

Another solution is to wrap all tables within a `<div>` and make that scrollable instead of the whole section, but that would require us to write a transform for markdown.

bebraw:
> Another solution is to wrap all tables within a <div> and make that scrollable instead of the whole section, but that would require us to write a transform for markdown.

There are a couple in place already so one more wouldn't hurt I guess.

bebraw:

skipjack:
Closing, however I'm going to pull some of these changes into a new PR after #875 is merged.

PR #441, opened 2016-11-26
Mobile enhancements

pastelsky:
# Changes
- Sidebar open is a _pull_ transition rather than a linear one.
- New styles for sidebar to indicate active state, use colors from scheme.
- Added a theme color meta tag for chrome.

|  Before | After  |
|---|---|
|  |   |


SpaceK33z:
Nice work again!

Issue #444, opened 2016-11-27
Vote app


NejcZdovc:
Based on a talk with @sokra I am creating an issue for a new vote mechanism.

Here is an [initial implementation](https://output.jsbin.com/qunevoyuhe), just to demonstrate the behaviour (it's not working on mobile yet).

What I wanted to achieve here is to reduce number of arrows for voting and to be mobile friendly. With this new system there would be only two buttons, up and down. If you click on a button you would increment your vote for 1. But if you would like to vote for a lot more, you can just hold down a button and amount for a vote will increase over time.

Current implementation (left number is your current vote, right number is increment):
1-5      =>  +1
5-10    =>  +2
10-40  =>  +5
40-70  =>  +10
70+     =>  +15

What do you think?

SpaceK33z:
I like this. It reminds me a bit more of UserVoice (or services like that), where you get 3 points you can divide between feature requests. They also allow you to spend those 3 points on 1 feature, and have a simple up and down vote button. If you vote twice, it will say "+2" between the vote buttons.

jhnns:
I also like this 👍

NejcZdovc:
Great will implement it tomorrow

NejcZdovc:
I think that I should wait for #447 to be merged and then work based on this. @pksjce already assumed in her design that we will have one button.

NejcZdovc:
I added this vote app logic to PR #447

PR #447, opened 2016-11-28
Some cleanup of voting page

pksjce:
- [x] Cleaned up login part.
- [x] Voting arrows are only one per side now. Maybe @NejcZdovc  could implement #444  on these?

Screenshots
**Logged In**
screen shot 2016-11-28 at 7 29 34 pm

**Logged out**
screen shot 2016-11-28 at 7 31 46 pm

**On Desktop**
screen shot 2016-11-29 at 8 15 43 am
screen shot 2016-11-29 at 8 16 01 am


pksjce:
Sure will do that.

NejcZdovc:
@pksjce sure I can implement into this design

components/vote/app-style.scss
```diff
@@ -5,7 +5,20 @@
 .vote-app {
   margin: 1.5em;
 
-  &__influence-descriptions {
+  &__top {
+      display: flex;
```
sokra: spacing

components/vote/button/button.jsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 require("./button-style.scss");
 
 export default (props) => {
-  let { value, myValue, maxDown, maxUp, color, onVote, className } = props;
+  let { value, myValue, maxDown, maxUp, color, onVote, className, isLoginActive } = props;
+  let {voteAppToken} = localStorage;
```
sokra: We should not let this component know about the login system (`localStorage`, `voteAppToken`). Could this information just be passed via props i. e `loggedIn = true` or `myValue != null`?

components/vote/button/new-button.jsx
```diff
@@ -0,0 +1,75 @@
...
+      </a>;
+    } else {
+      return <a
+        className={className}>
+          {fn({size: size, color: "#eee"})}
+      </a>;
+    }
+  }
+  render() {
+    const {voteAppToken} = localStorage;
```
sokra: We should not let this component know about the login system (localStorage, voteAppToken). Could this information just be passed via props i. e loggedIn = true or myValue != null?

components/vote/app.jsx
```diff
@@ -5,7 +5,9 @@ import * as api from "./api";
 import './app-style';
 import './influence-style';
 import VoteButton from './button/button';
```
sokra: Remove this if it's no longer used. Also delete the file if no longer used.

sokra:
Could you include a non-mobile screenshot?

components/vote/button/new-button.jsx
pksjce: Sure. Will do that

sokra:
We should merge this into a separate branch so @NejcZdovc can do his changes.

pksjce:
I have given @NejcZdovc collaborator permissions to my fork. Does that help?

NejcZdovc:
@pksjce I think this will work. I will start working on `voting/logoutstate` branch.

NejcZdovc:
I added my new vote app logic and also I added some mobile modifications, so that it's easy to read on mobile. One thing that is problematic is that arrow is too small for mobile and it's hard to vote. Any suggestion?

What if we place arrows on the side of the vote number (up arrow on the left and down arrow on the right)?

pksjce:
@NejcZdovc - Your changes look good!

bebraw:
Good to merge now?

pksjce:
I like the behaviour right now, but may have to change the up and down buttons to have better haptic feedback.

bebraw:
Yeah, another PR would work.

pksjce:
Thanks!

PR #448, opened 2016-11-28
Fix mixed quotation marks and backticks

soulchainer:
All `filename` properties in `code-splitting-libraries.md` are using quotation marks for open their filename string and «closing» them with backticks. Yeah, is a silly edit, but I'm just reading the documentation and it disturbes me ^_^U.


SpaceK33z:
Thanks, this would disturb me too haha

PR #450, opened 2016-11-29
Fix asset handling

bebraw:
Now it generates a single CSS file and handles assets through webpack. Instead of multiple CSS files, you end up with a single now so the result should be faster as well.

Interactive components cannot build to styling anymore so I moved the reference to places where `Interactive` is used instead. This has to do with the fact that those are evaluated separately from the main build. Maybe I could stub this in the future somehow but for now doing this extra step seemed like the smartest move.

Please test locally and let me know how it goes. Note that there's a new dependency (copy plugin) and also antwar received a couple of updates in the process (simpler now).


bebraw:
Yeah. I set it up so that it generates a separate font file instead of inlining. Easy to adjust.

PR #453, opened 2016-11-30
Remove incorrect apostrophe usage in "it's".

eventualbuddha:
The word should be "its" since the intention is to convey possession, not a contraction of "it is".


SpaceK33z:
Thanks!

PR #456, opened 2016-12-01
Remove unclear qoutes in configuration/target/ around "node"

ndelangen:
They seem unnecessary to me and might cause confusion.


SpaceK33z:
Good catch, thanks!

PR #459, opened 2016-12-01
Adds a line about nosources

pksjce:
With reference to #133

### DIFF
--- a/content/configuration/devtool.md
+++ b/content/configuration/devtool.md
@@ -50,7 +50,7 @@ W> There are some issues with Source Maps in Chrome. [We need your help!](https:
 
 `cheap-module-source-map` - A SourceMap without column-mappings that simplifies loaded Source Maps to a single mapping per line.
 
-`nosources-source-map` - A SourceMap is created without the sourcesContent in it.
+`nosources-source-map` - A SourceMap is created without the `sourcesContent` in it. It can be used to map stack traces on the client without exposing all of the source code.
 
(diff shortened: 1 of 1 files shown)


SpaceK33z:
Do you know what the difference is with this and passing `false`?

content/configuration/devtool.md
```diff
@@ -49,6 +49,8 @@ W> There are some issues with Source Maps in Chrome. [We need your help!](https:
 
 `cheap-module-source-map` - A SourceMap without column-mappings that simplifies loaded Source Maps to a single mapping per line.
 
+`nosources` - This option prevents the generation of a SourceMap file for your project.
```
sokra: the description is wrong. it generates a SourceMap but don't include the sources in the SourceMap.

pksjce: I tried this out in my project
 - `devtool: 'nosources'` - No source map files were created.
 - `devtool: false` - No sourcer
 - `devtool : <Any other option>` - Source map file is created

Should it behave differently?

pksjce:
@sokra - How is the sourcemap used if it doesnt have `sourcesContent`? Is there a way to provide `sourceRoot` in the webpack config?

sokra:
A SourceMap without `sourcesContent` can be used to map stack traces on client without exposing too much source code.

bebraw:
@pksjce Good to go?

pksjce:
@bebraw @sokra I think so, Checked all the cases in [devtool-example](https://github.com/kalcifer/webpack-examples/tree/master/devtool)

PR #461, opened 2016-12-01
The 'debug' property was removed in webpack 2.

loftyduck:
I tried using the debug option and got an Invalid configuration object error message so it seems strange to keep this Webpack 1 config in the v2.1 documentation.  I see there are some details about how to properly specify debug here: https://webpack.js.org/guides/migrating/#debug so I figured that's sufficient.


sokra:
Thanks

PR #462, opened 2016-12-01
Update dependency-management.md

ndelangen:
ADD documentation about dynamic requires, context api & require.context() usage.

### DIFF
--- a/content/guides/dependency-management.md
+++ b/content/guides/dependency-management.md
@@ -1,6 +1,8 @@
 ---
 title: Dependency Management
 sort: 6
+contributors:
+  - ndelangen
 ---
 
 > es6 modules
--- a/content/guides/dependency-management.md
+++ b/content/guides/dependency-management.md
@@ -13,7 +13,7 @@ contributors:
 
 ## require with expression
 
-A context is created if your request contains expressions, so the _exact_ module is not known on compile time.
+A context is created if your request contains expressions, so the **exact** module is not known on compile time.
 
 Example:
 ```javascript
@@ -29,7 +29,7 @@ Regular expression: /^.*\.ejs$/
 
 **context module**
 
-A context module is generated. It contains references to _all modules in that directory_ that can be required with a request matching the regular expression. The context module contains a map which translates requests to module ids.
+A context module is generated. It contains references to **all modules in that directory** that can be required with a request matching the regular expression. The context module contains a map which translates requests to module ids.
 
 Example:
 ```javascript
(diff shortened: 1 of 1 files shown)


content/guides/dependency-management.md
```diff
@@ -1,10 +1,96 @@
...
+
 > amd
-> require with expression
-> require.context
+
+## require with expression
+
+A context is created if your request contains expressions, so the _exact_ module is not known on compile time.
+
+Example:
```
bebraw: You could use `**Example:**` to make the formatting a little nicer here.

content/guides/dependency-management.md
```diff
@@ -1,10 +1,96 @@
...
+```
+Directory: ./template
+Regular expression: /^.*\.ejs$/
+```
+
+**context module**
+
+A context module is generated. It contains references to _all modules in that directory_ that can be required with a request matching the regular expression. The context module contains a map which translates requests to module ids.
+
+Example:
```
bebraw: Same.

content/guides/dependency-management.md
```diff
@@ -1,10 +1,96 @@
...
+    "./table.ejs": 42,
+    "./table-row.ejs": 43,
+    "./directory/folder.ejs": 44
+}
+```
+The context module also contains some runtime logic to access the map.
+
+This means dynamic requires are supported but will cause all possible modules to be included in the bundle.
+
+## require.context
```
bebraw: Maybe add formatting to `require.context` since it's a code fragment.

content/guides/dependency-management.md
```diff
@@ -1,10 +1,96 @@
...
+
+You can create your own context with the require.context function.
```
bebraw: Add formatting so you get `require.context`.

content/guides/dependency-management.md
```diff
@@ -1,10 +1,96 @@
...
+A context module exports a (require) function that takes one argument: the request.
+
+The exported function has 3 properties: `resolve`, `keys`, `id`.
+
+- `resolve` is a function and returns the module id of the parsed request.
+- `keys` is a function that returns an array of all possible requests that the context module can handle.
+
+  This can be useful if you want to require all files in a directory or matching a pattern, Example:
+
+  ```javascript
```
bebraw: Looks like there are some extra whitespaces in these examples (beginning/end). Better to drop although the parser probably doesn't choke on those.

ndelangen: This whitespace is to keep it inside the list-item, It's intended.

TheLarkInn: Yeah @bebraw that passes our mdlint tbh.

bebraw:
Thanks. 👍

PR #467, opened 2016-12-03
Update module-resolution.md

pikeshawn:
improved wording. removed 'is' instead of 'which is can be' it is 'which can be'


SpaceK33z:
Thanks!

Issue #280, opened 2016-10-26
Organization page table looks broken


SpaceK33z:
The [organization](https://webpack.js.org/organization/) page is a bit broken:

![webpack](https://cloud.githubusercontent.com/assets/533616/19729706/3639cc16-9b98-11e6-89ba-0f241a02d456.png)

NejcZdovc:
The only solution without changing tables completely is to set fixed width of the cells.

I personally would change this to card layout. We don't need a table here. We just need to display all projects. This can be easily solved by different layout.

http://thenextweb.com/wp-content/blogs.dir/1/files/2015/06/Screen-Shot-2015-06-16-at-08.57.56.png

https://material.google.com/components/cards.html

SpaceK33z:
Using cards is fine with me, but is there no way to fix it so that it uses the minimum width available?

skipjack:
I just pushed fa8c6616e which contains a fix for this. The issue was a `max-width` declaration on all images.

I agree with @NejcZdovc that we should consider using something other than a table to lay this out. This relates to our discussion in #246.

Maybe a custom `Stats` component and an `organizations.json` sheet would make the most sense for this. It could map the data to cards, a table, or whatever else... I'm willing to open a PR to experiment with this idea. This would mean moving the custom logic for the `stats` language from our `markdown.js` utility, which is already a bit cluttered, to the new component.

bebraw:
Moving the logic could be a good idea. 👍

skipjack:
This is fixed for now although I know tables still aren't perfect. Let's move this conversation back over to #246.

PR #474, opened 2016-12-06
Refactoring the Organization Page

skipjack:
As discussed a while back in #280, I stripped the custom `stats` lang logic out of `markdown.js` and created a component that maps out the projects to cards instead of a table:

This is still a work in progress but I think this view makes the information a bit more approachable, especially on smaller screens. Here's what's done and left to do:

- [x] Replace stats lang with a component
- [x] Move project list to json sheet
- [x] Replace table with card view
- [ ] Link to this page from somewhere in the site
- [x] Clean up the styling and fix Contributors hack
- [x] Discuss the configuration for this page with @bebraw 

The last one is the biggest hold up, **currently this only works on the dev-server**. If we like the direction of this then maybe @bebraw you might be able to give me an idea how best to configure this route?


bebraw:
Do you get some particular error when building?

skipjack:
@bebraw no build errors, it's just that nothing is created in the build folder for the organization page. I was reading through the antwar docs and it seems that mapping a route just to a component (not to a component that accepts content like Page) wasn't really the intended use case?

SpaceK33z:
@skipjack nice work! Looks good visually (based on your gif)! I'll test it when it's a bit further along.

bebraw:
> @bebraw no build errors, it's just that nothing is created in the build folder for the organization page. I was reading through the antwar docs and it seems that mapping a route just to a component (not to a component that accepts content like Page) wasn't really the intended use case?

Yeah, it has been designed so that it pushes found content through given layouts. We probably have to extend the system so that you can do something like `path: function() { ... return loaded jsx here ... }`. Then you can have custom views like this.

An alternative would be to extend the current `/` definition (it matches only markdown files) and build it there although I can see the value of functionality specific overrides like the one you did.

skipjack:
@SpaceK33z sounds good I'll let you know when it's in better shape.

@bebraw -

> We probably have to extend the system so that you can do something like `path: function() { ... return loaded jsx here ... }`. Then you can have custom views like this.

Ah, didn't think of that, that would then just pass it to a simple wrapper component I guess? I was also thinking maybe the json could be loaded through the path configuration and be passed to the component that way instead of directly `import`ed. Might be a good solution for auto-generated json that we want to feed to multiple pages with the same general layout.

> An alternative would be to extend the current `/` definition (it matches only markdown files) and build it there although I can see the value of functionality specific overrides like the one you did.

Yeah, that might work. If necessary maybe we have to do a quick hack to get it in for now and then figure out something more long term (not sure what effort any of these solutions would take on the antwar side). There's probably a few other places, e.g. the voting app, that could use a similar kind of configuration. Anyway, let's take this offline in gitter or something. I'll ping you once I have a bit more time to play around with it.

bebraw:
@skipjack I implemented something you might find useful. You can write

```javascript
standalone: {
  path() {
    return require('./layouts/Standalone').default;
  }
}
```

now. It's important that `path` is wrapped like that as otherwise it won't work. But this should give you overrides.

If you want to give it a go, you can find the feature in `22838a12` release.

skipjack:
@bebraw awesome, I'll try this out as soon as I can.

bebraw:
Alright. `efa30336` is the freshest version (I had to do a couple of fixes/tweaks). You can reach me through Gitter if there's something funky going on.

skipjack:
Ok with @bebraw's help, I got this into a better state (now it's passing the build). Working on wrapping it up but I have a few questions/comments:

@webpack/documentation-team 

1. @sokra @NejcZdovc can one of you test the voting app to make sure everything is working as you would expect? I updated the config for it and made a fix in the `List` component. Seems to be ok, but I haven't messed with it much before so I don't know if I'm missing something.

2. Any ideas where we should link to this page from? The footer's getting kinda full...

3. For the projects that didn't have maintainer's associated with them, I was going to link [here](https://github.com/webpack/webpack/issues/2734) although that issue is now closed. Is there somewhere better to point people to?

4. I'm going to finish the JSON sheet manually, but at some point we could probably auto-generate this like we are for plugins and loaders. @bebraw @sokra let me know once the plan for auto-generation is more fleshed out and I'll try to follow suit for this page.

bebraw:
> I'm going to finish the JSON sheet manually, but at some point we could probably auto-generate this like we are for plugins and loaders. @bebraw @sokra let me know once the plan for auto-generation is more fleshed out and I'll try to follow suit for this page.

If we stick with the current scheme for loaders and plugins, that could be easily generalized to work across the whole organization (certain `filter` needs to be skipped for this page).

NejcZdovc:
@skipjack Tried it out and vote app is working correctly as far as I can see.

skipjack:
@NejcZdovc great, thanks!

@SpaceK33z should be ready for a some testing now, when you have a chance. We still need to decide where to link to it from but for now you can just hit `localhost:8080/organization` on `build-test`.

bebraw:
@skipjack Poke me if you want to get rid of that `components/organization/projects.json`. It can be generated with a little bit of effort. That `maintainer` bit might be harder, but we can probably assume it's in *package.json* of each project or so.

skipjack:
@SpaceK33z great, I'll find a place to add a link to this page tonight and then merge it.
@bebraw I'll merge it the way it is for now bc I'd like to discuss automation a bit more before implementing. I'll ping you soon re that and the other stuff we've been discussing.

PR #482, opened 2016-12-09
Update why-webpack.md

Vitegor:


SpaceK33z:
webpack should always be written in lower case, as documented in the webpack/media repo

bebraw:
@SpaceK33z That would be a good thing to lint (mdlint plugin?). No need to worry after that.

bebraw:
Here's a related issue if you want to track: https://github.com/DavidAnson/markdownlint/issues/39 .

bebraw:
I think I'll close this. We'll get a linting rule in place eventually to force the writing style in this particular case.

PR #483, opened 2016-12-10
Notice about stats config

danez:
The stats config does not have any effect in the Node.JS API

I added a warning on configuration/stats
And also mentioned it on the Node.JS API page

### DIFF
--- a/content/api/node.md
+++ b/content/api/node.md
@@ -8,7 +8,7 @@ contributors:
 
 webpack provides a Node.js API which can be used directly in Node.js runtime.
 
-The Node.js API is useful in scenarios in which you need to customize the build or development process since all the reporting and error handling must be done manually and webpack only does the compiling part.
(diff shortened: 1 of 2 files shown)


SpaceK33z:
Hmm, it's not entirely correct that the `stats` options don't have any effect. It is correct that using e.g. `stats: { chunks: false }` in your **webpack config** has no effect, but the `stats` options are still relevant if you want to display stats, e.g. like this:

```js
compiler.watch({...}, function(err, stats) {
    console.log(stats.toString({ chunks: false }));
});
```

danez:
Yeah true. What I actually meant is that if this options is supplied to the webpack constructor, it won't print anything and calling `stats.to{String/Json}` won't respect these.

bebraw:

bebraw:
Thanks.

PR #493, opened 2016-12-14
Update/Fix Get-Started with ES6 standard import

Sieabah:
Changed 
```javascript
import _ from 'lodash';
```
to
```javascript
import * as _ from 'lodash';
```

### DIFF
--- a/content/concepts/index.md
+++ b/content/concepts/index.md
@@ -88,7 +88,7 @@ const config = {
 };
 ```
 
-In the configuration above we have defined a rules which used our loader with its two required properties: `test`, and `use`. This tells webpack's compiler the following:
+In the configuration above we have defined a rule which used our loader with its two required properties: `test`, and `use`. This tells webpack's compiler the following:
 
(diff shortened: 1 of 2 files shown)


asapach:
It should work both ways: https://github.com/lodash/lodash/blob/4.17.2-es/lodash.js. Why change it?

Sieabah:
@asapach It _should_ but for some reason doesn't when you follow the guide. My guess is it's trying to find '_' in the package to import.

bebraw:
@Sieabah That sounds like a webpack bug esp. if it works through Babel. Any thoughts on the import issue @sokra?

bebraw:

bebraw:
@Sieabah I've changed the PR to be applied against master. I think you need to do something like this:

```bash
git rebase master
git push --force
```

It is possible you may have to pull the latest master.

bebraw:
Ah, it's looking pretty good already. You could squash the commits, but apart from that it's fine.

bebraw:
`import _ from 'lodash';` works for me against the current RC. Can you verify?

bebraw:
Given `import _ from 'lodash';` should work, I'm going to close this one. Note that with lodash, the recommended way seems to be `import { omit } from 'lodash';`. There are also specific plugins that can help with it etc.

PR #496, opened 2016-12-14
Use import() for lazy-loading

bondz:
Use import() to lazy load files as it is now supported in webpack 2 as opposed to previously using System.import


bebraw:
Nice. Can you sign the CLA so we can get this merged? You may need to rebase and force push after that to make it catch the signing.

SpaceK33z:
Thanks!

bondz:
@bebraw done. Had to close and reopen the PR.

PR #500, opened 2016-12-15
docs(performance): Add performance config documentation

### LINKED ISSUE #435: Document new performance budgets feature
Now that most of the details are stamped out for https://github.com/webpack/webpack/pull/3350#pullrequestreview-9743611, we need to ensure that we are documenting the new configuration features, as well as specify what they represent. 

Here is the schema for the new information: 

```
    "performance": {
      "description": "Configuration for web performance recommendations.",
      "additionalProperties": false,
      "properties": {
        "hints": {
          "description": "Turn hints on or off",
          "type": "boolean"
        },
        "maxInitialChunkSize": {
          "description": "Total size of all initial chunks (in bytes)",
          "type": "number"
        },
        "maxAssetSize": {
          "description": "Filesize limit (in bytes) when exceeded, that webpack will provide performance hints",
          "type": "number"
        },
        "errorOnHint": {
          "description": "Throw errors instead of warnings when performance hints are found",
          "type": "boolean"
        }
      },
      "type": "object"
    }
```

Performance property: 
Default recommendations about web practice and performance. 

Reference https://github.com/webpack/webpack/issues/3216 for more details.

TheLarkInn:
Fixes #435

### DIFF
--- a/content/configuration/performance.md
+++ b/content/configuration/performance.md
@@ -5,7 +5,7 @@ contributors:
   - thelarkinn
 ---
 
-The options allows you to control how webpack notifies you of assets and entrypoints that exceed a specific file limit. 
+These options allows you to control how webpack notifies you of assets and entrypoints that exceed a specific file limit. 
 This feature was inspired by the idea of [webpack Performance Budgets](https://github.com/webpack/webpack/issues/3216).
 
 ## `performance`
--- a/content/configuration/performance.md
+++ b/content/configuration/performance.md
@@ -1,0 +1,97 @@
+---
+title: Performance
+sort: 14
+contributors:
+  - thelarkinn
+---
+
+The options allows you to control how webpack notifies you of assets and entrypoints that exceed a specific file limit. 
+This feature was inspired by the idea of [webpack Performance Budgets](https://github.com/webpack/webpack/issues/3216).
+
+## `performance`
+
+`object`
+
+Configure how performance hints are shown. For example if you have an asset that is over 250kb, webpack will emit a warning notifiying you of this.
+
+
+## `performance.hints`
+
+`boolean | "error" | "warning"`
+
+Turns hints on/off. In addition, tells webpack to throw either an error or a warning when hints are found. This property is set to `"warning"` by default.
+
+Given an asset is created that is over 250kb:
+
+```js
+performance: {
+  hints: false
+}
+```
+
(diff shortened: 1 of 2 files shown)


content/configuration/performance.md
```diff
@@ -0,0 +1,97 @@
+---
+title: Performance
+sort: 14
+contributors:
+  - thelarkinn
+---
+
+The options allows you to control how webpack notifies you of assets and entrypoints that exceed a specific file limit.
```
SpaceK33z: s/The/These/

content/configuration/performance.md
```diff
@@ -0,0 +1,97 @@
...
+
+No hint warnings or errors are shown.
+
+```js
+performance: {
+  hints: "warning"
+}
+```
+
+A warning will be displayed notifying you of a large asset. We recommend something like this for development environments.
```
bebraw: Let's say you are inlining assets during development, missing minification etc., wouldn't this yield warnings even if they aren't relevant? Isn't the best value in production build?

TheLarkInn: Yes, but also gives perspective to users as they are developing. 

If a bundle grows from 200->350kb after a certain feature, I'd want to know specifically about it while developing so I can trace my steps back a few changes and investigate more into what is causing the bloat.

bebraw: Alright. I guess it takes more experience with the feature for me to tell any better. Hitting the limits is very easy by inlining just a few images.

TheLarkInn: Yes true. Maybe that's where they could filter assets, etc.

I mean at the most these are recommendations but I 100% agree where you are coming from.

bebraw: Would it make sense to have something like `--json` kind of flag for the perf data? Can you get it in JSON to track?

TheLarkInn: Aww that would actually be a really cool feature but the log might be built. I'm going to add that to the original issue.

TheLarkInn: Might be big*

TheLarkInn:
Any other looks? Want to get this in since Perf is in now.

TheLarkInn:
Any other objections? Going to merge this.

bebraw:
Yeah, merge. More tweaks can be done in separate PRs.

PR #504, opened 2016-12-16
Updated Title to include "2"

gregoryagu:
Added "2" to the title.  The reason for this is that I found this page via a deep link (google search) and I thought it was talking about Webpack 1.  So the import statement did not work.  So I finally went on Stackoverflow where it was pointed out to me that it would work with Webpack 2.  So this simple update will help future travelers who do not realize the code will not work in V1.

### DIFF
--- a/content/get-started/index.md
+++ b/content/get-started/index.md
@@ -1,5 +1,5 @@
 ---
-title: Getting Started
+title: Getting Started with Webpack 2
 sort: 3
 contributors:
   - bebraw


SpaceK33z:
All webpack.js.org docs are for v2 exclusively. It also says so in the sidebar:
webpack

Note that I agree a little with you here, but as soon as we release webpack 2, which will be very soon™, this will no longer be an issue.

bebraw:
Yeah, agreed. Let's close this one. Thanks, though.

Issue #502, opened 2016-12-16
Migrating Guide: Code Splitting inaccuracy?


sunyang713:
Hi, I was reading over the migrating documentation, and under the section for code-splitting it says:

```javascript
function onClick() {
  import("./module").then(module => {
    module.default;
  }).catch(err => {
    console.log("Chunk loading failed");
  });
}
```

https://github.com/webpack/webpack.js.org/blob/develop/content/guides/migrating.md#code-splitting-with-es2015

Shouldn't the function inside the `.then()` return `module.default`?
```diff
  import("./module").then(module => {
-   module.default;
+   return module.default;
  }).
```

TheLarkInn:
That is correct!!! Would love a PR for this!!!!

sunyang713:
okay, super simple so I'm on it.

bebraw:
Merged.

PR #507, opened 2016-12-16
Update migrating.md: code-splitting section typo

sunyang713:
Under "Code Splitting with ES2015," the function inside `then()` should _return_ `module.default`.

Addresses https://github.com/webpack/webpack.js.org/issues/502


bebraw:
Thanks.

PR #510, opened 2016-12-17
New Look for homepage

pastelsky:
Have replaced the temporary png assets with svg, and added back the animated cube and also a new text rotation animation.

optim


pastelsky:
I guess at some point the cube component was removed from source, why did we do that?

SpaceK33z:
Woah, this... this is so beautiful!! This might be one of the best landing pages I've ever seen for a tool.

Two feedback points:

- When the title switches from styles -> assets, the text jumps a bit. You can also notice this in your GIF.
- The CI seems to be failing

sokra:
:+1: 

The logo must look equal to the original logo. i. e. the white borders...

sokra:
also great: http://codepen.io/jotavejv/details/MbZQzX/

skipjack:
@pastelsky 

> The logo must look equal to the original logo. i. e. the white borders...

That's why I removed the rotating cube -- I couldn't exactly replicate the new logo/icon without any funkiness. Somebody would be welcome to take a stab at getting the current Cube component closer, however I also like pen that @sokra linked to which might fit your graphic a bit better.

pastelsky:
Here's what I could achieve:

optim1

(_notice how the thicker border always stays on the outside_)

I think this looks pretty close to the original logo (it looks better in actual than the gif). 

As for the transparency, I kind of liked it since it signifies how webpack is not this "blackbox" that does stuff magically, but instead is transparent layer through which you can peek (read configure) into the core.

Also I think the rotating animation is more suited to the modules graph we have, since it represents the continuous "churn" that spits out static assets.

SpaceK33z:
@pastelsky that looks great, I agree that this is very suited since the rotating cube represents the "churning" of modules -> static assets.

Can you fix the CI and the merge conflict?

pastelsky:
@SpaceK33z Done. Could you check?

SpaceK33z:
@pastelsky, I fixed the merge conflict because GitHub was still complaining about it, but now the build fails again. Do you have any idea what this is:

```
ERROR in ./components/cube/cube-style.scss
Module parse failed: /home/travis/build/webpack/webpack.js.org/components/cube/cube-style.scss Unexpected character '@' (1:0)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected character '@' (1:0)
```

`@import 'functions';` is also used in other files, so why wouldn't it work in this file?

NejcZdovc:
@SpaceK33z SCSS files that contains import should be specified for this scenario in `components/splash/splash.jsx` and not in child components

SpaceK33z:
@NejcZdovc thanks, didn't know that!

I'm going to merge this after the CI finishes. There might be some things we still want to iterate on, but this is wayyyy better than what we have now.

NejcZdovc:
@SpaceK33z sure not a problem

PR #523, opened 2016-12-19

bonham000:
Added several suggested corrections to typos. Also, I'm not sure what the intended meaning in this sentence (line 33) is: In general the module developer writes handlers that are called when a dependency of this module is updated.

### DIFF
--- a/content/concepts/hot-module-replacement.md
+++ b/content/concepts/hot-module-replacement.md
@@ -24,19 +24,19 @@ In addition to the normal assets the compiler need to emit the "Update" to allow
 
 The manifest contains the new compilation hash and a list of all update chunks (2.).
 
-The update chunks contains code for all updated modules in this chunk (or a flag if a module was removed).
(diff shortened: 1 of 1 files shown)


bebraw:
Thanks for the fixes.

> Also, I'm not sure what the intended meaning in this sentence (line 33) is: In general the module developer writes handlers that are called when a dependency of this module is updated.

I would rewrite

```
HMR is an opt-in feature, so it only affects modules that contains HMR code. The documentation describes the API that is available in modules. In general the module developer writes handlers that are called when a dependency of this module is updated. The developer can also write a handler that is called when this module is updated.
```

as

```
HMR is an opt-in feature that affects only modules containing HMR code. You can consider patching styling through *style-loader*  as a good example. To make this work, *style-loader* implements the HMR interface. Implementing it you describe what should happen when a module is updated. In this case we simply replace the styling with the one received through HMR. 
```

Would that make more sense? If so, feel free to PR. Adjust as you like.

bonham000:
Yes, that looks good to me. I just submitted a pull request with that change.

PR #525, opened 2016-12-19
Removed unnecessary references to __dirname. 

vprasanth:
According to the `path.resolve` [documentation](https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths), and from testing, if an absolute path is not determined from the arguments, it automatically uses the current working directory, which is what `__dirname` ends up being when executing the script. Anyways.. that's my understanding, please do yell at me if I'm overlooking something obvious! 

Btw, if you folks agree, I can review the rest of the docs for occurrences of this.

### DIFF
--- a/content/configuration/index.md
+++ b/content/configuration/index.md
@@ -33,7 +33,7 @@ T> Notice that throughout the configuration we use Node's built-in [path module]
   [output](/configuration/output): {
     // options related how webpack emits results
 
-    [path](/configuration/output#output-path): path.resolve(__dirname, "dist"), // string
+    [path](/configuration/output#output-path): path.resolve("dist"), // string
     // the target directory for all output files
     // must be an absolute path (use the Node.js path module)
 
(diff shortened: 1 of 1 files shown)


bebraw:
I think the point with having `__dirname` there has to do with the fact that sometimes people like to move the configuration around. So you could have something like `config/dev.js`. Even though more verbose, the current setup would work in this case.

That said, there are neater ways to manage paths. Usually I push them to a single place and resolve there. This keeps it simple while giving the aforementioned advantage.

vprasanth:
Hey @bebraw thanks for the reply :) However, can you expand on that scenario, I don't think I'm understanding it clearly :( 

From my understanding, are saying because the configuration file(s) can be located in any arbitrary location within a project (in this case `config/dev.js`), having `__dirname` ensures that relative paths within the configuration continue to resolve correctly? That if `__dirname` is removed, paths relative to the config file would break?

bebraw:
@vprasanth That's the idea, correct. It's true we might be able to improve the examples somehow, though.

vprasanth:
@bebraw no that makes sense now, I confused myself with a false positive when testing out that scenario (wasn't running the config I thought I was) which confused me. Thanks again! Will close PR.

bebraw:
It would make a nice PR to mention why `__dirname` is used. Maybe there's a good place for that.

PR #536, opened 2016-12-21
Mention running the command on home page

Flimm:
Add one sentence to the home page, to make it clear that you need to run a webpack command to actually build the thing.

### DIFF
--- a/content/index.md
+++ b/content/index.md
@@ -46,5 +46,7 @@ module.exports = {
 </html>
 ```
 
+Then run `webpack` on the command-line to create `bundle.js`.
+
 ## It's that simple.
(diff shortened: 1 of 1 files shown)


bebraw:
I agree with the basic idea that the copy is too abstract. It took a while for me at least to see what's going on. I imagine it's only harder for people not familiar with the idea of bundling.

Adding a sentence like that would help a notch, but it would not remove the underlying issue. If possible, get people not familiar with webpack to take a look at it and see what they think. The copy can be improved based on that feedback.

bebraw:
Thanks. 👍

Flimm:
Thanks for merging!

PR #538, opened 2016-12-21
Fix #533

### LINKED ISSUE #533: A minor mistake with 'Getting Started'
image

image

Use `./node_modules/.bin/webpack -h` instead?

cntanglijun:


content/get-started/index.md
```diff
@@ -17,7 +18,8 @@ Create a demo directory to try out webpack. [Install webpack](/get-started/insta
 mkdir webpack-demo && cd webpack-demo
 npm init -y
 npm install --save-dev webpack
```
TheLarkInn: Ironically I've never tried this before, do you really need to reference the full bin path? @bebraw @pksjce

bebraw: Assuming there's no global install in place, yes.

bebraw:
Let's merge.

Interestingly running `webpack --help` locally in a project picks it up from **node_modules** if there's a global install in place (it picks up the local version over global). But if the global one is missing, this won't work. So I would say this PR is correct.

PR #542, opened 2016-12-22
add note about use with Babel [skip ci]

hzoo:
Should we be mentioning https://github.com/airbnb/babel-plugin-dynamic-import-webpack separately as well?


addyosmani:
This LGTM. Thanks @hzoo!

PR #548, opened 2016-12-25
Fix scrolling marquee text overlapping on safari

CasperVonPfaler:
Setting `flex-shrink: 0;` fixes the text overlapping problem on safari. Used `* >` in case someone uses the component with something else than `<span>` as children.

Also set an initial delay on the text rotator, because the first transition did not work on safari if none was set.


CasperVonPfaler:
Closed this pull request because I committed the changes with the wrong git settings. 😅 Will make a new one.

Issue #540, opened 2016-12-21
Fix vert scrolling marquee on splash page for Safari Mobile


TheLarkInn:
Looks like this has issues on mobile safari. ![image](https://cloud.githubusercontent.com/assets/3408176/21407911/84740e8e-c797-11e6-86e1-6e48b74f13f4.png)

CasperVonPfaler:
The same issue is also on desktop safari.
<img width="1141" alt="screen shot 2016-12-24 at 13 23 30" src="https://cloud.githubusercontent.com/assets/15709469/21466705/474e3fd8-c9dc-11e6-8a59-d52eb76d067a.png">
I could take a crack at fixing it over the holidays.

skipjack:
Pretty sure this was fixed in #549 as I remember reviewing that. Feel free to re-open if anyone is still seeing this bug.

PR #549, opened 2016-12-25
Fix scrolling marquee text overlapping on safari

CasperVonPfaler:
See #540 

Setting `flex-shrink: 0;` fixes the text overlapping problem on safari. Used `* >` in case someone uses the component with something else than `<span>` as children.

Also set an initial delay on the text rotator, because the first transition did not work on safari if none was set.


bebraw:

CasperVonPfaler:
@bebraw yeah, will do. @skipjack No problem! Excited to get my first real OSS contribution 😄

PR #553, opened 2016-12-26
Fix note about default extensions

evenfrost:
See [#3043](https://github.com/webpack/webpack/issues/3043#issuecomment-249314455) for reference.


bebraw:
Note that [all examples](https://github.com/webpack/webpack/search?p=3&q=extensions&utf8=%E2%9C%93) use `['.js', '.jsx']` kind of format so maybe this can be clarified further. I don't see `*` there. I don't understand why that comment is missing the `.` from extensions.

Maybe this needs more clarification.

PR #564, opened 2016-12-29
Update production-build.md - split configurations

rajagopal4890:
A simple approach to have Splitting configurations for the application and manage for different environments

### DIFF
--- a/content/guides/production-build.md
+++ b/content/guides/production-build.md
@@ -71,9 +71,11 @@ module.exports = {
   /*...*/
 };
 ```
+##Configuring webpack for multiple environments
+
 When we do have multiple configurations in mind for different environments, the easiest way is to write seperate js files for 
 each environment. For example:
-##config/dev.js
+####config/dev.js
 ```js
 module.exports = function (env) {
     debug: true,
@@ -94,7 +96,7 @@ module.exports = function (env) {
     }
 }
 ```
-##config/prod.js
+####config/prod.js
 ```js
 module.exports = function (env) {
     debug: false,
@@ -123,7 +125,7 @@ module.exports = function (env) {
     ]
 }
 ```
-Have our webpack.config.js has the following snippet:
+Have the following snippet in our webpack.config.js:
 ```js
 function buildConfig(env) {
   return require('./config/' + env + '.js')({ env: env })
@@ -134,7 +136,7 @@ module.exports = buildConfig(env);
 And from our package.json, where we build our application using webpack, the command goes like this:
 ```js
  "build:dev": "webpack --env=dev --progress --profile --colors",
- "build:dist": "webpack --env=qa --progress --profile --colors",
+ "build:dist": "webpack --env=prod --progress --profile --colors",
 ```
 
 You could see that we passed the environment variable to our webpack.config.js file. From there we used a simple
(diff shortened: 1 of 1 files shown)


content/guides/production-build.md
```diff
@@ -71,8 +71,87 @@ module.exports = {
...
+module.exports = function(env) {
+debug: false,
+    output: {
+      path: path.join(__dirname, '/../dist/assets'),
+      filename: '[name].bundle.js',
+      publicPath: publicPath,
+      sourceMapFilename: '[name].map'
+    },
+    plugins: [
+      new DedupePlugin(),
```
bebraw: Webpack 2 doesn't have `DedupePlugin` anymore.

rajagopal4890: Thanks for pointing out. Removed it. The idea was just to provide a sample config and the example was straight out of my code, which am still updating.

content/guides/production-build.md
```diff
@@ -71,8 +71,90 @@ module.exports = {
...
+      noInfo: false,
+      stats: 'minimal',
+      publicPath: publicPath
+    },
+ });
+```
+##config/prod.js
+```js
+module.exports = function(env) {
+debug: false,
```
bebraw: Looks like the indentation is off in this code example.

content/guides/production-build.md
```diff
@@ -71,8 +71,90 @@ module.exports = {
...
+        comments: false
+      })
+    ]
+});
+```
+Have our webpack.config.js has the following snippet:
+```js
+function buildConfig(env) {
+  var config;
+  switch (env) {
```
bebraw: I'm not sure if the `switch` is needed. Basically you could do `return require('./config/' + env + '.js')({ env: env })` or so to replace the structure. If it doesn't match, it will fail hard, but that's fine.

rajagopal4890: You are right. The switch-case is just a fail-safe, which I felt was necessary for my project, just in case. As for the example, am with your suggestion. It is much simpler. Thanks.

bebraw:
@SpaceK33z, @TheLarkInn Any thoughts?

TheLarkInn:
I like this approach. I also like using `webpack-merge`. I think we should remind people that there are multiple strategies somewhere in the doc.

rajagopal4890:
@TheLarkInn ,
Do you think, I should put up a snippet for 'webpack-merge' as well? Am using it in mine and It made my life a lot easier and made my code a lot cleaner.
I wanted to keep things simpler, but at the same time, wanted people to explore advanced configurations. That's why I just dropped a note on it.

TheLarkInn:
I think it would be quite valuable. Show an example of a composed pair of configs.

base => dev
base => prod

bebraw:
I might add a single file variant about merging at the end. Tha feels like a good section to add and it's actually the approach I use myself as then you have a single file to dig into to understand what's going on.

In addition I push lower level parts to `webpack.parts.js` to keep the level of abstraction a notch but I guess that can be skipped here as I discuss the approach in the book.

rajagopal4890:
@bebraw ,
As for the single file variant, the approach I demonstrated here, is actually an evolution from a single file approach, with which I started first. However, I wasn't using merge back then. I just stuck to the basics of if-else and switches. I totally agree that it would be easier to understand to start with. But then again, I felt, a split file system would instantly help you understand what plugins go where. For example, to understand how my final prod build would look like, I could just go to my prod file and get a feel of what goes into my prod build and what are the plugins am using to achieve it. Again, it is totally a developer, rather , the team's choice. Do you think, we should drop a note at the end on a single file approach as well?

bebraw:
@rajagopal4890 Ideally we should a couple of directions and let people pick whatever fits their purposes. To some extent these are up to personal preferences (think tabs vs. spaces kind of discussion) and it's hard to find absolute truths.

Give it a pass if you want. We'll merge later today. 👍

PR #566, opened 2016-12-30
Fix broken sublinks in sidebar

Shayac:
There is a bug in the sidebar links. Only the top level links (eg. `Module`) will link to another page. The sublinks (eg. `module.noParse`) only append the fragment to the url instead of linking to the correct page. 
For example, if you are at `/configuration`, open the `Module` category in the sidebar, and click the `module.noparse` anchor, it'll navigate to 
`/configuration/#module-noparse`
instead of 
`/configuration/module#module-noparse`.
I fixed it to include the entire link using react-router.


Shayac:
I just realized that this breaks the scrolling. It no longer automatically scrolls to the correct section of the docs. Gonna see if I can fix this and reopen.

Issue #563, opened 2016-12-29
Wrong line highlighting in source code


kennyjacob:
On https://github.com/webpack/webpack.js.org/blob/develop/content/concepts/plugins.md at the bottom there is a link to a code sample highlighting a specific section of code (https://github.com/webpack/webpack/blob/master/bin/webpack.js#L290-L292). Due to changes in that file, the highlight has become wrong.

You might need a way to link to a commit's version of the file or else update these doc links whenever the file changes.

bebraw:
Yeah, it would be far better to link to a specific commit. GitHub allows that. PR welcome.

bdougie:
I can PR real quick, already have a cloned version locally.

PR #567, opened 2016-12-30
Updates link to be commit specific

bdougie:
# What is this?

I saw #563 which states that a link is broken

# Solution

I pointed to a specifc commit message instead of the branch. Sorry for the PR snipe @kennyjacob, but it was quick and I was already forked and cloned repo this morning to check out the new site setup. 

see it live here http://typist-minimums-31187.netlify.com/concepts/plugins/

### DIFF
--- a/content/concepts/plugins.md
+++ b/content/concepts/plugins.md
@@ -87,4 +87,4 @@ module.exports = config;
   });
 ```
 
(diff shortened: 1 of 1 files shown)


bebraw:
Thanks!

PR #572, opened 2016-12-31
Fixed sidecar overflowing content

### LINKED ISSUE #382: Github etc buttons cover the article text
Loving the new docs, excellent work. One small issue I noticed though is that the buttons to the side slightly obscure the article text. The attached screenshot is taken from Safari on an iPad.

img_0033

NejcZdovc:
Resolves #382

@skipjack What do you think?

I see two fixes for this problem:

1. Increase break point (implemented in this PR)
2. Add padding to the content so that we can still display sidecar to the 1024px break point

I think that first option is better, that's why it's also part of this PR :smiley:

### DIFF
--- a/components/sidecar/sidecar-style.scss
+++ b/components/sidecar/sidecar-style.scss
@@ -4,7 +4,7 @@
 .sidecar {
   display: none;
 
-  @include break(large) {
+  @media (min-width: 1085px) {
     display: block;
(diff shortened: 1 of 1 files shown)


bebraw:

NejcZdovc:
@bebraw rebased

PR #573, opened 2016-12-31
Added split code blocks to home page

### LINKED ISSUE #350: Code blocks on homepage
The four code blocks on the homepage should be two in one row, instead of wasting a bunch of space:

screen shot 2016-11-17 at 18 03 09

Those code blocks could easily fit next to each other instead of having a massive amount of horizontal space, like this:

| Write your code | |
| ------------- | ------------- |
|  import bar from 'foo';       | export function bar()           |
| **Bundle with webpack** | |
| module.exports = {}      | &lt;html&gt; |

NejcZdovc:
Resolves #350

Now you can write HTML directly into the markdown files.

### DIFF
--- a/utilities/markdown.js
+++ b/utilities/markdown.js
@@ -139,11 +139,42 @@ function parseContent(data) {
   return tokens;
 }
 
+function handleHTMLSplit(tokens, htmlArray, merging) {
+  const htmlItem =  htmlArray[0];
+  htmlArray = htmlArray.slice(1);
+  const tickSplit = htmlItem.split('`');
+  const tickLength = tickSplit.length;
+
+  // detect start of the inline code
+  if(merging.length === 0 && tickLength%2 === 0) {
+    merging = htmlItem;
+  }
+  // append code inside the inline code
+  else if(merging.length > 0 && tickLength === 1) {
+    merging += htmlItem;
+  }
+  // finish inline code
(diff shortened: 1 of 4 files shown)


bebraw:
Looks good to me. Do you want to add anything else to the PR?

bebraw:

NejcZdovc:
@bebraw rebased

utilities/markdown.js
```diff
@@ -112,23 +113,63 @@ module.exports = function(section) {
...
+    // Split code in markdown, so that HTML inside code is not parsed
+    var codeArray = t.text.split(/(```(.|\n)*```)/g).filter(v => (v !== '' && v !== '\n'));
+
+    // if only one item in codeArray, then it's already parsed
+    if(codeArray.length == 1) {
+      return t;
+    }
+
+    codeArray.forEach(item => {
+      // if item is not code, then check for html tags and parse accordingly
```
SpaceK33z: What about code in backticks (e.g. \`<script>\`)?

NejcZdovc: I still need to cover this scenario, when we use only single backtick

NejcZdovc: Single backtick is now covered as well.

NejcZdovc:
@SpaceK33z can you look at it again please

PR #579, opened 2017-01-02
change filename to bundlje.js

kasperpeulen:
Seems consistent with the following information:

> On running `webpack` in your application, if you inspect the resulting bundle, you will see that `moment` and `index.js` have been bundled in `bundle.js`.

### DIFF
--- a/content/guides/code-splitting-libraries.md
+++ b/content/guides/code-splitting-libraries.md
@@ -37,7 +37,7 @@ module.exports = function(env) {
     return {
         entry: './index.js',
         output: {
-            filename: '[chunkhash].[name].js',
+            filename: 'bundle.js',
             path: './dist'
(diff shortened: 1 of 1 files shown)


bebraw:
I'm not sure. `'[chunkhash].[name].js'` is a generic form and it gives you cache busting as well thanks to `chunkhash` placeholder.

It's probably better to wait more opinions on this before merging.

Note that I changed the PR against `master` as we are are phasing out `develop` branch.

kasperpeulen:
@bebraw 

In that case, I think the tutorial text should change, as there is no `bundle.js` file now.

bebraw:
@kasperpeulen Yeah, sounds fair. Feel free to PR a change to the tutorial. Let's close this one.

Issue #557, opened 2016-12-27
Searching for "CLI" does not show you the CLI page


marcofugaro:
Here is the page I was trying to access to: https://webpack.js.org/api/cli/

I don't know how search works, but I think just adding `(CLI)` to the page title would suffice

dzello:
APIs, loaders and plugins weren't being picked up. That's been [fixed](https://github.com/algolia/docsearch-configs/commit/fa6c750cdd4a8673a0f5e1597dbf6f325fd3056e) and deployed.

Searching for `command line interface` works as expected now. Adding `(CLI)` to the header tag on that page should make it a top match for the short `CLI` form also.

bebraw:
Thanks for the fix. Closing.

PR #580, opened 2017-01-02
added (CLI) to CLI page title following discussion in #557

marcofugaro:
To access the page when searching for "cli"
> Adding (CLI) to the header tag on that page should make it a top match for the short CLI
https://github.com/webpack/webpack.js.org/issues/557


bebraw:
Thanks. Please develop against `master` in the future. 👍

PR #585, opened 2017-01-03
Antwar: live build also loaders/plugins docs

simon04:


bebraw:
Hi,

It actually points to `./generated/loaders` on purpose. The build process copies `content` below `generated` and then generates the rest of the files based on the loaders/plugins that exist.

simon04:
I see. Currently, the disadvantage is that one cannot update docs on loaders/plugins with automatic rebuild/reload.

bebraw:
@simon04 Ah, yeah. It's a little unfortunate during development, that's true.

This could be solved by handling it with two `require.context`s. The first one would do what the current one does and the latter would work against `content` like in your PR.

Composing contexts is a little tricky, but it should work as long as it exposes the same API (`keys()` etc.).

PR #591, opened 2017-01-04
Update install-webpack.md

johanjordaan:
I don't think its appropriate to have @beta in the documentation anymore.


bebraw:
It's on purpose as otherwise people get webpack 1 and the tutorial will fail. This can be changed only after there's a stable version out there.

johanjordaan:
Ahhh ... Thats makes so much sense :) Thanks.

Issue #301, opened 2016-11-05
Repo Structure - Modularizing and Consuming other Webpack Repos


skipjack:
Opening a discussion for how we can break parts of this repo out (as we were discussing with the components) and consume other repos like `webpack/media`. 

From #297:

> Adding `package.json` to `webpack/media` would allow that to be consumed as a npm dependency. That might be a nice way to manage it.

@bebraw that's a good idea, I was thinking via a submodule but maybe sticking with npm for everything is simpler... 

I'm also interested in what the team's plans are for the analyze tool. Ideally maybe we could have shared repo of components and utilities that this site, the analyze tool, and others all consume for a consistent interface?

Or maybe things like the voting app and analyze tool can be separate SPAs that this site consumes and displays under one of it's routes?

bebraw:
There's [an effort going on for renewing the analyse tool](https://github.com/webpack/analyse-tool). Integration is going to be interesting. Maybe we could consume it as a dependency (React component) and push it within an `Interactive` container somehow? That would allow us to retain main site navigation and all that.

Any thoughts @Bartekus?

Bartekus:
@bebraw I definitely agree about keeping things modular and the way the current analyze tool is structure should allow the whole component to work as either part of a bigger set or as an individual SPA.
Certainly flexibility, modability and extensibility are the ways of the webpack and as such, they should be our aim too. 
Keeping the webpack brand consistent and coherent to itself is in my opinion a great strategy to take.

bebraw:
@Bartekus Awesome. Good design would make it possible to run it through Electron and so on. Actually wrapping the site within Electron would be feasible (not sure of the exact benefit unless we add some features), but still. 👍

Bartekus:
@bebraw I think the benefit in and of itself would be to push the envelope further and further and to keep trying new ways of utilizing the power that webpack based development affords so that we continue to lead and advance the javascript community.

skipjack:
@bebraw @webpack/documentation-team I'm going to close this for now and try to focus mainly on knocking out documentation tasks for the next month or so. Once we get further along with the actual content, I think we can then take more time to clean everything up. This issue has so many parts to it that I think we should instead create smaller more targeted issues and re-open it once more pieces are in place.

Some notes in the meantime:

- I've already extracted and cleaned up our [banner/navigation][1] component into a separate repo called [react-banner][2]. 
[...]

PR #593, opened 2017-01-04
navigation: flatten menu hierarchy

simon04:
One says (e.g. [here](https://www.orbitmedia.com/blog/website-navigation/)) that it's okay to have up to 7 menu entries. I find the current splitting rather confusing (why is "guides" not part of "documentation"?)

### DIFF
--- a/components/navigation/navigation.jsx
+++ b/components/navigation/navigation.jsx
@@ -8,22 +8,22 @@ let Sections = [
   {
     title: 'Concepts',
     url: 'concepts'
-  },
-  {
+  }, {
     title: 'Guides',
     url: 'guides'
-  },
-  {
-    title: 'Documentation',
-    url: 'configuration',
-    children: [
-      { title: 'API', url: 'api' },
-      { title: 'Configuration', url: 'configuration' },
-      { title: 'Loaders', url: 'loaders' },
-      { title: 'Plugins', url: 'plugins' }
-    ]
-  },
-  {
+  }, {
+    title: 'API',
+    url: 'api'
+  }, {
+    title: 'Configuration',
+    url: 'configuration'
+  }, {
+    title: 'Loaders',
+    url: 'loaders'
+  }, {
+    title: 'Plugins',
+    url: 'plugins'
+  }, {
     title: 'Donate',
     url: '//opencollective.com/webpack'
   }


SpaceK33z:
Hmm, not sure about this. If later we want to add more pages then we'll need to redo something like this again.

Curious what @pastelsky and @bebraw think about this.

bebraw:
Losing "Documentation" category and flattening it up might not be a bad idea so 👍 from me. I'll leave the final decision to @skipjack and @pastelsky, though.

skipjack:
I split it up because we were running up against the limit. Although 7 isn't really a UX issue, as your article states, we have plans to add other top-level links (e.g. to the blog) there as well. And in our setup on certain screen sizes we were starting to run up against other elements. 

So as @SpaceK33z mentioned I think we would end up going back to something like this pretty quickly. What about renaming the Documentation link to Reference or otherwise re-organizing the links to address the confusion?

simon04:
For instance: while rewriting the shimming docs, I found myself jumping around between the following pages:
* /concepts/loaders/
* /configuration/
* /configuration/module/
* /guides/shimming/
* /loaders/
This required using both the menu (right-aligned) and the documentation submenu (left-aligned).

Another approach would be to list /concepts, /guides also in the submenu of /documentation. Whether we then keep their top menu entries is a matter of taste.

bebraw:
I don't see any harm in flattening now esp. due to that jumping issue (annoying extra clicks). Maybe secondary links, like blog, can go to some other place.

skipjack:
> This required using both the menu (right-aligned) and the documentation submenu (left-aligned).

Also aligning the submenu to the right or parts of the main menu to the left doesn't look so bad:

image

bebraw:
It is good to note how [ESLint](http://eslint.org/) tackles it. You can access the hierarchy from the front page without having to visit the pages themselves.

Based on that I would say if you could hover on top of documentation and see the submenu below it while you are on some other page, that would go a long way.

simon04:
What about (with improved spacing):
* 
* All docs are below Documentation
* The sidecar is merged into the menu

skipjack:
@simon04 yea something like that but without the multi-colored backgrounds and we'd probably leave the gitter link out and add some kind of fixed "Find Help" button or something for it instead. Personally, I'd prefer "Concepts", "Guides" and "Reference" (instead of Documentation) with the hover behavior @bebraw mentioned. 

Either way though, I'm going to close this for now and open this discussion again in an issue soon. We're working abstracting the banner (and other stuff, see #301), so I'll make sure to keep all this stuff in mind (icons, hover, etc.) as I get [react-banner](https://github.com/skipjack/react-banner) ready for it's first *working* release. 

If others bring up the navigation as a problem before then, then we can push one of the quick fixes we've discussed.

PR #607, opened 2017-01-05
Antwar: combine content+generated contexts

simon04:
Affects /loaders, /plugins.

Copying content to generated is no longer required.
Live build now also work for loaders/plugins docs.

Fixes #585.


bebraw:
Great work. Thanks.

PR #620, opened 2017-01-09
Add documentation for EnvironmentPlugin

einarlove:
Includes new API from https://github.com/webpack/webpack/pull/3161 which is not yet merged.


einarlove:
I did not see [content/plugins](https://github.com/webpack/webpack.js.org/tree/master/content/plugins) folder until now. It seems incomplete and not yet published to the site.

Should this be moved to a separate file in that folder?

einarlove:
I've moved the documentation to `content/plugins/environment-plugin.md` and adopted the writing style used in the other plugin documentations.

content/plugins/environment-plugin.md
```diff
@@ -0,0 +1,61 @@
...
+  - einarlove
+---
+
+The EnvironmentPlugin lets you list specific variables from `process.env` to be replaced on **compile** allowing your code to be run in the browser or later with the same configuration when built.
+
+```js
+new webpack.EnvironmentPlugin(keys)
+```
+
+### How to use
```
bebraw: `##` over `###`

bebraw:
> I did not see content/plugins folder until now. It seems incomplete and not yet published to the site.

The setup is partially dynamic. Any files placed there will be picked up by the build (it's aggregate of plugin readmes + this kind of data).

einarlove:
Beware that this documents a feature which isn't available before webpack@2.4. I'm not sure how long until it's released, so if you want this documentation merged before it arrives, we can add a notice about when that feature becomes/became available.

bebraw:
Maybe it would work if you documented the old behavior first (there's `EnvironmentPlugin` in webpack 1 even) and added a big fat warning (`W>`) to the section about the new work. Or we can just merge the docs when a new version of webpack goes out. Either is fine by me.

einarlove:
I gave it a try, but since the example is also based on the new feature, I would be a large rewrite. Suggest we wait for weback@2.4 instead.

bebraw:
Yeah, let's do that. One more option would be to push `EnvironmentPlugin` out of the core to webpack-contrib (pair with `DefinePlugin`?), but that would take a bigger decision as it might have other implications (uglify etc.). 👍

bebraw:
This should be good to go according to https://github.com/webpack/webpack/releases/tag/v2.2.1 .

Can you fix the conflict so we can merge?

einarlove:
Fixed

content/plugins/environment-plugin.md
```diff
@@ -1,24 +1,62 @@
 ---
 title: environment-plugin
 contributors:
-  - simon04
+  - einarlove
```
simon04: Do not delete previous contributors.

content/plugins/environment-plugin.md
```diff
@@ -1,24 +1,62 @@
...
+  NODE_ENV: 'development',
+  VERSION: 33,
+})
 ```
 
-This is equivalent to the following define-plugin application:
+T> Warning is gived if the value is undefined. This can be suppressed with a default value or a value of `null`.
 
-```javascript
-new webpack.DefinePlugin({
```
simon04: Do not remove the equivalent define-plugin application. This is the key in understanding this plugin.

content/plugins/environment-plugin.md
```diff
@@ -1,24 +1,62 @@
...
-  'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV)
+## Example
```
simon04: Do we need yet another example? We just can refer to https://webpack.js.org/plugins/define-plugin/

content/plugins/environment-plugin.md
simon04: Do not remove previous contributors.

content/plugins/environment-plugin.md
```diff
@@ -1,24 +1,62 @@
...
+
+```js
+new webpack.EnvironmentPlugin({
+  NODE_ENV: 'development',
+  VERSION: 33,
+})
 ```
 
-This is equivalent to the following define-plugin application:
+T> Warning is gived if the value is undefined. This can be suppressed with a default value or a value of `null`.
```
bebraw: /s/gived/given

content/plugins/environment-plugin.md
bebraw: Yeah, point to the other doc if possible.

content/plugins/environment-plugin.md
einarlove: 😨  Sorry @simon04, It was not intentional. I merged this obviously to quick without much thought.

content/plugins/environment-plugin.md
einarlove: I believe the docs for EnvironmentPlugin should describe how it works and be understood without knowing how DefinePlugin works. There's no need to be sparse on examples.

einarlove: The point of the example is also to show how EnvironmentPlugin stringify values which DefinePlugin does not.

bebraw: Ok, cool.

content/plugins/environment-plugin.md
```diff
@@ -2,23 +2,67 @@
...
-on [`process.env`](https://nodejs.org/api/process.html#process_process_env) keys.
+The EnvironmentPlugin is a shorthand for using the [DefinePlugin](/plugins/define-plugin) on [`process.env`](https://nodejs.org/api/process.html#process_process_env) keys.
+```js
+new webpack.DefinePlugin({
+  'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV)
+})
+```
+can be replaced with
+```js
+new webpack.EnvironmentPlugin([ 'NODE_ENV' ])
```
simon04: Drop the spaces after `[` and before `]` for consistency.

content/plugins/environment-plugin.md
```diff
@@ -2,23 +2,67 @@
...
-This is equivalent to the following define-plugin application:
+When given an object, the values are used as defaults for when they are not defined in `process.env`.
 
-```javascript
-new webpack.DefinePlugin({
-  'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV)
+```js
+new webpack.EnvironmentPlugin({
+  NODE_ENV: 'development',
+  VERSION: 33,
```
simon04: Maybe show also for this example how it corresponds to `DefinePlugin`.
```javascript
new webpack.DefinePlugin({
  'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'development')
  // ...
})
```

bebraw: That could be just

```javascript
new webpack.DefinePlugin({
  'process.env.NODE_ENV': `"${process.env.NODE_ENV || 'development'}"`
  // ...
})
```

using ES6 syntax.

bebraw: You already have `import` etc. in a lot of places so going with the current standard doesn't feel too bad. The gotcha is that you will need to process through Babel or Buble to support older browsers, but that's a separate problem.

einarlove: Either way, I think it would be unwise to bring back DefinePlugin syntax since it's already a large example.

einarlove: The whole introduction explains this is a shorthand for DefinePlugin and shows equivalent examples.

simon04: Right now, we have 3 different examples involving `NODE_ENV`, `NODE_ENV`+`VERSION`, and `NODE_ENV`+`DEBUG`+`TITLE`, respectively. Can we stick to only `NODE_ENV`+`VERSION` and explain everything on this example?

Still, I'd like `DefinePlugin` analogy for the object version somewhere.

einarlove: I disagree. I'll be gone for a week for a surgery so I'm unable to resolve this. You could add another PR amending current docs with the new options.

bebraw: @simon04 Maybe merge as is and then do tweaks you want as a separate PR?

@einarlove Good luck with the surgery. 👍

simon04: I'm not overly happy with the current state, but I don't stand in the way merging it (I might submit a followup-PR) :neutral_face:

bebraw: Yeah. Do a follow up. No need to get perfect first time around. 👍

bebraw:
I merged it as a version of webpack with the functionality has been released. Thanks again.

PR #584, opened 2017-01-03
production-build: enhance documentation

simon04:
- Copy over docs of DefinePlugin from webpack 1
- Add `webpack -p`
- Clarify `NODE_ENV` and the role of DefinePlugin

Addresses #353.
Incorporates parts of @markerikson's blog post https://www.packtpub.com/books/content/building-better-bundles-why-processenvnodeenv-matters-optimized-builds

PR #621, opened 2017-01-09
Fix production string literal

hpurmann:
See https://github.com/facebook/react/issues/7032


hpurmann:
Signed

simon04:
Well, for #584, I've taken this string from https://github.com/webpack/webpack/blob/8a07d4c1dd6a78a983de9945ef54108e61881a19/bin/config-yargs.js#L268
The outer quotes are required when invoking the command from a shell, but not for `system(3)` or similar.

PR #622, opened 2017-01-10
Update lazy-load-react.md

wmyers:
I think these `import`s should be `require`s


bebraw:
It's `import` on purpose as dynamic `import`s (returns a Promise) are supported now.

wmyers:
OK thanks for response, I did not know about that.

PR #626, opened 2017-01-11
Fix typo in migration guide

JamesHenry:
I _think_ it currently says the opposite of what it means because of a one letter typo - figured it would quicker to just submit a PR than an issue 🚀


sokra:
Oh yeah, that's important

JamesHenry:
Cool, CLA signed!

sokra:
Thanks

PR #630, opened 2017-01-13
config: tales of jQuery and the early days of modules

terinjokes:
In the distant past, long before React, ES Modules, fancy bundlers, or
even bower, the web was a very different place. Some might say, it was,
simpler.

Back in those days, to include a dependency in your project you had
primarily had two options:

- Download the required files and place them into a vendor directory.
- Hotlink the JavaScript from another source.

In either case, script tags had to be added to your site's HTML in the
correct order—manually handling dependency order—and the scripts
modified the window scope.

As web pages were beginning to become more complicated, this method of
doing things presented two major issues: it was common to include
resources that weren't being used, and errors attempting to use
a resource that hadn't yet been loaded.

AMD, the Asynchronous Module Definition, was created to define an API
that allowed JavaScript to be written as modules that defined their
dependencies which a loader could load asynchronously. It also had
a killer feature: used modules could be rolled into one file for
production!

While not common for most modules, some very popular libraries would
define with a specific name, to make it easy to consistently depend on
them by other libraries.

But, modules were still in the early days. Third party scripts, not
loaded by your AMD loader, might include the same popular library, but
with a different incompatible version. Including the third party script,
that didn't use AMD, might change versions of the popular library you
were using.

So AMD loaders implemented a way to signal to these popular libraries
that they had some way to protect the application developer, through
specialized loading logic, namespacing, or even warnings. AMD loaders
that implemented these protections could signal by setting a property on
define.amd to a truthy value, primarily for jQuery.

And while that was long ago, you may want to continue using modules that
expect this special handling. The “amd” configuration option allows you
to set the properties that should be exported.

### DIFF
--- a/content/configuration/other-options.md
+++ b/content/configuration/other-options.md
@@ -4,6 +4,7 @@ sort: 16
 contributors:
   - sokra
   - skipjack
+  - terinjokes
 ---
 
 ?> These are all the other options that might not need an entire page. Either we need to create new pages for them, move them to an existing page, or keep an **Other Options** section like this and replace this TODO with a short description/lead-in paragraph.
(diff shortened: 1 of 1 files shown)


bebraw:
Did you mean to include the issue description to some part of the docs?

terinjokes:
@bebraw No, I got a bit [carried away](https://twitter.com/terinjokes/status/819768913743355904). If there's any part of the commit message here you think should be part of the docs, I'm happy to send another PR, but I felt like being more verbose there wouldn't provide much benefit to the audience.

bebraw:
Ah, ok. Feel free to turn that into a blog post or some such so it doesn't get lost then. 👍

ev1stensberg:
@terinjokes You're my Open Source hero

PR #640, opened 2017-01-15
config:remove none-existent resolveToContext prop description. fix #634

### LINKED ISSUE #634: resolve.resolveToContext should be removed from docs cause webpack doesn't use this prop.
resolve.resolveToContext should be removed from docs cause webpack doesn't use this prop.

See response to [issue](https://github.com/webpack/webpack/issues/3954) opened in webpack repo.

stsiarzhanau:
Non-existent prop resolve.resolveToContext should be removed.


SpaceK33z:
Thanks!

PR #641, opened 2017-01-15
update(guides/lazy-load-react): fiber friendly side-effects

iammerrick:
Refactor LazilyLoad.js example to use latest React recommendations to be more defensive for Fiber rendering. This moves side effects into different lifecycle hooks. 

I actually jumped on to update the documentation to `import()` away from `System.import` but open source is awesome and others have already done that! So instead I made it more React-future friendly.

cc @TheLarkInn

### DIFF
--- a/content/guides/lazy-load-react.md
+++ b/content/guides/lazy-load-react.md
@@ -77,29 +77,26 @@ class LazilyLoad extends React.Component {
     };
   }
 
-  componentWillMount() {
-    this.load(this.props);
-  }
-
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


bebraw:
Could you re-open this against `master`? `develop` isn't used anymore and I accidentally merged against it (getting late).

bebraw:
@iammerrick No need to re-open. I cherry-picked your commit through #653. 👍

iammerrick:
Great! Thanks @bebraw!

PR #642, opened 2017-01-15
DOCS: Expand content/loaders 

mikeerickson:
Extended `content/concepts/loaders` section

- [x] Expand documentation, providing additional instructions
- [x] Added examples to accompany expanded points
- [x] Added `mikeerickson` to list of contributors (also updated output section)

@TheLarkInn Please review and let me know if you like the format used for the `loaders` section.  I basically expanded on what you have done elsewhere so it should flow the same.

### DIFF
--- a/content/concepts/loaders.md
+++ b/content/concepts/loaders.md
@@ -66,49 +66,8 @@ $ npm install --save-dev babel-loader css-loader style-loader sass-loader
 ```
 ## Options
 
-The following is a list of values you can pass to the `module.rules` array.  Each element shall contain an object defining the desired loader.
-
-### `test (required)`
-
-The `test` property contains a regex of the files you wish to transform
-
-`eg /\.js$/` will transform all files which end with `.js`
-
-### `loader (either loader or loaders required)`
-
-The `loader` property contains the name of the desired loader (eg `babel-loader`). If you wish to apply more than one transformation, use `loaders` property.
-
-`eg. loader: 'babel-loader'` will apply babel transformation to all `.js` files
-
-### `loaders (either loader or loaders required)`
-
-The `loaders` property shall contain an array of loaders (reads right to left) which you wish to apply.
-
-`eg. ['style-loader','css-loader']` will apply `css` then `style` tranformation
-
-### `use`
-
-The `use` property can be used instead of the `loaders` -or- `loaders` properties, where each array element contains a `loader` object
-
-```javascript
-  module: {
-    rules: [
-      ...
-      {test: /\.js$/, use: [{loader: 'babel-loader'},{loader: 'eslint-loader'}]}
-      ...
-    ]
-  },
-```
-
-### `include`
-
-The `include` property will contain a regex of path patterns you wish to use during transformation.  If property is not supplied, the default webpack `context` pathname will be used.
-
-### `exclude`
-
-The `execlude` property will contain a regex path patterns you wish to ignore during transformation.
-
-`eg. exclude: /bower_components|node_modules` will ignore supplied paths
+The above example demonstrates the different approaches which can be used when defining the `module.rules` section of your `webpack.config.js` file.
+For complete details, refer to [loaders section](/loaders).
 
 #### Omitting Loader Error
 
--- a/content/concepts/loaders.md
+++ b/content/concepts/loaders.md
@@ -6,6 +6,7 @@ contributors:
   - ev1stensberg
   - SpaceK33z
   - gangachris
+  - mikeerickson
 ---
 
 Loaders are transformations that are applied on a resource file of your application. They are functions (running in Node.js) that take the source of a resource file as the parameter and return the new source.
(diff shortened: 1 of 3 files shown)


simon04:
This duplicates content from https://webpack.js.org/configuration/module/. Do we want to go into detail in the "concept" section? IMO, also https://webpack.js.org/concepts/output/ is too long/detailed for the "concept" section.

bebraw:
Agreed. The content is valid, but it can be laid out better. It's enough if the concept article can describe the basic idea and link to other articles that dig into greater detail. You can have `## References` with links at the end.

mikeerickson:
@bebraw @simon04 my approach (and you are free to say otherwise) is to document the minimum examples to get started.  

I am teaching a new JR to start using webpack and was met with friction as soon as getting to concepts/loaders as there was literally not even simple examples 

It's your call, but I still believe examples where users are getting started

bebraw:
Have you seen https://webpack.js.org/loaders/ ? Probably the main thing missing is a reference between the concept page and that one. The concept page could contain a very basic example (the first one you added) and then defer to that article for more.

mikeerickson:
@bebraw no, I didn't even know about  this section until I read this initial reply.  I will review this section and revise my edits and provide link accordingly

bebraw:
@mikeerickson Ok, cool.

mikeerickson:
@bebraw One quick question before I proceed.   In review of this page example (https://webpack.js.org/loaders/) it uses the `use` long syntax.   Is this the "preferred" method, or is it safe to use the shorter approach as documented in my edits?

```
  module: {
    rules: [
      {
        test: /\.css$/,
        use: [
          { loader: 'style-loader'},
          {
            loader: 'css-loader',
            options: {
              modules: true
            }
          }
        ]
      }
    ]
  }
```

vs 

```
module: {
  rules: [
    {test: /\.js$/, loader: 'babel-loader'},
    ... 
  ]
}
```

bebraw:
@mikeerickson Both forms have their use I think. `use` is handy if you want to pass multiple loaders within a single definition. I have written a [little chapter](http://survivejs.com/webpack/understanding-loaders/loader-definitions/) with the key ideas.

mikeerickson:
@bebraw Would it be prudent and example worthy, to demonstrate the `short` syntax in the loaders section (along with the `use` format)?   Or, just link to it from `concepts/loaders` and be done with it.   Just trying to make it as clear for beginners as possible.

mikeerickson:
@bebraw In short, in the `Via Configuration` section, provide both short and extended examples, since this is what is used in the `concepts/loaders` section

bebraw:
@mikeerickson I would go with a mini example and a "call to action" to encourage to read more from the in-depth article that digs into further details.

mikeerickson:
@bebraw cool, that is the approach I have just applied  :-)

mikeerickson:
@bebraw I have updated my PR, please have a look and let me know what you think.

content/concepts/loaders.md
```diff
@@ -32,6 +33,64 @@ Loaders are [resolved similar to modules](/concepts/module-resolution/). A loade
...
+    rules: [
+      {test: /\.js$/, loader: 'babel-loader', exclude: /node_modules/}, // transform js files
+      {test: /\s[a|c]ss$/, loader: 'sass-loader'}                       // transform sass files
+      {test: /\.css$/, loaders: ['style-loader','css-loader']}          // transform css files
+    ]
+  }
+  ...
+};
+```
+#### Install Webpack Loader modules
```
bebraw: Can you check the header hierachy for both? I think you need ##/### at tops.

mikeerickson:
@bebraw Whoops, a product of a copy/paste error.  All fixed :-)

content/concepts/loaders.md
```diff
@@ -32,6 +33,64 @@ Loaders are [resolved similar to modules](/concepts/module-resolution/). A loade
...
+## Options
+
+The above example demonstrates the different approaches which can be used when defining the `module.rules` section of your `webpack.config.js` file.
+For complete details, refer to [loaders section](/loaders).
+
+### Omitting Loader Error
+
+If you fail to define a loader for the desired files within your project, you will receive an error similar to the following:
+
+`Module not found: Error: Can't resolve 'json-loader' in ...`
```
bebraw: json-loader is integrated to webpack 2. It would be a good idea to replace this example with some other format.

mikeerickson: @bebraw ah, I didnt know that :-)   It was already used as a basic example, so I just kept it around.  I will adjust to continue using the css stuff I have demonstrated (or maybe sass-loader)

content/concepts/loaders.md
```diff
@@ -32,6 +33,64 @@ Loaders are [resolved similar to modules](/concepts/module-resolution/). A loade
...
+
+**webpack.config.js**
+
+```javascript
+const config = {
+  ...
+  module: {
+    rules: [
+      {test: /\.js$/, loader: 'babel-loader', exclude: /node_modules/}, // transform js files
+      {test: /\s[a|c]ss$/, loader: 'sass-loader'}                       // transform sass files
```
bebraw: Why before the other styling loader? Shouldn't sass bits get evaluated first (evaluation order for loaders is from bottom to top if I remember right).

bebraw:
@simon04 Do you want to have a look? I could use a second opinion here.

simon04:
I'm not convinced:
I would rather take the _simple_, textual example from the intro paragraph ("you can use loaders to tell webpack to load a CSS file or to convert TypeScript to JavaScript") and develop a code example from there. Done in #736.
Also, it's good to know that there are various options to specify loaders, but I would just quickly sketch them and refer to the configuration page. Done in #736.

mikeerickson:
@bebraw @simon04 would you like me to adjust my changes, or just let it go (and accept #736)?  I am keen to do as you wish, just tell me what you want (since you are not convinced by my contribution).

bebraw:
@mikeerickson I think we'll accept #736 first and then figure out how to revise. I'll get back to you after that.

bebraw:
If you want, you could rebase it so that the PR retains `Omitting Loader Error` and `Webpack Rules Shorthand Syntax` sections. There's a typo in the latter (sytnax) to fix.

PR #644, opened 2017-01-16
Added section on BannerPlugin

doberkofler:
`BannerPlugin` no longer accept two parameters but rather only a single options object


simon04:
The migration to an options object affects a few other plugins:
* https://webpack.js.org/guides/migrating/#extracttextwebpackplugin-breaking-change
* https://webpack.js.org/plugins/commons-chunk-plugin/#options

It would be great to document all those changes in a similar fashion.

bebraw:
@simon04 Can you open separate PRs on those?

PR #650, opened 2017-01-16
Add simple Production section

### LINKED ISSUE #649: HMR add simple Production section
In order to get it work without errors and warns in console on production, it is required to add simple Production section for [HMR page](https://webpack.js.org/guides/hmr-react/).

In process of doing pull request...

AlexanderTserkovniy:
Production section explains how to get rid out of HMR error and warn logs after production compilation.

It contains plain recommendations for production bundling.

Fix #649

### DIFF
--- a/content/guides/hmr-react.md
+++ b/content/guides/hmr-react.md
@@ -6,6 +6,7 @@ contributors:
   - jhnns
   - sararubin
   - aiduryagin
+  - AlexanderTserkovniy
 ---
 
 As explained in detail on the [concept page](/concepts/hot-module-replacement), Hot Module Replacement (HMR) exchanges, adds, or removes modules while an application is running without a page reload.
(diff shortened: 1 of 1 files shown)


content/guides/hmr-react.md
```diff
@@ -296,3 +297,24 @@ dev-server.js:27 [HMR] App is up to date.
 
 Note that HMR specifies the paths of the updated modules.
 That's because we're using `NamedModulesPlugin`.
+
+### Production
+
+Minimal steps in order to get rid out of the HMR on production. As webpack sets `NODE_ENV` for working files only, we need to specify it before webpack for production is onvoked.
```
bebraw: /s/onvoked/invoked

SpaceK33z: "to get rid out of the" -> "to get rid of"

AlexanderTserkovniy:
Have fixed everything.

The `Alexander Tserkovniy` user was used by mistake, please check that in the commit there is no such user.

bebraw:
I wonder if it would make sense to link to [the production article](https://webpack.js.org/guides/production-build/) at the end and merge this information there. That would keep this on topic.

AlexanderTserkovniy:
Up to you guys, but in my opinion if I want precisely HMR for production, I would look for solution in HMR article.

bebraw:
Ah, I understand now. I handle the situation quite differently myself. I go with an inverse approach and **add** HMR to the development configuration avoiding the problem entirely.

But better wait for further comments.

SpaceK33z:
@bebraw I also do it like you. Resetting the `plugins` array is very tricky and only works if you don't have any other plugins. Same with the `entry` property. This might cause very weird bugs. I think the inverse approach (only adding it in dev mode) makes more sense.

bebraw:
Yeah, let's skip this one. That said, inverse approach or hinting about it would work.

PR #654, opened 2017-01-17
`RequireJS` in the title is misleading.

shidhincr:
`RequireJS` in the title is misleading ( confused with [RequireJS](http://requirejs.org/) ). It should be just `require.ensure`


bebraw:
👍

PR #665, opened 2017-01-18
[Vote Style] fix issue in Safari mobile

agirton:
In smaller screen sizes of Safari (e.g. Mobile) the user and score sections were overlapping the content. Which made the it harder to read the copy. Move the position to be medium on up. 

This wasn't an issue in Chrome or Firefox. Confirmed this still looks as expected in Safari medium on up, Chrome medium on up and Firefox medium on up.

**Before**
screen shot 2017-01-17 at 8 04 12 pm

**After**
screen shot 2017-01-17 at 8 03 02 pm

### DIFF
--- a/components/vote/app-style.scss
+++ b/components/vote/app-style.scss
@@ -28,10 +28,10 @@
   }
 
   &__user-section {
-    flex: 0 0 25%;
     padding: 0 0 30px;
 
     @include break(medium) {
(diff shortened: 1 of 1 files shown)


montogeek:
Confirmed it looks better on Mobile Safari, there is still space to show the score to the left, just like desktop, instead of the top

PR #672, opened 2017-01-18
Fix: Enable proselint and jslint for PRs.

pksjce:
Proselint is pretty useful in checking prose text. Had added this a long time back but hadn't completely figured out how it works. This here is at a state where all the rules pass.
We can now add nicer rules in the future. 
I couldn't find a way to ignore certain files or lines from a particular rule aka eslint. Hence, had to disable some rules.


bebraw:
Thanks. Can you add custom rules to Proselint? Linting that webpack is always lowercase would be valuable (comes up too often).

pksjce:
Yeah, I saw that comment and thought this could help. Now that it is setup properly, I ll investigate custom rules.

PR #682, opened 2017-01-19
Update migrating.md

divye01:
Replacing `use` with `loader`

### DIFF
--- a/content/guides/migrating.md
+++ b/content/guides/migrating.md
@@ -155,7 +155,7 @@ You may remove some hacks to work around this:
       {
         // ...
 -       loader: require.resolve("my-loader")
-+       loader: "my-loader"
++       use: "my-loader"
       }
     ]
   },
@@ -173,7 +173,7 @@ You may remove some hacks to work around this:
       {
         test: /\.js$/,
 +       enforce: "pre",
-        loader: "eslint-loader"
+        use: "eslint-loader"
       }
     ]
   }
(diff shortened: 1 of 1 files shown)


content/guides/migrating.md
```diff
@@ -261,7 +261,7 @@ module: {
   rules: [
     test: /.css$/,
 -    loader: ExtractTextPlugin.extract("style-loader", "css-loader", { publicPath: "/dist" })
-+    loader: ExtractTextPlugin.extract({
++    use: ExtractTextPlugin.extract({
```
bebraw: I think this fails now. It's a extract-text-plugin bug that should be fixed.

content/guides/migrating.md
```diff
@@ -368,7 +368,7 @@ module.exports = {
   module: { 
     rules: [{ 
       test: /\.tsx?$/,
-      loader: 'ts-loader'
+      use: 'ts-loader'
```
bebraw: `use` and `options` cannot be used together. If I remember right, validation should complain. A little confusing, I know.

content/guides/migrating.md
```diff
@@ -595,7 +595,7 @@ Having an `ident` on the options object means to be able to reference this optio
 ``` js
 {
   test: /.../,
-  loader: "...",
+  use: "...",
```
bebraw: Same here.

simon04:
@divye01 Are you still working on this PR?

content/guides/migrating.md
bebraw: Nm. `use` is fine now (I fixed it).

PR #690, opened 2017-01-20
Concepts tweaking

rouzbeh84:
Initially I just wanted to clarify a bit on what's the absolute minimum to get going and noticed that it stated only an `output.filename` is needed but since the only other **required** field is fairly far down below I thought it may be useful to have that upfront as well. 

Also, tried to expand upon the `targets` concept a bit as requested in the docs and ask if there should be further examples or clarification made.

Finally, I saw a spelling error in the `hot-moduele-replacement` file so while I was in the directory I just went through and updated some other inconsistencies and cleaned up some other inconsistencies. Please feel absolutely free to change any of these back however you see fit. My apologies in advance on this part 😳


.gitignore
```diff
@@ -3,4 +3,4 @@ npm-debug.log
 build
 generated
 .antwar
-
+.vscode
```
SpaceK33z: You should put this in your global gitignore instead.

rouzbeh84: Absolutely. Sorry about that. Was meant to be a nod/joke to @TheLarkInn. I can cancel and resubmit though.

SpaceK33z: No worries ;), you can just add an extra commit to this PR and it'll be fine.

rouzbeh84: Will do! Should be home shortly 👍🏽

rouzbeh84: Also, please let me know if any other adjustments are needed. I was a bit hesitant to add examples but thought they were appropriate where I did (hopefully).

SpaceK33z: I've not yet reviewed this fully, but one other thing: there seem to be two conflicting files. Merge or rebase to fix it.

rouzbeh84:
@SpaceK33z should be all set now. Even caught something I hadn't before so worked out even better in the end 👍

PR #692, opened 2017-01-21
Update output.md

chyipin:
Use `output.path` in example instead of `output.output`.


rouzbeh84:
thanks! I was just going through and adding my name to the contrib table and saw that. my apologies.

bebraw:
No probs. Feel free to PR more.

Issue #645, opened 2017-01-16
The BannerPlugin has no documentation in webpck 2


doberkofler:
There is no documentation of the BannerPlugin on https://webpack.js.org/plugins. How would I add a new plugin page as a PR?

simon04:
Go to https://github.com/webpack/webpack.js.org/new/master/content/plugins and input `banner-plugin.md` as filename. Make sure to include a minimal yaml header at the top of the file:
```yaml
---
title: banner-plugin
---
```

simon04:
BannerPlugin docs have been added in #779.

WouterVanherck:
Followed the above link and instructions. Have only found the commit for the banner-plugin.md file which actually points to older documentation of the banner-plugin? Searched it myself which resulted in the following link:  https://github.com/webpack/webpack.js.org/blob/master/src/content/plugins/banner-plugin.md

Posting this so it could help others in the future when searching for the banner-plugin docs.

PR #694, opened 2017-01-21
content/plugins directory cleanup

rouzbeh84:
adding sort order of plugin files and minor cleanup as well as adding a `[content/plugins/banner-plugin.md](content/plugins/banner-plugin.md` file with a yaml header thanks to @doberkofler via issue #645.

Is the index of this file supposed to be a running list of plugins?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ npm-debug.log
 build
 generated
 .antwar
-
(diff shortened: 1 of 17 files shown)


rouzbeh84:
closing due to letting an unset git user.name on a new laptop of mine get thru on commit 		49b046e. Will try to get back to this asap!

PR #699, opened 2017-01-21
Added instructions to disable safe write in the WebStorm IDE

miguelduarte42:
Just a simple addition to the list of instructions on how to disable safe write in WebStorm


SpaceK33z:
Thanks!

PR #707, opened 2017-01-22
Homepage cleanup

chrisdothtml:
I love what the homepage has turned into! It's really looking great. I have just a couple tweaks to the intro section that I think make it look a bit more fluid:

1. Adding a darker background to separate it visually from the 'support' section
2. Removing out of place 'webpack' title

Before/after preview:
before/after


bebraw:
Looks good to me. I'll let Greg do the final call on this, though.

PR #713, opened 2017-01-23
Added Babel to user get the final result working

DanZeuss:
Without Babel, the user cannot get the final result because we're using import * without transpiling it.

### DIFF
--- a/content/get-started/index.md
+++ b/content/get-started/index.md
@@ -116,8 +116,15 @@ index.js  1.56 kB       0  [emitted]  main
 
 T> If you created a local `webpack@beta` build, be sure to reference `webpack` with `./node_modules/.bin/webpack` on the command line.
 
-Open `index.html` in your browser to see the result of a successful bundle.
-You should see a page with the following text: 'Hello webpack'.
+if you try to open `index.html` file in your browser, you will see a blank page, or a error in the console with the following message: `import declarations may only appear at top level of a module`. This means that webpack is bundling the non-transpiled ES6 code, which is why these import/export statements are being found. babel-loader must therefore not be transpiling what you expect.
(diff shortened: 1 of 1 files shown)


content/get-started/index.md
```diff
@@ -116,8 +116,15 @@ index.js  1.56 kB       0  [emitted]  main
 
 T> If you created a local `webpack@beta` build, be sure to reference `webpack` with `./node_modules/.bin/webpack` on the command line.
 
-Open `index.html` in your browser to see the result of a successful bundle.
-You should see a page with the following text: 'Hello webpack'.
+if you try to open `index.html` file in your browser, you will see a blank page, or a error in the console with the following message: `import declarations may only appear at top level of a module`. This means that webpack is bundling the non-transpiled ES6 code, which is why these import/export statements are being found. babel-loader must therefore not be transpiling what you expect.
```
bebraw: /s/if/If

bebraw:
I wonder if it would be a better fix to drop *? Webpack should transpile imports for sure. It won't touch `const` etc. so perhaps having Babel around is good.

DanZeuss:
@bebraw Thank you for giving me an opportunity to try.

bebraw:
@SpaceK33z Do you want to have a quick look? Maybe introducing Babel at this point isn't a bad idea.

bebraw:
@simon04 Any thoughts on this?

simon04:
For inexperienced users this adds the need to dig into ES6, babel in addition to learning webpack. So I'd rather make the example ES5 compliant and maybe add a "Using webpack with ES6" just above the "Conclusion" section.

bebraw:
@simon04 Great point. Pushing the Babel bits to a separate article like that would be ideal.

PR #735, opened 2017-01-26
Fetch only webpack-contrib repositories

simon04:
The loader/plugin repository locations have changed again: Now all relevant repositories are in the webpack-contrib project.

At the moment, https://webpack.js.org/loaders/ does not list any loader.


bebraw:
I fixed this separately. It takes a peek at both webpack and webpack-contrib now although we can probably drop the webpack org at some point if any plugins/loaders won't be maintained there. But they are there as a fallback for now.

PR #736, opened 2017-01-26
concepts/loaders: improve, add code example

simon04:
* Start from the textual example, and add usage instructions (`npm install`, `webpack.config.js`).
* Sketch the different configuration options.
* Clarify loader resolution.


bebraw:
I think you may have to rebase. Loaders were moved below `webpack-contrib` and I had to tweak the config a little to take that into account.

bebraw:
Never mind. It was good. 👍

PR #739, opened 2017-01-27
Add missing require to code in hmr-react.md

mayani:
In Hot Module Replacement API section, the code is missing a require statement.
const App = require("./components/App").default;

### DIFF
--- a/content/guides/hmr-react.md
+++ b/content/guides/hmr-react.md
@@ -192,6 +192,7 @@ render(App);
 // Hot Module Replacement API
 if (module.hot) {
   module.hot.accept('./components/App', () => {
+    const App = require("./components/App").default;
     render(App)
   });
 }


bebraw:
This shouldn't be needed in webpack 2. Here's [reference](https://github.com/gaearon/react-hot-boilerplate/blob/9609a8e567b940ff341176638fb1169f70efd461/src/index.js).

That said, we could add a comment like that there.

mayani:
@bebraw You are right. 
Require was not needed with Webpack 2 ES modules. My set did not work because of an error in my .babelrc.

[Hot Module Replacement](https://webpack.js.org/guides/hmr-react/#code) Code section does reiterate this in point 3. So extra code comment is probably not necessary.

```json
"presets": [
    [ "latest", { "modules": false } ],
    "react"
]
```

instead of

```json
"presets": [
    [ "latest", { "es2015": { "modules": false } } ],
    "react"
]
```

bebraw:
Ok, cool. Thanks.

PR #740, opened 2017-01-27
remove confusing vendor entry in CommonsChunkPlugin example

axe312ger:
I found it pretty confusing to have the vendors chunk set to moment when using the `CommonsChunkPlugin`. Shouldn't the plugin do this for us, so there is no need to add `moment` explicitly?

> On running `webpack` now, we see that two bundles have been created. If you inspect these though, you will find that the code for `moment` is present in both the files!


axe312ger:
oops, edited wrong code snipped. Reopening ... :)

PR #741, opened 2017-01-27
remove confusing vendor entry in CommonsChunkPlugin example

axe312ger:
I found it pretty confusing to have the vendors chunk set to moment when using the `CommonsChunkPlugin`. Shouldn't the plugin do this for us, so there is no need to add `moment` explicitly?

> On running `webpack` now, we see that two bundles have been created. If you inspect these though, you will find that the code for `moment` is present in both the files!

### DIFF
--- a/content/guides/code-splitting-libraries.md
+++ b/content/guides/code-splitting-libraries.md
@@ -92,8 +92,7 @@ var path = require('path');
 module.exports = function(env) {
     return {
         entry: {
-            main: './index.js',
-            vendor: 'moment'
+            main: './index.js'
         },
         output: {
             filename: '[chunkhash].[name].js',


bebraw:
The line is there on purpose as it illustrates what `CommonsChunkPlugin` achieves. The goal is to end up with two bundles, show duplication, and then show how to get rid of the duplication in the next section.

bebraw:
That said, there's a [better technique](http://survivejs.com/webpack/building-with-webpack/splitting-bundles/#loading-dependencies-to-a-vendor-bundle-automatically) for achieving the same result with less complications. It might be worth expanding on that idea there.

simon04:
The doc for minChunks is already there: https://webpack.js.org/plugins/commons-chunk-plugin/#passing-the-minchunks-property-a-function

So close this PR?

bebraw:
Yeah, let's close. 👍

PR #746, opened 2017-01-28
Title for vote page

pllearns:
1. Created Title for page "Voting and Influence" - any suggestions are welcome. 
2. Styled the title to the vote page.

### DIFF
--- a/components/vote/app-style.scss
+++ b/components/vote/app-style.scss
@@ -16,7 +16,7 @@
 
   &__title {
     font-size: 30pt;
-    font-weight: bold;
+    font-weight: 800;
     margin-bottom: 10px;
     margin-left: 5px;
   }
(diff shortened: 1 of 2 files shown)


bebraw:
Can you split this into two PRs so that the styling bit goes to its own? Easier to review/merge. There's also some CLA issue (probably committed from different accounts). Thanks.

johnstew:
Having the styling bit on its own would be easier to review/merge. Also my original ticket was in reference to the actual HTML title of the page.

Right now it looks like this.

screen shot 2017-01-28 at 12 31 55 pm

That title for the page should match what you have in your styled title that you added to the page.

pllearns:
Thanks for the feedback. I'll take care of this tomorrow.

Issue #743, opened 2017-01-27
Getting Started cannot be accessed from the menu nor search


simon04:
The whole documentation part https://webpack.js.org/get-started/ can only be accessed from the "Get started" link from the index page. Neither the navigation menu (on non-mobile browsers) nor the documentation search will bring you there.

bebraw:
The footer link is a bit hidden. Maybe we should just add a new entry below "Guides" (new subnavigation). Guides could be more granular anyway as it has grown quite a bit.

johnstew:
@bebraw I think I follow. So should guides be broken up into different parent sections?

If thats the case then there should probably be a ticket dedicated to figuring out what those sections are. 

In the short term to solve this ticket, should this be added to the Guides navigation for now?

bebraw:
@johnstew Yup. That would be enough short term.

I think we may want to split the guides between development and production at least. This would mean three subs once you add "get started" there.

simon04:
Related: Right now, we have two installation pages:
* https://webpack.js.org/get-started/install-webpack/
* https://webpack.js.org/guides/installation/
My suggestion: keep the content from the first one.

bebraw:
@simon04 Great point. PR welcome. 👍 Btw, you should have access to the Slack now (check your gmail).

bebraw:
@simon04 I would merge the latter article with the first.

skipjack:
Closing this as @johnstew's PR (#750) added this to the sidebar. It sounds like some of the other problems discussed may be moved to new tickets.

PR #750, opened 2017-01-28
adding in get started to Guides page for #743

johnstew:


bebraw:
@skipjack Any thoughts on this? Fine by me.

skipjack:
Yea, I'm good with this for now. I'd like to clean up our routing structure a bit so hacks like this and others won't be necessary but that can wait a bit.

johnstew:
@skipjack Felt hacky when I was doing it but hopefully just temp solution.

PR #754, opened 2017-01-29
chore - Set up yarn for travis

bebraw:
Safer builds.


bebraw:
I tried without first. It won't work as we have `install:` block due to proselint. That seems to constrain the setup.

PR #755, opened 2017-01-29
Adding documentation for implicit common vendor chunk

rafde:
From the [Stack Overflow answer](http://stackoverflow.com/questions/30329337/how-to-bundle-vendor-scripts-separately-and-require-them-as-needed-with-webpack/38733864#38733864
) I gave a while back.

I updated the example to use `module.context` since it's more reliable than `module.userRequest`.

@bebraw you might want to [update](http://survivejs.com/webpack/building-with-webpack/splitting-bundles/#loading-dependencies-to-a-vendor-bundle-automatically)

Where I have this example and then seeing
https://webpack.js.org/plugins/commons-chunk-plugin/#explicit-vendor-chunk

I am starting to think that maybe this and all examples in this area should migrate to the plugin example section to have one area for all related common chunk info.

### DIFF
--- a/content/guides/code-splitting-libraries.md
+++ b/content/guides/code-splitting-libraries.md
@@ -5,6 +5,7 @@ contributors:
   - pksjce
   - chrisVillanueva
   - johnstew
+  - rafde
 ---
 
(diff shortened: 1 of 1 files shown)


bebraw:
Thanks. I did the book update as well (going to the next release). 👍

PR #760, opened 2017-01-29
Drop empty development using Docker

simon04:
webpack/webpack does not provide a Dockerfile, and most likely Docker is
not needed for development on webpack.


SrisailamKakurala:
@simon04 i'm having dependency conflicts when i'm trying to set up the project.. probably docker is required for webpack.js.org repo so we don't have to do `npm install`

SrisailamKakurala:
help me with the setup please!!

PR #774, opened 2017-01-31
Fetch babel/babel-loader README into loaders

simon04:


bebraw:
We may want to rework fetchers to work against a npm search in addition to the current ones. Maybe this info could be cached somewhere to speed things up a bit.

Issue #749, opened 2017-01-28
Break up Guides into different parent sections 


johnstew:
Breaking out the conversation from #743 into its own ticket to discuss possible parent sections and if this is a good idea. Plus need to decide the UI/UX around having a parent section.

**Parent Section Ideas:**
- Introduction
- Getting started
- Production
- Caching
- Migration
- Code Splitting
- Tree Shaking
- Contributing to Webpack
- React
- Angular
- Extras

rouzbeh84:
So is this one a wrap?

skipjack:
@johnstew so just to make sure I understand this correctly: you're talking about grouping the page links in sidebar and potentially creating folders within `/content/guides` (or any of the other `/content/...` directories) for each group?

For the UI/UX it would be good to tie this into #438 which, if I understand correctly, is a similar discussion.

johnstew:
@skipjack yes exactly

Looks like #438 is talking about the same thing as well.

skipjack:
@johnstew ok great, closing this in favor of that one and quoting you there.

PR #776, opened 2017-01-31
List get-started content under guides

simon04:
Related: #743, #749.

Supersedes #773 (wrong source branch).


bebraw:
It might be a good idea to write redirects for the pages. The syntax is like this:

```javascript
    redirects: {
      'get-started': '/guides/get-started'
    }
```

That should go below the path definition next to `path()` etc. Consider building only that portion of the site when testing the build (comment out everything else).

simon04:
I added the redirects, and extracted the syntax from [here](https://github.com/antwarjs/antwar/blob/619fa298abc8db0771104623d18e68610e0d49eb/smoketest/antwar.config.js#L51-L55).

PR #794, opened 2017-02-02
Clarify documentation of EnvironmentPlugin

simon04:
Follow-up of #620.

### DIFF
--- a/content/plugins/environment-plugin.md
+++ b/content/plugins/environment-plugin.md
@@ -6,63 +6,74 @@ contributors:
 ---
 
 The EnvironmentPlugin is a shorthand for using the [DefinePlugin](/plugins/define-plugin) on [`process.env`](https://nodejs.org/api/process.html#process_process_env) keys.
-```js
+
+## Usage
+
+The EnvironmentPlugin accepts either an array of keys.
+
+```javascript
+new webpack.EnvironmentPlugin(['NODE_ENV', 'DEBUG'])
+```
+
+This is equivalent to the following DefinePlugin application:
+
+```javascript
 new webpack.DefinePlugin({
-  'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV)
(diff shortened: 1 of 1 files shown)


content/plugins/environment-plugin.md
```diff
@@ -6,63 +6,74 @@ contributors:
...
+The EnvironmentPlugin accepts either an array of keys.
 
-```js
-new webpack.EnvironmentPlugin(['NODE_ENV', 'VERSION'])
+```javascript
+new webpack.EnvironmentPlugin(['NODE_ENV', 'DEBUG'])
 ```
 
-When given an object, the values are used as defaults for when they are not defined in `process.env`.
+This is equivalent to the following DefinePlugin application:
```
bebraw: `DefinePlugin` over DefinePlugin. You could do the same change for the rest as it reads a little better then.

simon04: We are not very consistent on that in other guides:
* https://webpack.js.org/plugins/ – lowercase w/o ticks
* http://localhost:3000/plugins/loader-options-plugin – lowercase w/ ticks

Shall we go for `CamelCase` for all plugins?

bebraw: Yeah, that's a good addition to the writer's guide too. I think it's a good idea as then you have a bit of meta in the text (plugins pop out).

simon04: I'll do a separate PR then.

bebraw: Ok. Poke me when you want me to look into something. Just adding as a reviewer seems to do the trick.

simon04: Can we merge this PR first to avoid merge conflicts?

bebraw: Yup.

simon04: Will you?

bebraw: Done. Sorry for the delay (dinner time).

PR #799, opened 2017-02-03
Updating the fixed sidebar behavior

skipjack:
Using `position: fixed` to prevent lagginess caused on certain displays.

Before:

After:

### DIFF
--- a/components/page/page-style.scss
+++ b/components/page/page-style.scss
@@ -12,6 +12,7 @@
 .page > div:first-of-type {
   flex: 0 0 auto;
   position: relative;
+  overflow: hidden;
 
   @include break {
     flex:0 0 30%;
--- a/components/sidebar/sidebar-style.scss
+++ b/components/sidebar/sidebar-style.scss
@@ -2,7 +2,6 @@
 @import 'mixins';
 
 .sidebar {
-  position: absolute;
   display: none;
   width: 100%;
   max-height: 100%;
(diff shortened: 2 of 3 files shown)


skipjack:
@SpaceK33z not sure how we missed this the first time around. On external monitors everything was working perfectly smooth, but on my laptop it lagged like crazy. @bebraw noticed as well.

Feel free to review, there's one other tiny fix I want to add before merging though.

skipjack:
@SpaceK33z wanna give this one more test? I got it to a good place and I think I ironed out a few other quirks along the way. 

@bebraw seeing as you noticed that glitchy-ness too if you could give it quick test that would be awesome.

The last thing I want is for this behavior to introduce new bugs we have to deal with as there's still a lot else to tackle.

bebraw:
It's perfect for me now. Nice work.

Btw, text rotater gives `Cannot read property 'classList' of null`. That's pointing to `./components/text-rotater/text-rotater.jsx?:52`. I think that `textRotatorWrap` bit is missing a little existence check. It might be good to resolve that in another PR.

SpaceK33z:
@skipjack I'm not on my own PC until monday (forgot my charger... 😕) so I can't test it out. But if @bebraw tested it already, it should be good to go!

skipjack:
@SpaceK33z sounds good, I'll merge then. @bebraw I'll look into that error, is it only in production or on the dev-server as well?

bebraw:
@skipjack I saw it in development mode after navigating a while. Not sure of production.

PR #804, opened 2017-02-03
/loaders: add index of important loaders

simon04:
* Move all loader documentation to /concepts/loaders
* List various loaders at /loaders, partially based on https://github.com/webpack/webpack/blob/2ed6448/README.md#loaders

### DIFF
--- a/content/concepts/loaders.md
+++ b/content/concepts/loaders.md
@@ -6,6 +6,7 @@ contributors:
   - ev1stensberg
   - SpaceK33z
   - gangachris
+  - TheLarkInn
   - simon04
 ---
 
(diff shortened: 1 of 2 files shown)


bebraw:
I wonder what to do with those license links. I guess we should resolve them to the packages somehow as now it's trying to look up against the site itself.

simon04:
The idea of those [license links/checks](https://travis-ci.org/webpack/webpack.js.org/builds/198179476#L678-L716) is to make sure that the included content is under a compatible license?

bebraw:
The links are broken in npm too due to the way they have been written. I guess the best fix would be to fix the links at each package for this reason rather than work around here.

simon04:
Hopefully those PRs will be merged quickly :smile:

bebraw:
@simon04 Merged.

simon04:
Thank you! Now the Travis build succeeds.

PR #809, opened 2017-02-03
guides: tune sorting of pages

simon04:
→


TheLarkInn:
Only thought here take a look at the Google Analytics page traffic we have for all pages filtered by /guides/: 
screen shot 2017-02-04 at 12 03 23 am

A topic that I find myself having to answer for folks in the wild, and on stack overflow quite common is "shimming" broken modules. Seeing the page rank, I wonder if perhaps we could bump up shimming a bit hire on the list to see if this helps remedy this.

TheLarkInn:
Furthermore, I'll let you decide if based on that data you see any other things that might sweeten the tune.

bebraw:
It's a good start. I think it would be a good idea to see if we could add some headers in between (just "development" and "production" entries that don't link anywhere would add hierarchy).

simon04:
Thanks for the comments, and the insights in usage statistics. I changed the sorting as follows:
* getting started
* install, migrate
* code splitting
* production build + caching
* development
* shimming

Everything else is sorted alphabetically below.

PR #820, opened 2017-02-04
Add CLI reference

simon04:


simon04:
Ahh, there's already an (possibly outdated) CLI doc: https://webpack.js.org/api/cli/ – *sigh*. However the search does not find this one, see #822. :confused:

bebraw:
Yeah, better merge with the api doc (really hidden). Also moving the resulting doc below root level (needs a rewrite) would be a good move. Then we get "api, cli, configuration, ..." at navigation with a tiny amount of tweaking.

bebraw:
What do you want to do with this? I would do `/cli` and make it visible in the navigation (easier to find). Push all CLI bits there and call it a day.

simon04:
I'll open a separate PR to merge the new bits from this PR with the existing docs.

PR #828, opened 2017-02-05
code-splitting-css: rewrite guide in a concise way

simon04:
Focus on the relevant points. Link to corresponding loaders/plugins for
details. Consistent typesetting for loaders/plugins.


bebraw:
A tiny intro could still be good so you get a quick overview.

simon04:
Good point. I added an intro, and introduced another section on "Importing CSS".

PR #837, opened 2017-02-06
Consistent typesetting ES6→ES2015

simon04:


bebraw:
Good after rebase.

simon04:
Rebased.

Issue #814, opened 2017-02-04
Document NormalModuleReplacementPlugin


gonzoyumo:
Add documentation for NormalModuleReplacementPlugin

gonzoyumo:
As suggested by @TheLarkInn I'm willing to help with this. I'll submit a PR soon.

simon04:
Added in 7a6540b via #840.

PR #840, opened 2017-02-06
normal-module-replacement-plugin: add documentation

gonzoyumo:
My humble contribution to that great project. Refs #814 

I tried to keep the same structure used on other plugin documentations but it doesn't look like there is a template to follow. Please advise me if changes are needed.

### DIFF
--- a/content/plugins/normal-module-replacement-plugin.md
+++ b/content/plugins/normal-module-replacement-plugin.md
@@ -5,7 +5,7 @@ contributors:
 ---
 ## Install
 
-The `NormalModuleReplacementPlugin` is a built-in  Webpack plugin. There is nothing to do except requiring webpack.
+The `NormalModuleReplacementPlugin` is a built-in webpack plugin.
 
 
 ## Usage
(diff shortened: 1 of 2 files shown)


content/plugins/normal-module-replacement-plugin.md
```diff
@@ -0,0 +1,88 @@
...
+The `NormalModuleReplacementPlugin` is a built-in  Webpack plugin. There is nothing to do except requiring webpack.
+
+
+## Usage
+
+``` javascript
+new webpack.NormalModuleReplacementPlugin(resourceRegExp, newResource)
+```
+
+The `NormalModuleReplacementPlugin` allows you to replace resources that match `resourceRegExp` with `newResource`. If `newResource` is relative, it is resolved relative to the previous resource. If `newResource` is a function, it is expected to overwrite the **request** attribute of the supplied object.
```
simon04: - s/**request**/`request`/
- s/supplied object/supplied `resource`/

content/plugins/normal-module-replacement-plugin.md
```diff
@@ -0,0 +1,88 @@
...
+      })
+    ]
+  }
+  
+}
+```
+
+Create the two config files:
+
+app/config-VERSION_A.js:
```
simon04: Enclose filenames in a pairs of `**` for bold typesetting.

content/plugins/normal-module-replacement-plugin.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Advanced example
+
+Conditional build depending on an environment var.
+
+Say you want a configuration with specific values for different build targets.
+
+``` javascript
+// Usual config omitted for brevity
+module.exports = env => {
```
simon04: We've mostly stuck to ES5 syntax for webpack config to avoid overloading beginners with all sorts of new things. I guess we can discuss on that. I would prefer ES5 since the config does usually not benefit a lot when using ES2015.

PR #844, opened 2017-02-07
Fixed 404 fail resource in plugins page

garmjs:
Removed `section` variable from `renderer.image` in file markdown.js because it will fail to load resource ( jpg, gif, svg, mp4 )

url with failed resources can be found here.
https://webpack.js.org/plugins/npm-install-webpack-plugin
https://webpack.js.org/plugins/uglifyjs-webpack-plugin/

### Actual Behavior in pages
screen shot 2017-02-07 at 4 24 09 pm

screen shot 2017-02-07 at 4 24 38 pm

screen shot 2017-02-07 at 5 13 01 pm

### DIFF
--- a/utilities/markdown.js
+++ b/utilities/markdown.js
@@ -9,7 +9,7 @@ module.exports = function(section) {
   var renderer = new marked.Renderer();
 
   renderer.image = function(href, title, text) {
-    return '<img src="' + section + href + '" alt="' + text + '">';
+    return '<img src="' + href + '" alt="' + text + '">';
   };
 
   // patch ids (this.options.headerPrefix can be undefined!)


garmjs:
@simon04 review required

simon04:
@bebraw can we drop the `section` part of the href of images?

If yes: the line [`section = …`](https://github.com/garmjs/webpack.js.org/blob/c114914c2ef680ba6648a02683cf233d9f2254de/utilities/markdown.js#L5-L7) plus comment should be removed as well.

bebraw:
@simon04 Yeah, most likely. At least the link integrity check passes so it should be fine. I would still check locally to be sure, though.

garmjs:
@bebraw checked localy everything looks good without `section`

bebraw:
Ok. 👍

Issue #777, opened 2017-01-31
Add a bar to announce new releases/important info


dmitriid:
It would be nice to have a small bar below the top menu for feature/milestone/release/etc. announcements. It may include relevant links to blog posts or to documentation.

A similar feature is implemented on http://www.typescriptlang.org:

![](http://dmitriid.com/i/gm2tknzvgu4dqmzy.png)

The reason is simple: increase visibility of certain types of announcements (especially important new releases and critical fixes).

I found myself a bit intimidated by the new site's structure, so I didn't create a pull request. Esp. because this feature probably needs design and discussion.

skipjack:
I can PR this over the weekend if nobody does sooner.

dmitriid:
Awesome!

skipjack:
Started playing around with something:

![image](https://cloud.githubusercontent.com/assets/8988697/22616661/7f513d86-ea7f-11e6-8a40-8ac5c31597b2.png)

If this looks good I can open a PR. The way I implemented this for now, it would only be displayed on landing page. I could do something similar for all pages (like how TypeScript does it) but it would have to go elsewhere for the landing page as our top navigation menu blends in with the background there.

skipjack:
P.S. the "announcement" link points to @TheLarkInn's [post](https://medium.com/webpack/webpack-2-and-beyond-40520af9067f#.ojp0x5ls1) and "install it" points to [npm](https://www.npmjs.com/package/webpack).

cc @bebraw @TheLarkInn what are your thoughts on this?

bebraw:
The new header looks good.

Btw, we could push all the blog bits to the site itself. That would avoid the pain of linking to Medium etc. It would not be hard to add a section just for that and we could do custom things like tagging easily. A lot what's needed is in place already.

TheLarkInn:
How about for mobile? Would this appear as dismissible top banner above toolbar or down below still. Having at the top feels like a great call to action on mobile.

skipjack:
Yea let's do that site wide -- that is, a similar bar, at the top, mobile-friendly and dismissible. I'll create a PR soon.

skipjack:
Closing... added in #847.

PR #847, opened 2017-02-08
Notification Bar

skipjack:
This is intended to address #777. Added the bar as discussed there...

Mobile:

image

Desktop/laptop:

image

### DIFF
--- a/components/notification-bar/notification-bar-style.scss
+++ b/components/notification-bar/notification-bar-style.scss
@@ -1,0 +1,44 @@
+@import 'functions';
+@import 'mixins';
+
+.notification-bar {
+  color: getColor(white);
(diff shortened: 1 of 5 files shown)


skipjack:
@TheLarkInn adding you as a reviewer as you'll probably be the one updating this component down the road. Basically the flow for updating would go something like this:

1. Update the content
2. Instead of just saving `true` we'd save a "notification version"
3. Done, if a user has seen and dismissed the most recent version it will be saved in `localStorage` and we can determine whether or not to show them the current notification.

Let me know what you think. I'm assuming once v2 has been out for a while you'll want to change this to another call to action as you mentioned in #777.

dmitriid:
This looks very nice!

PR #850, opened 2017-02-08
Fix typos related to rules in migration guide

joshwcomeau:
Hi! Two small typo fixes in the migration guide, related to the new `module.rules` syntax.

### DIFF
--- a/content/guides/migrating.md
+++ b/content/guides/migrating.md
@@ -269,7 +269,7 @@ module: {
     {
       test: /.css$/,
 -      loader: ExtractTextPlugin.extract("style-loader", "css-loader", { publicPath: "/dist" })
-+      loader: ExtractTextPlugin.extract({
++      use: ExtractTextPlugin.extract({
 +        fallback: "style-loader",
 +        use: "css-loader",
 +        publicPath: "/dist"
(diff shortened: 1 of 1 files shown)


content/guides/migrating.md
```diff
@@ -266,13 +266,15 @@ It's no longer necessary to specify it in configuration.
...
--    loader: ExtractTextPlugin.extract("style-loader", "css-loader", { publicPath: "/dist" })
-+    loader: ExtractTextPlugin.extract({
-+      fallback: "style-loader",
-+      use: "css-loader",
-+      publicPath: "/dist"
-+    })
+    {
+      test: /.css$/,
+-      loader: ExtractTextPlugin.extract("style-loader", "css-loader", { publicPath: "/dist" })
++      loader: ExtractTextPlugin.extract({
```
bebraw: `use: ExtractTextPlugin...` would work too and that's the preferable form.

joshwcomeau: is there any syntactic difference except `loader` -> `use`?

I'm new to Webpack 2 so I don't wanna assume.

bebraw: `use` is more powerful and it's the preferred option as we'll likely drop `loader` in a future release. That makes it a good recommendation.

PR #858, opened 2017-02-09
Font anti aliasing

semlette:
This (semi-opinionated) commit adds CSS font anti aliasing on devices that support it.

I added a mixin `fontantialias` that will add anti aliasing properties or not. The mixin is applied to the body element and disabled on code and the footer (the text becomes too small/thin). 

# Comparison
*Open in fullscreen!*

## Before

before

## After

after


SpaceK33z:
Thanks! Tested this on my retina and non-retina screen, especially on non-retina it looks better :).

Issue #178, opened 2016-09-20
Reference Style Links | Re-consider Marked


skipjack:
For some reason [marked](https://github.com/chjj/marked) isn't parsing these correctly (see the [markdown spec](https://daringfireball.net/projects/markdown/syntax#link) if you're not familiar with them). I find them very useful to keep paragraphs uncluttered for easy review and editing. [This issue](https://github.com/chjj/marked/issues/310) says they they should work, however this:

``` markdown
... here's a [link][1] ...

[1]: https://github.com
```

always gets parsed incorrectly to `...here's a [link][1]...`

I'm assigning this to myself for now but anyone's welcome to take a look.

bebraw:
We might want to look into some other parser as marked isn't that maintained. It would have to be possible to customize the alternative, though, as there's certain custom syntax in place.

bdougie:
I have had problems with marked in the past for an app at work, but figured out a workaround. I was unable to find a better solution than marked and curious to know if there is something better.  

Is there an example of reference linking already being used somewhere, if not I can play around with this and look into [remarkable](https://github.com/jonschlinkert/remarkable) or [common mark](https://github.com/jgm/CommonMark)?

bebraw:
The problem is that we've extended the parser with some custom syntax. That would need to be ported over to some other solution. That said, replacing marked with something maintained could be a good move.

skipjack:
Yea I was thinking about closing this as it doesn't seem to bother others too much and I haven't had the time to dig into it myself. However, @bdougie if you want to take a stab at using another parser or debugging this, that would be awesome. As @bebraw mentioned though, I would take a look at the [markdown utility](https://github.com/webpack/webpack.js.org/blob/develop/utilities/markdown.js) we've been using and make sure the other parsers you try out can be customized in a similar fashion.

bdougie:
I first tried Remarkable and found it works fine with reference linking, I did have some trouble trying to match the Markdown Utility customizations and did not proceed with it.

I did find the point of failure in [Marked](https://github.com/chjj/marked/blob/master/lib/marked.js#L635) and it correctly parses that there are reference links, if just can't parse the reference link list, specifically `this.links.length` is always 0. I have been tinkering with MarkedJS code to find out why the list is being overlooked. The fix in this [issue](https://github.com/chjj/marked/issues/310) addresses finding reflinks but not the list of links. 

```raw
  [1]: http://google.com/
  [2]: http://search.yahoo.com/
[...]

PR #875, opened 2017-02-12
Mobile friendly tables

### LINKED ISSUE #246: Fix Table Styling (esp. on mobile)
As mentioned in #244, we need to figure out a way to better handle markdown tables.

I'm pushing a quick fix in 6a95c83 to #225 but if others have better ideas please discuss here.

NejcZdovc:
Added mobile friendly tables.

Resolves #246

### DIFF
--- a/styles/markdown.scss
+++ b/styles/markdown.scss
@@ -121,43 +121,94 @@
     }
   }
 
-  table {
-    display:block;
-    width:100%;
-    overflow:auto;
+  .table {
+    &-wrap {
+      display:block;
+      width:100%;
+      overflow:auto;
+    }
 
-    tr {
-      border-top: 1px solid #cccccc;
+    &-tr {
       background-color: white;
       margin: 0;
       padding: 0;
+      border-left: 1px solid #cccccc;
+      display: block;
 
-      &:nth-child(2n) { background-color: #f8f8f8; }
-
-      th {
-        font-weight: bold;
-        border: 1px solid #cccccc;
-        text-align: left;
-        margin: 0;
-        padding: 6px 13px;
+      @include break {
+        display: flex;
       }
 
-      td {
-        border: 1px solid #cccccc;
-        text-align: left;
-        margin: 0;
-        padding: 6px 13px;
+      &:nth-child(2n) {
+        background-color: #f8f8f8;
+      }
 
-        img {
-          max-width:none;
+      &:last-child {
+        border-bottom: 1px solid #cccccc;
+      }
+    }
+
+    &-th {
+      font-weight: bold;
+      border: 1px solid #cccccc;
+      border-bottom: none;
+      text-align: left;
+      margin: 0;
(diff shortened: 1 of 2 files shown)


NejcZdovc:
@skipjack what do you think?

skipjack:
@NejcZdovc my inclination is to tie in some of the styling changes from #440, then merge this and come back to it more in-depth later on if we ever get to moving away from `marked`. 

I think the cleanest way to implement all these customizations in a react environment is to have something parse the markdown and pass that data to react components for any kind of special behavior. The only projects I have come across that seem to offer this functionality are:

- [markdown-to-react-components](https://github.com/christianalfoni/markdown-to-react-components)
- [reactdown](https://andreypopp.github.io/reactdown/)

My only worry would be trying to pick something that will continue to be maintained.

cc @bebraw @bdougie - interested in both of your thoughts here as well. I know we've discussed this a bit on #178.

skipjack:
@NejcZdovc we should try to delineate the the two "rows" better here, they can blend together a bit.

image

In your screenshot in #246 you had removed the horizontal borders within each row:

image

which would be a good start although I'm not sure if this is tough to implement or not.

NejcZdovc:
@skipjack yes you are correct. I only changed table parser, everything else was copied from an original library. I didn't find a way to override only table renderer so I needed to override `tok` function.

Regarding design, in this stage I only added basic styling, so that if you like the approach we can go one step further. So let's first check if we all agree to like this change and then invest some more time into CSS changes. Personally I like this approach, because it's a lot more readable on mobile and like you said it's not a perfect solution, but I am not aware of a perfect one :smile:

NejcZdovc:
@skipjack I removed borders now for mobile

bdougie:
@skipjack, having these customizations and overrides would be interesting in a React Component. I like the idea of generally making hi-jacks or overrides more declarative than implied, which is what the markdown.js does a lot of. TBH, it took me a bit of time just to get up to speed with the file, but that might just be a limit of my experience.

I also have reservations to adding more options to the markdown.js but only from the experience trying to swap/update marked to a moreup to date and maintained remarkable or showdown.

bebraw:
@bdougie Yup, agreed. Ideally we should be able to push these kind of bits to React side. I would love to replace marked with something actively maintained btw.

That said, it's fine by me if we go with a solution that makes sense from user point of view even if the solution is not technically perfect. So keep up good work. 👍

skipjack:
@NejcZdovc looks much better with the removed borders on mobile... one last question:

image

It seems more than just the table contents is wrapped in `table-wrap` class, is this intentional?

Aside from that I think this is good to go, as you mentioned I'll probably be submitting another PR shortly after we get this one merged to do a bit more styling (somewhat based off of #440).

NejcZdovc:
@skipjack No it was not intentional, I forgot to close `table-wrap`. It's fixed now.

PR #876, opened 2017-02-12
Consistent use of NODE_ENV "production"

tschaub:
This suggests using `"production"` (instead of `"prod"`) everywhere for the value of `NODE_ENV`.


SpaceK33z:
Thanks!

PR #889, opened 2017-02-13
Active arrow toggles the list

mikezaby:
When we have active sidebar, the arrow toggler it does't work, this PR is fixing this issue.
click1

### DIFF
--- a/components/sidebar-item/sidebar-item.jsx
+++ b/components/sidebar-item/sidebar-item.jsx
@@ -6,17 +6,16 @@ export default class SidebarItem extends React.Component {
     super(props);
 
     this.state = {
-      open: false
+      open: this.isOpen(props)
     };
   }
 
   render() {
-    let { index, url, title, anchors = [], currentPage } = this.props;
+    let { index, url, title, anchors = [] } = this.props;
 
     let emptyMod = !anchors.length ? 'sidebar-item--empty' : '';
-    let active = `/${currentPage}` === url;
-    let openMod = (active || this.state.open) ? 'sidebar-item--open' : '';
-    let anchorUrl = (active) ? '#' : url + '#';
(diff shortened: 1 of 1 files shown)


mikezaby:
Hello @skipjack, it seems that something going wrong with this https://github.com/webpack/webpack.js.org/pull/889/commits/4a1714a024eaaa959aab4d8c4e014e5bb7d00365, what is the purpose of merging master in the branch?

skipjack:
@mikezaby what exactly is going wrong? I'm seeing not seeing any issues with CI or your fix... although you are right in this case merging master was probably unnecessary.

mikezaby:
I mean the problem with the merge commit. I'll remove this commit and I'll rebase from master. After that I think that the PR is ready for merge.

skipjack:
@mikezaby ok I'm still not sure what issue you're talking about, but I see no issue with merging it as is. @bebraw is there a reason behind merging master before merging to master (even when there's no conflicts)? I know I can override this with my admin privileges but the way we have the flow makes it seem like it's recommended to do that first:

image

I don't remember this always popping up so I'm not sure if it was github change or something one of us enforced for all PRs.

simon04:
@skipjack, I _guess_ the idea is to trigger another Travis build to verify that all links are set correctly. But sometimes I'm too impatient waiting for the build to finish, or feel certain enough not braking anything and use the admin merge.

skipjack:
@simon04 makes sense, thanks for the clarification. I'm gonna merge this but yea maybe that's something we should discuss more going forward.

@mikezaby if I do that again on another PR then please just consider it a part of our process and leave it as is. If you are seeing it actually cause an issue in the build or in what you've implemented, please be specific as to what the issue is. That said, thanks again for contribution! 

> Side note: I'll review your PR later today (iirc there was a reason I used `parentHeight` instead of `innerHeight` but it doesn't hurt to double check).

PR #892, opened 2017-02-14
Compare sidebar height with window height

mikezaby:
This PR is fixing the issue about the following sidebar when there aren't enough space.

We need to compare the height of sidebar with window inner height instead of parent element.
With innerHeight we check if the sidebar has enough space to follow, the check for the offsetHeight of the parent sidebar element is irrelevant.

Example:
follower

innerHeight: 719px
parent node of sidebar (`<div class='interactive'>`): 3994px

### DIFF
--- a/components/sidebar/sidebar.jsx
+++ b/components/sidebar/sidebar.jsx
@@ -77,14 +77,14 @@ export default class Sidebar extends Component {
     let { scrollY, innerHeight } = window;
     let { scrollHeight } = document.body;
     let { offsetHeight: sidebarHeight } = this._container;
-    let { offsetWidth: parentWidth, offsetHeight: parentHeight } = this._container.parentNode;
+    let { offsetWidth: parentWidth } = this._container.parentNode;
     let headerHeight = document.querySelector('header').offsetHeight;
     let footerHeight = document.querySelector('footer').offsetHeight;
     let distToBottom = scrollHeight - scrollY - innerHeight;
     let availableSpace = innerHeight + distToBottom - footerHeight;
     
     this.setState({ 
-      fixed: scrollY >= headerHeight && sidebarHeight < parentHeight,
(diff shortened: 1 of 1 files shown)


skipjack:
@mikezaby no the `parentHeight` check is what we want. Consider the case where the sidebar outgrows the window on a large page:

image

In this case we still want it to catch on scroll (i.e. `fixed:true`) but if we compare `sidebarHeight` against `innerHeight` this won't be the case. Does your change fix some other bug that I'm not seeing or was it just a refactor?

skipjack:
I'm closing this but please let me know if I missed something (with details on how to replicate).

mikezaby:
The problem is when the sidebar is bigger than window height, it follows on scroll, as a result the last links are not clickable.
fixed_sidebar

mikezaby:
Ok i found some strange behavior in my previous implementation, so i changed to this https://github.com/mikezaby/webpack.js.org/commit/d617257bb8c674df03aaf8e55f57baab2f367d36. I keep the existing implementation and I add an additional check for handling the case that Ι mention in the previous comment.

skipjack:
That's definitely a fair point and something one other person (@bebraw) has mentioned to me as well. I think the way to solve this would be to dynamically set a `max-height` and make the sidebar content scrollable when it outgrows the visible space. We could calculate that space using the `innerHeight` of the window minus any of the footer and header that are currently visible.

The current implementation you have would work ok if you're starting at the top of the page but if you start expanding the sidebar once you've scrolled a bit, it would all of a sudden snap back up to the top. I can share a screenshot tomorrow to demonstrate my point.

mikezaby:
I understand what you mean and I agree. Maybe I will try the concept that you described in the next days.

skipjack:
Cool... I might PR it along with some other things I'm working on tonight. But, if you don't see one in the next few hours, then you'd be more than welcome to take it on. Thanks again for your other contribution and pointing this out.

Issue #401, opened 2016-11-23
Documentation: Passing a Configuration page


SpaceK33z:
https://webpack.js.org/configuration/passing-a-config/

This page is not finished yet.

simon04:
I'm struggling to associate the content of extended configuration definitions with "Passing a Configuration". What his "passed"? Alternative suggestion: "Configuration Types" (relating to Javascript types function, array, Promise).

See also my comment: https://github.com/webpack/webpack.js.org/pull/895#issuecomment-280034635

bebraw:
Configuration types would be better, yeah. Better rename the page, add the `env` bits there and include multi-compiler mode (return an array) too. That covers it.

simon04:
Documenting multi-compiler mode is covered in #898.

It's just that https://webpack.js.org/configuration/environment/#using-environment-variables does not fit into the new title. Maybe keep this section on a separate page titled "environment variables"? Or keep `--env` on both pages?

bebraw:
`Using --env` would fit the suggested page. `Using environment variables` could become a guide. It's out of place there.

simon04:
Good idea. PR #898 is updated.

PR #895, opened 2017-02-15
Config can return a promise

kbariotis:
Config can return a promise documentation added

### DIFF
--- a/content/configuration/passing-a-config.md
+++ b/content/configuration/passing-a-config.md
@@ -4,12 +4,30 @@ sort: 2
 contributors:
   - sokra
   - skipjack
+  - kbariotis
 ---
 
-...
+Besides exporting a single config object, there are a few more ways that cover other needs as well.
 
 ?> exporting a function and --env
 
-?> returning a Promise
+## Returning a Promise
+
+Webpack will run the function exported by the configuration file and wait for a Promise to be returned. Handy when you need to asyncronously load configuration variables.
+
+```js
+const path = require('path');
+
+module.exports = () => {
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


content/configuration/passing-a-config.md
```diff
@@ -4,12 +4,30 @@ sort: 2
...
 
-...
+Besides exporting a single config object, there are a few more ways that cover other needs as well.
 
 ?> exporting a function and --env
 
-?> returning a Promise
+## Returning a Promise
+
+Webpack will run the function exported by the configuration file and wait for a Promise to be returned. Handy when you need to asyncronously load configuration variables.
```
simon04: s/Webpack/webpack/, see https://webpack.js.org/writers-guide/.

content/configuration/passing-a-config.md
```diff
@@ -4,12 +4,30 @@ sort: 2
 contributors:
   - sokra
   - skipjack
+  - kbariotis
 ---
 
-...
+Besides exporting a single config object, there are a few more ways that cover other needs as well.
 
 ?> exporting a function and --env
```
simon04: Since you're already working on this file, can you replace this TODO with a nice link to https://webpack.js.org/configuration/environment/? :smile:

simon04:
Relates to #401.

kbariotis:
@simon04 since there is already the environment page, what should we note in this page? Another example of `--env`?

simon04:
The TODO should be removed since the documentation exists.

I'm unclear whether the docs on "config function / `--env`" should be in /passing-a-config or /environment. I do not find neither title very concise. @kbariotis, @bebraw what's your opinion on that?

PR #897, opened 2017-02-16
Bad `.jsx` test may load the JSON file

a8568730:
When `json-loader` has not been configured, bad `.jsx` test may load the JSON file as a javascript module.
JSON files must be filtered out by the test rule for `.jsx`.

```
test: /\.jsx?/,  // Bad

test: /\.jsx$/,  // Good

test: /\.jsx?$/, // Good
```

### DIFF
--- a/content/guides/migrating.md
+++ b/content/guides/migrating.md
@@ -135,7 +135,7 @@ file with the [`json-loader`](https://github.com/webpack/json-loader).
   module: {
     rules: [
 -     {
--       test: /\.json/,
+-       test: /\.json$/,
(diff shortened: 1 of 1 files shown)


simon04:
Why is this important to consider during the migration from webpack 1 to 2?

As I understand it, using `test: /\.jsx?/` also caused problems in webpack 1. Is this a common pitfall? If yes, it might go to a troubleshooting page. If no, it might better go to a personal blog post somewhere else.

bebraw:
Can you give an example of a bad JSX file?

PR #915, opened 2017-02-19
Added var webpack = require("webpack");

joetidee:
Added `var webpack = require("webpack");` to the example webpack.config.js


simon04:
Why is this import necessary? We do not use the variable `webpack` afterwards?

joetidee:
Used in the 'plugins' block.

simon04:
The `content/guides/get-started.md` guide does not use any plugins. It might be true for another guide or an arbitrary webpack config somewhere else.

PR #917, opened 2017-02-20
Fixed CSS that breaks tutorial

leebradley:
The purpose of this line is to allow copying the entire example into code and conveniently also denote the appropriate file name.

Unfortunately, a single-line comment here breaks the next CSS selector and thus post-css aspect of the tutorial. Running the tutorial as-is, the header appears to ignore all styles in `.app`.

Single-line comments are not supported by run-of-the-mill CSS. Replace with a multiline comment.

### DIFF
--- a/content/guides/hmr-react.md
+++ b/content/guides/hmr-react.md
@@ -211,7 +211,7 @@ export default App;
 ```
 
 ```css
-// ./src/components/App.css
+/* ./src/components/App.css */
 .app {
     text-size-adjust: none;
     font-family: helvetica, arial, sans-serif;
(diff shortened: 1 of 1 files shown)


bebraw:
It would be even better to do it like this:

**./src/components/App.css**

```css
... css goes here ...
```

If you could port the entire article to this scheme, that would be great.

leebradley:
@bebraw I agree. I found this is also the format used in the [Getting Started](https://webpack.js.org/guides/get-started/) tutorial.

The latest commit includes some formatting changes to make the page fall in line with the general layout of that tutorial.

bebraw:
Thanks. 👍

Issue #911, opened 2017-02-17
Table display: rows not aligned


simon04:
The rows are not aligned. Tested /w latest Firefox, Chrome.

![2017-02-17-195529_1440x900_scrot](https://cloud.githubusercontent.com/assets/782446/23079023/50ae2a58-f54b-11e6-8c90-17cdf0d37bca.png)

skipjack:
@NejcZdovc can you look into this?

Dajust:
Hi @skipjack, @NejcZdovc,

I'm currently looking into this issues and will love to send a PR if I'm allowed. 

### Mobile First

![webpack js org-guides-comparison- iphone 5](https://cloud.githubusercontent.com/assets/8015514/23102177/a3ddadb6-f6a4-11e6-9462-2320d542bbb7.png)

**Issue**: On screen below 400px, the class `table-td-content` and `table-td-title` collides/overlaps. 
Solution:  Since the parent of both classes is` displayed:flex,` instead of `display:block `on both classes, we could simply utilize the flex property on t`able-td-content`, give it a fixed with `flex:0 0 150`, then progressively adjust the `flex-grow` to 40%. This will prevent them from ever overlapping, even on a screen below 320px, at the same time shrink/grow as needed.
### Desktop
The major course of the dis-aligned table appears to be the <code> element inside each table-td-content. Since the` <code>` is rightly styled `white-space:nowrap`, it forces the `table-td-content` to grow beyond its width.
To fix this, we must first turn off the flex property, then, give all `table-td-content` fixed width. Then set the `overflow-x` to auto. This is the image of how it looks like after edit.
![table-layout-fix-2](https://cloud.githubusercontent.com/assets/8015514/23102130/a5b2c9c4-f6a3-11e6-9bec-3ce9d57daae0.png)

![table-layout-fix](https://cloud.githubusercontent.com/assets/8015514/23102135/ca7bb5d6-f6a3-11e6-83b6-a864d5836295.png)

Note we can further customize the scroll-bar for chrome users :stuck_out_tongue_winking_eye:

NejcZdovc:
@Dajust I am already working on some fixes, so let me finish it first and let's talk afterwords.

PR #928, opened 2017-02-21
Added scroll table for desktop

NejcZdovc:
Based on #911 I added basic table css for desktop version. 

Because the comparison table is just to big I added option to display scroll for table on desktop and then mobile version for mobile devices.

### DIFF
--- a/content/guides/comparison.md
+++ b/content/guides/comparison.md
@@ -10,8 +10,6 @@ contributors:
 
 webpack is not the only module bundler out there. If you are choosing between using webpack or any of the bundlers below, here is a feature-by-feature comparison on how webpack fares against the current competition.
 
-<div class="scroll-table">
-
 | Feature | webpack/webpack | jrburke/requirejs | substack/node-browserify | jspm/jspm-cli | rollup/rollup | brunch/brunch |
 |---------|-----------------|-------------------|--------------------------|---------------|---------------|---------------|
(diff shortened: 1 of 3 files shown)


skipjack:
@NejcZdovc aside from my two review comments I did test and this does seem to resolve the issue from #911.

NejcZdovc:
@skipjack I don't see your review comments

skipjack:
@NejcZdovc sorry looks like I forgot to submit, you should be able to see them now.

content/guides/comparison.md
```diff
@@ -10,6 +10,8 @@ contributors:
 
 webpack is not the only module bundler out there. If you are choosing between using webpack or any of the bundlers below, here is a feature-by-feature comparison on how webpack fares against the current competition.
 
+<div class="scroll-table">
```
NejcZdovc: Yes we can remove that and scroller to all tables on desktop.

styles/markdown.scss
```diff
@@ -121,11 +121,23 @@
     }
   }
 
+  .scroll-table {
+    @include break {
+      overflow-x: auto;
```
NejcZdovc: Answered above.

content/guides/comparison.md
skipjack: Does it change anything to apply it to all? I'm not saying this is necessarily the best approach, it just seems like applying the same styles everywhere wouldn't change or hurt the styling of the other tables.

NejcZdovc: I think it will not change anything. But I would need to test it out to be sure

skipjack: Ok let me know once you've pushed the changes, I'll test again and we should be good to go.

NejcZdovc:
@skipjack added overflow to all tables for desktop and looks good

PR #937, opened 2017-02-22
Add documentation for `overlay` in devServer

SpaceK33z:
In webpack-dev-server 2.3.0, `overlay` was introduced. I forgot to document this, but better late than never.

### DIFF
--- a/content/configuration/dev-server.md
+++ b/content/configuration/dev-server.md
@@ -259,6 +259,24 @@ With `noInfo` enabled, messages like the webpack bundle information that is show
 noInfo: true
 ```
 
+## `devServer.overlay`
+
+`boolean` `object`
+
+Shows a full-screen overlay in the browser when there are compiler errors or warnings. Disabled by default. If you want to show only compiler errors:
+
+```js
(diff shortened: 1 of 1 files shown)


bebraw:
I wonder if it would make sense to push webpack-dev-server docs to its readme and consume from there. That would give higher quality docs for npm users too. Most likely we still need some heavier docs to support that, but it could contain the basic API at least.

SpaceK33z:
Yeah that's probably a good idea. At the moment WDS doesn't really have docs in the readme, but we could easily move those to there.

A small problem is that WDS inherits a lot of options from webpack-dev-middleware. It would be weird to document them again in the WDS readme, but maybe we could solve this by making a kind of Table of Contents and list the WDS + WDM options in there. For the WDM options, it would link to the WDM readme instead of an anchor tag.

bebraw:
If we do a `WebpackDevServerPlugin`, that might help with the latter concern. It could also tell that webpack-dev-middleware should become a package of its own (monorepo below wds?). But it's basically an organization issue.

PR #941, opened 2017-02-22
Minor wording change

shogunsea:
`but` definition: "used to introduce something contrasting with what has already been mentioned."

It doesn't seem to me the sentence "Webpack will yell at you when doing this incorrectly" is adversative of previous lines.


shogunsea:
@simon04 good point, thanks for pointing out

PR #945, opened 2017-02-23
Update module-resolution.md

adam-beck:
Fixes small grammatical error

### DIFF
--- a/content/concepts/module-resolution.md
+++ b/content/concepts/module-resolution.md
@@ -16,7 +16,7 @@ require('path/to/module')
 ```
 
 The dependency module can be from the application code or a third party library. The resolver helps
-webpack finds the module code that needs to be included in the bundle for every such `require`/`import` statement.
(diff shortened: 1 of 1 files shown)


simon04:
"webpack" is a singular noun. Thus "finds" is correct.

adam-beck:
@simon04 I agree that "webpack" is a singular noun. However, `find` is still the correct word to use here. Unfortunaley I don't know the exact name of the rule or why it is that way.

```
Jack helps Jill tests the code

 **VS**

Jack helps Jill test the code
```

adam-beck:
@simon04 It's called an infinitive. I found an article but unfortunately I can't link exactly where at on the page since it isn't using #ids.

http://www.chompchomp.com/terms/infinitive.htm

Then scroll to: **Recognize an infinitive even when it is missing the to**

simon04:
Sorry, I overlooked your comment since the ticket was closed. Due to majority vote, I'll merge this one. 😉

PR #948, opened 2017-02-23
docs(loaders): Clarify how to include CSS modules and fix grammar. 

maecapozzi:
This clarification is necessary because users might try to copy and paste the below code, and will run into bugs that do not have an obvious solution. If users set `modules: true` without meaning to, they will run into an issue where the classes will be replaced with generated ones, but this isn't made clear by the documentation. 

This PR also fixes some grammar mistakes for clarity.

### DIFF
--- a/content/concepts/loaders.md
+++ b/content/concepts/loaders.md
@@ -60,17 +60,17 @@ There are three ways to use loaders in your application:
 ### Via `webpack.config.js`
 
 [`module.rules`](https://webpack.js.org/configuration/module/#module-rules) allows you to specify several loaders within your webpack configuration.
-This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader. You can find documentation for `css-loader` [here](/loaders/css-loader), and documentation for `style-loader` [here](/loaders/style-loader)  
+This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader.
 
-```js
+```js-with-links-with-details
   module: {
     rules: [
       {
         test: /\.css$/,
         use: [
-          { loader: 'style-loader'},
+          { loader: ['style-loader'](/loaders/style-loader)},
           {
-            loader: 'css-loader',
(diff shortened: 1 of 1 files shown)


content/concepts/loaders.md
```diff
@@ -60,8 +60,7 @@ There are three ways to use loaders in your application:
 ### Via `webpack.config.js`
 
 [`module.rules`](https://webpack.js.org/configuration/module/#module-rules) allows you to specify several loaders within your webpack configuration.
-This is a concise way to display loaders, and helps to have clean code as
-well as you have a full overview of each respective loader.
+This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader. If you would like to enable css modules, set `modules: true.` Otherwise, be sure to set `modules: false`.
```
simon04: I'd rather keep the text in the concepts part as generic as possible. `modules` is specific to the `css-loader` and documented at https://webpack.js.org/loaders/css-loader/.
I'm fine with the grammatical change.

maecapozzi: Hi, @simon04! Is it possible to link to [the css-loader documentation](https://webpack.js.org/loaders/css-loader/) here? Something like..."You can find more details regarding the `css-loader` [here](https://webpack.js.org/loaders/css-loader/)."  

I think the specification would add clarity because developers that copy and paste the code below might find themselves running into an issue where they are accidentally loading a CSS module without meaning to. 

[Here](http://stackoverflow.com/questions/42402151/when-loading-css-with-webpack-i-can-write-css-for-tags-but-not-for-classes-or) is a stack overflow question that I asked that illustrates my point.

simon04: There are already two links in the [example section above](https://webpack.js.org/concepts/loaders/#example). However, I do not object a third link in the other section if that helps. For consistency, please also add a link for `style-loader` in the same example.

maecapozzi: Let me know if these changes better fit what you are describing!

PR #953, opened 2017-02-24
Make grammatical fixes to index.md

maecapozzi:
Hello! These are just a couple of grammatical fixes I came across as I was reading your documentation. Please let me know if making these fixes is helpful to you, and I will continue to do so. Otherwise, I will stop!

### DIFF
--- a/content/concepts/index.md
+++ b/content/concepts/index.md
@@ -34,7 +34,7 @@ There are multiple ways to declare your `entry` property that are specific to yo
 
 ## Output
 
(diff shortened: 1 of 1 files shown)


simon04:
Any improvement is appreciated! :+1: 
Should we stick to either "you" or "we" in this file (now both forms are mixed)?

maecapozzi:
Good point. I think "you" is better. Would it be helpful if I made that fix for you?

simon04:
Sure, go ahead :smile:

PR #963, opened 2017-02-27
remove unnecessary comma

monkindey:


simon04:
Trailing commas are often used with intent: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

Therefore I'm reluctant removing it.

bebraw:
I think the underlying problem is that we don't lint the examples. Either is fine by me. It's easier to manage diffs with trailing commas, though.

monkindey:
Thanks for your reply, I learn a lot from this conversation. There are the good point using the trailing commas. I just be used to clear the trailing commas to be consistent with the JSON syntax.

bebraw:
Yeah, no way to avoid them in vanilla JSON. 😞

simon04:
Let's keep it as it is.

PR #968, opened 2017-03-01
Fix CLI sample mistake.

potato4d:
I found probrem in the loader sample.
This sample was allowed in webpack 1.x, but there was a sample that was not allowed in 2.x.


simon04:
Good catch! Would you please also update the example [here](https://webpack.js.org/api/cli/#module-options) (column "usage")?

potato4d:
@simon04
Okay, I finished it.

PR #969, opened 2017-03-01
Clean Up and Fix Error in TypeScript Guide

kkamali:
Submitting a correction to the webpack and typescript guide. Saw an error in the configuration which includes the source map loader. The configuration wasn't running anything through the typescript loader, which throws errors when you introduce typescript specific features to our code. To address this error - we run our code through the ts-loader and get js code that can go through the webpack bundler correctly.


simon04:
This does not look right. The section you are editing is called "Enabling source maps" and the text below the code also refers to installing/using `source-map-loader`.

Maybe @bebraw or @morsdyce can comment on that.

bebraw:
Yeah, the original text is correct. It expects you have configured TypeScript already. Maybe that could be made clearer, though.

nitishdayal:
Not sure I understand why this closed; the config provided in the documentation under the sourcemaps section doesn't work; it doesn't provide any loader for TypeScript, so while it might work with .ts files without any TypeScript specific syntax (ie; types), as soon as types or interfaces or anything of the sort are introduced it fails.

Config provided by @kkamali generates a valid bundle + has functioning source map configuration. 

Demo repo: https://github.com/nitishdayal/wbts . Master branch is config as recommended by docs, branch kkamali is config as recommended by @kkamali.

skipjack:
@dmitriid could you take a look at this one as I know you have some experience with typescript. It does look like something about this guide is off... it looks like maybe it was intended to be 

``` js
use: [ 
  'source-map-loader',
  'ts-loader'
]
```

but it seems correct the the `ts-loader` would be necessary somewhere in there to load `.tsx` files.

skipjack:
@kkamali @nitishdayal thanks for bringing this up btw and sorry we haven't gotten to it earlier. I don't think we have too many contributors/maintainers using typescript regularly so this guide likely hasn't gotten as much attention as it might need.

dmitriid:
@skipjack Looks like you are right.

skipjack:
@kkamali would you be willing to test that and update the pr? i.e.

``` js
use: [ 
  'source-map-loader',
  'ts-loader'
]
```

The reason for this is to @simon04's point:

> The section you are editing is called "Enabling source maps" and the text below the code also refers to installing/using source-map-loader.

So I do think this section should contain source-mapping documentation. Also I'm a little unsure as to what that `pre` rule is there for before the `.tsx` rule:

``` js
     {
       enforce: 'pre',
       test: /\.js$/,
       loader: "source-map-loader"
     },
```

Is that supposed to somehow take the output of the `ts-loader` and source map it? Otherwise, if it's not typescript related, it can probably go for simplicity sake.

skipjack:
@morsdyce we're you the original author of this guide? Any thoughts on all this?

morsdyce:
Not mentioning a ts loader in the sourcemap section was probably a mistake when preparing isolated examples or I just didn't wanted to repeat the same configuration over and over (I really can't remember at this point).

the source-map-loader is used to load the typescript generated sourcemaps into webpack so they can display properly in the browser, without this loader the sourcemaps from typescript do not work well in my experience.

Apart from these comments I can't help allot anymore as I've stopped using TypeScript so i'm not up to date with things that may have changed since then.

skipjack:
It seems to me that based on this pr and #1257, the whole guide can be simplified and the __Enabling Source Maps__ section can either be removed or simplified.

skipjack:
Updated this PR based on my last comment and cleaned up much of rest of the guide as well.

PR #976, opened 2017-03-05
chore - Add a stub for contrib docs

bebraw:


bebraw:
Just got a better idea... We can pull this info from a separate repo.

PR #977, opened 2017-03-05
chore - Update yarn dependencies

bebraw:


bebraw:
I had to disable Yarn cache (bugged for some reason). I still have to add a relative check for links since some will fail always.

PR #987, opened 2017-03-06
add meta tag description to website

raspo:
Added the description meta tag to the `<head>` section.
The content is taken from the old webpack website ([webpack.github.io](webpack.github.io)), feel free to suggest edits.

When googling "webpack" I currently get this:
screen shot 2017-03-06 at 11 08 47 am
This change should hopefully fix that problem.

### DIFF
--- a/template.ejs
+++ b/template.ejs
@@ -3,6 +3,8 @@
   <head>
     <meta charset="UTF-8">
     <title><%= webpackConfig.template.title %></title>
+    <meta name="description" content="webpack is a module bundler. It packs CommonJs/AMD modules i. e. for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand.">
+
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="theme-color" content="#2B3A42">
     <link rel="shortcut icon" href="/assets/favicon.ico">
--- a/template.ejs
+++ b/template.ejs
@@ -3,8 +3,7 @@
   <head>
     <meta charset="UTF-8">
     <title><%= webpackConfig.template.title %></title>
-    <meta name="description" content="webpack is a module bundler. It packs CommonJs/AMD modules i. e. for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand.">
-
+    <meta name="description" content="<%= webpackConfig.template.description %>">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="theme-color" content="#2B3A42">
     <link rel="shortcut icon" href="/assets/favicon.ico">
--- a/antwar.config.js
+++ b/antwar.config.js
@@ -7,6 +7,7 @@ var highlight = require('./utilities/highlight');
 module.exports = {
   template: {
     title: 'webpack',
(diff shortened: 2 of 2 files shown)


template.ejs
```diff
@@ -3,6 +3,8 @@
   <head>
     <meta charset="UTF-8">
     <title><%= webpackConfig.template.title %></title>
+    <meta name="description" content="webpack is a module bundler. It packs CommonJs/AMD modules i. e. for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand.">
+
```
simon04: Please drop the newline.

template.ejs
```diff
@@ -3,6 +3,8 @@
   <head>
     <meta charset="UTF-8">
     <title><%= webpackConfig.template.title %></title>
+    <meta name="description" content="webpack is a module bundler. It packs CommonJs/AMD modules i. e. for the browser. Allows to split your codebase into multiple bundles, which can be loaded on demand.">
```
simon04: * CommonJs/AMD modules → ES2015/CommonJS/AMD modules
* "i. e." does not make any sense here?
* Maybe also mention "loaders" for CSS, Images, JSON, LESS, …? (see project summary on https://github.com/webpack/webpack)

bebraw:
There might be a better way. If I remember right, it should be possible to map meta from a page headmatter/jsx to a template. I'll check that out and get back to you tomorrow.

template.ejs
raspo: What about the intro found in the readme?
> webpack is a module bundler. Its main purpose is to bundle JavaScript files for usage in a browser, yet it is also capable of transforming, bundling, or packaging just about any resource or asset.

raspo:
@bebraw [antwar.config.js](https://github.com/webpack/webpack.js.org/blob/master/antwar.config.js#L9) maybe?

template.ejs
bebraw: This should basically become `<meta name="description" content="<%= webpackConfig.template.description %>">`. The setup should be able to pick it up from YAML headmatter if you include descriptions there.

raspo:
Thanks for the feedback. 
I have updated the PR to pull the description from `antwar.config.js`.
Changed the text to be the same as the description found inside the [readme](https://github.com/webpack/webpack#webpack).

bebraw:
Ok, cool. I have to inspect this locally (most likely tomorrow) to see the exact result. You should be able to adjust this below `content` as well through the headmatter at each page.

bebraw:
Thanks.

If you want to work on this further, consider the following:

* Document how to write meta descriptions at the writer's guide.
* Start adding meta to the remaining pages. A good fallback would be to use `preview` shortcut like this:

**antwar.config.js**

```javascript
...

function processPage() {
  return {
    ...
    preview: function(o) {
      var previewLimit = 300;
      var content = o.file.__content.split('##')[0].split('\n').slice(1).join('\n');
      var stripped = removeMd(content);

      if(stripped.length > previewLimit) {
        return stripped.substr(0, previewLimit) + '…';
      }

      return stripped;
    }
  };
}
```

You may have to tweak this, but that's the basic idea. `removeMd` comes from `remove-markdown` package.

You probably should also push the current description from Antwar config to `content/index.md` YAML headmatter as the system picks it up from there.

PR #988, opened 2017-03-06
code-splitting: expand Promise polyfill

simon04:


bebraw:
It would be better to have that info in one place and link there. Otherwise it will get out of sync eventually. Maybe the documents should go into one (needs redirect)?

simon04:
Good point. I removed the docs in the `require.ensure` guide and added a link.

bebraw:
Ok. The only tweak I can think of would be to merge **code-splitting-import** and **code-splitting-require** as a **code-splitting** doc. The advantage of doing this is that then you can discuss the differences in one place. You would have to reroute both to the resulting document.

simon04:
Good point. I opened issue #992 for that since this goes beyond the scope of this PR.

bebraw:
Ok. Let's merge this now. If you want, do another PR later.

PR #993, opened 2017-03-07
Add Rule.resourceQuery property to module.md

jhnns:
The `Rule.resourceQuery` property can be used to match different loader pipelines depending on the query string, like this: https://github.com/webpack/loader-utils/issues/69#issuecomment-284725071

---

@sokra @bebraw Was this omitted intentionally? Is it ok to document it as a public API?

### DIFF
--- a/content/configuration/module.md
+++ b/content/configuration/module.md
@@ -5,6 +5,7 @@ contributors:
   - sokra
   - skipjack
   - jouni-kantola
+  - jhnns
 ---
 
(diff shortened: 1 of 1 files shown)


bebraw:
It's in the public API I think. You can also find mentions in the site elsewhere.

jhnns:
Yes, here for example: https://webpack.js.org/guides/migrating/#advanced-loader-matching

PR #999, opened 2017-03-08
add note about import() paths

kennetpostigo:
cc @TheLarkInn

### DIFF
--- a/content/guides/code-splitting-import.md
+++ b/content/guides/code-splitting-import.md
@@ -26,6 +26,7 @@ function determineDate() {
 
 determineDate();
 ```
+T> Keep in mind that `import([path])` path cannot be fully dynamic. Rather either completely static or partially static.
 
 ## Promise polyfill
 
--- a/content/guides/code-splitting-import.md
+++ b/content/guides/code-splitting-import.md
@@ -26,7 +26,7 @@ function determineDate() {
 
 determineDate();
 ```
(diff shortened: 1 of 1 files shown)


content/guides/code-splitting-import.md
```diff
@@ -26,6 +26,7 @@ function determineDate() {
 
 determineDate();
 ```
+T> Keep in mind that `import([path])` path cannot be fully dynamic. Rather either completely static or partially static.
```
simon04: Please give examples for 
* fully dynamic: ?
* partially static: `import('./locale/' + language + '.json')`
* completely static: `import('./locale/de.json')`

TheLarkInn: Oh yes that's a great idea. I like having clear examples of them like this.

kennetpostigo: Okay do I add those examples into the same info block?

simon04: Maybe this way:
> Keep in mind that `import()` path cannot be fully dynamic (e.g., `import(Math.random())`). Rather either completely static (e.g., `import('./locale/de.json')`) or partially static (e.g., `import('./locale/' + language + '.json')`).

Maybe you can think of an example from real life instead of `Math.random()` :smile:

kennetpostigo:
Added the examples you suggested @simon04 . I think this is good now?

PR #1002, opened 2017-03-09
Fix a few typos

selbekk:
Walked through this guide as I was porting a project from webpack 1 to 2 - and spotted a few typos.

### DIFF
--- a/content/guides/migrating.md
+++ b/content/guides/migrating.md
@@ -77,9 +77,9 @@ The new naming conventions are easier to understand and are a good reason to upg
 
 ## Chaining loaders
 
-Like in webpack v1, loaders can be chained to pass results from loader to loader. Using the [rule.use](/configuration/module#rule-use)
+Like in webpack 1, loaders can be chained to pass results from loader to loader. Using the [rule.use](/configuration/module#rule-use)
  configuration option, `use` can be set to an array of loaders.
-In webpack v1, loaders were commonly chained with `!`. This style is only supported using the legacy option `module.loaders`.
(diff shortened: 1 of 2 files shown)


simon04:
Thanks for the PR. Looks good, but I'd stick with `webpack 1` instead of `webpack v1` since it is far more common (using [ag](https://github.com/ggreer/the_silver_searcher)):
```
$ ag 'webpack [123]' | wc -l 
30
$ ag 'webpack v[123]' | wc -l 
3
```

selbekk:
Ah good point, just looked at the one document - I'll correct it tomorrow 😊

simon04:
Please rather get rid of the 3 instances of `webpack v[123]` than changing 30 the other way. Feel free to open a separate PR on that. Thanks.

selbekk:
There you go - a global search replace for `webpack v[123] -> webpack [123]`

selbekk:
Not sure about this CLA issue - I pushed from my work computer (instead of my personal computer as I did yesterday), and even though I've now added that email to my github-account it still seems slightly broken.

simon04:
Please fix the CLA check. Your commits are done with different mail addresses, see the `From:` lines in https://patch-diff.githubusercontent.com/raw/webpack/webpack.js.org/pull/1002.patch

selbekk:
@simon04 - I've added the email in question to my user profile, and then I revoked the access tokens for the js.foundation CLA app before logging in again - but I can't figure out how to re-sign it with my new email.

Any tips? =)

simon04:
I rebased your commit following https://stackoverflow.com/a/750182. After the Travis build we should be set.

PR #1003, opened 2017-03-10
Rework loader api documentation

jhnns:
- Remove deprecated features like `this.options` and `this._compilation` that make it heard to implement a multi-process pipeline
- Add `this.callback`
- Add `this.async`
- Add some explanatory comments
- Prepend `this` to all properties to make it obvious that these properties are accessible via `this` inside the loader

@sokra do you agree with these changes?

### DIFF
--- a/content/api/loaders.md
+++ b/content/api/loaders.md
@@ -7,18 +7,13 @@ contributors:
 
 ---
 
-Loaders allow you to preprocess files as you `require()` or “load” them. Loaders are kind of like “tasks” in other build tools,
-and provide a powerful way to handle front-end build steps. Loaders can transform files from a different language (like CoffeeScript to JavaScript), or inline images as data URLs. Loaders even allow you to do things like `require()` css files right in your JavaScript!
-
-To tell webpack to transform a module with a loader, you can specify the loader in the webpack [configuration](/configuration) file (preferred) or in the module request, such as in a `require()` call.
-
-?> When /concepts/loaders merges, we should link to the many usages of loaders found there (require vs configuration) from this page.
+Loaders are transformations that are applied on the source code of a module. They are functions (running in Node.js) that take the source of a resource file as the parameter and return the new source.
 
 ## How to write a loader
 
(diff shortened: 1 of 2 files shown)


content/api/loaders.md
```diff
@@ -93,33 +104,60 @@ In `/abc/file.js`:
...
+2. The second argument a `string` or a [`Buffer`](https://nodejs.org/api/buffer.html).
+3. Optional: The third argument must be a source map that is parsable by [this module](https://github.com/mozilla/source-map).
+4. Optional: `AST` can be an Abstract Syntax Tree of the given language, like [`ESTree`](https://github.com/estree/estree). This value is ignored by webpack itself, but useful to speed up the build time if you want to share common ASTs between loaders.
+
+In case this function is called, you should return undefined to avoid ambigious loader results.
+
+### `this.async`
 
-### `data`
+A function than returns either `callback` or a falsy value depending whether asynchronous loader results are supported by the loader pipeline or not.
```
jhnns: @TheLarkInn Native speakers: Do you say `depending on whether...` or can `on` be omitted? 😁

bebraw: There's actually a typo at `than`.

I would invert the problem and go with something like this:

The results of `this.async` depend on the loader pipeline. If the pipeline supports asynchronous behavior, it returns a callback. Otherwise it returns a falsy value.

---

Then refer to that loader pipeline idea so people can check out what that means. Now there isn't enough context. Ideally you would have two different examples too to cover all the bases.

content/api/loaders.md
```diff
@@ -129,14 +167,14 @@ By default, loader results are cacheable. Call this method passing `false` to ma
 
 A cacheable loader must have a deterministic result, when inputs and dependencies haven't changed. This means the loader shouldn't have other dependencies than specified with `this.addDependency`. Most loaders are deterministic and cacheable.
```
bebraw: `this.cacheable` is set by default now. That might be worth mentioning. Check with Tobias.

jhnns:
Thanks for the feedback @bebraw I will add your requested changes.

jhnns:
Ok, this is ready for merge now (you should squash and merge 😉)

Things I've done:

- Address requested changes
- Move the introduction paragraph from loader API documentation to the loader concepts. The `content/api/loader.md` page is intended for loader authors that are already familiar with the concepts. I felt like the introduction was targeted at webpack beginners.
- Remove `this.exec`
- Add `this.loadModule`
- Move all removed stuff into a deprecated section. This way, it's still documented but discouraged.

PR #1009, opened 2017-03-13
grammatical suggestion to loaders.md

shailendher:
Fixes grammatical error.
from:

> In most cases **you** will be loaders from

to:

> In most cases **it** will be loaders from

Alternative option:

> In most cases you will be **loading** from..


bebraw:
👍

PR #1012, opened 2017-03-14
Fix module-bind usage

yuffiy:
```
webpack --module-bind /\.js$/=babel-loader
```

`test` value will be `/\.\/\\\.js\$\/$/`


yuffiy:
@bebraw 

Hi, has something module `exclude` or `include` option support plan about cli ?

bebraw:
It's better to bring that up at [webpack-cli](https://github.com/webpack/webpack-cli).

yuffiy:
Yeap, thanks

PR #1016, opened 2017-03-14
Update shimming.md

cerlestes:
Add example of ProvidePlugin's capability of providing certain children of a module by configuring it with an array instead of a string.


scottdj92:
Looks good.

bebraw:
Can you fix the typo? Good to go after that.

Issue #812, opened 2017-02-03
Changelog


aight8:
I like to read more about webpack and how it's internal works.
But because the docs are in heavy development I never know if some new docs are added.
Is there a way to make some notes in a changelog file or so some bigger changes?

SpaceK33z:
You could read the commits? I don't think we'll ever have time to make a changelog for docs...

aight8:
Yeah sure I can read the commits on every repo ;) but okey, I understand it that this would be to much extra work than

rouzbeh84:
Is there a tagging structure or minimum required amount of updates that would involve a full change log? Perhaps chunking issues into block of 10 or so that are tagged as 'major' or 'significant' updates or possible even just a 'last updated' auto block that goes under the title block and for sections that involve `loaders || modules || plugin` with a 'last updated' comparison next to one another? 

For example: 

[css-loader](/plugins/commons-chunk-plugin/)

Latest Updates:

| This Page | [loader](https://www.npmjs.com/package/css-loader) |
|--- |--- | 
| [date] | [date] | 

Although, I do see how this could become quite a mess to maintain but just throwing it out there...

skipjack:
Yea idk about this one, @rouzbeh84 something like that might not be bad but, as @SpaceK33z mentioned, there's a ton of other work to be done that I would say is much higher priority than this. To get "high-level" overview of the most recent stuff going on in the docs I would say options include:

* Watching this repository
* Subscribing to issues that concern some piece of documentation you're waiting on
* Checking the most recent commits

You can also use the github/git history of an individual file such as [this one](https://github.com/webpack/webpack.js.org/commits/master/content/concepts/entry-points.md) to get an idea of what's changed in any particular place. 

Personally, I also wouldn't be opposed to standardizing my commits to this repository and using some tool to auto-generate a changelog but that might be fairly hard to manage as we have many different contributors opening all different kinds of PRs. 

I'm going to close this for now but maybe it's something we can loop back to down the road.

PR #1017, opened 2017-03-14
feat(automatic-release): Generation of automatic release

wardpeet:


bebraw:
We don't do official releases for the docs. How would this work exactly?

bebraw:
Yeah, I understand the point of changelog generation. It's just that we don't bump the version of the site ever.

wardpeet:
I can change the package to use [conventional-changelog-cli](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli) so it only generates a changelog without bumping the npm version.

Or if you're not interested than just close the PR 👍. Just added this to all repos 😄

skipjack:
I think this is interesting... I'm about to start using standard-version/conventional-changelog on a few other projects. That said though I'm 100% not sure it would work here, some thoughts:

- We have so many contributors that it would probably be next to impossible to get everyone to follow the [conventional-changelog][1] standard. 
- As @bebraw mentioned we're not versioning now but it may be worthwhile to try to keep the version of these in sync with the current webpack version (or maybe just with majors and minors). However to do this I think we'd really have to get caught up with all the documentation issues that relate directly to webpack and webpack-dev-server. The plugin/loader documenation should be ok (although we probably should pull in the package version to each of those pages as well).
- Not sure how useful a changelog on the docs would be. We did have one other person bring this up in an issue (#812). I feel like the more important thing would be to make sure that this site is as up to date as possible with the current webpack release.

I'd be interested to see some changelogs for other large documentation sites if anyone has an example.

[1]: https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md

wardpeet:
the thing with PR is you can rewrite the commit message when you're going to merge so you can fix commits that way but it extra work for the maintainers.

I do get the point that it doesn't make much sense for documentation but never the less it might be handy to see how the repo is changing without having to go through each commit message?

skipjack:
@wardpeet @bebraw I'm going to close this for now as I think it might still be a little ways off before we implement. However I think we should...

- continue trying to follow the commit convention
- consider adding [conventional-changelog-lint][1] to enforce it
- start doing "squash and merge" for PRs that don't conform
- implement this in the new [voting-app repo][2]

@bebraw feel free to re-open if you think I'm being too hasty here, I think we should implement this but first see if we can follow the convention for a bit. Also how would we version this repo? According to commits and changes or maybe have it trail the current webpack version (e.g. once a new webpack release has been fully documented, cut a release here as well with the same version #).

[1]: https://github.com/marionebl/conventional-changelog-lint
[2]: https://github.com/webpack-contrib/voting-app

bebraw:
Most likely we would sync releases with webpack versioning. Done right, this would be a chance to write archives for older versions online (little gh-pages script).

The hardest part is forcing all commits to follow a certain convention. That rules out casual commits in PRs but that's not necessarily a bad thing.

skipjack:
> Most likely we would sync releases with webpack versioning. Done right, this would be a chance to write archives for older versions online (little gh-pages script).

Yea I was thinking the same thing. In my mind the tasks we need to accomplish before this can happen are:

- Knock out the majority of open documentation issues so the repo is actually caught up with the current webpack version.
- Figure out what to do with loaders and plugins. These either could just be the latest version, or could have version dropdowns as well. Not saying we should do that necessarily, but we should at least make it clear what's what. (maybe something like `v0.15.3 compatible with webpack versions [range]` would be enough)
- We should also consider how this will play with the internationalized docs/locki.

I think if we can get this repo to a point where we don't have a whole bunch of core doc issues open then we can start looking into this more. Workflow would be something like:

 webpack release => new issues and prs to document the release => webpack.js.org release

> The hardest part is forcing all commits to follow a certain convention. That rules out casual commits in PRs but that's not necessarily a bad thing.

True but as @wardpeet mentioned we can use the squash and merge option on PRs to rewrite commits. I think it's more that we as the maintainers must consistently remember to do it but with something like conventional-changelog-lint (soon to be commitlint) in the CI process maybe it wouldn't be too hard.

bebraw:
> Figure out what to do with loaders and plugins. These either could just be the latest version, or could have version dropdowns as well. Not saying we should do that necessarily, but we should at least make it clear what's what. (maybe something like v0.15.3 compatible with webpack versions [range] would be enough)

I would go with the latest version only. Implementing dropdowns for loaders/plugins feels like expensive to implement given how many pages you would have to write during generation (unless archives work in a smart way).

skipjack:
@bebraw yep I agree. Well at least we've landed on a general plan. Let's revisit this once our issue count is down and more of the core content is fleshed out.

skipjack:
@wardpeet I think we're slowly starting to follow the convention more and more. The "Squash and Merge" option you mentioned has been very useful indeed. Out of curiosity, have you implemented anything like [conventional-changelog-lint][1] into a build/ci process? We could enforce the convention this way but I'm not sure how that would handle "Squash and Merge" where we're actually rewriting the commit prior to merging. 

Just interested in your thoughts on this...

[1]: https://github.com/marionebl/conventional-changelog-lint

marionebl:
@skipjack [commitlint](https://github.com/marionebl/commitlint) as a CLI to be executed during CI is not suited very well to the "Squash and Merge" case.

@paulirish and @ahmed-taj came up with early commitlint bots that add a status to an PR based on the PR title - using those you end up conforming squash commits "by default". 

* https://github.com/ahmed-taj/commitlint-bot
* https://github.com/paulirish/commitlintbot

wardpeet:
it's used at [lighthouse](https://github.com/GoogleChrome/lighthouse)

skipjack:
> Is this something you are still interested in?

@marionebl yes it definitely is, I've been using `standard-version` and `commitlint` on a variety of other projects and love the workflow.

> commitlint as a CLI to be executed during CI is not suited very well to the "Squash and Merge" case...

Currently, the process I follow is basically:

- Are all these commits related to the fix/feat/refactor?
  - Squash & Merge.
- Else, do these commits follow the commit standard?
  - __Yes__: Rebase & Merge
  - __No__: Squash & Merge OR interactive rebase and merge

It looks like the first of the two bots you mentioned would be perfect as it would quickly answer that second question for us.

wardpeet:
@skipjack would you like me to set this up in a PR?

skipjack:
@wardpeet that would be great but wouldn't you need admin access to set up the bot? Also, which bot were you thinking about setting up. I think the first one @marionebl pointed out would be perfect:

https://github.com/ahmed-taj/commitlint-bot

wardpeet:
@skipjack i can do it on my fork to start from ;)

skipjack:
Great, yeah please feel free to submit one then.

marionebl:
One disclaimer: the bots don't support custom config yet, I am preparing support for this on the commitlint side of things: https://github.com/marionebl/commitlint/pull/169

Idea how this could work here: https://github.com/ahmed-taj/commitlint-bot/issues/1#issuecomment-342957711

skipjack:
@marionebl that’s fine, we pretty much just follow the conventionalcommits standard but we would probably customize it a little down the road.

PR #1020, opened 2017-03-15
Add note about `babel-preset-env`

montogeek:
Added a small note about configuring `babel-preset-env`


montogeek:
Uhmm, sorry, it should be "env" instead of "es2015"

bebraw:
Ah, yeah. Missed that. Another PR. :)

PR #1024, opened 2017-03-16
adding limit-chunk-count-plugin page

rouzbeh84:
This one I'm _very_ iffy on as I basically ported over webpack1 info on this over and hope it's the same. The code looks very similar in the commit history for the file within the last year or so but please take an extra second on this one as I'd really hate to mislead folks


bebraw:
Let's check again after we get your other PR in.

PR #789, opened 2017-02-01
Corrects bundle filepath in index.html

gabefinch:
Bundle filepath wasn't relative to location of index.html. Changes to "../dist/bundle.js" to correct this

1. [Read and sign the CLA](https://cla.js.foundation/webpack/webpack.js.org). This needs to be done only once. PRs that haven't signed it won't be accepted.
2. Make sure your PR complies with [the writer's guide](https://webpack.js.org/writers-guide/).
3. Read through the PR diff carefully as sometimes this can reveal issues. The work will be reviewed, but this can save some effort.
4. Remove these instructions from your PR as they are for your eyes only.

PR #1025, opened 2017-03-16
Fix link path in code example for getting started.

jecoopr:
Fixes a link path in the code example for getting started. It's currently an absolute link, which would work, except the instructions say to just open the index.html file in your browser. Without running a server the absolute path fails.

Change: look up one directory "../" to find js files. Otherwise, if you directly open index.html in your browser the files won't be found.


simon04:
I don't think this change is necessary, see https://github.com/webpack/webpack.js.org/pull/789#discussion_r99062362.

But since this is the second PR on paths in "get started", maybe we should clarify the directory structure in the guide.

jecoopr:
@simon04 Thanks for pointing that out. I took a shot at clarifying the directory instead.

PR #1034, opened 2017-03-18
Adding some info/content and fixing links

rouzbeh84:
This is quite an overhaul but I thought it would be appropriate to provide this info here as well as these are all methods of contribution.

The only thing I see that can cause confusion is the footer link for 'support' to 'gitter'?

### DIFF
--- a/content/contribute.md
+++ b/content/contribute.md
@@ -2,16 +2,52 @@
 title: Contribute
 ---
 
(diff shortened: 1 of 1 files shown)


skipjack:
@rouzbeh84 it looks like you also added these changes in #1035? I'm going to close this and if you could make the requested change over there then I'd be happy to review again and merge.

Issue #1030, opened 2017-03-17
404 links in contribute.md


S-Kaushik:
Hi, 

The Writers Guide and Development Guide on [this page](https://github.com/webpack/webpack.js.org/blob/master/content/contribute.md) are 404 links.

rouzbeh84:
ah snap, didn't realize this was assigned already. i'll close out #1034

skipjack:
@rouzbeh84 no worries, I'll review your PR shortly. I'm happy to have one less thing to PR today!

skipjack:
Closed by #1035.

PR #1035, opened 2017-03-18
Hotfix/footer support link

rouzbeh84:
this would fix the link text in the footer from 'support' to 'gitter' to not confuse the support links in header v footer as mentioned in #1030


skipjack:
@rouzbeh84 thanks for this, the "Support" => "Gitter" change looks good. Maybe instead of copying the whole support page and having a contribute and support page we can just update the support page to include the writers-guide and development guide links (potentially in the "How can I help?" section)?

If we do that then we drop the contribute page and just link "Contribute" in the footer to the support page at that section. @bebraw what do you think?

bebraw:
@skipjack Yeah, sounds like a good change to make.

PR #1043, opened 2017-03-21
Update get-started.md

kulakowka:


bebraw:
Hi,

It should be lowercase always in these docs as per https://github.com/webpack/webpack.js.org/blob/master/content/writers-guide.md#typesetting .

We should write a linting rule to capture that. Thanks for the PR, though.

Issue #1039, opened 2017-03-19
Documentation site improvements


aight8:
## Mark active submenu automatically

Regarding: all pages
Go for example to this page: https://webpack.js.org/api/plugins/compiler/

**Compiler** *<- active*
- Watching
- MultiCompiler
- Event Hooks
- Examples

It's confusing because the sub menus can never be active, they are anchors. This could assume a reader it's a separate page. In my opinion the subpages must 1) be a separate page 2) or mark the current submenu as active while scrolling or clicking on the anchor.

**Currently**
<img width="316" alt="bildschirmfoto 2017-03-19 um 13 18 04" src="https://cloud.githubusercontent.com/assets/1138623/24080808/97c1bf3c-0ca6-11e7-8f4d-ddc54e8cef6e.png">

**Would be better**
<img width="385" alt="bildschirmfoto 2017-03-19 um 13 17 58" src="https://cloud.githubusercontent.com/assets/1138623/24080810/995a19a2-0ca6-11e7-908b-65e95d1c4891.png">

Near that:
- If you just open a main menu it, it will mark the main menu item as blue. So now your current active main menu item is blue and the main menu item which you opened. In other words: It marks just opened menu items instead of only the current active one.

## Prevent content scrolling if scroll in menu (usability)

The content start scrolling if you overscroll the menu scroll area. This should be prevented since if someone start scrolling the menu he want scroll in the menu and never the content. This ends up that the current content position is scroll to much down or up the menu.

## Cluttered plugin related pages

WIP
https://webpack.js.org/plugins/

## Documentation -> CLI Menu entry

Click **Documentation** -> **CLI**

CLI never could be active. It redirects to **Documentation** -> **API** -> **Command Line Interface (CLI)**

I think redirecting is more confuses then helps - beside that the CLI menu point never can get active. One concept of navigation would be better.

## Sticky navigation

There are currently 3 menus. 1) Header, 2) Sub-header (only for documentation), 3) Sidebar-Menu (with an toggle anchor submenu)

The Sidebar-Menu is sticky, and you see most time all menus (if it not overflows), but you lost the menus from the header and sub header. It's irritating if you start search something and the header menus disappears every time but a part of the menu is sticking.

## Menu item overflows

Reading: Some menu items with overflow submenu items
Go for example to this page: https://webpack.js.org/api/loaders/

[...]

Issue #438, opened 2016-11-26
Dev - Information Architecture (Page Groups)


pastelsky:
The sidebar is getting quite long in a few pages. It would be nice if related items are grouped into some kind of a collection. I'm not sure if this requires a change in antwar or it's just a config change.

**eg.**
<img width="236" alt="screen shot 2016-11-26 at 5 04 24 pm" src="https://cloud.githubusercontent.com/assets/8946207/20640037/a801dab2-b3fa-11e6-98a5-14e42c8a1af0.png">

<img width="224" alt="screen shot 2016-11-26 at 5 04 46 pm" src="https://cloud.githubusercontent.com/assets/8946207/20640038/a8470c40-b3fa-11e6-82c7-ff83b6d4a4e8.png">

skipjack:
Adding @pastelsky's design from #285 for reference:

![](https://cloud.githubusercontent.com/assets/8946207/19885971/86dea91c-a047-11e6-9654-022e5c6f63ad.jpg)

I really like the look of the sidebar here, and I think between this design and some way to better group the pages, we'd definitely have a more fluid ux. Will try to get started on this soon.

skipjack:
@johnstew from #749 (closed in favor of this issue):

> Breaking out the conversation from #743 into its own ticket to discuss possible parent sections and if this is a good idea. Plus need to decide the UI/UX around having a parent section.

**Parent Section Ideas:**
- Introduction
- Getting started
- Production
- Caching
- Migration
- Code Splitting
- Tree Shaking
- Contributing to Webpack
- React
- Angular
- Extras

skipjack:
I'm gonna start working on implementing @pastelsky's design (shown above) but I think we still need to discuss our routing structure a bit and agree on an approach for going forward. Here's what I would propose as a "spec" for how we structure `/content`...

Within the `/content` directory we have our top-level routes:

1. Top Level: **Concepts**, **Guides**, **Documentation**, etc.

There are also some others in the top level of the directory but that are currently displayed at the second level of our routing structure even though the directory structure doesn't match. If we're happy with the way it's displayed I think we should create a Top Level **Documentation** folder and these underneath:

2. Second Level: **Configuration**, **Development**, **Loaders**, **Plugins** etc.

> Note: I'm only listing the ones under **Documentation** but any directories nested under the other Top Level ones like **Concepts** would be allowed as well.

[...]

PR #1046, opened 2017-03-22
Sidebar Redesign

skipjack:
Redesigning the sidebar to match @pastelsky's comp (see #438).

This is phase one of two for getting everything discussed in #438 (and things mentioned in some other issues like #1039) knocked out. My next task is updating the structure of the `/content` directory, the antwar config, etc to match our actual hierarchy in the site. This should allow us to map the banner links more dynamically and to allow grouping in the sidebar (with a few more tweaks).

Changes:

- Changing webpack version in sidebar to an image shield
- Updating sidebar design to come closer to the comp
- Doing some code cleanup/refactoring

Here's a gif of the updates:

### DIFF
--- a/components/page/page-style.scss
+++ b/components/page/page-style.scss
@@ -10,22 +10,19 @@
 
 // XXX: Temporary hack to fix sidebar width
 .page > div:first-of-type {
-  flex: 0 0 auto;
+  display: none;
   position: relative;
   overflow: hidden;
 
   @include break {
-    flex:0 0 30%;
-  }
-
-  @include break('large') {
-    flex:0 0 25%;
+    display: block;
+    flex: 0 0 280px;
   }
 }
 
 .page__content {
+  flex: 1 1 auto;
   overflow-x: hidden;
-  width: 100%;
   padding: 1.5em 1em;
 
   @media break {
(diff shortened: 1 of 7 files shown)


skipjack:
@SpaceK33z if you have time could you give this a quick go and see if anything needs fixing/tweaking?

scottdj92:
Is there a reason why you switched from % based values to px based?

skipjack:
@scottdj92 you mean in `page-style.scss`? I just wanted to ensure the sidebar always had enough room to look decent however I don't feel too strongly about that change... if you want to propose some widths that you think look better on more screens I'm all ears.

scottdj92:
Oh no I trust you. I was just asking about it, because I know % based widths scale a bit better than direct px.

skipjack:
Yea there are some screens in between our largest width and smallest width (e.g. like around 750px - 900px) that look a bit tight both in the sidebar and content areas). One option would be to break to our mobile view a bit earlier. 

Probably worth playing around with a bit more.

skipjack:
Merged as I think this is a pretty significant improvement over what we have now.

@SpaceK33z @webpack/documentation-team feel free to to test and ping me or create a ticket if you notice any issues.

PR #1051, opened 2017-03-24
Configuration TODOs

skipjack:
Hitting some of the outstanding configuration TODO's that were left hanging around. Added a call to action on the "Other Options" page to try to draw contributors who might be more familiar with some of the undocumented options there than I am.


bebraw:
Related idea - `?>` could generate instructions on how to proceed with fixing (just a sentence + link to the right place). Maybe easier to contribute then.

skipjack:
@bebraw yea i like that idea.

PR #1057, opened 2017-03-27
Fixed resolveLoader defaults

Kamshak:
This part is outdated, c.f.
- https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L98
- https://github.com/webpack/enhanced-resolve/blob/master/lib/ResolverFactory.js#L38
- https://github.com/webpack/webpack/issues/2532


Kamshak:
@sokra could you confirm if this is correct? Just pinging you since you answered the linked issue as well :)

PR #1064, opened 2017-03-30
Fix issue with sidebar-item with duplicate key attributes

ryandrewjohnson:
I discovered an issue on this [page](https://webpack.js.org/plugins/commons-chunk-plugin/#components/sidebar/sidebar.jsx) on the production site where the Example link in the sidebar would continually duplicate itself as you scrolled down the page.

This [tweet](https://twitter.com/ryandrewjohnson/status/847083931094659078) has a screen cap of the issue.

When running the site locally I could see that there was a React error complaining that there were duplicate `key` attributes being used on elements in the sidebar. I have updated the sidebar-item component to use lodash uniqueId to generate key which prevented the error.

### DIFF
--- a/components/sidebar-item/sidebar-item.jsx
+++ b/components/sidebar-item/sidebar-item.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { uniqueId } from 'lodash';
 import Link from '../link/link';
 
 const block = 'sidebar-item';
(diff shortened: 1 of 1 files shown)


skipjack:
@bebraw maybe we should add IDs to anchor the anchor link objects when we extract them? This PR fixed the problem but it kind of navigates around the diff-ing algorithm (although in this case maybe it doesn't matter).

ryandrewjohnson:
@bebraw @skipjack happy to help. If there is anything else you want to throw my way feel free.

bebraw:
@skipjack One option - `anchors.map(anchor => (` -> `anchors.map((anchor, i) => (` and then use `i` as a part of key (guaranteed to be unique, no need to generate unique ids). That would solve the issue too.

PR #1072, opened 2017-04-02
Color palette cleanup

skipjack:
Fixing, sorting, and cleaning up the color palette section of `branding.md`.

From:

image

to this:

image

@NejcZdovc I'll still be following up with another more targeted table styling PR (at some point 😆 , i know i said that a while ago too but it will come!). I wanted to check with you to make sure that padding I removed wouldn't break anything you did (I did check the mobile view which looked fine as far as I could tell).

@rouzbeh84 feel free to take a look, I know you worked on this file as well.


skipjack:
@NejcZdovc thanks and yea html should be fine just not table elements themselves as was used here before.

PR #1074, opened 2017-04-03
Update targets.md

joe-glasgow:
Gives an example of multiple deployment targets using Electron.


bebraw:
The problem is that the post is on webpack 1 syntax. Unless the original author updates the post, a good way to handle this would be to paraphrase and refer to the original for credit.

PR #1077, opened 2017-04-05
Explanation of module argument in CommonsChunkPlugin & add advanced implicit common vendor chunk example

### LINKED ISSUE #902: [CommonChunksPlugin] minChunks funciton docs should have a link to module description
This page: https://webpack.js.org/plugins/commons-chunk-plugin/#passing-the-minchunks-property-a-function

>
> You also have the ability to pass the minChunks property a function. This function is called by the CommonsChunkPlugin and calls the function with module and count arguments.
> 
> **The `module` property represents each module in the chunks you have provided via the `names` property.**
>

This should mention that each module is a Webpack representation of a module and contains such and such properties (e.g. `context`). Or, possibly, a link to a page which details these properties.

kevinzwhuang:
This PR focuses on improvements to `CommonsChunkPlugin`.

### Summary:
There currently is no description of what shape the `module` argument provided by `minChunks` takes - which makes it hard for new users to understand how to use it for custom common chunks. (This is mentioned in #902)

### Changes
This PR adds a description of two particularly important properties for this use case (`context` and `resource`), and links to the `NormalModule` class definition that the `module` argument takes the shape of. 

It also adds an additional advanced implicit common chunks example that shows how to use the `resource` property to prevent specific files (such as stylesheets needed in a separate file with `ExtractTextWebpackPlugin`) from being moved to the common chunk.

---

Fixes #902

### DIFF
--- a/content/plugins/commons-chunk-plugin.md
+++ b/content/plugins/commons-chunk-plugin.md
@@ -187,7 +187,7 @@ new webpack.optimize.CommonsChunkPlugin({
 })
 ```
 
-In order to obtain a single CSS file containing your application and vendor CSS, use the following `minChunks` function together with [`ExtractTextPlugin`](plugins/extract-text-webpack-plugin/):
+In order to obtain a single CSS file containing your application and vendor CSS, use the following `minChunks` function together with [`ExtractTextPlugin`](/plugins/extract-text-webpack-plugin/):
 
 ```javascript
 new webpack.optimize.CommonsChunkPlugin({
--- a/content/plugins/commons-chunk-plugin.md
+++ b/content/plugins/commons-chunk-plugin.md
@@ -194,9 +194,9 @@ specific modules from being included in the vendor chunk (such as stylesheets yo
 new webpack.optimize.CommonsChunkPlugin({
   name: "vendor",
   minChunks: function (module) {
+    // This prevents stylesheet resources with the .css or .scss extension
+    // from being moved from their original chunk to the vendor chunk
     if(module.resource && (/^.*\.(css|scss)$/).test(module.resource)) {
-      // This prevents stylesheet resources with the .css or .scss extension
-      // from being moved from their original chunk to the vendor chunk
       return false;
     }
     return module.context && module.context.indexOf("node_modules") !== -1;
--- a/content/plugins/commons-chunk-plugin.md
+++ b/content/plugins/commons-chunk-plugin.md
@@ -153,8 +153,8 @@ You also have the ability to pass the `minChunks` property a function. This func
 The `module` argument represents each module in the chunks you have provided via the `name`/`names` property.
 `module` has the shape of a [NormalModule](https://github.com/webpack/webpack/blob/master/lib/NormalModule.js#L57), which has two particularly useful properties for this use case:
 
-- `resource`: The name of the file being processed. For example: `'/my_project/node_modules/example-dependency/index.js'`
-- `context`: The directory that stores the file. For example: `'/my_project/node_modules/example-dependency'`
+- `module.context`: The directory that stores the file. For example: `'/my_project/node_modules/example-dependency'`
+- `module.resource`: The name of the file being processed. For example: `'/my_project/node_modules/example-dependency/index.js'`
 
 The `count` argument represents how many chunks the `module` is used in.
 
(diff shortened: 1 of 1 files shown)


content/plugins/commons-chunk-plugin.md
```diff
@@ -149,9 +150,13 @@ new webpack.optimize.CommonsChunkPlugin({
...
 
-The `module` property represents each module in the chunks you have provided via the `names` property.
+The `module` argument represents each module in the chunks you have provided via the `name`/`names` property.
+`module` has the shape of a [NormalModule](https://github.com/webpack/webpack/blob/master/lib/NormalModule.js#L57), which has two particularly useful properties for this use case:
 
-The `count` property represents how many chunks the `module` is used in.
+- `module.context`: The directory that stores the file. For example: `'/my_project/node_modules/example-dependency'`
+- `module.resource`: The name of the file being processed. For example: `'/my_project/node_modules/example-dependency/index.js'`
+
+The `count` argument represents how many chunks the `module` is used in.
```
simon04: Great to see the function arguments documented! :+1:

content/plugins/commons-chunk-plugin.md
```diff
@@ -182,6 +187,23 @@ new webpack.optimize.CommonsChunkPlugin({
 })
 ```
 
+To add further control to an implicit common vendor chunk, you can additional logic before the `node_modules` context check to prevent
+specific modules from being included in the vendor chunk (such as stylesheets you want to keep in the original chunk):
```
simon04: Reading through this example leaves the question open as why this would be useful. I mean from the nice documentation of the module object from above this should be easy to achieve. But I do not see the importance in general to keep the CSS files in a different chunk. If there's a good use case, please elaborate. Otherwise, I'd drop this example.

kevinzwhuang: The rationale for adding this example comes from a gotcha from using the implicit vendor chunk with the `ExtractTextPlugin`.

`ExtractTextPlugin` separates your CSS into its own separate CSS bundle - but when combined with implicit common vendor chunk logic alone, it will generate another separate vendor-only CSS bundle, so you end up with `vendor.css` and an `application.css` for example.

This separated vendor CSS chunk might not be immedately obvious to users (which is confusing when you intend to just have the single separate CSS chunk(`application.css`) from `ExtractTextPlugin`). It also complicates the ordering of CSS `require` statements, since the CSS modules will be split up across the vendor and original chunk - this can affect style precedence. I think most `ExtractTextPlugin` users would prefer to have 1 `application.css` chunk instead of `application.css` + `vendor.css` chunks.

It took me some time to debug why our styles were being affected in production after we changed from an explicit to implicit vendor chunk, so I thought this might be worth pointing out in case anybody else gets caught up by their `ExtractTextPlugin` setup. (Having the function arguments for `minChunks` also helps a lot here to too!)

I'm fine with dropping this example but I do think it would help users who might have a similar setup with `ExtractTextPlugin`. Let me know what you think @simon04!

content/plugins/commons-chunk-plugin.md
```diff
@@ -149,9 +150,13 @@ new webpack.optimize.CommonsChunkPlugin({
 
 You also have the ability to pass the `minChunks` property a function. This function is called by the `CommonsChunkPlugin` and calls the function with `module` and `count` arguments.
 
-The `module` property represents each module in the chunks you have provided via the `names` property.
+The `module` argument represents each module in the chunks you have provided via the `name`/`names` property.
+`module` has the shape of a [NormalModule](https://github.com/webpack/webpack/blob/master/lib/NormalModule.js#L57), which has two particularly useful properties for this use case:
```
simon04: Linking to a specific line on master might break on a new commit. Either drop `#L57` or s/master/70e48d2/.

content/plugins/commons-chunk-plugin.md
simon04: Sounds reasonable. I'd drop the rather obvious existing two lines and write something similar to "In order to obtain a single CSS file containing your application and vendor CSS, use the following `minChunks` function together with `ExtractTextPlugin`" (and add a link to the ExtractTextPlugin docs).

kevinzwhuang:
@simon04 Sorry for the wait! Had a busy week, so it took a bit longer than I wanted to find time to get back to this PR.

- I updated the link to the `NormalModule` to no longer have the line reference - I assume it should be fairly easy for users to find the constructor themselves.
- Also updated the note explaining the `ExtractTextPlugin` example with the note you wrote.

PR #1080, opened 2017-04-05
Update ProvidePlugin usage

re-fort:
### Summary
Updated the content of the following PR
https://github.com/webpack/webpack/pull/3597

### DIFF
--- a/content/plugins/provide-plugin.md
+++ b/content/plugins/provide-plugin.md
@@ -3,13 +3,16 @@ title: ProvidePlugin
 contributors:
   - sokra
   - simon04
+  - re-fort
 ---
 
 ```javascript
 new webpack.ProvidePlugin({identifier1: 'module1', /* ... */})
+// or
+new webpack.ProvidePlugin({identifier1: ['module1', 'property1'], /* ... */})
 ```
 
-Automatically loads modules. Whenever the `identifier` is encountered as free variable in a module, the `module` is loaded automatically and the `identifier` is filled with the exports of the loaded `module`.
+Automatically loads modules. Whenever the `identifier` is encountered as free variable in a module, the `module` is loaded automatically and the `identifier` is filled with the exports of the loaded `module`(of `property`).
 
 ## Typical use-cases
 
@@ -36,5 +39,13 @@ Angular looks for `window.jQuery` in order to determine whether jQuery is presen
 ```javascript
 new webpack.ProvidePlugin({
   'window.jQuery': 'jquery'
+})
+```
+
+### Use Vue.js(ES Modules)
+
+```javascript
+new webpack.ProvidePlugin({
+  Vue: ['vue/dist/vue.esm.js', 'default'] // v2.2.6
 })
 ```
--- a/content/plugins/provide-plugin.md
+++ b/content/plugins/provide-plugin.md
@@ -12,7 +12,7 @@ new webpack.ProvidePlugin({identifier1: 'module1', /* ... */})
 new webpack.ProvidePlugin({identifier1: ['module1', 'property1'], /* ... */})
 ```
 
-Automatically loads modules. Whenever the `identifier` is encountered as free variable in a module, the `module` is loaded automatically and the `identifier` is filled with the exports of the loaded `module`(of `property` in order to support named exports).
+Automatically loads modules. Whenever the `identifier` is encountered as free variable in a module, the `module` is loaded automatically and the `identifier` is filled with the exports of the loaded `module` (of `property` in order to support named exports).
 
 W> For importing the default export of an ES2015 module, you have to specify the default property of module.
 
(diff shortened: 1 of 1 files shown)


content/plugins/provide-plugin.md
```diff
@@ -3,13 +3,16 @@ title: ProvidePlugin
...
 ---
 
 ```javascript
 new webpack.ProvidePlugin({identifier1: 'module1', /* ... */})
+// or
+new webpack.ProvidePlugin({identifier1: ['module1', 'property1'], /* ... */})
 ```
 
-Automatically loads modules. Whenever the `identifier` is encountered as free variable in a module, the `module` is loaded automatically and the `identifier` is filled with the exports of the loaded `module`.
+Automatically loads modules. Whenever the `identifier` is encountered as free variable in a module, the `module` is loaded automatically and the `identifier` is filled with the exports of the loaded `module`(of `property`).
```
simon04: s/of property/of property in order to support named exports/ (give the reader a clue why this is useful)

content/plugins/provide-plugin.md
```diff
@@ -38,3 +41,11 @@ new webpack.ProvidePlugin({
   'window.jQuery': 'jquery'
 })
 ```
+
+### Use Vue.js(ES Modules)
```
simon04: "(ES Modules)"? Why is this information necessary?

content/plugins/provide-plugin.md
```diff
@@ -38,3 +41,11 @@ new webpack.ProvidePlugin({
...
+
+```javascript
+new webpack.ProvidePlugin({
+  Vue: ['vue/dist/vue.esm.js', 'default'] // v2.2.6
```
simon04: Referring to the `'default'` export does not show the power of this feature since it is the default anyway? For a more practical example, you could use `_map: ['lodash', 'map']`, this should work, right?

What is `v2.2.6` referring to?

content/plugins/provide-plugin.md
re-fort: I think this feature is also effective to load ES Modules. 
(ES modules use ES2015 `export` syntax)
Please let me explain in detail with the comments below.

content/plugins/provide-plugin.md
re-fort: > What is v2.2.6 referring to?

`Vue.js` version which I use when I wrote the document.

> Referring to the 'default' export does not show the power of this feature since it is the default anyway?

If modules use ES2015 `export` syntax(like ES Modules), need to specify the `default` property.
Here's the sample.

```js
// webpack.config.js

new webpack.ProvidePlugin({
  esmodules_default: './esmodules.js',
  esmodules_default2: ['./esmodules.js', 'default'],
  esmodules_foo: ['./esmodules.js', 'foo'],
  commonjs_default: './commonjs.js',
  commonjs_foo: ['./commonjs.js', 'foo']
})
```

```js
// esmodules.js

export default function() { console.log('Hello from ES Modules of default') }
export { foo }

const foo = function() { console.log('Hello from ES Modules of foo') }
```

```js
// commonjs.js

bar = function() { console.log('Hello from CommonJS of default') }
bar.foo = function() { console.log('Hello from CommonJS of foo') }
module.exports = bar
```

```js
// main.js

esmodules_default() // Uncaught TypeError: esmodules_default is not a function
esmodules_default2() // Hello from ES Modules of default
esmodules_foo() // Hello from ES Modules of foo
commonjs_default() // Hello from CommonJS of default
commonjs_foo() // Hello from CommonJS of foo
```

Since this is a problem related to [babel](https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/), should not it not be taken up?
If so, I'll remove the usage of loading ES Modules.
I'll add an example you mentioned in any case.

simon04: Confirmed that this is an unexpected behaviour of webpack, see https://github.com/webpack/webpack/issues/4729. Since this is independent of Vue or Babel, I'd suggest to have a [warning box](https://webpack.js.org/writers-guide/#tip) just above the "Typical use-cases" explaining the current need to specify the `"default"` for ES2015 modules.

If you want to keep the Vue section, please remove the "(ES Modules)" from the caption and the version comment.

re-fort: Thank you for your confirmation.
I'll re-update the document after webpack/webpack#4729 is solved.

> If you want to keep the Vue section, please remove the "(ES Modules)" from the caption and the version comment.

Okay.
if it is determined that `ProvidePlugin` behaves like `require` as is, I'll add the Vue section  with displaying a warning box.

simon04:
Any update on this PR?

re-fort:
Updated

re-fort:
Thank you for your review!

PR #1082, opened 2017-04-06
Update tapable.md

briandipalma:
I presume this is a typo?


simon04:
@TheLarkInn, is the asterisk intended here (to refer to any of the `apply…` function).

briandipalma:
After looking through the code I think that's the intention. I'll close the PR.

PR #1087, opened 2017-04-07
Update index.md

ecarlisle:
Writer's Guide link has a relative path `./writers-guide` pointing to the wrong location.
Changed link to instead to point to correct location, `/writers-guide`.


bebraw:
@skipjack Open a separate issue at antwar repo so I'll fix the discrepancy. The dev mode is too lax in that case.

We could also tighten up link checker to catch these better. Now it lets ten fails through but that's starting to feel like a bad idea.

PR #1089, opened 2017-04-07
Significantly expanded the node options docs

oneforwonder:
The Node configuration page was still basically a stub, so I fleshed it out.


skipjack:
@oneforwonder added one minor comment but aside from that this looks great, thanks! Once that's changed I'd be happy to merge unless @webpack/documentation-team has any other thoughts before hand.

PR #1092, opened 2017-04-07
document warningsFilter 

timse:
document warningsFilter as introduced in https://github.com/webpack/webpack/pull/4381


bebraw:
It's better to merge this only after a new stable version is available.

PR #1094, opened 2017-04-08
Fix OccurrenceOrderPlugin section

rohmanhm:
Fix OccurrenceOrderPlugin typo


simon04:
No. The comment "webpack 1" is referring to an upgrade from webpack 1. Thus, the plugin shall be removed.

PR #1103, opened 2017-04-13
Update module.md phrasing

pedromartinez11:

### DIFF
--- a/content/configuration/module.md
+++ b/content/configuration/module.md
@@ -259,7 +259,7 @@ For compatibility a `query` property is also possible, which is an alias for the
 }
 ```
 
-Note that webpack needs to generate a unique module identifier from resource and all loaders including options. It tries to do this with a `JSON.stringify` of the options object. This is fine in 99.9%, but may be not unique if you apply the same loaders with different options to the same resource and the options have some stringified values. It also breaks if the options object cannot be stringified (i. e. circular JSON). Because of this you can have a `ident` property in the options object which is used as unique identifier.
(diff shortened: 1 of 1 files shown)


content/configuration/module.md
```diff
@@ -259,7 +259,7 @@ For compatibility a `query` property is also possible, which is an alias for the
 }
 ```
 
-Note that webpack need to generate an unique module identifier from resource and all loaders including options. It tries to do this with a `JSON.stringify` of the options object. This is fine in 99.9%, but may be not unique if you apply the same loaders with different options to the same resource and the options have some stringified values. It also breaks if the options object cannot be stringified (i. e. circular JSON). Because of this you can have a `ident` property in the options object which is used as unique identifier.
+Note that webpack needs to generate a unique module identifier from resource and all loaders including options. It tries to do this with a `JSON.stringify` of the options object. This is fine in 99.9%, but may be not unique if you apply the same loaders with different options to the same resource and the options have some stringified values. It also breaks if the options object cannot be stringified (i. e. circular JSON). Because of this you can have a `ident` property in the options object which is used as unique identifier.
```
skipjack: @pedromartinez11 thanks for your contribution -- the first sentence looks better now definitely.

Would you mind also changing:

`This is fine in 99.9%` => `This is fine in 99.9% of cases`

and:

`(i. e. circular JSON)` => `(i.e. circular JSON)`

and lastly:

`Because of this you can have a ...` => `Because of this you can have an ...`

If you don't have the time, just let me know and I'll commit these separately.

PR #1109, opened 2017-04-14
Clarify use of `import()` with ES2015 modules.

knpwrs:
The examples on the page show usage with commonjs modules but not ES2015 modules, which was confusing for me when I tried to import ES2015 modules.

See https://github.com/webpack/webpack/issues/3829. My provided examples could probably use some work, though I was trying to come up with something generic that doesn't introduce too many concepts at once. My personal use case is for HMR with a React component and Redux reducers:

```js
if (module.hot) {
  module.hot.accept('./root', async () => {
    mount((await import('./root')).default);
  });
  module.hot.accept('./reducers', async () => {
    store.replaceReducer((await import('./reducers')).default);
  });
}
```

### DIFF
--- a/content/guides/code-splitting-async.md
+++ b/content/guides/code-splitting-async.md
@@ -11,7 +11,7 @@ contributors:
 
 This guide documents how to split your bundle into chunks which can be downloaded asynchronously at a later time. For instance, this allows to serve a minimal bootstrap bundle first and to asynchronously additional features later.
 
-webpack supports two similar techniques to achieve this goal: using `import()` (preferred, ECMAScript proposal) and `require.ensure()` (legacy, webpack specific). 
+webpack supports two similar techniques to achieve this goal: using `import()` (preferred, ECMAScript proposal) and `require.ensure()` (legacy, webpack specific).
 
 
 ## Dynamic import: `import()`
@@ -174,9 +174,10 @@ module.exports = {
 };
 ```
 
-## `import` imports the entire module namespace
 
-Consider the following two examples:
+### `import()` imports the entire module namespace
+
+Note that the promise is [resolved with the module namespace](https://github.com/tc39/proposal-dynamic-import#proposed-solution). Consider the following two examples:
 
 ```js
 // Example 1: top-level import
@@ -185,7 +186,7 @@ import * as Component from './component';
 import('./component').then(Component => /* ... */);
 ```
 
-`Component` in both of those cases resolves to the same thing, meaning in the case of using `import()` with ES2015 modules you have to explicity access default and named exports:
+`Component` in both of those cases resolves to the same thing, meaning in the case of using `import()` with ES2015 modules you have to explicitly access default and named exports:
 
 ```js
 async function main() {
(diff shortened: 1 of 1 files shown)


content/guides/code-splitting-import.md
```diff
@@ -144,6 +144,28 @@ module.exports = {
 };
 ```
 
+## `import` imports the entire module namespace
+
+Consider the following two examples:
```
simon04: The statement "The promise is resolved with the module namespace" from https://github.com/tc39/proposal-dynamic-import/issues/21#issuecomment-258127680 brought me right on track. Please add this at the beginning and link to the spec (2.1.1, step 8.c.iv).

content/guides/code-splitting-import.md
```diff
@@ -144,6 +144,28 @@ module.exports = {
...
+
+```js
+// Example 1: top-level import
+import * as Component from './component';
+// Example 2: Code-splitting with import()
+import('./component').then(Component => /* ... */);
+```
```
simon04: Maybe show a three liner of `component` exporting `default`+`foo`+`bar` to illustrate the context.

content/guides/code-splitting-import.md
```diff
@@ -144,6 +144,28 @@ module.exports = {
...
+
+`Component` in both of those cases resolves to the same thing, meaning in the case of using `import()` with ES2015 modules you have to explicity access default and named exports:
+
+```js
+async function main() {
+  // Destructuring example
+  const { default: Component } = await import('./component');
```
jhabdas: If this is for dynamic imports, why the synchronous operation?

simon04: @jhabdas Have you noted the `async`/`await` keywords?

jhabdas: > Have you noted the `async`/`await` keywords?

I did see the `async`/`await`. But I also saw the line below it, which won't run until this promise resolves and the desctructuring assignment occurs, resulting in a synchronyous operation. Perhaps it would be better to split the two examples into separate methods or find a way to run them all in parallel for better performance.

knpwrs: This was just to show the same thing two different ways. One gets a reference to the component, and the other passes it inline. This can be broken apart if confusing.

content/guides/code-splitting-import.md
knpwrs: Haven't gotten a response here yet: https://github.com/tc39/proposal-dynamic-import/issues/21#issuecomment-294293874

Do you know where to find the actual spec?

simon04: https://github.com/tc39/proposal-dynamic-import#proposed-solution

simon04:
I rebased this PR to resolve the conflicts (due to 11ee3885ad5d6c2bdb6e745a6d5ea962f1782481), and incorporated some changes.

Yet to be clarified+incorporated is the review https://github.com/webpack/webpack.js.org/pull/1109#pullrequestreview-32927159

PR #1110, opened 2017-04-15
Clarify Dynamic Import Limitations

### LINKED ISSUE #1083: More Clarity In Docs Needed on Limitations of Dynamic Imports in Webpack 2
[This page in the docs](https://webpack.js.org/guides/code-splitting-import/) only has a quick sentence about how import statements cannot be "fully" dynamic, but can be partly.

>  Keep in mind that import() path cannot be fully dynamic (e.g., import(Math.random())). Rather either completely static (e.g., import('./locale/de.json')) or partially static (e.g., import('./locale/' + language + '.json')).

Ideally the docs would provide details on when and how variable names can be used in import statements.  For example, something like this doesn't work.

```js
export function foo(deps){
  deps.forEach(dep => import(dep).then(doSomething))
}
//usage:
foo(['bar', 'baz']);
```

but this does (according to the docs at least)

```js
import('./locale/' + language + '.json')).
```

it would be helpful if there was more detail about when dynamic import paths are allowed and what parts of them can be dynamic (it seems like the beginning needs to be static?). 

Thanks for the help! I appreciate all of the work you all do on this.

skipjack:
This is an attempt to resolve #1083 and could also potentially address #1099, however #1099 is in relation to `require.resolve`. Maybe this can be tweaked somehow to accommodate both cases.

This will conflict with some other PRs that are currently open but I'd be happy to rebase once those are merged. In the meantime, hoping someone can give this a quick review to make sure I have everything right.

### DIFF
--- a/content/guides/code-splitting-async.md
+++ b/content/guides/code-splitting-async.md
@@ -34,7 +34,11 @@ function determineDate() {
 
 determineDate();
 ```
-T> Keep in mind that `import()` path cannot be fully dynamic (e.g., `import(Math.random())`). Rather either completely static (e.g., `import('./locale/de.json')`) or partially static (e.g., `import('./locale/' + language + '.json')`).
+
+Note that fully dynamic statements, such as `import(foo)`, __will fail__ because webpack requires at least some file location information. This is because `foo` could potentially be any path to any file in your system or project. The `import()` must contain at least some information about where the module is located, so bundling can be limited to a specific directory or set of files.
+
+For example, ``import(`./locale/${language}.json`)`` will cause every `.json` file in the `./locale` directory to be bundled into the split-point. At run time, when the variable `language` has been computed, any file like `english.json` or `german.json` will be available for consumption. So keep in mind that when using `import()`, the path must contain some path information or be completely static (as is `'moment'` in the example above).
+
 
 ### Promise polyfill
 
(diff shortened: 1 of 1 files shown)


content/guides/code-splitting-import.md
```diff
@@ -27,7 +27,11 @@ function determineDate() {
 
 determineDate();
 ```
-T> Keep in mind that `import()` path cannot be fully dynamic (e.g., `import(Math.random())`). Rather either completely static (e.g., `import('./locale/de.json')`) or partially static (e.g., `import('./locale/' + language + '.json')`).
+
+Note that fully dynamic statements, such as `import(foo)`, __will fail__ because webpack requires at least some file location information. This is because `foo` could potentially be any path to any file in your system or project. The `import(...)` must contain at least some information about where the module is located, so bundling can be limited to a specific directory or set of files.
+
+For example, `` `import(`./locale/${language}.json`)`` will cause every `.json` file in the `./locale` directory to be bundled into the split point. At run time, when the variable `language` has been computed, any file like `english.json` or `german.json` will be available for consumption. So keep in mind that when using `import()`, the path must contain some path information or be completely static.
```
simon04: * There's one extra backtick before `import(…locale…)`.
* "completely static" → "completely static (as `'moment'` in the example above)"

content/guides/code-splitting-import.md
```diff
@@ -27,7 +27,11 @@ function determineDate() {
 
 determineDate();
 ```
-T> Keep in mind that `import()` path cannot be fully dynamic (e.g., `import(Math.random())`). Rather either completely static (e.g., `import('./locale/de.json')`) or partially static (e.g., `import('./locale/' + language + '.json')`).
+
+Note that fully dynamic statements, such as `import(foo)`, __will fail__ because webpack requires at least some file location information. This is because `foo` could potentially be any path to any file in your system or project. The `import(...)` must contain at least some information about where the module is located, so bundling can be limited to a specific directory or set of files.
```
simon04: Two minor remarks right at the beginning:
* "split point" → "split-point" (same typesetting as in the intro above the example)
* `import(...)` → `import()`? (We've been referring to this always without arguments. But I see that here you want to stress the argument. Decide yourself. :smile:)

skipjack:
@simon04 fixed those issues, all valid points.

PR #1122, opened 2017-04-18
Sidebar: show latest release badge

simon04:


content/guides/installation.md
```diff
@@ -15,6 +15,8 @@ The next section tells you how to install webpack locally in a project.
 
 ### Local Installation
 
+The latest released webpack version is: [![GitHub release](https://img.shields.io/github/release/webpack/webpack.svg?style=flat-square)](https://github.com/webpack/webpack/releases)
```
skipjack: Maybe `The latest webpack release is: ...` would sound slightly better. Feel free to ignore though, somewhat nitpicky.

components/sidebar/sidebar.jsx
```diff
@@ -35,7 +35,6 @@ export default class Sidebar extends Component {
           height={ availableHeight } />
 
         <div className="sidebar__inner">
-          <img src="https://img.shields.io/badge/webpack-v2.3.3-blue.svg" />
```
skipjack: Not sure if we should entirely remove it from here without having it somewhere else that's consistently displayed. This came up a few times a while back which is why we added it to the sidebar for lack of a better place. @bebraw @TheLarkInn what do you think?

bebraw: Leave it for now. You still want to show the version globally I think and sidebar is a nice place for that.

skipjack: @simon04 maybe just update this one to be the dynamic version you used in the installation guide? Unless we want to bump this manually as new docs are added...

simon04: Alright. I'll move the badge back to the sidebar.

skipjack: I think you could have it in both places.

simon04: I agree. The badge in the installation guide is back with 28c89cd.

PR #1023, opened 2017-03-16
adding environment-plugin to list of plugins

rouzbeh84:
adding environment-plugin info to plugin index with a snippet of what it does.

PR #1129, opened 2017-04-19
LimitChunkCountPlugin: remove unrelated content

simon04:
@rouzbeh84 I'm concerned with the focus of the `LimitChunkCountPlugin` docs from https://github.com/webpack/webpack.js.org/pull/1023. I really like the introduction, which nicely explains the context where this plugin is to be used.

However, the parts under the headings "Single-Page-App", "Multi-Page-App" do not mention the `LimitChunkCountPlugin` once. I also considered extracting the text as a guide; however I couldn't think of a good title and it seems that there's little information added to the [existing CommonsChunkPlugin docs](https://webpack.js.org/plugins/commons-chunk-plugin/).

What do you think?


rouzbeh84:
Let me look at it in the morning and see if I can add it (find any info) on the limiting chunks and see if it warrants its own section as I think it's just a maxval specified as an object literal but it is 5am where I live so my mind is mush haha. Thanks for bringing this up as some things even I write sometimes sound confusing to me but I've mostly only worked on the documentation side of things.

skipjack:
@rouzbeh84 any update on this?

(side note: i sent you a collaborator invitation as i know you've made significant contributions)

rouzbeh84:
Ah yes, i've lost track of time lately so my apologies! I'll revisit this tonight (pst) and see if I can add anything more useful and keep you posted.

Also, thanks for the recognition as well as your own contributions! oddly enough i'm wearing my [webpack shirt](https://i.imgur.com/tJaoIsD.jpg) today of all days too haha

skipjack:
🎉  haha perfect timing! Re this pr: sounds good no rush.

rouzbeh84:
This can be closed for now. I will try to revisit and am open to suggestions as I do agree that it would be nice for there to be some explanation/usage of the of the limiting put to good use between SPA/MPA. 

I would like to try it out that uses it in both situations to ref or find projects utilizing it to provide a better reference rather than arbitrary usage of it for users to see the benefit. If anyone has what they feel is a decent example I can look at or even include (pending project licensing) that would be great but until otherwise this should get merged to remove the extraneous content.

If nobody has anything to add at this time please do either here or I can add a new issue to include these if it doesn't exist yet..

rouzbeh84:
sorry again for the lack of activity lately. Little Ellahe Jasmine Sarrafieh thanks you very much for your patience too! Will send out pic once she grows into her onesie 👍

rouzbeh84:
Any update on this? I'm good with it being drilled down to reduce any confusion

simon04:
@rouzbeh84, thanks for looking into this documentation page again!

I rebased to get round the conflict. Now, we only need an approving review. :smile:

PR #1133, opened 2017-04-19
SourceMapDevToolPlugin: review, add example

simon04:


content/plugins/source-map-dev-tool-plugin.md
```diff
@@ -2,24 +2,34 @@
 title: SourceMapDevToolPlugin
 contributors:
     - johnnyreilly
+    - simon04
 ---
 
-?> Review this content
-
-Adds SourceMaps for assets.
+Adds source maps for assets.
```
skipjack: @simon04 we may want to flesh this description out a bit more... initial questions/thoughts that come to mind are:

- How does this play with `devtool` source-mapping?
- Why would I use this over just `devtool`? Does it give the user more control over source map generation?
- Linking to the other sections of the docs that discuss source-mapping throughout this description would be nice as well.

Aside from that, the documentation changes and example you added look fairly straightforward.

simon04: Good point. I tried to improve the introduction.

skipjack:
@simon04 perfect, nice work!

PR #1135, opened 2017-04-19
Add NoEmitOnErrorsPlugin documentation

simon04:


skipjack:
@simon04 on the assumption that at some point we may implement something like #1017 (and also the convention helps with consistency/readability), I think @bebraw and I are both slowly shifting our commit messages over to the [standard defined here](https://conventionalcommits.org/). 

I'm starting to use the squash and merge option to rewrite prs that don't follow the convention on merge, so if you could try that as well it would be helpful. You don't necessarily have to follow the convention on each branch but we should at least attempt to stick to it on master.

PR #1140, opened 2017-04-21
add style-loader to tutorial "Code Splitting - CSS"

### LINKED ISSUE #1131: incomplete css-tutorial?
the tutorial for bundling css with webpack (https://webpack.js.org/guides/code-splitting-css/) does not quite tell you all the steps necessary in my opinion. is it just me and i'm missing the point of the tutorial?

if the point of the tutorial is indeed to bundle css so that it is rendered on the website, i'd like to submit a PR which adds the missing steps (installing and using the style-loader-module). thoughts?

TomTasche:
fixes #1131

### DIFF
--- a/content/guides/code-splitting-css.md
+++ b/content/guides/code-splitting-css.md
@@ -7,6 +7,7 @@ contributors:
   - johnstew
   - simon04
   - shinxi
+  - tomtasche
 ---
 
 To bundle CSS files with webpack, import CSS into your JavaScript code like [any other module](/concepts/modules), and use the `css-loader` (which outputs the CSS as JS module), and optionally apply the `ExtractTextWebpackPlugin` (which extracts the bundled CSS and outputs CSS files).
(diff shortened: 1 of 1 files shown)


content/guides/code-splitting-css.md
```diff
@@ -36,13 +37,13 @@ module.exports = {
...
             test: /\.css$/,
-            use: 'css-loader'
+            use: [ 'style-loader', 'css-loader' ]
         }]
     }
 }
 ```
 
-As a result, the CSS is bundled along with your JavaScript.
+As a result, the CSS is bundled along with your JavaScript and applied to the page after it loaded.
```
skipjack: "...and applied to the page via '<style>' tag injection after the initial load." to elaborate a bit.

TomTasche:
thanks for the feedback, implemented!

skipjack:
Awesome, great work!

PR #1144, opened 2017-04-22
Update hmr-react.md

Mr-Binary:
I did everything as the docs said, but didn't work. Finally, solve the problem by removing a split configuration of Babel in the webpack.config.js. I think it worths adding a note here.

### DIFF
--- a/content/guides/hmr-react.md
+++ b/content/guides/hmr-react.md
@@ -159,6 +159,7 @@ module.exports = {
 Important to note:
 
 If you have a `query` section for the `babel-loader` in your `rules`, remember to turn off the `modules` tranformation as well, otherwise HMR won't work.
+
 ```js
 loader: "babel-loader",
 query: {
@@ -168,6 +169,7 @@ query: {
     ]
 }
 ```
+
 A better solution is to remove the `query` part here, manage all you babel configuration in `.babelrc`.
 
 ### App code
--- a/content/guides/hmr-react.md
+++ b/content/guides/hmr-react.md
@@ -156,6 +156,7 @@ module.exports = {
   ],
 };
 ```
+
 Important to note:
 
 If you have a `query` section for the `babel-loader` in your `rules`, remember to turn off the `modules` tranformation as well, otherwise HMR won't work.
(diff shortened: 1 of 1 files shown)


content/guides/hmr-react.md
```diff
@@ -157,6 +157,21 @@ module.exports = {
 };
 ```
 
+Important to note:
+
+If you have a `query` section for the `babel-loader` in your `rules`, remember to turn off the `modules` tranformation as well, otherwise HMR won't work.
+
+```js
+loader: "babel-loader",
+query: {
```
skipjack: @Mr-Binary thanks for your contribution! 

This guide is targeted towards latest version of webpack where I believe `query` is deprecated in favor of `options`. Also, this guide already mentions this in the [babel config](https://webpack.js.org/guides/hmr-react/#babel-config) section... maybe just a short warning pointing users back to the config section would be enough? For example:

```md
W> If you are using the `options` key for your babel configuration, make sure you turn babel's `modules` feature off as mentioned in the [config section](https://webpack.js.org/guides/hmr-react/#babel-config).
```

However, I'm not sure this is even necessary with the whole config section being there. I think we usually tend to recommend using separate configs (like a `.babelrc`) instead trying to inline everything with `options`. @bebraw @simon04 thoughts?

Mr-Binary: Hi, Thanks for reading @skipjack 
I read that [part](https://webpack.js.org/guides/hmr-react/#babel-config) before. The whole document is very clear and smooth.

But it is good for a new user but not an existing config. That's why it cost me some time to figure it out. Yes, it's a mistake by the user rather than the documents. But it would be better if it covers that part.

But I think your suggestion does make sense. If you guys think this still needs to be added, I could do a new commit. :)

bebraw: Maybe you could rewrite the idea as a warning (`W>`). Perfect for secondary info.

Mr-Binary: @bebraw Thanks, have pushed.

Mr-Binary:
@skipjack  Thanks :)

PR #1146, opened 2017-04-23
Resolve Anchor Issue

### LINKED ISSUE #1126: Sidebar: anchors with links
Fetched content from `webpack-contrib` plugins/loaders does not show up nicely when a header contains a link, such as https://webpack.js.org/loaders/coffee-loader/#-literate-http-coffeescript-org-literate-

2017-04-19-124925_1198x947_scrot

Related: https://github.com/webpack-contrib/bundle-loader/pull/47

skipjack:
Fix #1126 

Remove markdown syntax and link content from ID...

image

 and title...

image

Also including some cleanup like moving all anchor parsing to a single function, so when the regex needs updating, we won't have to add it in two places.

### DIFF
--- a/components/sidebar-item/sidebar-item.jsx
+++ b/components/sidebar-item/sidebar-item.jsx
@@ -38,9 +38,10 @@ export default class SidebarItem extends React.Component {
           <ul className={ `${block}__anchors` }>
             {
               anchors.map(anchor => (
-                <li 
-                  className={ `${block}__anchor` } 
-                  key={ `anchor-${title}-${uniqueId()}` }>
+                <li
+                  key={ `anchor-${title}-${uniqueId()}` }
+                  className={ `${block}__anchor` }
+                  title={ anchor.title }>
                   <a href={ this._generateAnchorURL(anchor) }>
                     { anchor.title }
                   </a>
--- a/components/sidebar-item/sidebar-item.jsx
+++ b/components/sidebar-item/sidebar-item.jsx
@@ -21,15 +21,15 @@ export default class SidebarItem extends React.Component {
     return (
       <div className={ `${block} ${openMod} ${disabledMod}` }>
         { anchors.length > 0 ? (
-          <i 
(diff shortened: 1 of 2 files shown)


simon04:
Works great for the first scenario (Markdown links), but I identified a second one (HTML links: `<a>`): https://github.com/webpack/webpack.js.org/issues/1126#issuecomment-296439570. I like the code extraction to the `parseAnchor` function.

skipjack:
@simon04 stripped HTML out as well and added a `title` attribute to each `SidebarItem` so that if there is overflow, it is at least accessible.

I'm a bit wary of these fixes and I feel like there might be a more stable solution by either:

- parsing it to full HTML using marked and then extracting all text nodes from that html (although I feel this could get pretty hacky as well)
- or maybe trying to set some general guidelines for all webpack (or webpack-contrib) markdown at some point... I was reading into [another webpack repo][1] the other that seemed to be dealing with enforcing some general guidelines so maybe a conversation about this over there would be appropriate

Anyway this should be good to go for now, but probably something we could revisit down the road (maybe when doing some general cleanup on our markdown processing).

[1]: https://github.com/webpack-contrib/webpack-defaults

skipjack:
@simon04 this good to go or did you have anything else in mind?

simon04:
Awesome. Sorry that I haven't responded earlier, I didn't find time in the past days.

skipjack:
@simon04 no worries. I really want to start a move to begin knocking down the main issues (mostly content related) now. I feel once we get more of the core documentation in place, we can start cleaning things up a bit so this repo is easier to maintain and add to going forward.

For this reason, I may start merging things a bit quicker if I can't get a review within a few days. Definitely let me know if I jump the gun on anything.

simon04:
This is perfectly fine with me. Unless we completely break the website (the CI is a good indicator), we can always fine tune or fix minor glitches a few days layer. :smile:

skipjack:
Yeah, at the very least I'll always make sure that CI passes before merging.

Issue #722, opened 2017-01-25
Chinese doc translation issue.


lcxfs1991:
Hi @bebraw and @TheLarkInn.

We are recently having trouble with managing our translation docs. As we know, `webpack` doc repo does not keep loader and plugin docs itself. So we have to translate docs under `generated` folder which  makes us harder to make a diff.

Would you think of keeping those docs under `content` folder ? i.e., run a program to regularly update docs from each loader and plugin.

TheLarkInn:
There's also the possibility to doing Kron jobs on Travis. Thoughts?

bebraw:
A cron job would have to perform sync commits against master. I'm not sure if that's a smart idea.

Would it make sense to set up an intermediate repository like that? You would generate the directory there, and whenever you want to update the translation, you would trigger the generation script against that, get a diff, and then translate. I think this might give you the information you need to keep it up to date without complicating the setup of the main site.

lcxfs1991:
We are having this [repo](https://github.com/webpack-china/webpack.js.org) for translation. 
That would be great if we could have such an intermediate repo. Then we could run a script to update the content ourselves.  I think it would also help if other developers want to translate the doc into other languages.

bebraw:
@lcxfs1991 Ok, I'll try to set up one below `webpack-contrib` once I find a bit of time. I'll add you to the project too so you can update it etc.

lcxfs1991:
Thank you so much. 

Please also add `dear-lizhihua` who also leads the translation project.

TheLarkInn:
Thanks @bebraw. 🙇

lcxfs1991:
hey @bebraw.

We have another suggestion  that a new branch `translation` can be created under `webpack.js.org` based on branch `master`. 

In branch `translation`, we can manage a script to update docs from loaders/plugins manually. On the one hand, we can update major docs of `webpack` if `master` branch is updated. On the other hand, once there is an update from loaders/plugins doc, we can merge content from `translation` to our translation repository.

Managing another branch may be easier than another new repository.

bebraw:
@lcxfs1991 Is it enough if I set up a branch with proper rights for you to handle then? That sounds simpler, yeah.

lcxfs1991:
yes man.

bebraw:
@lcxfs1991 Ok. I set up the branch and sent you an invite. You should be able to write to the branch.

lcxfs1991:
thanks!

Issue #295, opened 2016-11-03
Translation - Use GitLocalize


chikathreesix:
Hi, I am Ryo, a creator of Locki.
Thanks @bebraw for allowing me to open this issue.

Locki(https://locki.io) is a website localization tool that enables your localization community to easily translate your docs on your website. Check out [this video](https://youtu.be/c21-ac_26sk) to learn more about how it works.

Now Node.js is also considering using Locki(nodejs/node#8798) and [here is the test integration](https://locki-nodejs-test.herokuapp.com/api/) so try this out.

Feel free to ask me any questions!
We'd be happy if we can help Webpack's localization challenges.

lcxfs1991:
@chikathreesix hi Ryo, this is interesting.

1. I have a question for this. Is the owner has the right to approve or reject the translation? 
2. Where is the translation data? Is it stored in locki's server?

chikathreesix:
Thanks @lcxfs1991!

> 1.
We don't have an approval process for everytime. But you can take 2 types of approaches for this.

a) The owner can restrict the edit only to the specified members so it will be fully under the control of the team. And we will enable users out of the team to suggest translations so that the owner and the team can approve or reject those suggestions (this is under the development now).

b) We will introduce a rollback feature soon to reject a translation that the owner doesn't want.

> 2.
These are stored on our server but you can fetch the JSON anytime from API. We will enable you to easily download them too.

sokra:
Any plan to tackle the FOUC (here: Flash Of Unlocalized Content)? (You see the english version for a second)

---

Have you considered using a git repo as backing store for the translation. Each change is commited to a repo of the users choice, permissions from repo are used. This would make intergration pretty easy and looks safer as data storage from user perspective.

lcxfs1991:
@sokra  @sokra 
Agree.

Guess we can create a folder/content/ lang/zh, /content/ lang/jp for contributors to translate the doc.  And nominate someone responsible for reviewing the translation version.

Then after Locki resolve the above problems, we can consider migrate to that system.

bebraw:
Organization-wise it might be interesting to push the translations to repositories of their own and consume them as a dependency to the main project. Easier to handle perhaps as you can keep language specific issues there.

sokra:
[...]

PR #1148, opened 2017-04-23
add multiple language selection

lcxfs1991:
Preview:

language-webpack

Fixes #1148


lcxfs1991:
@bebraw Please take a look

skipjack:
@bebraw @sokra @TheLarkInn @lcxfs1991 also, from a higher level, how does internationalization play out in terms of issue management, prs, etc?

I know right now there's a `translation` branch in this repo and there's been some talk of using locki, #295, but I'm not sure where we're at with that. It seems there's [a separate repo](https://github.com/webpack-china/webpack.js.org) for the chinese fork (which I think makes sense for management purposes) but then what's the local `translation` branch for?

lcxfs1991:
Local translation branch is for a separate management just like webpack-china since we have to fetch plugin/loader readme and put them under `/content/` folder. However, the master branch generate directly. It is a bit hard for us to make a diff if we simply follow what master do.

I think we can try locki. But we need to ensure the performance and we'd better keep a local copy of data.

skipjack:
@lcxfs1991 ah ok thanks for the clarification, I did see something about that in #722 I think. In terms of the getting this PR merged please see my comments.

@bebraw any reason we can't just keep that change in the master branch? I actually kind of like having all site content, including fetched content, in the `/content` directory. Is this to keep a visual distinction for contributors so they know not to edit it?

lcxfs1991:
Let me fix this tonight.

skipjack:
Ok great, no rush.

lcxfs1991:
width: 768

width: 1024

hover: 
language

skipjack:
@lcxfs1991 awesome work, I will review again (and hopefully merge) shortly!

PR #980, opened 2017-03-05
Port to webpack 3 and Upgrade Antwar

bebraw:
Port site to webpack 3 and a new version of antwar that fixes a variety of bugs. What's left:

- [x] fix `<title>` tags, none of which are populated
- [x] fix `introduction` sidebar links
- [x] fix activation of current page in sidebar (i.e. displaying anchors)
- [x] fix mobile sidebar (pretty much empty for some reason, maybe due to title issue)
- [x] fix terminated build issue (not sure what's going on there 😕)

Resolves #720 (once titles are fixed, remember to double check vote title)

PR #1153, opened 2017-04-25
Synchronize Content and Routing Structure

skipjack:
Still a WIP...

Currently our `/content` directory doesn't exactly reflect our routing structure. This is an attempt to fix that and clean up a variety of other oddities and hacks related to content and routing. It won't be a perfect fix, as there are a few antwar tickets open that would allow us to simplify things even further and add other functionality like the fully dynamic routing/content structure proposed [here][1] but it will take us one step closer to that so things are easier to migrate later on.

Here are the things I'd like to accomplish with this PR:

- [ ] Rebase on master once #980 is finished
- [x] Synchronize content and routing structure... the big one being nesting _api_, _configuration_, _development_, loaders, and _plugins_ all under a new `/documentation` route.
- [x] Update all internal links broken by the above change.
- [ ] Add redirects for all external links that would be broken by the above change.
- [x] Update navigation links both in the `Navigation` and `SidebarMobile` components... the mobile sidebar is probably in the worse state of these two as some links are just blank. I'm hoping to achieve a more dynamic result but this is another one that will be facilitated by certain antwar updates like antwarjs/antwar#124.
- [ ] Add link to new starter-kits guide which is still hidden.

This will take some serious review when finished as it is a major change to the site. However it will knock out some things from #1039 and should provide a more intuitive layout for future (and present) contributors.

[1]: https://github.com/webpack/webpack.js.org/issues/438#issuecomment-282508376


skipjack:
Pretty much just edit and generated links failing now... I'm going to start on the UI work and redirects.

skipjack:
Closing as this has become fairly dated and I'm not sure when #980 will be completed -- I'll either re-open or make similar changes down the road.

PR #1154, opened 2017-04-25
Remove Obsolete `ident` Section

### LINKED ISSUE #1150: Guides(migrating/#complex-options): remove ident
The `ident` (Complex Options) is obsolete since `webpack >= v2.2.1` and gets automatically added internally by webpack

skipjack:
Added a warning instead that points to the v2 version in which this option
became obsolete. We could also reference an older version of the docs if
necessary for people to jump back to, though that may be overkill.

Fixes #1150

### DIFF
--- a/content/guides/migrating.md
+++ b/content/guides/migrating.md
@@ -622,4 +622,17 @@ __webpack v1__ only supports `JSON.stringify`-able options for loaders.
 
 __webpack v2__ now supports any JS object as loader options.
 
-W> Before [v2.2.1](https://github.com/webpack/webpack/releases/tag/v2.2.1) (i.e. from v2.0.0 through v2.2.0), using complex options required using `ident` for the `options` object to allow its reference from other loaders. This was removed in v2.2.1 and thus current migrations do not require any use of the `ident` key.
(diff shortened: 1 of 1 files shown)


content/guides/migrating.md
```diff
@@ -618,35 +618,8 @@ Loaders are now cacheable by default. Loaders must opt-out if they are not cache
...
+__webpack v1__ only supports `JSON.stringify`-able options for loaders.
 
-Using complex options comes with one restriction. You may need to have a `ident` for the option object to make it referenceable by other loaders.
+__webpack v2__ now supports any JS object as loader options.
 
-Having an `ident` on the options object means to be able to reference this options object in inline loaders. Here is an example:
-
-`require("some-loader??by-ident!resource")`
-
-``` js
```
michael-ciniawsky: Maybe move this example under the note, so folks know how it looked like :). Does \```diff work on the docs page ?

**webpack.config.js**
```diff
{
  test: /\.ext/
  use: { 
    loader: '...',
    options: {
-     ident: 'id',
      fn: () => require('./foo.js')
    }
  }
}
```

skipjack: > Does ```diff work on the docs page ?

It does, I'll update.

skipjack:
@michael-ciniawsky done 👍 

image

Issue #1100, opened 2017-04-11
Building for production guide error


maasha:
I am following the guide: https://webpack.js.org/guides/production-build/

The

```
function buildConfig (env) {
  return require('./config/' + env + '.js')(env)
}

module.exports = buildConfig
```

Results in

```
/Users/maasha/install/src/rpncalc/webpack.config.js:2
  return require('./config/' + env + '.js')(env)
                                           ^

TypeError: require(...) is not a function
    at buildConfig (/Users/maasha/install/src/rpncalc/webpack.config.js:2:44)
    at requireConfig (/Users/maasha/install/src/rpncalc/node_modules/webpack/bin/convert-argv.js:102:15)
    at /Users/maasha/install/src/rpncalc/node_modules/webpack/bin/convert-argv.js:109:17
    at Array.forEach (native)
    at module.exports (/Users/maasha/install/src/rpncalc/node_modules/webpack/bin/convert-argv.js:107:15)
    at Object.<anonymous> (/Users/maasha/install/src/rpncalc/node_modules/webpack/bin/webpack.js:153:40)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
```

skipjack:
Are you passing `--env=dev` and `--env=prod`? Can you log out the what's returned from `require('./config/' + env + './js')`? The guide is expecting this will resolve correctly will return a function in both cases.

That said, I think this guide needs some work and could probably be simplified a bit. For example, the `env` parameter is never used in `dev.js` or `prod.js`.

skipjack:
@bebraw @simon04 do we have example/boilerplate repos backing any of these guides? Maybe something we should try out or encourage going forward. This way people would have something to go off when things break down and we could point them to a working example when something goes wrong. 

However keeping the guide and repo in sync may be tough as well, so I'm not sure this is necessarily the perfect answer either.

maasha:
Ah, sorry, I have wiped the code since I couldn't get it to work. But would `require('./config/' + env + './js')` not cause the same err?

skipjack:
[...]

PR #1155, opened 2017-04-25
Update Production Build Guide

skipjack:
Rework and simplify the production build guide to address some of the issues discussed in #1100 as well as just doing some general cleanup of titles and whatnot.

Also sneaking a UI/UX fix for the contributors section in here...

image

to...

image

This puts some space between rows and prevents them from being overlapped by the gitter button.

### DIFF
--- a/components/contributors/contributors-style.scss
+++ b/components/contributors/contributors-style.scss
@@ -1,15 +1,17 @@
 @import 'functions';
+@import 'mixins';
 
 .contributors__list {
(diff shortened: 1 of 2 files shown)


skipjack:
@bebraw thanks for the review, yeah on a normal day I would push it separately but I was trying to hit a lot of things last night. It is somewhat related to this PR as the guide I edited has a ton of contributors.

PR #1158, opened 2017-04-26
Added info about pitfall for externals

richardsimko:
Added some info about a pitfall mentioned in webpack/webpack#4771 which is good to keep in mind until the issue is fixed.


richardsimko:
I will! The change in webpack will most likely only only add an error message and makanother issue easier to discover.

PR #1159, opened 2017-04-27
Support Cleanup

### LINKED ISSUE #411: Randomize Open Collective sponsors and backers
The [current implementation](https://github.com/webpack/webpack.js.org/blob/0fe4e32067940ebd106ed73b591a56ea2d912048/components/support/support.jsx) just asks Open Collective for the first 5 sponsors and the 20 first backers. Unfortunately, the order is unclear. It's not alphabetical, it's not chronological – but it's also not random.

Although this might be a small detail, it's quite important for corporate sponsors to be visible on the page. We should randomize the occurrence.

skipjack:
Cleaning up the support component by... 

- updating the way sponsors are retrieved using the method mentioned in #411 
- abstracting the additional sponsors into a separate json sheet that follows the same model
- concatenating the additional sponsors on to those fetched from open collective

This gives us full control over styling and allows us to...

- easily add new additional sponsors, not from open collective, without bloating the component file
- refactor how we display sponsors, e.g. using one of the methods discussed in #411 

I think this should resolve #411 as it has been inactive for a while __and__ with this in place changing the display algorithm should be a piece of cake as we have more data to key off.

Thanks to @rafaelrinaldi for pinging opencollective to find out about the json feature!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,6 +4,7 @@
     "react"
   ],
   "plugins": [
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "transform-class-properties"
   ]
 }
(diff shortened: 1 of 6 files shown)


skipjack:
Should've run a build-test ☹️ , this broke the support section as I think it needed to be an `Interactive` component. The `fetch` is never executed and therefore no sponsors are displayed. Testing a fix now -- if that doesn't work, I'll revert this pr.

PR #1165, opened 2017-04-28
Update migrating.md

davidctj:
seems like UglifyJsPlugin default `warnings` is true... is the documentation outdated?


simon04:
https://github.com/webpack/webpack/blob/87702a2/lib/optimize/UglifyJsPlugin.js#L87-L89 suggests the opposite, i.e., the current statement seems true. See also https://github.com/webpack/webpack/commit/ffd2d37efd3f258be2eb15271b9e8d018c6481fc

PR #1168, opened 2017-04-29
Supporters and PrismJS Lang Fixes

skipjack:
Temporarily fix\ing a bug created by #1159. Need to discuss w @bebraw before adding that back in.
Also removing an unsupported prismjs language.


skipjack:
@bebraw I tried to fix the `<Support>` issue by using two `<Interactive>` components but it seems that isn't supported by antwar-interactive (I guess because both use the same `id`?). Are any of the lifecycle methods called where I could call the `fetch` so this can execute during the build? If so, I guess I could use that in combination with the [isomorphic-fetch package](https://github.com/matthew-andrews/isomorphic-fetch)?

Or is there any way to bypass the `<Interactive>` issue?

Issue #1167, opened 2017-04-28
Sponsor/backer icons ↔ links are off by one


simon04:
Clicking on the second icon (with dotted border in the screenshot; having id 1 in the URL), I reach the opencollective profile page corresponding to the first icon.

![](https://cloud.githubusercontent.com/assets/782446/25549214/ea91b444-2c70-11e7-927e-ed912dab8e8e.png)

skipjack:
@simon04 can you test #1159 and see if this resolves that issue? (I bet it does)

simon04:
Yes, this issue is fixed with #1159.

PR #1170, opened 2017-04-29
Support Refactor

skipjack:
Still a WIP...

Re-implementing the reverted refactor/fix of the `Support` component. 

@bebraw moving the conversation from #1168 to here, my attempts so far have failed. Please let me know any suggestions, if we don't get this in somehow we'll likely have to re-open #1167 and #411.

### DIFF
--- a/components/support/support-additional.json
+++ b/components/support/support-additional.json
@@ -1,0 +1,18 @@
+[
+  {
+    "id": 20000,
+    "name": "MoonMail",
+    "username": "moonmail",
+    "tier": "sponsor",
+    "avatar": "https://static.moonmail.io/moonmail-logo.svg",
+    "website": "https://moonmail.io/?utm_source=webpack.js.org"
+  },
+  {
+    "index": 20001,
+    "name": "MONEI",
+    "username": "monei",
+    "tier": "sponsor",
+    "avatar": "https://static.monei.net/monei-logo.svg",
+    "website": "https://monei.net/?utm_source=webpack.js.org"
+  }
+]
--- a/components/support/support-additional.json
+++ b/components/support/support-additional.json
@@ -8,7 +8,7 @@
     "website": "https://moonmail.io/?utm_source=webpack.js.org"
   },
   {
-    "index": 20001,
+    "id": 20001,
     "name": "MONEI",
     "username": "monei",
     "tier": "sponsor",
--- a/components/support/support-additional.json
+++ b/components/support/support-additional.json
@@ -1,6 +1,5 @@
 [
   {
-    "id": 20000,
     "name": "MoonMail",
     "username": "moonmail",
     "tier": "sponsor",
@@ -8,7 +7,6 @@
     "website": "https://moonmail.io/?utm_source=webpack.js.org"
   },
   {
-    "id": 20001,
     "name": "MONEI",
     "username": "monei",
     "tier": "sponsor",
--- a/components/support/support.jsx
+++ b/components/support/support.jsx
@@ -1,17 +1,17 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+// import PropTypes from 'prop-types';
 import Additional from './support-additional.json';
-import 'whatwg-fetch';
+import 'isomorphic-fetch';
 import './support-style';
 
 export default class Support extends React.Component {
-  static propTypes = {
-    number: PropTypes.number.isRequired,
-    type: PropTypes.oneOf([
-      'sponsors',
-      'backers'
-    ]).isRequired
-  };
+  // static propTypes = {
+  //   number: PropTypes.number.isRequired,
+  //   type: PropTypes.oneOf([
+  //     'sponsors',
+  //     'backers'
+  //   ]).isRequired
+  // };
 
   state = {
     supporters: [],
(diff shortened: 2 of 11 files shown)


bebraw:
That `fetch` doesn't feel right. I wonder if it would be possible to push the problem outside of React and manage it on build like we do for loaders/plugins/etc.

skipjack:
Ah like generate a json file on every build cycle that contains data?

bebraw:
Yeah, it doesn't have to be more complex than that. Also avoids some requests for the client.

skipjack:
I'll update with that change and re-test.

skipjack:
@bebraw just moved it to the `npm run fetch` process. Please take a look when you have a chance to make sure everything looks sound.

skipjack:
@bebraw note, we should definitely squash and merge this one, once you've reviewed.

package.json
```diff
@@ -98,6 +98,7 @@
     "yaml-frontmatter-loader": "0.0.3"
   },
   "dependencies": {
+    "isomorphic-fetch": "^2.2.1",
```
bebraw: Please drop this as the implementation doesn't need it anymore.

skipjack: Good catch.

skipjack:
@bebraw should be good to go now.

components/support/support.jsx
```diff
@@ -1,56 +1,50 @@
 import React from 'react';
+// import PropTypes from 'prop-types';
+import Additional from './support-additional.json';
+import 'isomorphic-fetch';
```
bebraw: No need for `import 'isomorphic-fetch';`.

skipjack: ah right again!

skipjack:
@bebraw ok now it should be good to go (hopefully 😁 )

components/support/support-additional.json
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "id": 20000,
```
bebraw: Where do the ids come from? Is there any chance these could clash in the future?

components/support/support-additional.json
```diff
@@ -0,0 +1,18 @@
...
+    "name": "MoonMail",
+    "username": "moonmail",
+    "tier": "sponsor",
+    "avatar": "https://static.moonmail.io/moonmail-logo.svg",
```
bebraw: Would it be better to host the svgs on webpack site instead? I can see pros/cons to both and we can stick with the current solution.

components/support/support.jsx
```diff
@@ -1,56 +1,49 @@
 import React from 'react';
+// import PropTypes from 'prop-types';
```
bebraw: Maybe drop the commented bits?

components/support/support-additional.json
skipjack: Idk, I kind of like keeping these external for now. I think our asset management still has some room for improvement so I'd like to avoid adding any more than there already are for now. Plus, this stays consistent with the rest of the supporter images which are also fetched externally from the opencollective urls.

components/support/support.jsx
skipjack: Sure... I was leaving it in for now as kind of a TODO. I'd like to add prop type checking to some of our components at some point but I'm fine with curbing this for now.

components/support/support-additional.json
skipjack: > Where do the ids come from?

Hardcoded as I was trying to stay consistent with the opencollective data. And they're used for the `key` prop.

> Is there any chance these could clash in the future?

Potentially, yes. I tried to bump them pretty high to avoid that.

In general I like to avoid `key={ index }` everywhere, hence the ids. In our case, it currently doesn't matter too much for non-interactive components as a `render()` will never be called again on the client side. I'm ok with `key={ supporter.id || index }` for now and maybe shifting back to something like this, except using higher ids or uuids to prevent collisions, down the road.

skipjack:
@bebraw pushed some more tweaks based on your comments.

skipjack:
Seeing the same error now as on #1177... @bebraw any idea what's going on there?

bebraw:
`failure.diag` can be `undefined` at `scripts/check-links.js` apparently. I think you have to add a check against that. I'm not sure if that fixes everything, though.

skipjack:
Yep, debugging the issue now... hopefully once this is resolved we can get this PR (and the other) merged finally. I'm also going to remove that `--exclude` as the support urls have changed now that we're pulling in the raw data.

skipjack:
@Munter any idea what the `FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory` is about? Seems to be getting thrown from somewhere within `hyperlink`.

Munter:
@skipjack Haven't seen that one before. But since hyperlink is not particularly clever about unloading the assets it has visited and out of memory error seems probable. @papandreou is this specific error known to you?

papandreou:
Yeah, that's how V8 dies when the JavaScript heap exceeds the maximum allowed size (even after performing a full GC). The maximum size is specified using the `node --max-old-space-size=<numMegaBytes>` switch and defaults to 2000 (2 GB).

That could either indicate that the combined size of the assets that hyperlink is loading has increased so it doesn't fit anymore, or that there is a bug that means it'll run off and follow external relations/redirects infinitely.

As the first thing I would try invoking it as `node --max-old-space-size=5000 hyperlink ...` to rule out the latter.

Munter:
There's a bunch of things that hyperlink could do to free up memory during the run. Once an asset has been visited it's never used again. So I could potentially unload it. Similarly I might be able to avoid loading assets that are known to not contain external relations and are huge, like videos.

PR #1175, opened 2017-04-29
Shimming Node Functionality

### LINKED ISSUE #153: Shimming: Node Built-Ins
_From @wearhere on August 14, 2016 22:53_

**I'm submitting a feature request**

**Webpack version:**
1.13.1

**Please tell us about your environment:**
OSX 10.11.5

**Current behavior:**
Webpack's automatic shimming of Node built-ins is not documented [so far as I can tell](https://www.google.com/search?espv=2&q=webpack+shim&oq=webpack+shim&gs_l=serp.3..0i71k1l8.0.0.0.1514941.0.0.0.0.0.0.0.0..0.0....0...1c..64.serp..0.0.0.9gqCe882Img).

**Expected/desired behavior:**
- **What is the expected behavior?**
  
  It would be documented, specifically (my stabs at documentation follow the colons):
  - how it's triggered: you `require` a Node built-in or use a Node global anywhere in
    your input to Webpack
  - what modules/globals are shimmed: so far I have seen `os`, `crypto`, and `Buffer`
  - how to disable this shimming: define the module as an [external](https://webpack.github.io/docs/library-and-externals.html)
    or use `DefinePlugin` to wrap the `require` in a [dead branch](https://github.com/webpack/webpack/issues/99#issuecomment-24985486)
- **What is the motivation / use case for changing the behavior?**
  
  I'm actually not advocating changing the automatic shimming, I just think it should be documented. It can be very nice, but is somewhat surprising, and you might accidentally bundle a huge shim using code like:
  
  ``` js
  var os = require('os');
  var crypto;
  if (os.platform() === 'browser') {
    crypto = window.crypto;
  } else {
    crypto = require('crypto');
  }
  ```
  
  not realizing that you need to either mark `'crypto'` as an external or `define`-out the latter branch.
  
  Also if the shims used aren't documented, it's not clear where you'd go to file bugs against them (is Webpack responsible for them per se?)
- **Browser:** all
- **Language:** all

_Copied from original issue: webpack/webpack#2871_

skipjack:
Add note to shimming guide about polyfilling node.

Fixes #153

cc @wearhere @jrvidal

### DIFF
--- a/content/guides/shimming.md
+++ b/content/guides/shimming.md
@@ -27,6 +27,7 @@ module.exports = {
 };
 ```
 
+
 ## `ProvidePlugin`
 
 The [`ProvidePlugin`](/plugins/provide-plugin) makes a module available as a variable in every other module required by `webpack`. The module is required only if you use the variable.
@@ -43,7 +44,7 @@ module.exports = {
 };
 ```
 
-This plugin is also capable of providing only a certain export of a module by configuring it with an array path using this format: `[module, child, ...children?]`. 
+This plugin is also capable of providing only a certain export of a module by configuring it with an array path using this format: `[module, child, ...children?]`.
 
 The following configuration will correctly import function `__assign` from TypeScript's `tslib` package, and provide it wherever it's invoked.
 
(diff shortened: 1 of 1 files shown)


wearhere:
Cool thanks!

PR #1176, opened 2017-04-30
fixtypo

pwnn:
`Condition.not` only accept array of `Condition`s.


skipjack:
@pwnn can you sign the CLA agreement?

bebraw:
Good to go after CLA.

PR #1181, opened 2017-05-02
Fix HMR module.hot.accept example code

jasonnutter:
The example shows the `module.hot.accept` callback re-rendering the original component, instead of re-requiring the component in order to render the updated component.

### DIFF
--- a/content/guides/hmr-react.md
+++ b/content/guides/hmr-react.md
@@ -190,7 +190,8 @@ render(App);
 // Hot Module Replacement API
 if (module.hot) {
   module.hot.accept('./components/App', () => {
-    render(App)
+    const updatedApp = require('./components/App').default;
+    render(updatedApp)
   });
 }
 ```


bebraw:
I think it's on purpose that way. It should work without a new `require`. I found this surprising in the past but it's correct.

jasonnutter:
@bebraw I don't believe that's correct. I just watched a coworker try setting up HMR without re-requiring and the components didn't update until the component was re-required.

See: https://gist.github.com/gaearon/a1c079f4d985015abe84/

Which was posted in this:
https://medium.com/@dan_abramov/hot-reloading-in-react-1140438583bf

bebraw:
The setup was simplified with a bit in webpack 2 as it manages `import`. Without having specific code to run, it's hard for me to say what's going on. In my [own example](https://survivejs.com/webpack/appendices/hmr-with-react/) just forcing `render` is enough.

skipjack:
@bebraw is right, I've implemented HMR with react recently and a new `require`/`import` was not necessary. It's not [what's recommended](https://github.com/gaearon/react-hot-loader/tree/master/docs#webpack-2) in react-hot-loader's documentation either (although it was necessary for webpack 1 I believe).

PR #1185, opened 2017-05-03
Asset Management

### LINKED ISSUE #615: Document Asset Bundling (Files/Images/Fonts)

skipjack:
Added the basic scaffolding and introductory content for this guide. Still need to flesh out the rest of the content.

Resolves #615

### DIFF
--- a/content/guides/asset-management.md
+++ b/content/guides/asset-management.md
@@ -79,17 +79,87 @@ T> The next step is minifying and optimizing your images. Check out the [image-w
 
 ## Loading Fonts
 
-...
+So what about other assets like fonts? The file and url loaders will take any file you load through them and output it to your build directory. This means we can use them for any kind of file, including fonts:
+
+__webpack.config.js__
+
+``` js
+module.exports = {
+  //...
+  module: {
+    rules: [
+      {
+        test: /\.(woff|woff2|eot|ttf|otf)$/,
+        use: [
+          'file-loader'
+        ]
+      }
+    ]
+  },
+  //...
+}
+```
+
+With this in place you can define a font like so...
+
+``` css
+@font-face {
+  font-family: 'MyFont';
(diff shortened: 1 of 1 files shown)


skipjack:
@michael-ciniawsky take a look and let me know if this is along the lines of what you were thinking. The one section I'm probably still missing is something like "Directory/Project Structure" where we could discuss the two approaches (old and new).

content/guides/asset-management.md
```diff
@@ -0,0 +1,57 @@
...
+        test: /\.css$/,
+        use: 'css-loader'
+      }
+    ]
+  },
+  //...
+}
+```
+
+This enables you to `import './style.css'` into the file that depends on that styling. Now, when that module is run, a `<style>` tag with the stringified css will be inserted into the `<head>` of your html file.
```
michael-ciniawsky: Only if `style-loader` is used, not be default :)

content/guides/asset-management.md
```diff
@@ -0,0 +1,57 @@
...
+
+T> Note that you can also [split your CSS](/guides/code-splitting-css) for better load times in production. On top of that, loaders exist for pretty much any flavor of CSS you can think of -- [postcss](), [scss](), and [less]() to name a few.
```
michael-ciniawsky: [postcss-loader](https://github.com/postcss/postcss-loader) (I'm on the finishing line for `postcss-loader` v2.0.0 and it will use our README template, so maybe we should add it to the docs page aswell sonce it's a common loader, but separate PR 😛 )

`less` and `scss-loader` are on the loaders docs page

content/guides/asset-management.md
skipjack: Ah jeez, what was I thinking! Good catch!

skipjack:
@michael-ciniawsky I think this is pretty much good go, let me know your thoughts.

PR #1191, opened 2017-05-04
Fix dev-server config docs typo

sw-double:


content/configuration/dev-server.md
```diff
@@ -122,7 +122,7 @@ T> `filename` has no effect when used without **lazy mode**.
 
 `object`
 
-Adds headers to all requests:
+Adds headers to all responses:
```
sw-double: _requests_ could be misleading as dev-server actually makes requests in proxy mode

PR #1194, opened 2017-05-04
Improve lint:links - exclude support images/links

simon04:
* Ignore failures on sponsor/backer images/links
* Do not allow any other failure to not miss broken links such as https://github.com/webpack/webpack.js.org/pull/1188 or https://github.com/webpack-contrib/less-loader/pull/194

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "fetch": "scripts/fetch.sh",
     "init:generated": "mkdirp ./generated/loaders && mkdirp ./generated/plugins ",
     "lint": "run-s lint:*",
-    "lint:links": "hyperlink -r build/index.html | ./scripts/check-links.js --skip 10",
+    "lint:links": "hyperlink -r build/index.html | ./scripts/check-links.js",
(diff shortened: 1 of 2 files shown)


skipjack:
This is great, unfortunately we still have to figure out this `FATAL ERROR: ...` thing happening in the hyperlink package. I may test the suggestion [here](https://github.com/webpack/webpack.js.org/pull/1170#issuecomment-298858925) if that works as a temporary solution. @Munter maybe we should create an issue for this over at the hyperlink package to track a more stable fix?

skipjack:
@Munter thanks for being so responsive and helpful with this package... just created Munter/hyperlink#117 so we can track this.

simon04:
Currently the check correctly fails due to one broken link. Maybe we should wait for https://github.com/webpack-contrib/less-loader/pull/194 to be merged?

skipjack:
@simon04 yea I'm good with that, I'll leave this to you then.

simon04:
Since the other PR is merged, I'll merge this one as well. :smile:

PR #1195, opened 2017-05-06
global ==> root

jingsam:


simon04:
Please name a source for this change! https://github.com/webpack/webpack/blob/6afc397/lib/ExternalModule.js#L78-L95 contradicts the use of `"root"`.

skipjack:
Closing... @jingsam feel free to re-open if you can name a source for your change as @simon04 mentioned.

PR #1196, opened 2017-05-06
Fixed examples for CLI-only DevServer options

ingver:
I have changed all JS examples for CLI-only DevServer options to Bash to avoid confusion.
I've also noticed that `port` option despite CLI-only flag in documentation could be used inside JS-config too.


simon04:
Thanks for your contribution! I agree that makes examples more understandable.

However, the "CLI only" tags seem to be greatly outdated. When reading the source code, firstly the [`webpack.config.js` is being parsed](https://github.com/webpack/webpack-dev-server/blob/d2f579c/bin/webpack-dev-server.js#L192-L208), and that config is [updated based on the commandline arguments](https://github.com/webpack/webpack-dev-server/blob/d2f579c/bin/webpack-dev-server.js#L210-L324). So I think, most/all "CLI only" tags should be droppen. @webpack/documentation-team can you confirm?

bebraw:
@SpaceK33z Any thoughts on this?

skipjack:
> However, the "CLI only" tags seem to be greatly outdated. When reading the source code, firstly the webpack.config.js is being parsed, and that config is updated based on the commandline arguments. So I think, most/all "CLI only" tags should be droppen. @webpack/documentation-team can you confirm?

@simon04 I agree... also this is all within the `/configuration` section of our documentation so either it should be removed or moved to a true `/cli` section of our docs once we have one.

skipjack:
@SpaceK33z should we just drop all these CLI flags and simply link to somewhere within the `webpack-dev-server` repository?

skipjack:
Closing in favor of #1468 which was just merged.

PR #1199, opened 2017-05-09
Update yarn.lock file

silvenon:
Not sure why, but this happened when I ran `yarn`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -492,12 +492,29 @@ babel-helper-function-name@^6.22.0, babel-helper-function-name@^6.23.0:
     babel-traverse "^6.23.0"
     babel-types "^6.23.0"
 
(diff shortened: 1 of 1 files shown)


skipjack:
@bebraw do you know what's going on with this? I'm not too familiar with yarn.

bebraw:
Looks like it pulled some new deps. It *might* be ok but you would have to check locally with a fresh install (nuke `node_modules` first).

silvenon:
I ran `yarn` when I cloned the repo, so it didn't have an existing `node_modules`.

simon04:
I can confirm this. Running `yarn` on a fresh clone of this repository, I get a different diff: https://github.com/webpack/webpack.js.org/compare/master...simon04:yarn

PR #1204, opened 2017-05-10
Fix Markdown Issue

### LINKED ISSUE #1187: The markdown table is parsed unexpectedly in css-loader documentation page
In the "Options" section of the [css-loader](https://webpack.js.org/loaders/css-loader/#options) documentation.

image

skipjack:
Fix issue with escaped `|` not being taken into account.

This is allowed in GFM but not supported by marked. See this issue:
https://github.com/chjj/marked/issues/595

Fixes #1187

cc @Alex1990 @NejcZdovc

### DIFF
--- a/utilities/markdown.js
+++ b/utilities/markdown.js
@@ -98,7 +98,7 @@ module.exports = function() {
       tokens.links = [];
 
       marked.Parser.prototype.tok = function () {
-        if(this.token.type === 'table') {
+        if (this.token.type === 'table') {
           return handleTable.call(this, this.token);
         } else {
           return handleTok.call(this);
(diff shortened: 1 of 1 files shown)


utilities/markdown.js
```diff
@@ -234,6 +234,23 @@ function handleTable(t) {
     let row = t.cells[i];
     cell = '';
 
+
+    // Fix escaped '|' characters
+    // See https://github.com/chjj/marked/issues/595
+    let erroneous = row.map(item => item.endsWith('\\') ? item : null).filter(item => item !== null)
```
bebraw: I might write this as

```javascript
let erroneous = row.map(item => item.endsWith('\\') && item).filter(item => item)
```

Same result.

NejcZdovc: ++

PR #1205, opened 2017-05-10
Add postcss-loader

### LINKED ISSUE #1202: Loaders: Add PostCSS Loader
https://github.com/postcss/postcss-loader

skipjack:
Add postcss-loader to the loaders documentation. Did a few things:

- Added a new `fetch` line to pull in the readme (similar to the others)
- Had to drop html comments as that seems to be another thing not supported by our markdown process 😢  (just for reference #178)
- Updated and added some links to the new page

Fixes #1202


michael-ciniawsky:
Yep these particular comments are not necessary, but it isn't a bad idea to sanitize them in general :D

PR #1209, opened 2017-05-13
Module API

### LINKED ISSUE #1099: Critical dependency warning on a require.resolve

> bug

> Critical dependency warning on `require.resolve` call.

> Create a file with following contents and process it with default webpack config:
> 
> ```
> let resolvedPath = require.resolve(`${somepath}`);
> ```

> No warnings expected due to this is not a `require` statement and no external modules are loaded.

–

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**

> macOS 10.12.4 (16E195)
> nodeJS v7.7.2
> webpack v2.3.3

---

This issue was moved from webpack/webpack#4681 by @sokra. Orginal issue was by @dkfiresky.

Document behavior for dynamic expressions.

webpack is a static bundler and requires statically analysable expressions.

skipjack:
Port module api page from old docs.

Took the content from https://webpack.github.io/docs/api-in-modules.html and
started re-organizing it. This is still a work in progress.

Fixes #1099 and adds a significant bit of missing documentation.


skipjack:
@simon04 @bebraw this is still a first draft, but I'm interested in your thoughts on the structure which I changed a bit from the [original page in the old docs](https://webpack.github.io/docs/api-in-modules.html). I think it will read better this way as most things are now grouped by style usage.

skipjack:
@bebraw @simon04 I think this is good to go... can one of you give it another quick skim before I merge? I tried to simplify some of the examples and clean up some of the grammar in the descriptions.

content/api/module.md
```diff
@@ -0,0 +1,399 @@
...
+
+``` javascript
+require: 'some-dependency';
+console.log(answer);
+method(...);
+```
+
+
+
+## Webpack
```
bebraw: Would it make sense to split here and push webpack specifics to another article? It's quite long otherwise.

skipjack: I think maybe it would make more sense to split at the `## Module Variables` section? Until we have that a nicer page grouping (#438) maybe something like `module-api-methods.md` and `module-api-variables.md`?

skipjack: Idk I guess it could work to split out webpack specific functionality as well, I just think it may be odd having to jump around the two articles to see all module variables (methods would probably be ok as people would most likely choose one method and stick with it).

bebraw: Variables is a good split, yeah.

PR #1214, opened 2017-05-14
Change externals module contexts documentation from `global` to `root`

dxu:
The currently deployed documentation indicates that there are 4 module contexts you can specify under the externals config - `global`, `commonjs`, `commonjs2`, and `amd`, which contradicts both the examples that surround it, and the actual library functionality. 

This just changes the word "global" to "root" in the section that lists the module contexts.

### DIFF
--- a/content/configuration/externals.md
+++ b/content/configuration/externals.md
@@ -48,7 +48,7 @@ $('.my-element').animate(...);
 
 The bundle with external dependencies can be used in various module contexts, such as [CommonJS, AMD, global and ES2015 modules](/concepts/modules). The external library may be available in any of these forms:
 
-* __global__ - An external library can be available as a global variable. The consumer can achieve this by including the external library in a script tag. This is the default setting for externals.
+* __root__ - An external library can be available as a global variable. The consumer can achieve this by including the external library in a script tag. This is the default setting for externals.
 * __commonjs__ -  The consumer application may be using a CommonJS module system and hence the external library should be available as a CommonJS module.
 * __commonjs2__ -  Similar to the above line but where the export is `module.exports.default`.
 * __amd__ - Similar to the above line but using AMD module system.


skipjack:
Related #1195. @dxu can you see @simon04's question on that PR?

@simon04 @bebraw it is interesting that two people have brought this up now. Even if this change doesn't make sense, there's probably something that needs to be re-phrased.

dxu:
I'm going off of my own usage in [this repository](https://github.com/dxu/matter-collision-events).

If you change the externals from `root` to `global`, it will break the outputted code:

```
// OUTPUT:
// when specifying {root: "Matter"}
-               root["MatterCollisionEvents"] = factory(root["Matter"]);
// when specifying {global: "Matter"}
+               root["MatterCollisionEvents"] = factory(root[undefined]);
```

I'm pretty sure it's due to [these lines of source code in the webpack repository](https://github.com/webpack/webpack/blob/23a2a159cdd0a5a3b9e2b39a99a52460012fe7bd/lib/UmdMainTemplatePlugin.js#L153-L155). I'm not familiar with the webpack codebase, so I don't really know what the switch statement that @simon04 linked to in the other diff does.


Regardless of what you guys end up changing, there is a discrepancy between the examples in the documentation and the documentation itself.

Take a look at these screenshots from [this page](https://webpack.js.org/configuration/externals/):
screen shot 2017-05-14 at 3 56 52 pm
screen shot 2017-05-14 at 3 56 48 pm

Hope this helps!

skipjack:
@dxu good catch, thanks for the detailed explanation! That code block with the `root` definition vs. what's shown the list is definitely the cause of the confusion. I'll defer to @simon04 and @bebraw for this as I'm not as familiar with this section of webpack either. It's good to know that `root` is the only one that's been working for you in practice.

sokra:
`root` is correct. `global` only exists in these configuration options: `output.libraryTarget: "global"` and `node.global`. Everywhere else it's `root`. (Maybe a bit weird naming...)

> This is the default setting for externals.

The default depends on the `output.libraryTarget` option.

> ``` js
> externals : {
>   lodash : {
>     commonjs: "lodash",
>     amd: "lodash",
>     root: "_" // indicates global variable
>   }
> }
> ```

This is only valid with `output.libraryTarget: "umd"`. It should moved as it's weird in the `object` section.

> ``` js
> externals: /^(jquery|\$)$/i
> ```

Here should be a note added, that this map input to identical output.

i. e. `jquery` maps to `jquery`, `jQueRy` maps to `jQueRy`.

This also makes it a weird example because it doesn't make sense of jquery.

A good example would be i. e.:

``` js
output: { libraryTarget: "commonjs2" }
externals: /^(react|lodash)/
```

which maps: `react` to `react`, `react-dom/server` to `react-dom/server`, ...

PR #1215, opened 2017-05-14
Add link to react-markdown loader list

fernandopasik:

### DIFF
--- a/content/loaders/index.md
+++ b/content/loaders/index.md
@@ -40,6 +40,7 @@ Loaders are activated by using `loadername!` prefixes in `require()` statements,
 * `pug-loader` Loads Pug templates and returns a function
 * `jade-loader` Loads Jade templates and returns a function
 * `markdown-loader` Compiles Markdown to HTML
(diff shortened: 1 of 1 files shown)


fernandopasik:
We developed for use only in webpack, it doesn't work as standalone.

skipjack:
Is it a loader itself though, or just used in conjunction with the `json-loader`? Just to clarify, I'm not saying we can't include it somewhere in the documentation -- we love to promote useful tools in the webpack ecosystem -- I'm just not sure the loaders page is best.

fernandopasik:
Thanks for your help @skipjack! It is a loader itself, it's not used in conjunction with any other loader.
It's used with markdown files, all the content gets into a react component, and as a bonus if there is a react code block it gets also executed. It's ideal for building styleguides. That is why we added it in the previous webpack site in the templating loaders list.

skipjack:
Ah I see, I was looking at the [wrong package](https://github.com/rexxars/react-markdown)!

fernandopasik:
Thank you @skipjack !!

PR #1220, opened 2017-05-16
Add "Enabling tree-shaking" section

### LINKED ISSUE #995: Typescript + Tree shaking configuration
Relevant page: https://webpack.js.org/guides/webpack-and-typescript/

If I understand this correctly, this setup will not trigger webpack's tree shaking algorithm, because the loader will produce a CommonJS-stye `require`, not static `import`s required by [tree shaking](https://webpack.js.org/guides/tree-shaking/).

If I'm correct, something like this should be used for tree shaking:

----

1. `npm install babel-core babel-preset-es2015`
2. Add `.babelrc` with 

```
{
  "presets": [
    [
      "es2015",
      {
        "modules": false
      }
    ]
  ]
}
```

3. `tsconfig.json` should be set to `es6`

```
{
  "compilerOptions": {
    "outDir": "./dist/",
    "sourceMap": true,
    "noImplicitAny": true,
    "module": "es6",
    "target": "es6",
    "jsx": "react",
    "allowJs": true
  }
}
```

4. webpack.config should invoke both loaders:

```
module.exports = {
 entry: './index.ts',
 output: {
   filename: 'bundle.js',
   path: __dirname
 },
 module: {
   rules: [
     {
       test: /\.tsx?$/,
       use: ['babel-loader', 'ts-loader']
       exclude: /node_modules/,
     },
   ]
 },
 resolve: {
   extensions: [".tsx", ".ts", ".js"]
 },
};
```

(Note, `loaders` should be replaced by `use`)

dmitriid:
Add "Enabling tree-shaking" section as discussed in #995.

Resolves #995.

### DIFF
--- a/content/guides/webpack-and-typescript.md
+++ b/content/guides/webpack-and-typescript.md
@@ -15,16 +15,19 @@ If you didn't do so already please check out [webpack installation guide](/guide
 To start using webpack with Typescript you need a couple of things:
 
 1. Install the Typescript compiler in your project.
-2. Install a Typescript loader (in this case we're using ts-loader).
-3. Create a __tsconfig.json__ file to contain our TypeScript compilation configuration.
+2. Install a Typescript loader (in this case we're using `ts-loader`).
+3. Create a __tsconfig.json__ file for our TypeScript configuration.
 4. Create __webpack.config.js__ to contain our webpack configuration.
 
 You can install the TypeScript compiler and the TypeScript loader from npm by running:
- `npm install --save-dev typescript ts-loader`
+
+ ``` bash
+ npm install --save-dev typescript ts-loader
+ ```
 
 __tsconfig.json__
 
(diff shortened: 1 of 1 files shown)


dmitriid:
@skipjack 

> maybe we tackle shortening this guide via merging your content with the initial example in a separate PR.

I'm a bit on the fence about this one :) On the one hand it's better to have a shorter guide that immediately lets people do the intended thing. On the other hand this particular section requires additional tools (babel-loader, babel configuration) whereas previous sections only required a typescript loader and typescript configuration.

Hmmm... The rare case of a win-win situation though :)

skipjack:
> I'm a bit on the fence about this one :) ....

I'm in the same boat, I'm gonna merge as is but let's keep giving it some thought. I'm going to keep pinging you, if it's ok, when any changes with this guide come up. You can be our go-to typescript expert 🕶!

dmitriid:
@skipjack All right :)

PR #1226, opened 2017-05-18
fix(ad): make logo visible again

sokra:
For some weird reasons the ad was invisible. This PR fixes it.


bebraw:
Looks good enough. The ad won't show up in low resolutions but I think that's by design (hard to fit).

skipjack:
Yeah, so I'm not sure if this was the issue but the ad would get cut off on screens smaller than 1525px. It was hidden on screens smaller than that by design. Interested in what the issue was if someone could help me to reproduce but other than that I'm good with change if the rest of the team is happy with it.

PR #1229, opened 2017-05-18
fix(docs): Fix cut out text

glennreyes:
Should fix cut out text:
image


skipjack:
Can you share a larger screenshot of where this is cut off and let us know at about what screen size? Also the build is going to fail because for code blocks you need to specify a language e.g.

```` md
``` js
...
```
````

Other than that I think this change is good as those bullets did run on a bit too long, but I think there may be an underlying styling issue we should also fix.

glennreyes:
It's on this page:
https://webpack.js.org/guides/code-splitting-async/

image

PR #1233, opened 2017-05-22
Add webpack-dev-server troubleshooting tip

jcrben:
I suspect that this should probably go into its own section, but I'm not too sure, especially since the content is so thin.

I realize that this may seem like an unnecessary tip to experts, but I found it helpful - in my case, I had put a screwy output.filename in a quick project which had messed it up, so the logged output of `webpack output is served from /` wasn't quite right. Discovered this trick at http://stackoverflow.com/a/42592398/4200039

Hopefully a troubleshooting section can be expanded which in the long run will reduce the load of new issues opened.


jcrben:
Thanks for the tips. Updated.

PR #1223, opened 2017-05-16
Add Glossary Page

rouzbeh84:
Still a WIP...

Building out base for glossary per #970. Definitely needs more items so help would be much appreciated in checking/adding glossary items 👍

PR #1242, opened 2017-05-24
Glossary

### LINKED ISSUE #970: Concepts - Bundle vs Chunk
There is no definite definitions for the concepts `bundle` and `chunk` in the documentation of both V1 and V2.
After a learning for several days, I found that they guys use `bundle` and` chunk` in the documentation alternately. As if there isn't a clear difference between `bundle` and `chunk`.
Am I right?
If not, the effort for documentation improvement is needed.
If it is, I have two suggests:

1. use `chunk` instead of `bundle` everywhere, to avoid confusing
2. use `chunk` as a noun while `bundle` as a verb

rouzbeh84:
From #1223:

> Building out base for glossary per #970. Definitely needs more items so help would be much appreciated in checking/adding glossary items 👍

Ditching forked repo and clearing up git index for clarity sake. Thanks to @skipjack!

Resolves #970


rouzbeh84:
this also closes #1223

skipjack:
Added the frontmatter and a link to this page from the footer. @bebraw I also updated the cc-by license image to a slightly tighter one and pointed that to our `/license` page:

image

The other badge was causing some overflow and tightness on mobile.

If everyone's good with it, I think we should merge this sucker and then maybe add another pr to start linking to this page from various articles... and obviously if anyone else comes up with other key terms that would be great as well.

skipjack:
Ok let's merge, the build is only failing due to the fact that the edit link doesn't exist yet (but will once this is merged). If any issues slip in we can fix in a separate PR.

PR #1245, opened 2017-05-26
Babel-loader example is misleading

Maxim-Filimonov:
I have used that example to configure babel-loader and it wasn't working as by default Babel won't transpile the code without a preset passed to it.


simon04:
@Maxim-Filimonov, thanks for bringing this up. For fixing it, I'd like to take a different approach:

@webpack/documentation-team, maybe we should use a loader which works without any additional configuration? Such as `url-loader`? Or even more basic, the `raw-loader` for text files? The `concepts` page should focus on the concepts and not get lost in the configuration details of a specific loader.

bebraw:
Yup. `url-loader` or `raw-loader` would be enough.

simon04:
Let's go with `raw-loader` for `.txt` files.

@Maxim-Filimonov, would you be interested in PR this? The following lines would need to be changed:
* https://github.com/webpack/webpack.js.org/blame/623862481703ba35d14c6e3545e1615449c5bf9c/content/concepts/index.md#L88
* https://github.com/webpack/webpack.js.org/blame/623862481703ba35d14c6e3545e1615449c5bf9c/content/concepts/index.md#L88
* https://github.com/webpack/webpack.js.org/blame/623862481703ba35d14c6e3545e1615449c5bf9c/content/concepts/index.md#L127

Maxim-Filimonov:
@simon04 Sure I'll make required changes to use raw-loader.

Maxim-Filimonov:
See #1270

PR #1252, opened 2017-05-27
Dynamic `import` returns the promise itself

isiahmeadows:


skipjack:
Hmmm, I'm not sure. I think you have to call the `loadModule` function in order to execute the fetch. Otherwise how would the split bundle be retrieved?

bebraw:
It's the right fix since that can be rewritten as

```javascript
import('lodash').then(_ => ... do something with lodash ...)
```

I don't know why the original material had extra invocation in between.

skipjack:
Ah very true, that might have been my fault -- confusing this with the way the bundle-loader and promise-loader work. Good catch @isiahmeadows!

PR #1259, opened 2017-05-31
Update code-splitting-libraries

TheDutchCoder:
This tries to clear up some confusing aspects for beginners and makes the examples more in line with the `getting-started` examples.

### DIFF
--- a/content/guides/code-splitting-libraries.md
+++ b/content/guides/code-splitting-libraries.md
@@ -8,6 +8,7 @@ contributors:
   - rafde
   - bartushek
   - shaunwallace
+  - TheDutchCoder
 ---
 
 A typical application uses third party libraries for framework/functionality needs. Particular versions of these libraries are used and code here does not change often. However, the application code changes frequently.
@@ -24,7 +25,7 @@ Install `moment` as follows in your application directory.
 npm install --save moment
 ```
 
-The index file will require `moment` as a dependency and log the current date as follows
+The `app/index.js` file will require `moment` as a dependency and log the current date as follows
 
 __index.js__
 
@@ -40,20 +41,18 @@ __webpack.config.js__
 ```javascript
 var path = require('path');
 
-module.exports = function(env) {
-    return {
-        entry: './index.js',
-        output: {
-            filename: '[name].[chunkhash].js',
-            path: path.resolve(__dirname, 'dist')
-        }
+module.exports = {
+    entry: './app/index.js',
+    output: {
+        filename: '[name].[chunkhash].js',
+        path: path.resolve(__dirname, 'dist')
     }
 }
 ```
 
-On running `webpack` in your application, if you inspect the resulting bundle, you will see that `moment` and `index.js` have been bundled in `bundle.js`.
+On running `webpack` (e.g. [through npm](https://webpack.js.org/guides/get-started/#using-webpack-with-npm)), if you open the resulting output file (`main.[hashcode].js`) and search for `moment`, you will see that both `moment` and `index.js` have been bundled in `main.[hashcode].js`.
 
-This is not ideal for the application. If the code in `index.js` changes, then the whole bundle is rebuilt. The browser will have to load a new copy of the new bundle even though most of it hasn't changed at all.
+This is not ideal for the application. If the code in `index.js` changes, then the whole bundle is rebuilt. The browser will have to load the new bundle even though most of it (like `moment`) hasn't changed at all.
 
 
 ## Multiple Entries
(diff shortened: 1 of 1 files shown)


skipjack:
So let me ask a few questions actually before we even dive in to start editing guides...

- Was the `getting-started` guide useful and easy to follow for you?
- Does the order of the guides in the sidebar seem logical or like it gets advanced extremely quick?

I think a good approach might be to first get a good order, and then start going through them from the top down, especially if we want them to continue building on `getting-started` as you've started to do here. For example, it seems __Installation__ should be above __Getting Started__.

Maybe a _Prerequisites_ section at the top of every guide would also be useful?

TheDutchCoder:
Hi @skipjack !

I appreciate you take the time for this. I'd love to contribute and at the same time learn more about webpack this way.

To give you concrete answers:
 * Yes it was very low-level and complete. I especially liked the "diffs" in the examples to see what should be change din each step.
 * Do you mean in general, or specific to webpack? In the former case: I'm a big fan of the Vue documentation. Very thorough and hand-holding in a good way.
 * I haven't gone through all of them yet, but I think `Development` should move up in the tree, followed by `Production`, then more advanced stuff like caching, code splitting, tree shaking, etc.

The order itself is not very important (in my opinion) as it will heavily depend on what your goals are. But I think a good middle ground is to make sure people can set up a basic configuration, with a development and production step and with some static assets.

From there it would be up to the user to see which elements they might night.

skipjack:
Hey there, sorry for the delay.

> Yes it was very low-level and complete. I especially liked the "diffs" in the examples to see what should be changed in each step.

Good to know. If we're going to base all the other guides on that one, I would like to give it a quick review first (as well as starting to sort things a little better).

> Do you mean in general, or specific to webpack? In the former case: I'm a big fan of the Vue documentation. Very thorough and hand-holding in a good way.

I meant specific to webpack, i.e. on this site, but that's still helpful. I think there's a variety of very well written documentation we can take inspiration from. The general idea here is for people to go from `concepts` => `configuration || api` => `guides` if they want to get a full understanding of how things work. However, we still have a lot of work to do before that flow is clear to users.

> I haven't gone through all of them yet, but I think Development should move up in the tree, followed by Production, then more advanced stuff like caching, code splitting, tree shaking, etc.

Ok cool. I think I'll open up another PR soon to do a quick review of the getting-started guide and start sorting things slightly better.

> The order itself is not very important (in my opinion) as it will heavily depend on what your goals are.

I see your point but I do think there are some that are more obvious, kind of like how react has separated their guides into __Quick Start__ and __Advanced Guides__. Meaning I think there's some that almost everyone using webpack should read and some that are more advanced and based on use-case.

> The user would then incrementally learn about webpack specifics while actually building a working "app". For example, I learn a lot by "seeing" things happen, I'm not so good at absorbing documentation directly, I need to see the result.

Yeah I agree, maybe having a boilerplate repo with variations in it that correspond to each guide could be a good idea but then it's a whole other thing to maintain -- meaning every guide update would also require a boilerplate update. I took a stab at something like this a while ago [here](https://github.com/skipjack/webpack-boilerplate).

TheDutchCoder:
Hey @skipjack,

Go ahead and review the beginner's guide, maybe if we can agree on a "sample project" we want to use, and you provide the order of the guides, then I can take a stab at aligning the guides with our project and have you review it for correctness, consistence, tone, etc.

I think going the "Quick start" / "Advanced" route would be a good idea, the hardest part will be to determine what is considered "Advanced" ;)

Starting a new boilerplate repo is indeed extra overhead, so let's try and keep everything in one place. Worst case is that people have to re-create the files that we outline in the guides, which (if we do it well), is a one time thing, because all the other ones are derived from that starting "template".

I think adding in the html-loader would be beneficial btw, it clears things up a bout loaders and it helps people see how they can achieve a real world example.

The one you have in your repo (the clock) might be sufficient even, as long as we can extend it (i.e. it also lends itself for code splitting, managing assets, etc.).

Cheers!

skipjack:
Agreed on all counts. The one thing I would add is that we should simplify things as much as possible as we go. I've noticed there's some over-explanation and extraneous content in various guides that can either be removed, re-thought, or replaced with a link to some external site.

I'll ping you as soon as I have the initial pr in so we can discuss more. Should be sometime today or tomorrow.

TheDutchCoder:
Hold off on this, as we're rewriting stuff from the start.

skipjack:
@TheDutchCoder I'm going to close this but will take these changes into account in my new pr.

TheDutchCoder:
Perfect!

PR #1260, opened 2017-05-31
*trying* to make the sentence a bit more readable

cool88:
I was stumbled here.


content/guides/code-splitting-libraries.md
```diff
@@ -154,9 +154,10 @@ module.exports = function() {
 ## Manifest File
 
 But, if we change application code and run `webpack` again, we see that the hash for the vendor file changes. Even though we achieved separate bundles for `vendor` and `main` bundles, we see that the `vendor` bundle changes when the application code changes.
+
 This means that we still don't reap the benefits of browser caching because the hash for vendor file changes on every build and the browser will have to reload the file.
 
-The issue here is that on every build, webpack generates some webpack runtime code, which helps webpack do its job. When there is a single bundle, the runtime code resides in it. But when multiple bundles are generated, the runtime code is extracted into the common module, here the `vendor` file.
+The issue here is that on every build, webpack generates some webpack runtime code, which helps webpack do its job. When there is a single bundle, the runtime code resides in it. But when multiple bundles are generated, the runtime code is extracted into the common module - the `vendor` file in this example.
```
skipjack: I think a semicolon or `--` would be more appropriate here, e.g.

``` md
common module -- the `vendor` file in this example.
```

or

```md
common module; the `vendor` file in this example.
```

cool88: I intend to use the long dash, will these convert double '--' convert to an Em dash? IMO, semicolon most likely not because the sentence is still strongly connected to the previous sentence.

skipjack: Hmmm, I see your point. Idk if our system converts the double dash to an em dash yet but it's definitely something we can tackle down the road. You could use the html entity, `&#8212;`, but that'd look rather out of place mid sentence. I would just do `--` for now as I think most people understand that's meant to be an em dash. I'll look into the actual conversion later on.

PR #1265, opened 2017-06-01
Points out that outputFileSystem needs of methods

mc-zone:
the `mkdirp` and `join` are also needed by [`outputFileSystem`](https://github.com/webpack/webpack/blob/efc576c8b744e7a015ab26f1f46932ba3ca7d4f1/lib/node/NodeOutputFileSystem.js).

### DIFF
--- a/content/api/node.md
+++ b/content/api/node.md
@@ -277,4 +277,4 @@ compiler.run((err, stats) => {
 });
 ```
 
-T> The output file system you provide needs to be compatible with Node’s own [`fs`](https://nodejs.org/api/fs.html) module interface, and other two helper methods `mkdirp`, `join` are also needed.
+T> The output file system you provide needs to be compatible with Node’s own [`fs`](https://nodejs.org/api/fs.html) module interface, which requires the `mkdirp` and `join` helper methods.
--- a/content/api/node.md
+++ b/content/api/node.md
@@ -277,4 +277,4 @@ compiler.run((err, stats) => {
 });
 ```
 
-T> The output file system you provide needs to be compatible with Node’s own [`fs`](https://nodejs.org/api/fs.html) module interface.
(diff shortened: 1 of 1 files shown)


mc-zone:
OTOH, Should webpack auto make the polyfills if someone set an `outputFileSystem` without `mkdirp` or `join`? 

@sokra have a miunte to give opinions?

sokra:
> OTOH, Should webpack auto make the polyfills if someone set an outputFileSystem without mkdirp or join?

no, they could be different for each fs.

mc-zone:
>Guessing --root from input files: file:///home/travis/build/webpack/webpack.js.org/build/
should respond with HTTP status 200
404 http://cssnano.co/options/ at build/loaders/css-loader/index.html:507:182 <a href="http://cssnano.co/options/">...</a>

https://travis-ci.org/webpack/webpack.js.org/builds/238324866#L746

Can't pass travis CI check caused by this link is 404: http://cssnano.co/options/ . 
It's not result from this PR. @skipjack Would you give some advice please?

skipjack:
@mc-zone just opened a [pull request](https://github.com/webpack-contrib/css-loader/pull/550) to resolve this. css-loader's readme is dynamically pulled into this repo and the broken link was there.

content/api/node.md
```diff
@@ -277,4 +277,4 @@ compiler.run((err, stats) => {
 });
 ```
 
-T> The output file system you provide needs to be compatible with Node’s own [`fs`](https://nodejs.org/api/fs.html) module interface.
+T> The output file system you provide needs to be compatible with Node’s own [`fs`](https://nodejs.org/api/fs.html) module interface, and other two helper methods `mkdirp`, `join` are also needed.
```
skipjack: I think this would be better as:

``` md
module interface, which requires the `mkdirp` and `join` helper methods.
```

mc-zone: Right. Got it.

PR #1267, opened 2017-06-03
Clarify Available Options in `node` Configuration

### LINKED ISSUE #1264: Node shimming doesn't allow `empty` anymore
1. https://webpack.js.org/configuration/node/
2. Using `"empty"` yields the following:

```bash
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.node.process should be one of these:
   false | true | "mock"
```

It's either core bug or documentation bug. This might have changed in webpack 2.

skipjack:
Resolves #1264

### DIFF
--- a/content/configuration/node.md
+++ b/content/configuration/node.md
@@ -14,14 +14,14 @@ These options configure whether to polyfill or mock certain [Node.js globals](ht
 
 `object`
 
-This is an object where each property is the name of a NodeJS global or module and each value may be one of the following...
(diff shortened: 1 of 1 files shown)


skipjack:
It seems to be mostly NodeJS or Node.js throughout the rest of the docs... personally I think NodeJS looks a bit better but actually on their site they seem to consistently use Node.js or Node. I'll change to Node and maybe submit another PR to make it consistent throughout the rest of the documentation as well.

PR #1275, opened 2017-06-07
Fix path.resolve() import error in example config.

mhaagens:


skipjack:
Thanks for you contribution! What's the error you're seeing though? Simply destructuring `resolve` out of `path` doesn't seem like it would actually fix anything.

mhaagens:
This is the section where the error is located;
```
const path = require('path');

module.exports = {
  entry: './index.js',

  plugins: [
    new webpack.HotModuleReplacementPlugin() // Enable HMR
  ],

  output: {
    filename: 'main.js',
    path: path.resolve(__dirname, 'dist'),
    publicPath: '/'
  },

  devServer: {
    hot: true, // Tell the dev-server we're using HMR
    contentBase: resolve(__dirname, 'dist'), <-- THIS WON'T WORK
    publicPath: '/'
  }
};
```
```contentBase: resolve(__dirname, 'dist')``` is the issue, since resolve isn't declared (path is).
You could go the other way and change it to ```contentBase: path.resolve(__dirname, 'dist')```, but I just thought it was cleaner do destructure and use resolve directly instead.

**Comment by skipjack** [comment]
Ah, now I see thanks for the clarification. Yeah I understand what you mean, however I want to prevent people from confusing it with any other `resolve` methods like `require.resolve`. Can you go the other route you mentioned (i.e. simply referencing `contentBase: path.resolve(...)`)?

On a side note, did you find the guide clear and useful?

**Comment by mhaagens** [comment]
Changed it to ```const path = require("path");``` now. Made a bit of a mess with the commits, hope that's not a big issue. I find the guide pretty easy to follow, but I was actually there to grab the "boilerplate" for setting up react-hot-loader, which now seems to be gone. I do think a lot of people would benefit from having framework specific set up guides (HMR with React, HMR with Express etc.), but I also understand that that's not really related to Webpack.

File: content/guides/hot-module-replacement.md [hunk 1]
```diff
@@ -26,7 +26,7 @@ W> __HMR__ is not intended for use in production, meaning it should only be used
 Enabling this feature is actually fairly simple. Let's take a look at how to set it up with [webpack-dev-server](https://github.com/webpack/webpack-dev-server)...
 
 ``` js
-const path = require('path');
+const resolve = require('path');
```
simon04: Why change `path` into `resolve` here?

mhaagens: Whoops, my bad! 😓  Fixed in latest commit.

skipjack:
> I do think a lot of people would benefit from having framework specific set up guides (HMR with React, HMR with Express etc.), but I also understand that that's not really related to Webpack.

Yeah, there were a few reasons behind the decision to remove it. First of all, the old hmr-react guide covered a lot (much of which was not really directly related to HMR) and also had some unnecessary entry points that weren't required to make HMR work. Secondly, documenting integrations with specific frameworks is kind of a never ending spiral and increases the scale of our documentation tremendously (which is already pretty big even without additional third-party "integration" docs). Lastly, we had multiple people asking how to just get a basic HMR setup working without any frameworks.

If you have any other links, e.g. to blogs or somewhere that explain the react-hot-loader setup better you should feel free to add them to the __Other Code and Frameworks__ section at the bottom. TBH though I think the best approach would be to try to improve the docs over at the react-hot-loader repo, where they can now they can link to our simple guide as a pre-requisite.

mhaagens:
Yeah, I totally understand why the decision was made, it would just end up with you getting support requests for libraries other than Webpack as well. I removed react-hot-loader, now just using HotModuleReplacementPlugin with module.hot.accept() in my entry file, and HMR still works as far as I can tell. Thanks for merging the PR!

PR #1273, opened 2017-06-07
Initial Guides Cleanup and Review

skipjack:
@TheDutchCoder ok pretty much have this wrapped up -- just need to do a bit more sorting based on our previous conversation. The main change to `getting-started` was to clarify the directory structure, use `/src` instead of `/app` (as I think it's a bit more common), and consistently use `/dist` to house the entire production site (e.g. `index.html` as well). The last part will come into play specifically in our `output-management` guide but probably in a few others as well.

Look through the changes and let me know what you think. I think it'll give us a nice base for simplifying and updating the other guides. I'll ping a few others for review as well once we've agreed it's good to go.

PR #1276, opened 2017-06-08
Update Code Splitting - Libraries

cool88:
please feel free to discard

### DIFF
--- a/content/guides/code-splitting-libraries.md
+++ b/content/guides/code-splitting-libraries.md
@@ -14,7 +14,7 @@ A typical application uses third party libraries for framework/functionality nee
 
 Bundling application code with third party code would be inefficient. This is because the browser can cache asset files based on the cache header and files can be cached without needing to call the cdn again if its contents don't change. To take advantage of this, we want to keep the hash of the vendor files constant regardless of application code changes.
 
-We can do this only when we separate the bundles for vendor and application code.
+We can do this only when we separate the bundles for vendor and application code. We walk you through with an example file that uses one of the popular dependencies. We show how the typical bundling process happens as we bundle our code with and without a helpful plugin -- `CommonsChunkPlugin` -- that we introduce in the walk-through.
 
 Let's consider a sample application that uses [momentjs](https://www.npmjs.com/package/moment), a commonly used time formatting library.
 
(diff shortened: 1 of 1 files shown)


content/guides/code-splitting-libraries.md
```diff
@@ -86,7 +86,7 @@ It is for this reason, that we will need to use the [CommonsChunkPlugin](/plugin
 
 ## `CommonsChunkPlugin`
 
-This is a pretty complex plugin. It fundamentally allows us to extract all the common modules from different bundles and add them to the common bundle. If a common bundle does not exist, then it creates a new one.
+This is a pretty complex plugin. It fundamentally allows us to extract all the common modules from different bundles and adds them to a "common" bundle. If a common bundle does not exist, then it creates a new one.
```
skipjack: I like this change but I think the whole sentence should be rewritten:

```md
This plugin allows us to extract all duplicate (or "common") modules from different bundles into a "common" bundle. This "common" bundle can either be one that already exists (e.g. an entry bundle) or an entirely new one.
```

dropping that first bit about it being complex.

content/guides/code-splitting-libraries.md
```diff
@@ -14,7 +14,7 @@ A typical application uses third party libraries for framework/functionality nee
 
 Bundling application code with third party code would be inefficient. This is because the browser can cache asset files based on the cache header and files can be cached without needing to call the cdn again if its contents don't change. To take advantage of this, we want to keep the hash of the vendor files constant regardless of application code changes.
 
-We can do this only when we separate the bundles for vendor and application code.
+We can do this only when we separate the bundles for vendor and application code. We walk you through with an example file that uses one of the popular dependencies. We show how the typical bundling process happens as we bundle our code with and without a helpful plugin -- `CommonsChunkPlugin` -- that we introduce in the walk-through.
```
skipjack: I would drop this part for now as it feels a bit wordy -- we'll make sure everything's clear as we review and go through all the guides. #1273 is the start of the guides cleanup/review process.

skipjack:
This guide and other _Code Splitting_ guides were rewritten in #1338 so this no longer relevant. @cool88 thank you anyway for the effort, we appreciate it. The problem was that these guides had a variety of other issues that needed to be resolved and we had to synchronize it with _Getting Started_.

PR #1277, opened 2017-06-08
Update Writer's Guide

cool88:
I think the guideline is missing this important part.

### DIFF
--- a/content/writers-guide.md
+++ b/content/writers-guide.md
@@ -10,10 +10,11 @@ title: Writer's Guide
 
 ## Article Structure
 
-1. Brief introduction - a paragraph or two so you get the basic idea. Tell what you are going to tell.
-2. Main content - tell what you promised to tell.
-3. Conclusion - tell what you told and recap the main points.
-4. References - link to related articles and external resources so people can read and learn more about the topic.
+1. Brief introduction - a paragraph or two so you get the basic idea about - the what and why. 
+2. Structure description for the "Main Content" - the how - how you will tell it in a brief paragraph (e.g. a new plugin or a webpack concept)
+3. Main content - tell what you promised to tell.
+4. Conclusion - tell what you told and recap the main points.
+5. References - link to related articles and external resources so people can read and learn more about the topic.
 
 ## Running the Site
 
--- a/content/writers-guide.md
+++ b/content/writers-guide.md
@@ -2,19 +2,49 @@
 title: Writer's Guide
 ---
 
+The following sections contain all you need to know about editing and formatting the content within this site. Make sure to do some research before starting your edits or additions. Sometimes the toughest part is finding where the content should live and determining whether or not it already exists.
+
+
 ## Process
 
 1. Check related issue if an article links to one.
 2. Hit `edit` and expand on the structure.
 3. PR changes.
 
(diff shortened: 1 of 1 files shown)


content/writers-guide.md
```diff
@@ -10,10 +10,11 @@ title: Writer's Guide
 
 ## Article Structure
 
-1. Brief introduction - a paragraph or two so you get the basic idea. Tell what you are going to tell.
-2. Main content - tell what you promised to tell.
-3. Conclusion - tell what you told and recap the main points.
-4. References - link to related articles and external resources so people can read and learn more about the topic.
+1. Brief introduction - a paragraph or two so you get the basic idea about - the what and why. 
+2. Structure description for the "Main Content" - the how - how you will tell it in a brief paragraph (e.g. a new plugin or a webpack concept)
```
simon04: I find the "the how - how you will tell" confusing. What about "Outline for the remaining page – _how_ the content will be presented"?

content/writers-guide.md
```diff
@@ -10,10 +10,11 @@ title: Writer's Guide
...
+3. Main content - tell what you promised to tell.
+4. Conclusion - tell what you told and recap the main points.
+5. References - link to related articles and external resources so people can read and learn more about the topic.
```
simon04: This line is no longer true due to 96595246d823a7a7cddf4a204262eb92ae8742c3. It is written into the YAML frontmatter.

content/writers-guide.md
cool88: Hi, by "how" I mean : how will you present the solution? There are many ways that something can be explained. A brief intro of that process is what, I believe, should be placed as the second point. The 1st point (1. Introduction)  should only focus on developing the motivation of the post, what is it and why we care. Before jumping to the full explanation, I think, a couple of lines of about how the author intend to explain his point will be great. So the reader know, in advance, what to expect as a general instruction of the main content.

Please let me know if I should elaborate more.

cool88: The following example is not perfect, but this might help explain my point :
https://github.com/webpack/webpack.js.org/pull/1276/commits/cb37f0d09fd597600c03ca60e73be6cd182aa232

cool88: That said, it's up to you guys to decide if it is a good point or not for webpack -- if it adds more clutter than the value. As you know though, in most technical writing, it is a recommended step.

cool88: Oops .. I don't know how I didn't read this part "What about "Outline for the remaining page – how the content will be presented?"" 
That's exactly what I wanted to say. Thanks!

cool88: That's exactly what I wanted to say. Thanks!

content/writers-guide.md
skipjack: Yeah we should mention that somehow -- something like:

``` md
5. References - Use the `related` key in the YAML frontmatter at the top of the file. See the _Frontmatter_ section below for more information.
```

And then we should add a __Frontmatter__ section that documents `title`, `sort`, `contributors` and `related`.

skipjack:
@cool88 I made the updates so we could push this forward. Thanks for starting this!

PR #1278, opened 2017-06-08
In merge example, export the merged object.

dankh:
The current guide throws an error : 
```
Config did not export an object or a function returning an object.
```
When exporting the merged object instead of a function it works.


content/guides/production-build.md
```diff
@@ -250,8 +250,7 @@ __webpack.prod.js__
 const Merge = require('webpack-merge');
 const CommonConfig = require('./webpack.common.js');
 
-module.exports = function(env) {
```
simon04: Drop the matching `}` at the end. Fix indentation.

dankh: Done

PR #1284, opened 2017-06-11
Pulled the instructions out of the code

dnecklesportfolio:
Pulled the commented instructions out of the code to make it stand out more. I missed it the first few times since the comment was apart of the code. It doesn't have to be blockquotes like I have it. I couldn't fix that.


content/configuration/index.md
```diff
@@ -19,7 +19,8 @@ T> New to webpack? Check out our guide to some of webpack's [core concepts](/con
 T> Notice that throughout the configuration we use Node's built-in [path module](https://nodejs.org/api/path.html) and prefix it with the [__dirname](https://nodejs.org/docs/latest/api/globals.html#globals_dirname) global. This prevents file path issues between operating systems and allows relative paths to work as expected. See [this section](https://nodejs.org/api/path.html#path_windows_vs_posix) for more info on POSIX vs. Windows paths.
 
 ## Options
-
+ In the code below, click on the name of the option to get to the detailed documentation.
+ Also, you cam click on the items with arrows to show more examples / advanced options.
```
skipjack: I like this change though I would rephrase it slightly and you'll need to add line breaks to before and after to please our markdownlinter (used for consistent formatting). Something like:

``` md
## Options

Click on the name of each option in the configuration code below to jump to the detailed documentation. Also note that the items with arrows can be expanded to show more examples and, in some cases, more advanced configuration.
```

content/configuration/index.md
```diff
@@ -19,7 +19,8 @@ T> New to webpack? Check out our guide to some of webpack's [core concepts](/con
...
 ``` js-with-links-with-details
 const path = require('path');
```
skipjack: It's not showing it in the diff but can you also remove the code comments below now that the description exists above?

dnecklesportfolio:
ok I made the corrections, but let me know if it still appears as a blockquote on your end.

PR #1289, opened 2017-06-13
Fix typo in docs

tijwelch:
"The goal is **to have** .... **to be**"

should be

"The goal is **to have** ... **be**"


skipjack:
Thanks @tijwelch.. the build error was caused by something else which I'll fix shortly.

PR #1291, opened 2017-06-14
Fix some mistakes in the tutorial

supernavy:
Because of the mistakes, the "hello webpack" doesn't show up.

### DIFF
--- a/content/guides/getting-started.md
+++ b/content/guides/getting-started.md
@@ -59,7 +59,7 @@ __index.html__
     <script src="https://unpkg.com/lodash@4.16.6"></script>
   </head>
   <body>
-    <script src="/src/index.js"></script>
+    <script src="src/index.js"></script>
   </body>
 </html>
 ```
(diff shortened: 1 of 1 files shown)


skipjack:
@supernavy can you elaborate on what the mistakes were or what the errors you're seeing are? We'd like to keep `index.html` within the `/dist` directory.

supernavy:
There are 2 mistakes,
1. When not using webpack, the js path in index.html should be relative path "src/index.js" but not "/src/index.js". You can follow the step and see if the "hello webpack" shows up when you opening index.html by browser.
2. When using webpack and move index.html into dist folder, the js path in index.html should be "bundle.js" but not "dist/bundle.js". So need either moving index.html one level up or changing the js path.

skipjack:
@supernavy thank you for that clarification! Can you update this PR with these changes for each case...

1. `/src/index.js` => `src/index.js`
2. `dist/bundle.js` => `bundle.js`

This guide will be the base point for the rest of our guides, and keeping `index.html` within `/dist` is a key part of that.

supernavy:
Done.

PR #1292, opened 2017-06-14
revise the path error in the get start demo

HAOGRE:
the path is not correct to start the demo project!

### DIFF
--- a/content/guides/getting-started.md
+++ b/content/guides/getting-started.md
@@ -118,7 +118,7 @@ __dist/index.html__
   </head>
   <body>
 -   <script src="src/index.js"></script>
-+   <script src="dist/bundle.js"></script>
++   <script src="bundle.js"></script>
   </body>
  </html>
 ```


HAOGRE:
@jsf-clabot i have done the sign

skipjack:
Ah this is similar to #1291 -- please see my comment there. I think something else needs to be tweaked probably either one of the CLI commands or example config.

HAOGRE:
@skipjack ok! as your will ! but the now guide on the web can not work it out to show "hello webpack" when u open the index.html via brower!

skipjack:
> ok! as your will ! but the now guide on the web can not work it out to show "hello webpack" when u open the index.html via brower! 404 to find bundle.js

@HAOGRE I think you misunderstood me. You are 100% correct that there is an issue, I'm just saying we should resolve that issue in a slightly different way. This is definitely a high priority and we'll make sure to have a fix merged by tomorrow at the latest. Let's close this and move the discussion to #1291.

PR #1298, opened 2017-06-16
Improve sidebar and gitter components' performance (Was: 'Use passive event listeners for scroll events')

marcoms:
This PR adds the `passive: true` flag to `addEventListener` calls that monitor the scroll event, which indicates to the browser that the listener function will not call `preventDefault`. That lets the browser make some optimizations which improve performance.

As a note this would be much more performant and simpler using either `position: sticky` or alternating between `position: fixed` and `absolute` based on scroll position, as JS would be eliminated. On my machine the `_recalculate()` function called each scroll event consumes around 60ms (15fps)

### DIFF
--- a/components/gitter/gitter.jsx
+++ b/components/gitter/gitter.jsx
@@ -3,6 +3,11 @@ import testPassiveListeners from '../../utilities/test-passive-listeners';
 
 const supportsPassive = testPassiveListeners() !== false;
 
+let footerHeight;
+if (typeof window !== 'undefined') {
+    footerHeight = document.querySelector('footer').offsetHeight;
+}
+
 export default class Gitter extends React.Component {
   constructor(props) {
     super(props);
@@ -52,7 +57,6 @@ export default class Gitter extends React.Component {
     let { scrollY, innerHeight } = window;
     let { scrollHeight } = document.body;
     let distToBottom = scrollHeight - scrollY - innerHeight;
-    let footerHeight = document.querySelector('footer').offsetHeight;
 
     this.setState({
       offset: distToBottom < footerHeight ? footerHeight - distToBottom : 0
--- a/components/gitter/gitter.jsx
+++ b/components/gitter/gitter.jsx
@@ -4,9 +4,6 @@ import testPassiveListeners from '../../utilities/test-passive-listeners';
 const supportsPassive = testPassiveListeners() !== false;
 
 let footerHeight;
-if (typeof window !== 'undefined') {
-    footerHeight = document.querySelector('footer').offsetHeight;
-}
 
 export default class Gitter extends React.Component {
   constructor(props) {
@@ -57,6 +54,9 @@ export default class Gitter extends React.Component {
     let { scrollY, innerHeight } = window;
     let { scrollHeight } = document.body;
     let distToBottom = scrollHeight - scrollY - innerHeight;
+    if (!footerHeight) {
+      footerHeight = document.querySelector('footer').innerHeight;
+    }
 
     this.setState({
       offset: distToBottom < footerHeight ? footerHeight - distToBottom : 0
--- a/components/gitter/gitter.jsx
+++ b/components/gitter/gitter.jsx
@@ -32,14 +32,15 @@ export default class Gitter extends React.Component {
     );
 
     document.addEventListener(
-      'scroll', 
-      this._recalculate.bind(this)
+      'scroll',
+      this._recalculate.bind(this),
+      { passive: true }
     );
   }
 
   componentWillUnmount() {
     document.removeEventListener(
-      'scroll', 
+      'scroll',
       this._recalculate.bind(this)
     );
   }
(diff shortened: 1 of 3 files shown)


skipjack:
> This PR adds the `passive: true` flag to `addEventListener` calls that monitor the scroll event, which indicates to the browser that the listener function will not call `preventDefault`. That lets the browser make some optimizations which improve performance.

This change sounds great but see my comment below as well...

> As a note this would be much more performant and simpler using either `position: sticky` or alternating between `position: fixed` and `absolute` based on scroll position, as JS would be eliminated. On my machine the `_recalculate()` function called each scroll event consumes around 60ms (15fps).

To use `position: sticky` we'd have to include a polyfill, correct? I don't think that's supported by many browsers yet. Re the second option -- I'm not sure I follow completely but if you have a more performant implementation in mind using just CSS, we'd definitely be open to a PR to replace the current behavior. We tried using [react-sticky](https://github.com/captivationsoftware/react-sticky) to start but ran into some complications with our build process and use of [Preact](https://github.com/developit/preact) as a lighter-weight stand-in for React.

If you'd like to continue with this PR first, please correct the [build error](https://travis-ci.org/webpack/webpack.js.org/builds/243854171#L488).

marcoms:
Hopefully the build errors should be resolved now 👍 

To clarify the second option, basically it looks similar to what you are doing already - my mistake.

I have noticed though that the source of performance problems is probably due to the intense use of `innerHeight`, `scrollY`, etc. which trigger layout when accessed unfortunately (https://gist.github.com/paulirish/5d52fb081b3570c81e3a). I suppose that is a consequence of the current design constraints.

>When implemented the sticky behavior it seemed fairly performant but after seeing it on other screens/computers I noticed the glitchy-ness/lag as well. I guess the different refresh rates for various screens is the reason for the discrepancy?

Most displays are 60Hz, so I'd say the culprit is probably processing power.

skipjack:
> Edge and IE don't support it (Edge will soon apparently) so yes, but performance would not be the same.

🎉  I didn't even realize it was implemented in any browsers yet! Just saw the [MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/position?v=example) and I'm still recovering from shock. Do you know of a good polyfill that's not too heavy? I would love to go this route...

And actually, now that I think about it, unless it's going to break something we could probably just use `position: sticky` and tell any users that have issues to either upgrade their browser or bug Microsoft Edge to support it. I was kind of on the fence about even adding the sticky behavior with the current perf issues as it's more of a nice-to-have but we had a lot of ppl asking about it. Many documentation sites, even very popular ones, don't offer it. What I'm getting at is it's probably not the end of the world if it doesn't work for some small percentage of people (most of our audience -- devs, web devs in particular -- are probably on a supported browser anyway). @bebraw what do you think?

> To clarify the second option, basically it looks similar to what you are doing already - my mistake.

No worries and thanks for that link re `innerHeight` etc., definitely useful information.

It looks like the build passed but replacing all of this code with `position: sticky` is the way to go imo. If @bebraw's good with it, would you be willing to upgrade this PR accordingly? Sorry, I know it may mean the current changes will go to waste.

bebraw:
👍 for `position: sticky`. Simple over complex.

skipjack:
@marcoms I think we're good to go then... can you update the PR when you have a chance?

marcoms:
Sounds good - will have a go this evening

marcoms:
Update: my desk collapsed yesterday evening so I try and continue on my laptop - apologies

marcoms:
I'm beginning to think using sticky positioning isn't possible using the flex layout that is currently used, or at least I haven't found one - it seems to work when display is set to block. Perhaps for now we could merge the PR without the sticky changes, and maybe someone with a sharper mind might figure out a way in the future..? Failing that, an always fixed positioned sidebar may be a simple alternative, as it doesn't move by much when scrolling anyway.

skipjack:
> I'm beginning to think using sticky positioning isn't possible using the flex layout that is currently used, or at least I haven't found one - it seems to work when display is set to block.

Hmmm that's odd. I may take a stab at it as well.

> Perhaps for now we could merge the PR without the sticky changes

Sure, will test this out and hopefully get it merged sometime this weekend.

> Failing that, an always fixed positioned sidebar may be a simple alternative, as it doesn't move by much when scrolling anyway.

That is an option, but there are some other complications that arise with that, e.g. managing overflow.

skipjack:
@marcoms sorry for the delay on this -- I pulled your branch locally and started to test this out. So far I've only noticed one issue with the `Gitter` component...

image

which is not sticking to the bottom of the content container but instead overflowing the footer. Now that we've agreed `position: sticky` is probably the right way to go, I think the best thing to do would be to hold off on this PR to avoid accidentally introducing any other bugs or over-complicating the code. We're down to only __36 issues__ at this point, and once most of the content ones are done I will begin a more major cleanup of the site. At which point, I think we can figure out how to use `position: sticky` over any javascript.

I just threw together a [pen](https://codepen.io/gvenech/pen/PjRgYv?editors=1100) to test if it works within `flex: 1 1 auto` and it definitely seems to. Maybe you forgot the `top` / `bottom` declarations when you were playing around with it?

Anyway, I think we should close this and instead focus on moving to `position: sticky` which is much more stable. If you want to open a separate PR to do this before I have a chance to, by all means go ahead.

Issue #273, opened 2016-10-23
Config - Elaborate on Source Map Options


octref:
While browsing webpack's `devtool` section, I found this stub:

> This section is pretty much just copied over from existing docs... imo more background is needed on the different types of source mapping, maybe via links to external sites that discuss the different types of source maps in more detail.

I agree 100%. For general users, most of them only know sourcemap is a mapping between transpiled and un-transpiled code, and allows them to debug in un-transpiled code.

I come to this section with one question:

Which one should I use?

Which boils down to:
- Which one work with Chrome/FF debugger?
- What's the major difference between each one?

**However, I walked away with both questions unanswered, while having more questions**:

```
 devtool                      | build | rebuild | production | quality
------------------------------|-------|---------|------------|--------------------------
 eval                         | +++   | +++     | no         | generated code
 cheap-eval-source-map        | +     | ++      | no         | transformed code (lines only)
 cheap-source-map             | +     | o       | yes        | transformed code (lines only)
 cheap-module-eval-source-map | o     | ++      | no         | original source (lines only)
 cheap-module-source-map      | o     | -       | yes        | original source (lines only)
 eval-source-map              | --    | +       | no         | original source
 source-map                   | --    | --      | yes        | original source
```
- What's the difference between generated/transformed?
- What's "lines only", and how does it affect me?
- What's your definition of quality? And how does the options differ in that aspect?
- What it means for yes/no in "production"? If I were to put webpack in production, I surely want to know what each option does technically.
- What's "cheap"?

Also, I was seeing people using `webpack.SourceMapDevToolPlugin()` in webpack 2.0. How does that relate to this setting? Which one should I use?

I hope webpack team can keep these questions in mind while reworking on the doc.

SpaceK33z:
Note that there are a some open questions about source maps for us also, which is why it's hard to write these docs. With https://github.com/webpack/webpack/issues/3165 we hope to get more insights.

If you have anything specific to improv

panozzaj:
Thanks for writing up your thoughts on the documentation for `devtool`, I had similar thoughts while reading through it. My question ([original](https://github.com/webpack/webpack/issues/4543)) was more along the lines of speed and understanding the notation of the table. Restated here:

*What is the current behavior?*

Currently the documentation for devtool configuration is unclear to me, at least for the speed section. It seems like plusses mean that it runs faster. But maybe it means it takes more time? What does `o` mean? What does `-` mean?

*What is the expected behavior?*

[...]

PR #1303, opened 2017-06-18
Update Devtool

### LINKED ISSUE #686: Config - Add Missing Devtool Options
https://webpack.js.org/configuration/devtool/

`inline-source-map` is listed under "For development", but it's missing from the summary table above it.

skipjack:
Update devtool.md and add missing options.

Adds missing bits for the `inline-source-map`, `hidden-source-map` and `cheap-eval-source-map` options. Also did a bit of cleanup...

Resolves #686

### DIFF
--- a/content/configuration/devtool.md
+++ b/content/configuration/devtool.md
@@ -26,18 +26,18 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
 
 T> The webpack repository contains an [example showing the effect of all `devtool` variants](https://github.com/webpack/webpack/tree/master/examples/source-map). Those examples will likely help you to understand the differences.
 
-devtool                      | build | rebuild | production | quality
------------------------------|-------|---------|------------|------------------------------
-eval                         | +++   | +++     | no         | generated code
-inline-source-map            | ++    | ++      | no         | generated code
-cheap-eval-source-map        | +     | ++      | no         | transformed code (lines only)
-cheap-source-map             | +     | o       | yes        | transformed code (lines only)
-cheap-module-eval-source-map | o     | ++      | no         | original source (lines only)
-cheap-module-source-map      | o     | -       | yes        | original source (lines only)
-eval-source-map              | --    | +       | no         | original source
-source-map                   | --    | --      | yes        | original source
-hidden-source-map            | --    | --      | yes        | original source
-nosources-source-map         | --    | --      | yes        | without source content
+devtool                       | build | rebuild | production | quality
+----------------------------- | ----- | ------- | ---------- | -----------------------------
+eval                          | +++   | +++     | no         | generated code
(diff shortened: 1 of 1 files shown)


skipjack:
@bebraw @sokra can one of you take a look at these changes, particularly the `inline-source-map` addition in the table and the `cheap-eval-source-map` description in the __Development__ section? I looked [here](https://github.com/webpack/webpack/tree/master/examples/source-map#cheap-eval-source-mapjs) to try to figure out the description for `cheap-eval-source-map` but couldn't figure out the difference between it and `cheap-module-eval-source-map`.

sokra:
## `cheap-module` vs `cheap`

`-module` allows loaders to process source-map. i. e. babel-loader create a SourceMap which is used.

## `cheap-module-source-map` vs `source-map`

`cheap` uses `source-list-map` instead of `source-map` which is much faster but doesn't map columns resp. simplifies SourceMaps to one mapping per line.

## `cheap-eval-source-map` vs `eval`

`cheap-eval-source-map` shows source code before webpack replacements, `eval` after webpack replacements.

skipjack:
@sokra thanks for that clarification. Is the `inline-source-map` line I added to the table correct?

> 
devtool                      | build | rebuild | production | quality
-----------------------------|-------|---------|------------|------------------------------
eval                         | +++   | +++     | no         | generated code
inline-source-map            | ++    | ++      | no         | generated code

content/configuration/devtool.md
```diff
@@ -20,34 +26,46 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
- cheap-source-map             | +     | o       | yes        | transformed code (lines only)
- cheap-module-eval-source-map | o     | ++      | no         | original source (lines only)
- cheap-module-source-map      | o     | -       | yes        | original source (lines only)
- eval-source-map              | --    | +       | no         | original source
- source-map                   | --    | --      | yes        | original source
- nosources-source-map         | --    | --      | yes        | without source content
+devtool                      | build | rebuild | production | quality
+-----------------------------|-------|---------|------------|------------------------------
+eval                         | +++   | +++     | no         | generated code
+inline-source-map            | ++    | ++      | no         | generated code
```
sokra: `inline-source-map            | --    | --      | no         | original source`

sokra:
`inline-source-map` is similar to `source-map`

You could add the same for the `cheap` variants:

`cheap-module-inline-source-map | o | - | no | original source (lines only)`
`cheap-inline-source-map | + | o | no |  transformed code (lines only)`

skipjack:
@sokra updated.. let's hold off on adding any others to prevent new issues cropping up like #686. Tbh, in working on this PR, I think there may be a better way to lay out this documentation -- hopefully something we can discuss more in #273. I'll add a note there about those other options.

skipjack:
@sokra merging these fixes but let's keep discussing in #686.

Issue #966, opened 2017-03-01
add language selection menu


lcxfs1991:
webpack china doc is ready.  Shall we add a language selection menu on top navigation bar? 

link: https://doc.webpack-china.org/

Best regards,

Lee

skipjack:
Closed by #1148 and will be touched up further by #1171

PR #1310, opened 2017-06-19
Remove national flags from language selector.

sprrw:
As discussed on #1171, this PR removes national flags from language selector and introduces language change icon instead.


sprrw:
Related: #966, #1148.

components/navigation/navigation.jsx
```diff
@@ -93,22 +91,22 @@ export default class Navigation extends React.Component {
           <Link
             className="navigation__icon"
             title="GitHub Repository"
-            to="//github.com/webpack/webpack">
+          to="//github.com/webpack/webpack">
```
skipjack: Can you fix the formatting here?

components/navigation/navigation.jsx
```diff
@@ -93,22 +91,22 @@ export default class Navigation extends React.Component {
...
             <i className="sidecar__icon icon-github" />
           </Link>
 
           <Link
             className="navigation__icon"
             title="See Questions on Stack Overflow"
-            to="//stackoverflow.com/questions/tagged/webpack">
+          to="//stackoverflow.com/questions/tagged/webpack">
```
skipjack: And here too?

components/navigation/navigation.jsx
sprrw: Yeah, I fixed it. Strange thing is that my editor does favor editorconfig, but for an obscure reason prefers to format those fragments the way they were in the source (and doesn't even let me edit them back). Some misbehaving Atom package I guess.

sprrw:
@skipjack @bebraw I fixed formatting, it looks like it's good to go now.

skipjack:
Awesome work, thanks @sprrw! Will merge as soon as the build passes.

PR #1312, opened 2017-06-20
docs(config): conflate configuration languages

simon04:
- Move all configuration languages to the corresponding configuration page, away from the concepts page
- Reduce TypeScript example to a minimal working example

### DIFF
--- a/content/concepts/configuration.md
+++ b/content/concepts/configuration.md
@@ -3,6 +3,7 @@ title: Configuration
 sort: 6
 contributors:
 - TheLarkInn
+- simon04
 ---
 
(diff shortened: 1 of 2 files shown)


content/concepts/configuration.md
```diff
@@ -19,10 +20,12 @@ Use these features when appropriate.
 
 **You should NOT use the following things**. Technically you could use them, but it's **not recommended**:
```
skipjack: I think we should change this to:

``` md
__Avoid the following practices:__
```

or

``` md
While they are technically feasible, __the following practices should be avoided__:
```

simon04: I changed that. Thank you!

PR #1314, opened 2017-06-20
docs(config/output): add libraryExport option

simon04:
Docs for https://github.com/webpack/webpack/pull/4987


content/configuration/output.md
```diff
@@ -258,6 +258,13 @@ Note that `output.libraryTarget` defaults to `var`. This means if only `output.l
 T> Read the [authoring libraries guide](/guides/author-libraries) guide for more information on `output.library` as well as `ouput.libraryTarget`.
 
 
+## `output.libraryExport`
+
+`string` or `string[]` (since webpack 3.0.0)
```
skipjack: Just out of curiosity, is `string[]` supposed to mean "array of strings"? I've seen this used in various other places but didn't realize that was the standard way to document it.

PR #1309, opened 2017-06-19
updated to add the new 3.0 scope hoisting flag

maraisr:

PR #1320, opened 2017-06-20
docs(stats, cli): add new stats options

simon04:
Docs for https://github.com/webpack/webpack/pull/4143

I will rebase and fix the table after #1309 is merged.


simon04:
Rebased.

PR #1322, opened 2017-06-20
Updated getting-started.md

TheDutchCoder:
* Made the diffs line up better
* Small spelling mistakes fixed
* Edited minor wording

### DIFF
--- a/content/guides/getting-started.md
+++ b/content/guides/getting-started.md
@@ -41,7 +41,7 @@ __project__
 __src/index.js__
 
 ``` javascript
-function component () {
+function component() {
   var element = document.createElement('div');
 
   // Lodash, currently included via a script, is required for this line to work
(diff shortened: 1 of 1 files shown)


content/guides/getting-started.md
```diff
@@ -107,34 +107,34 @@ __src/index.js__
 ``` diff
 + import _ from 'lodash';
 +
-function component () {
-  var element = document.createElement('div');
+  function component () {
```
skipjack: `component () {` => `component() {`

TheDutchCoder:
Done.

I always use `foo () {` for declarations and `foo()` for invocations, but it's a style preference 😉

skipjack:
@TheDutchCoder definitely style but I think through the rest of the docs we're pretty consistent with no space. Thanks for updating!

TheDutchCoder:
Makes sense! No problem!

TheDutchCoder:
@skipjack could you check what's up with Travis? It doesn't seem to want to run :/

skipjack:
Yeah not sure what's going on here but there may be some limit on how many builds we can do on Travis in a day or something. We've been hitting it pretty hard this morning. I'm going to cancel some of the builds and see if that helps. I think this may also have to do with #1326 which I'm not able to replicate but has been reported before when a lot of PRs are merged.

PR #1327, opened 2017-06-20
Update sidebar-item-style.scss

nhantdn:
Reduce the maximum width of `.sidebar-item__title` to 85% to prevent text dropping when the scroll bar appears in sidebar.

This issue happens to the item named `NormalModuleReplacementPlugin` this page: https://webpack.js.org/plugins/


skipjack:
Can you share a screenshot of what your seeing? I'm seeing the ellipsis slightly cut off but aside from that nothing unexpected.

nhantdn:
It appears on Windows, the scroll bar takes a space instead of overlaying.
Here is the screenshot:
image

skipjack:
Ah thanks @nhantdn. Yep this should be good go then thanks!

PR #1334, opened 2017-06-23
swapping `Usage` content to not be header links

### LINKED ISSUE #1306: Dev - Link Usage in Headers
The DllPlugin doc puts links inside a heading. this breaks the navigation on the left:
webpack_dll_doc

rouzbeh84:
this is for #1306 

Unless I'm missing something Usage links but the Examples subheaders are being pulled into sidebar when they shouldn't be?

[screen recording of `Usage` working as expected](https://cl.ly/lIGm)

closes #1306

### DIFF
--- a/content/plugins/dll-plugin.md
+++ b/content/plugins/dll-plugin.md
@@ -97,7 +97,7 @@ new webpack.DllReferencePlugin({
 
 ## Examples
 
-### [vendor](https://github.com/webpack/webpack/tree/master/examples/dll) and [user](https://github.com/webpack/webpack/tree/master/examples/dll-user)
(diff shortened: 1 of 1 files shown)


skipjack:
So I think this will fix but I think maybe the title I used for that issue caused some confusion. By "Link Usage" I meant using links in headers shouldn't display the actual markdown in the sidebar:

image

rouzbeh84:
@skipjack totally agree on the header links thing. we'll see how it shakes out re:

> we'll be updating our whole markdown parsing process soon anyway (see [webpack-contrib/webpack-defaults#73](https://github.com/webpack-contrib/webpack-defaults/issues/73 "Markdown Syntax and Processing")).

But doc linking to a linked header could be as simple as adding the link title twice, once as an id as well as href when processed? Just a suggestion as I'm sure that'd probably have unintended implications in other parts of the documentation.

PR #1337, opened 2017-06-24
Guides Cleanup

skipjack:
This pull request...

- Removes two guides which were incomplete and only had TODOs
- Moves the `comparison.md` page as it doesn't really fall under guides

See the commit messages for more details on the "why" behind these changes.

cc @TheDutchCoder @jakearchibald


skipjack:
@bebraw the build errors can be ignored as the unpkg 500 should [now be fixed](https://unpkg.com/) and the other one is just a missing edit link for the moved comparison guide.

content/guides/getting-started.md
```diff
@@ -109,7 +109,7 @@ __src/index.js__
 +
   function component() {
     var element = document.createElement('div');
-  
+
 -   // Lodash, currently included via a script, is required for this line to work
```
TheDutchCoder: This comment was on purpose, as it shows the user that in this case lodash is imported through a `<script>` tag, whereas later on it's imported by webpack.

Would you mind leaving it in (or maybe clarify it)?

skipjack: Ah yeah sorry about that. I think I either misunderstood or took this out accidentally. Will remove this change tonight.

skipjack: @TheDutchCoder just double-checked this. It's a bit confusing because your viewing a code block of the syntax highlighting language `diff` within a GitHub change diff. It tripped me up too when I first saw your comment. The diff from source tree is easier to understand:

image

It's only the blank line above that comment that changed. Just a minor editorconfig autofix 👍 . Your comment should remain untouched.

TheDutchCoder:
facepalm!

Thanks! :)

Issue #652, opened 2017-01-16
Guides - Explain Hash Changes in Caching Guide


jouni-kantola:
I got a scenario here, that I would like to discuss and hopefully create a source for docs, for what a developer should really think about to prevent causing hash changes.
I'm making something, that might seem like a very small change in my code, which has quite large effect on long-term caching.

I'm writing this issue in docs, because I'm sure it's expected, just that I want to be able to understand why this is happening, and hopefully help others as well.

Example repo is here (using `webpack@2.2.0-rc.7` and `webpack-chunk-hash@0.3.0`):
https://github.com/jouni-kantola/webpack-output-by-build-type/tree/why-hash-change

Here goes the scenario:
1) I build:

| Asset                                  | Size      | Chunks |           | 
|----------------------------------------|-----------|--------|-----------| 
| 0.9bbf83cf4b0bde5cdd21.dev.js.map      | 512 bytes | 0, 6   | [emitted] | 
| 0.9bbf83cf4b0bde5cdd21.dev.js          | 756 bytes | 0, 6   | [emitted] | 
| 2.a01031186175de379984.dev.js          | 455 bytes | 2, 6   | [emitted] | 
| 3.64a0d1a6257d07bf9c8e.dev.js          | 455 bytes | 3, 6   | [emitted] | 
| vendor.72eb1607924636054b21.dev.js     |  254 kB   | 4, 6   | [emitted] | 
| app.d67c37259cfc27adf14e.dev.js        | 16.9 kB   | 5, 6   | [emitted] | 
| 1.b255a8ecec8dd054c7f3.dev.js          | 756 bytes | 1, 6   | [emitted] | 
| 1.b255a8ecec8dd054c7f3.dev.js.map      | 512 bytes | 1, 6   | [emitted] | 
| 2.a01031186175de379984.dev.js.map      | 424 bytes | 2, 6   | [emitted] | 
| 3.64a0d1a6257d07bf9c8e.dev.js.map      | 430 bytes | 3, 6   | [emitted] | 
| vendor.72eb1607924636054b21.dev.js.map |  335 kB   | 4, 6   | [emitted] | 
| app.d67c37259cfc27adf14e.dev.js.map    | 18.9 kB   | 5, 6   | [emitted] | 
| index.html                             | 6.35 kB   |        | [emitted] | 

2) I change `module-a.js` from:
```
import is from 'is-thirteen';

[...]

PR #1338, opened 2017-06-24
Rework Code Splitting Guides

### LINKED ISSUE #1333: Guides - Code splitting seems to miss a basic example
When reading https://webpack.js.org/guides/code-splitting-libraries/ I was trying to figure out how to create a new file based on the common parts of two entry points.

I guess this is technically what the "vendor" example is doing, but as a first-time reader I didn't realise this, because this is in a page specific to "libraries", whereas I was wanting to avoid duplication in code *I'd* written.

I don't know if this would help people other than me, but I would have understood the guide better if it was structured something like this:

(the following is just showing the order I think things should be introduced, not final copy)

# Splitting out common code

If two of your entry points import/require the same modules, you'll end up with that code duplicated in each output file. `CommonsChunkPlugin` is a plugin that can extract common modules:

```js
var webpack = require('webpack');
var path = require('path');

module.exports = function(env) {
    return {
        entry: {
            page1: './page1/index.js',
            page2: './page2/index.js'
        },
        output: {
            filename: '[name].[chunkhash].js',
        },
        plugins: [
            new webpack.optimize.CommonsChunkPlugin({
                name: 'common' // Specify the common bundle's name.
            })
        ]
    }
};
```

This will generate:

* `page1.ef0942b9ff95bd25c024.js` - the stuff unique to page1.
* `page2.9de799314948aa4d7686.js` - the stuff unique to page2.
* `common.a7b0b08b0c83bdccebbd.js` - the stuff used in both page1 and page2.

Note: The filename of the chunk will inherit the filename of the output. `[name].[chunkhash].js` in this case.

`page1` won't automatically load `common`, you must include both on your page, and `common` must execute first:

```html
<script src="common.a7b0b08b0c83bdccebbd.js" defer></script>
<script src="page1.ef0942b9ff95bd25c024.js" defer></script>
```

# Splitting out webpack's bootstrap

Webpack's bootstap code for dealing with importing modules is usually included with every output. However, `CommonsChunkPlugin` will see this duplication and move it into one file, `common` in this case.

However, since the bootstrap contains details about the other modules, such as their names & hashes, updating the code unique to `page1` will also update `common`, since `page1`'s hash will change. Adding a `page3` will also update `common` since it will be adding details of this new module. This means the user is having to download a fresh `common` even though only a small part of the bootstrap has changed.

To solve this, we can extract the bootstrap into its own file:

```js
var webpack = require('webpack');
var path = require('path');

module.exports = function(env) {
    return {
        entry: {
            page1: './page1/index.js',
            page2: './page2/index.js'
        },
        output: {
            filename: '[name].[chunkhash].js',
        },
        plugins: [
            new webpack.optimize.CommonsChunkPlugin({
                name: 'common' // Specify the common bundle's name.
            }),
            new webpack.optimize.CommonsChunkPlugin({
                name: 'manifest'
            })
        ]
    }
};
```

This will generate:

* `page1.ef0942b9ff95bd25c024.js` - the stuff unique to page1.
* `page2.9de799314948aa4d7686.js` - the stuff unique to page2.
* `common.a7b0b08b0c83bdccebbd.js` - the stuff used in both page1 and page2.
* `manifest.1302ca8eccf18a208f11.js` - webpack's bootstrap and module manifest.

Instances of `CommonsChunkPlugin` are handled in order. Our first instance extracts all the common code to `common`. The second instance doesn't find any common code (we've already extracted it), but it will generate a file containing the bootstrap, and remove it from the others.

Now `manifest` must execute first:

```html
<script src="manifest.1302ca8eccf18a208f11.js" defer></script>
<script src="common.a7b0b08b0c83bdccebbd.js" defer></script>
<script src="page1.ef0942b9ff95bd25c024.js" defer></script>
```

# Splitting out rarely-updated modules

We've removed all of the duplication, but you may find that a small part of one of your bundles is updating frequently. It's best to extract that, so the rest of the bundle, which perhaps updates infrequently, can cache for longer.

To do this, create an additional entry point for the parts that frequently update (or the parts that infrequently update – the important thing is they're separate):

```js
var webpack = require('webpack');
var path = require('path');

module.exports = function(env) {
    return {
        entry: {
            page1: './page1/index.js',
            page2: './page2/index.js',
            fequentlyupdates: './page1/frequentlyupdates.js'
        },
        output: {
            filename: '[name].[chunkhash].js',
        },
        plugins: [
            new webpack.optimize.CommonsChunkPlugin({
                names: ['common', 'fequentlyupdates']
            }),
            new webpack.optimize.CommonsChunkPlugin({
                name: 'manifest'
            })
        ]
    }
};
```

`CommonsChunkPlugin` sees that `fequentlyupdates` is the name of an entry point, so it removes its code from the other modules.

Note: I expected two instances of `CommonsChunkPlugin` to work. One for `common`, then one for `fequentlyupdates`, but it throws an error. I don't know if this is a bug with webpack or a bug in my mental model.

## Splitting out a library

Library modules are typically rarely updated as part of your build. So let's say instead of `fequentlyupdates`, you wanted to extract `moment`, the date formatting library:

```js
var webpack = require('webpack');
var path = require('path');

module.exports = function(env) {
    return {
        entry: {
            page1: './page1/index.js',
            page2: './page2/index.js',
            moment: 'moment'
        },
        output: {
            filename: '[name].[chunkhash].js',
        },
        plugins: [
            new webpack.optimize.CommonsChunkPlugin({
                names: ['common', 'moment']
            }),
            new webpack.optimize.CommonsChunkPlugin({
                name: 'manifest'
            })
        ]
    }
};
```

Here we've replaced the path to the `fequentlyupdates` script with the module name.

## Splitting out everything in `node_modules`

Assuming you update dependencies less-frequently than your application code, you may want to split everything within `node_modules`:

```js
var webpack = require('webpack');
var path = require('path');

module.exports = function(env) {
    return {
        entry: {
            page1: './page1/index.js',
            page2: './page2/index.js'
        },
        output: {
            filename: '[name].[chunkhash].js',
        },
        plugins: [
            new webpack.optimize.CommonsChunkPlugin({
                name: 'common'
            }),
            new webpack.optimize.CommonsChunkPlugin({
                name: 'vendor',
                minChunks: function (module) {
                    return module.context && module.context.indexOf('node_modules') !== -1;
                }
            }),
            new webpack.optimize.CommonsChunkPlugin({
                name: 'manifest'
            })
        ]
    }
};
```

skipjack:
This pr attempts to:

- Merge _Code Splitting_ guides by removing all caching discussion and other extraneous content.
    - Focuses on actual code splitting approaches instead.
    - Moves certain `import()` documentation to the _Module API_ page.
    - Uses links to prevent content duplication.
- Replace `lazy-load-react` with a short, framework-agnostic `lazy-loading` page.
    - Still has a place at the bottom for links to framework-specific approaches.

Resolves #1333 (though some of that discussion relates to #652)

cc @TheDutchCoder @jakearchibald -- I'll update this again once it's ready for review which should be within the next hour or two.

@TheDutchCoder I could use your help with a few sections in the examples that I left as TODOs for now. Before it's merged, or in a follow-up PR, we'll make sure they are fully in sync with those from `getting-started`.

@jakearchibald I read through your suggested edits and the other comments from #1333. Seeing as I changed the direction of this guide a bit, I think most of your discussed changes would make more sense in the _Caching_ guide (which would mention _Code Splitting_ as a pre-requisite). If you'd be up for tackling that guide in a follow up PR, which should take care of #652, that would be amazing.


skipjack:
Ok this is pretty much read for review if anyone want to start. I'm just working on the _Dynamic Imports_ section of _Code Splitting_ now.

TheDutchCoder:
This is looking good, though I find it a little hard to follow at some points.

Not everyone is familiar with CommonJS and AMD. There's no need to go into great detail about them, but it would be nice to provide links to their respective documentation.

Other than that I think this looks good!

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
-## Resource Splitting
 
-Vendor and CSS code splitting methods help with both caching and parallel loading.
+## Entry Points
 
-### Vendor Code Splitting
+This is by far the easiest, and most intuitive, way to split code. However, it is more manual and has a some pitfalls we will go over. Let's take a look at how we might split the core libraries and frameworks an application uses from the actual source code:
 
-A typical application can depend on many third party libraries for framework/functionality needs. Unlike application code, code present in these libraries does not change often.
+__webpack.config.js__
```
TheDutchCoder: Would it be a good idea to continue with the same files we created in the earlier guides?
If you want I can create a PR after this is all merged to try and tie them together.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+?> Add bash example of webpack output
 
-Learn [how to split CSS](/guides/code-splitting-css) using the `ExtractTextWebpackPlugin`.
+As mentioned there are some pitfalls to this approach:
 
+- If there are any duplicated modules between entry chunks they will be included in both bundles.
+- It isn't as flexible and can't be used to dynamically split code with the core application logic.
 
-## On Demand Code Splitting
+The first of these two points is definitely an issue for our example, as `moment` is also imported within `./src/index.js` and will thus be duplicated in the output. See the `CommonChunkPlugin` example below for a solution to this problem.
```
TheDutchCoder: Is `moment` not `lodash`? :)

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
 
-While resource splitting of the previous kind requires the user to specify the split points upfront in the configuration, one can also create dynamic split points in the application code.
 
-This can be used for more granular chunking of code, for example, per our application routes or as per predicted user behaviour. This allows the user to load non-essential assets on demand.
+## Plugins & Loaders
 
-Learn [how to split on demand](/guides/code-splitting-async) using `import()` or `require.ensure()`.
+There are multiple plugins and loaders in webpack's ecosystem that can help with splitting and managing split code. The most widely utilized of these is the [`CommonsChunkPlugin`](/plugins/commons-chunk-plugin), a fairly advanced tool that allows us to extract common dependencies into an existing entry chunk or an entirely new chunk. Let's use this to de-duplicate the `moment` dependency from the previous example:
```
TheDutchCoder: `lodash`?

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+    filename: '[name].bundle.js',
+    path: path.resolve(__dirname, 'dist')
+  }
+};
+```
+
+This will yield the following build result (notice how our `index.bundle.js` has shrunk in size):
+
+?> Add bash example of webpack output
+
```
TheDutchCoder: Elaborate/Illustrate a bit on what the plugin has done under the hood, so that it's clear what happened.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+Here are some other useful plugins and loaders provide by the community for splitting code:
+
+- [`ExtractTextPlugin`](/plugins/extract-text-webpack-plugin): Useful for splitting CSS out from the main application.
```
TheDutchCoder: 404 on this link

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+- [`bundle-loader`](/loaders/bundle-loader): Used to split code and lazy load the resulting bundles.
```
TheDutchCoder: 404 on this link

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+
+- document.body.appendChild(component());
++ getComponent().then(component => {
++   document.body.appendChild(component);
++ })
+```
+
+Now, when we run webpack, we should see lodash separated out to a separate bundle:
+
+?> Add bash example of webpack output
```
TheDutchCoder:
``` bash
Hash: 5f3b08906b5e7d8d0799
Version: webpack 2.6.1
Time: 527ms
           Asset       Size  Chunks                    Chunk Names
 index.bundle.js  542 bytes       0  [emitted]         index
vendor.bundle.js     547 kB       1  [emitted]  [big]  vendor
   [0] ./~/lodash/lodash.js 540 kB {1} [built]
   [1] (webpack)/buildin/global.js 509 bytes {1} [built]
   [2] (webpack)/buildin/module.js 517 bytes {1} [built]
   [3] ./src/index.js 401 bytes {0} [built]
   [4] multi lodash 28 bytes {1} [built]
```

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
-This can be used for more granular chunking of code, for example, per our application routes or as per predicted user behaviour. This allows the user to load non-essential assets on demand.
+## Plugins & Loaders
 
-Learn [how to split on demand](/guides/code-splitting-async) using `import()` or `require.ensure()`.
+There are multiple plugins and loaders in webpack's ecosystem that can help with splitting and managing split code. The most widely utilized of these is the [`CommonsChunkPlugin`](/plugins/commons-chunk-plugin), a fairly advanced tool that allows us to extract common dependencies into an existing entry chunk or an entirely new chunk. Let's use this to de-duplicate the `moment` dependency from the previous example:
+
+__webpack.config.js__
+
+``` js
+const path = require('path');
```
TheDutchCoder: We use `var` everywhere else still.
I'm okay with using ES6, but it would be good to keep it consistent.

Adjust to `var` for now?

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+Now, when we run webpack, we should see lodash separated out to a separate bundle:
+
+?> Add bash example of webpack output
+
+If you've enabled [`async` functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) via a pre-processor like babel, note that you can simplify the code as `import()` statements just return promises:
+
+__src/index.js__
+
+``` diff
+  async function getComponent() {
```
TheDutchCoder: Don't forget to diff the 1st line ;)

content/api/module-methods.md
```diff
@@ -53,7 +53,11 @@ export default {
 
 ### `import()`
 
-Dynamically load modules.
+`import('path/to/module') -> Promise`
```
TheDutchCoder: Can we remove the `-> Promise` part, as it looks like something that the user should actually type in their code.

Maybe something like:
`import('path/to/module') // returns a Promise`

content/api/module-methods.md
```diff
@@ -63,8 +67,33 @@ if ( module.hot ) {
 }
 ```
 
-The compiler treats this as a split point and will split everything from `lodash` into a separate bundle. This returns a promise that will resolve to the module once the bundle has been loaded. See the [async code splitting guide](/guides/code-splitting-async) for more information.
+W> This feature relies on [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) internally. If you use `import()` with older browsers, remember to shim `Promise` using a polyfill such as [es6-promise](https://github.com/stefanpenner/es6-promise) or [promise-polyfill](https://github.com/taylorhakes/promise-polyfill). See [Shimming](/guides/shimming) for more information.
+
+One problem with direct spec-based usage is that we have no control over the split chunk's name or other properties. Luckily webpack allows some special parameters via comments so as to not break the spec:
```
TheDutchCoder: "One problem with direct spec-based usage ..." This might be a bit difficult to understand for beginners. How about something like: "The spec for ìmport` doesn't allow control over the chunk's name or other properties ..."

content/api/module-methods.md
```diff
@@ -63,8 +67,33 @@ if ( module.hot ) {
...
+
+``` js
+import(
+  /* webpackChunkName: "my-chunk-name" */
+  /* webpackMode: "lazy" */
+  'module'
+);
+```
+
+`webpackChunkName`: A name for the new chunk. Since webpack 2.6.0, the placeholders `[index]` and `[request]` are supported within the given string to an incremented number or the actual resolved filename respectively.
```
TheDutchCoder: I would love to see an example + output for this as a beginner 😄

content/guides/getting-started.md
```diff
@@ -109,7 +109,7 @@ __src/index.js__
 +
   function component() {
     var element = document.createElement('div');
-  
+
 -   // Lodash, currently included via a script, is required for this line to work
```
TheDutchCoder: Can we leave this comment? It's to illustrate that in the basic example we explicitly depend on lodash.

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+## Example
+
+Let's take the example from [Code Splitting](/guides/code-splitting#dynamic-imports) and tweak it a bit to demonstrate this concept even more. The code there does cause a separate chunk, `lodash.bundle.js`, to be generated and technically "lazy-loads" it as soon as the script is run. The trouble is that no user interaction is required to load the bundle -- meaning that every time the page is loaded, the request will fire. This doesn't help us too much, and actually may impact performance negatively.
+
+Let's try something different. We'll add an interaction to log some text to the console when the user clicks a button. However, we'll wait to load that code until the actual interaction occurs for the first time. To do this we'll go back and extend the original example from [Getting Started](/guides/getting-started) and leave the `lodash` in the main chunk.
+
```
TheDutchCoder: Can you add the project folder's structure in here?

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,73 @@
...
+__src/index.js__
+
+``` diff
+  import _ from 'lodash';
+
+  function component() {
+    var element = document.createElement('div');
++   var button = document.createElement('button');
++   var break = document.createElement('br');
```
TheDutchCoder: `break` is a reserved word in a switch, better use `br` or something (linters may not like it).

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,73 @@
...
+
+-   // Lodash, now imported by this script
++   button.innerHTML = 'Click me and look at the console!';
+    element.innerHTML = _.join(['Hello', 'webpack'], ' ');
++   element.appendChild(break);
++   element.appendChild(button);
++
++   button.onclick = e => import(/* webpackChunkName: "console" */ './console').then(module => {
```
TheDutchCoder: Should be `'./print'`

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  document.body.appendChild(component());
+```
+
+W> Note that when using `import()` on ES6 modules you must reference the `.default` property as it's the actual `module` object that will be returned when the promise is resolved.
+
+Now let's run webpack and check out our new lazy-loading functionality:
+
+?> Add bash example of webpack output
```
TheDutchCoder: This will work, but the `index.html` file doesn't reference the right files. I think this will be solved in my guide, but just wanted to make a not of it.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+  }
+};
+```
 
-### CSS Splitting
+This will yield the following build result:
 
-You might also want to split your styles into a separate bundle, independent from application logic.
-This enhances cacheability of your styles and allows the browser to load the styles in-parallel with your application code, thus preventing a FOUC ([flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content)).
+?> Add bash example of webpack output
```
TheDutchCoder:
``` bash
Hash: d1e4eab63dc6ba09c930
Version: webpack 2.6.1
Time: 531ms
           Asset    Size  Chunks                    Chunk Names
vendor.bundle.js  544 kB       0  [emitted]  [big]  vendor
 index.bundle.js  544 kB       1  [emitted]  [big]  index
   [0] ./~/lodash/lodash.js 540 kB {0} {1} [built]
   [1] (webpack)/buildin/global.js 509 bytes {0} {1} [built]
   [2] (webpack)/buildin/module.js 517 bytes {0} {1} [built]
   [3] ./src/index.js 216 bytes {1} [built]
   [4] multi lodash 28 bytes {0} [built]
```

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+## Plugins & Loaders
 
-Learn [how to split on demand](/guides/code-splitting-async) using `import()` or `require.ensure()`.
+There are multiple plugins and loaders in webpack's ecosystem that can help with splitting and managing split code. The most widely utilized of these is the [`CommonsChunkPlugin`](/plugins/commons-chunk-plugin), a fairly advanced tool that allows us to extract common dependencies into an existing entry chunk or an entirely new chunk. Let's use this to de-duplicate the `moment` dependency from the previous example:
+
+__webpack.config.js__
+
+``` js
+const path = require('path');
+const webpack = require('webpack');
```
TheDutchCoder: Make this a diff

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+module.exports = {
+  entry: {
+    index: './src/index.js',
+    vendor: [
+      'lodash'
+    ]
+  },
+  plugins: [
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor' // Specify the common bundle's name.
```
TheDutchCoder: This will result in an error: `Uncaught ReferenceError: webpackJsonp is not defined`, I'm guessing there's a conflict between the `vendor` entry and the `vendor` name for the CommensChunckPlugin?

content/guides/code-splitting.md
```diff
@@ -1,41 +1,174 @@
...
+  output: {
+    filename: '[name].bundle.js',
+    path: path.resolve(__dirname, 'dist')
+  }
+};
+```
+
+This will yield the following build result (notice how our `index.bundle.js` has shrunk in size):
+
+?> Add bash example of webpack output
```
TheDutchCoder:
``` bash
Hash: bc0cb296200698aee722
Version: webpack 2.6.1
Time: 540ms
           Asset       Size  Chunks                    Chunk Names
 index.bundle.js  665 bytes       0  [emitted]         index
vendor.bundle.js     547 kB       1  [emitted]  [big]  vendor
   [0] ./~/lodash/lodash.js 540 kB {1} [built]
   [1] (webpack)/buildin/global.js 509 bytes {1} [built]
   [2] (webpack)/buildin/module.js 517 bytes {1} [built]
   [3] ./src/index.js 216 bytes {0} [built]
   [4] multi lodash 28 bytes {1} [built]
```

TheDutchCoder:
@skipjack Added some comments, but other than that this is looking really good.

When I'm tackling `Output Management` we can make the examples consistent probably. Will probably start work on that today.

content/guides/code-splitting.md
skipjack: You mean do the file diffs we did in other guides? Yeah we should add those here for consistency as well.

content/guides/code-splitting.md
skipjack: Yep, good catch.

content/guides/code-splitting.md
skipjack: 😆  ahh man... I'll correct this as well.

content/guides/code-splitting.md
skipjack: Will do 👍

content/guides/code-splitting.md
skipjack: Where are you seeing the 404? It [works fine](https://webpack.js.org/plugins/extract-text-webpack-plugin) for me.

content/guides/code-splitting.md
skipjack: This one [works fine](https://webpack.js.org/loaders/bundle-loader) for me as well.

skipjack: Plus the [CI build/test](https://travis-ci.org/webpack/webpack.js.org/builds/246645985?utm_source=github_status&utm_medium=notification) didn't catch any issues with either of them.

content/guides/code-splitting.md
skipjack: 👍

skipjack: Btw, are you maintaining a repo somewhere to test all these things from the guides? Would love to help out there and use it for testing...

content/guides/code-splitting.md
skipjack: Are you saying we use `var` specifically for `require`d packages in the other guides' _webpack.config.js_ examples? I'd rather use `const` here and I'd be happy to go update the others for consistency.

If you're talking about `var`/`const`/`let` usage in general I think it's ok to use them where they make sense as each are used for somewhat different things.

content/guides/code-splitting.md
skipjack: Ah good catch.

content/api/module-methods.md
skipjack: Yeah, I actually copied this from one of the other guides and modified it slightly. It's supposed to be a method/function "signature" I think. @bebraw is there an agreed upon standard syntax for this somewhere? I feel like there's a mix throughout the documentation that should probably be made consistent at some point.

content/api/module-methods.md
skipjack: Yep, that sounds better thanks!

content/api/module-methods.md
skipjack: There will be in both the `lazy-loading` and `code-splitting` guides. I want to avoid too many and too detailed examples in the actual reference documentation but we could potentially add a link to the guide here.

content/guides/lazy-loading.md
skipjack: Yep, as discussed above, we'll make sure to add the file/directory diffs throughout both guides for consistency.

content/guides/lazy-loading.md
skipjack: Ah good catch.

content/guides/lazy-loading.md
skipjack: 👍

content/guides/code-splitting.md
skipjack: 🎉

content/guides/lazy-loading.md
skipjack: > I think this will be solved in my guide, but just wanted to make a not of it.

Solved by the edits to the `output-management` guide?

content/guides/code-splitting.md
skipjack: 👍

content/guides/code-splitting.md
skipjack: 🎉

skipjack:
@TheDutchCoder awesome feedback, thank you! I will make these changes soon.

content/guides/code-splitting.md
TheDutchCoder: Okay sounds goo to me.

It's just that const/let is not implemented in all browsers yet and I think we shouldn't assume everyone uses babel for instance.

content/guides/code-splitting.md
TheDutchCoder: Ah, it's a relative URL, so it will 404 when I view this file for example.
Probably not an issue in the build then!

content/guides/code-splitting.md
TheDutchCoder: I actually just do it all locally, doing step by step to see the results, but maybe it would be an idea to setup a repo with the actual sources per guide or something?

TheDutchCoder:
@skipjack more than welcome!

content/api/module-methods.md
skipjack: I double-checked and it is used in both the _Dynamic Imports_ example in _Code Splitting_ as well as the example in _Lazy Loading_.

content/api/module-methods.md
bebraw: Nope. It should be specified at the writer's guide.

content/guides/code-splitting.md
skipjack: Ok, so yes you are correct that maybe within the `/src` files we shouldn't but within the configs it should be ok because Node has supported it for a little while now. So I would revise my earlier statement to be Node imports within `webpack.config.js` examples should be `const name = require('package'`. That's what we're using in the rest of the docs as well and I think it's the "recommended" practice until Node supports `import` and `export` syntax.

content/guides/code-splitting.md
skipjack: Yeah the more we go through this the more I think it wouldn't be a bad idea, at least for testing. Maybe at some point we expose it as an "official" boilerplate to follow along with the guides but for now should at least be useful to us.

skipjack: By the way I added this but at line `110` which is where I think you meant it to be. The example at `142` should yield a `lodash.bundle.js` file although things might be a little confusing without the file/directory diffs in place yet.

skipjack:
Ok applied the first round of changes and I'll give this another read through tomorrow night. @bebraw @jakearchibald you should both feel free to review as well.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,199 @@
...
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor' // Specify the common bundle's name.
+    })
+  ],
+  output: {
+    filename: '[name].bundle.js',
+    path: path.resolve(__dirname, 'dist')
+  }
+};
+```
```
jakearchibald: I'm not convinced this is the correct order to teach it. It might just be me, but I wanted `CommonsChunkPlugin` to extract all common modules from entry points into a new file.

From that point, it's easier to explain "but what if you only want to extract a specific module?" which is the example above.

That's the order I went for in https://github.com/webpack/webpack.js.org/issues/1333.

jakearchibald: Also, this guide is missing how to use these chunks.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,199 @@
...
+
+With the `CommonsChunkPlugin` in place, we should now see the duplicate dependency removed from our `index.bundle.js`. The plugin should notice that we've separated `lodash` out to a separate chunk and remove the dead weight from our main bundle. Let's do an `npm run build` to see if it worked:
```
jakearchibald: Will the reader know that `npm run build` will run webpack via a script in `packages.json`?

content/guides/code-splitting.md
```diff
@@ -1,41 +1,199 @@
...
+
+``` bash
+Hash: 5f3b08906b5e7d8d0799
+Version: webpack 2.6.1
+Time: 527ms
+           Asset       Size  Chunks                    Chunk Names
+ index.bundle.js  542 bytes       0  [emitted]         index
```
jakearchibald: A script coming out this small looks like an error to me. Maybe you need to make it clear that virtually nothing is happening in this script other than importing lodash.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,199 @@
...
+``` diff
+- import _ from 'lodash';
+-
+- function component() {
++ function getComponent() {
+-   var element = document.createElement('div');
+-
+-   // Lodash, now imported by this script
+-   element.innerHTML = _.join(['Hello', 'webpack'], ' ');
++   return import(/* webpackChunkName: "lodash" */ 'lodash').then(module => {
```
jakearchibald: The purpose of this comment isn't mentioned

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,73 @@
...
+---
+
+T> This guide is a small follow-up to [Code Splitting](/guides/code-splitting). If you have not yet read through that guide, please do so now.
+
+Lazy, or "on demand", loading is a great way to optimize your site or application. This practice essentially boils down to splitting your code at logical breakpoints, and then loading it once the user has done something that requires, or will require, a new block of code. This speeds up the initial load of the application and makes the lightens its overall weight as some blocks may never even be loaded.
+
+
+## Example
+
+Let's take the example from [Code Splitting](/guides/code-splitting#dynamic-imports) and tweak it a bit to demonstrate this concept even more. The code there does cause a separate chunk, `lodash.bundle.js`, to be generated and technically "lazy-loads" it as soon as the script is run. The trouble is that no user interaction is required to load the bundle -- meaning that every time the page is loaded, the request will fire. This doesn't help us too much, and actually may impact performance negatively.
```
jakearchibald: > may impact performance negatively.

*will*, surely?

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,73 @@
...
++   button.innerHTML = 'Click me and look at the console!';
+    element.innerHTML = _.join(['Hello', 'webpack'], ' ');
++   element.appendChild(br);
++   element.appendChild(button);
++
++   button.onclick = e => import(/* webpackChunkName: "print" */ './print').then(module => {
++     var print = module.default;
++
++     print();
++   })
```
jakearchibald: I realise the guide should be as simple as possible, but this should show a spinner since UI is blocked on network. Maybe just mention it as a note? May also be worth linking to `<link rel="preload">` https://developers.google.com/web/updates/2016/03/link-rel-preload, so you can handle the download in the background and have it ready by the time the user clicks.

content/guides/code-splitting.md
TheDutchCoder: Okay let me double check this, maybe I added the wrong output :)

TheDutchCoder: Nope this results in `index.bundle.js` and `vendor.bundle.js`, not `lodash.bundle.js`.

Ahhh, this is a webpack 2.4.0+ specific thing, I still run 2.6.1, so that's probably why.

We should probably make a note of which version to use for the guides. I'm assuming the latest stable build for the major version?

content/guides/code-splitting.md
```diff
@@ -1,41 +1,199 @@
...
+   [3] ./src/index.js 216 bytes {1} [built]
+   [4] multi lodash 28 bytes {0} [built]
+```
+
+As mentioned there are some pitfalls to this approach:
+
+- If there are any duplicated modules between entry chunks they will be included in both bundles.
+- It isn't as flexible and can't be used to dynamically split code with the core application logic.
+
+The first of these two points is definitely an issue for our example, as `lodash` is also imported within `./src/index.js` and will thus be duplicated in the output. See the `CommonChunkPlugin` example below for a solution to this problem.
```
TheDutchCoder: Good point, it's okay to point out what fails, but the guides should always show you what works.

jakearchibald:
I really like how these changes are looking btw!

content/guides/code-splitting.md
TheDutchCoder: Yes, it is explained in the first guide, but we should maybe link to it here (we can't assume everyone goes through all the guides sequentially, so cross-linking is a good idea)

content/guides/code-splitting.md
TheDutchCoder: Actually, it imports lodash, creates a new element, sets its innerHTML, and then appends it to the DOM.

It's a tiny script indeed, and just for illustration, but I don't think it would throw people off.

content/guides/lazy-loading.md
TheDutchCoder: That's a good idea, if we can keep it easy to understand.

content/guides/code-splitting.md
jakearchibald: It's probably ok.

When I was trying to figure out `CommonsChunkPlugin` I sometimes ended up with everything ending up in the "vendor" chunk, so I guess that's why I associate a small index bundle with "shit went bad" 😄

content/guides/code-splitting.md
skipjack: @TheDutchCoder I believe the discrepancy is because I haven't added the file diff and I may have forgotten to diff the removal of the extra `entry` and `CommonsChunkPlugin`. The `import()` behavior is smart enough not to split things out if the package has already been included elsewhere.

skipjack: I'll review and update this section tonight to make it clearer.

content/guides/code-splitting.md
skipjack: Ok I guess a better approach would be an entirely separate entry file that isn't imported anywhere else.

content/guides/code-splitting.md
skipjack: Yeah I think I forgot to add the...

> This guide is a follow up to [Getting Started]() and [Output Management]()...

at the beginning. Maybe I should link to just one of those two -- I guess you were thinking this would be a good follow up to _Output Management_ @TheDutchCoder?

skipjack:
@jakearchibald thanks so much for the thorough review! I will do my best to address your comments (and @TheDutchCoder's remaining comments) tonight.

content/guides/code-splitting.md
TheDutchCoder: Yes I think that this would be a good follow up Output Management once I start working on it haha

content/api/module-methods.md
```diff
@@ -63,8 +67,33 @@ if ( module.hot ) {
...
+
+`webpackMode`: Since webpack 2.6.0, different modes for resolving dynamic imports can be specified. The following options are supported:
+
+- `"lazy"` (default): Generates a chunk per request, meaning everything is lazy loaded.
```
skipjack: Yes I believe for an expression like ``import(`./locales/${code}.json`)``, `"lazy"` will generate a chunk for every possible import, so that once the `code` variable is processed, each JSON file can be loaded individually. And `"lazy-once"` would generate a single chunk containing all possible JSON files that would only be loaded on the first execution of `import()` (or what it's transformed to) at runtime.

skipjack: See my comment above.

content/api/module-methods.md
```diff
@@ -63,8 +67,33 @@ if ( module.hot ) {
...
+- `"lazy-once"`: Generates a single chunk for all possible requests. The first call initiates a network request for all modules, and any following requests are already fulfilled. This is only available for when importing an expression.
```
skipjack: See my comment above, "expression" means something like this: ``import(`./locales/${code}.json`)``. Meaning a partially dynamic statement that gives webpack enough information to do its bundling but will still be evaluated at runtime to get the remaining information (e.g. `code`).

skipjack: I'm working on updating this section now to clarify these things.

content/api/module-methods.md
```diff
@@ -63,8 +67,33 @@ if ( module.hot ) {
...
+- `"eager"`: Generates no chunk. All modules are included in the current chunk and no additional network requests are made. A `Promise` is still returned but is already resolved. In contrast to a static import, the module isn't executed until the request is made.
 
+T> Note that both options can be combined like so `/* webpackMode: "lazy-once", webpackChunkName: "all-i18n-data" */`. This is parsed as a JSON5 object without curly brackets.
+
+W> Fully dynamic statements, such as `import(foo)`, __will fail__ because webpack requires at least some file location information. This is because `foo` could potentially be any path to any file in your system or project. The `import()` must contain at least some information about where the module is located, so bundling can be limited to a specific directory or set of files.
+
+W> The entire module namespace is included. For example, ``import(`./locale/${language}.json`)`` will cause every `.json` file in the `./locale` directory to be bundled into the new chunk. At run time, when the variable `language` has been computed, any file like `english.json` or `german.json` will be available for consumption.
```
skipjack: By "namespace" I think the original author meant...

> every module that could potentially be requested by a statement like ``import(`./locales/${code}.json`)``

skipjack: I'll clarify.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,199 @@
...
+
+
+## Plugins & Loaders
+
+There are multiple plugins and loaders in webpack's ecosystem that can help with splitting and managing split code. The most widely utilized of these is the [`CommonsChunkPlugin`](/plugins/commons-chunk-plugin), a fairly advanced tool that allows us to extract common dependencies into an existing entry chunk or an entirely new chunk. Let's use this to de-duplicate the `lodash` dependency from the previous example:
```
skipjack: I renamed the section _Prevent Duplication_ and updated the list and descriptions above.

content/guides/code-splitting.md
skipjack: > I'm not convinced this is the correct order to teach it. It might just be me, but I wanted `CommonsChunkPlugin` to extract all common modules from entry points into a new file.

I tweaked the example so it demonstrates exactly what you describe as I agree that it's a more intuitive way to teach it. There are now two separate `src` modules that duplicate the `lodash` dependency and the _Prevent Duplication_ section resolves that duplication.

skipjack: > Also, this guide is missing how to use these chunks.

I'm not sure I understand what you mean? Do you mean how to include them in your `index.html` file?

content/guides/code-splitting.md
skipjack: Added some comments about this and a link to the reference in the paragraph below.

content/guides/lazy-loading.md
skipjack: Added a comment above the click handler to indicate this. Because this guide is relatively short, I actually wouldn't be opposed to you updating this example to really display one. Maybe a follow up PR?

skipjack:
Ok pretty much finished addressing the reviews although I think there still may be one or two file diffs and example updates needed. I'll work on finishing this up later today and merge #1337 in preparation for when this is good to go.

@TheDutchCoder note that because the examples changed, we'll need to update the bash output again 😞 , sorry! On the bright side they should all be a bit clearer and work as expected now 👍 . Definitely push that local demo you're maintaining to a repo at some point, I (and probably others as well) would love to help maintain that with you.

TheDutchCoder:
No worries, maybe we can add all the outputs when the guide is done, in one commit or PR.

Since we usually tweak our scripts/configs quite a bit while rewriting, it might be better to run them all consecutively at the end. This way we also catch any additional errors.

content/guides/code-splitting.md
jakearchibald: > do you mean how to include them in your index.html file?

Yeah. I know it's simple, but it isn't clear that you need to do that.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,226 @@
...
+    path: path.resolve(__dirname, 'dist')
+  }
+};
+```
 
-This can be used for more granular chunking of code, for example, per our application routes or as per predicted user behaviour. This allows the user to load non-essential assets on demand.
+This will yield the following build result:
 
-Learn [how to split on demand](/guides/code-splitting-async) using `import()` or `require.ensure()`.
+?> Update the bash output
```
TheDutchCoder:
``` bash
Hash: 309402710a14167f42a8
Version: webpack 2.6.1
Time: 570ms
            Asset    Size  Chunks                    Chunk Names
  index.bundle.js  544 kB       0  [emitted]  [big]  index
another.bundle.js  544 kB       1  [emitted]  [big]  another
   [0] ./~/lodash/lodash.js 540 kB {0} {1} [built]
   [1] (webpack)/buildin/global.js 509 bytes {0} {1} [built]
   [2] (webpack)/buildin/module.js 517 bytes {0} {1} [built]
   [3] ./src/another-module.js 87 bytes {1} [built]
   [4] ./src/index.js 216 bytes {0} [built]
```

content/guides/code-splitting.md
```diff
@@ -1,41 +1,226 @@
...
+
+As mentioned there are some pitfalls to this approach:
+
+- If there are any duplicated modules between entry chunks they will be included in both bundles.
+- It isn't as flexible and can't be used to dynamically split code with the core application logic.
+
+The first of these two points is definitely an issue for our example, as `lodash` is also imported within `./src/index.js` and will thus be duplicated in both bundles. See the `CommonsChunkPlugin` example below for a solution to this problem.
```
TheDutchCoder: I would rephrase the end of this sentence, maybe something like "and will thus be duplicated in both bundles. Let's remove this duplication by using the `CommonsChunkPlugin`.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,226 @@
...
+    output: {
+      filename: '[name].bundle.js',
+      path: path.resolve(__dirname, 'dist')
+    }
+  };
+```
+
+With the `CommonsChunkPlugin` in place, we should now see the duplicate dependency removed from our `index.bundle.js`. The plugin should notice that we've separated `lodash` out to a separate chunk and remove the dead weight from our main bundle. Let's do an `npm run build` to see if it worked:
+
+?> Update bash output.
```
TheDutchCoder:
``` bash
Hash: 70a59f8d46ff12575481
Version: webpack 2.6.1
Time: 510ms
            Asset       Size  Chunks                    Chunk Names
  index.bundle.js  665 bytes       0  [emitted]         index
another.bundle.js  537 bytes       1  [emitted]         another
 common.bundle.js     547 kB       2  [emitted]  [big]  common
   [0] ./~/lodash/lodash.js 540 kB {2} [built]
   [1] (webpack)/buildin/global.js 509 bytes {2} [built]
   [2] (webpack)/buildin/module.js 517 bytes {2} [built]
   [3] ./src/another-module.js 87 bytes {1} [built]
   [4] ./src/index.js 216 bytes {0} [built]
```

content/guides/code-splitting.md
```diff
@@ -1,41 +1,226 @@
...
+
+## Dynamic Imports
+
+Two similar techniques are supported by webpack when it comes to dynamic code splitting. The first and more preferable approach is use to the [`import()` syntax](/api/module-methods#import-) that conforms to the [ECMAScript proposal](https://github.com/tc39/proposal-dynamic-import) for dynamic imports. The legacy, webpack-specific approach is to use [`require.ensure`](/api/module-methods#require-ensure). Let's try using the first of these two approaches...
+
+Before we start, let's remove the `entry` and `CommonsChunkPlugin` from our config as they won't be needed for this next demonstration:
+
+__webpack.config.js__
+
+``` diff
```
TheDutchCoder: This diff is no longer in line with the previous code :)

content/guides/code-splitting.md
```diff
@@ -1,41 +1,226 @@
...
+
+- document.body.appendChild(component());
++ getComponent().then(component => {
++   document.body.appendChild(component);
++ })
+```
+
+Note the use of `webpackChunkName` in the comment. This will cause our separate bundle to be named `lodash.bundle.js` instead of just `[id].bundle.js`. For more information on `webpackChunkName` and the other available options, see the [`import()` documentation](/api/module-methods#import-). Let's run webpack to see lodash separated out to a separate bundle:
+
+?> Add bash example of webpack output
```
TheDutchCoder:
``` bash
Hash: a27e5bf1dd73c675d5c9
Version: webpack 2.6.1
Time: 544ms
          Asset     Size  Chunks                    Chunk Names
    0.bundle.js   541 kB       0  [emitted]  [big]  lodash
index.bundle.js  6.35 kB       1  [emitted]         index
   [0] ./~/lodash/lodash.js 540 kB {0} [built]
   [1] ./src/index.js 377 bytes {1} [built]
   [2] (webpack)/buildin/global.js 509 bytes {0} [built]
```

content/guides/code-splitting.md
```diff
@@ -1,41 +1,226 @@
...
++   }).catch(error => 'An error occurred while loading the component');
+  }
+
+- document.body.appendChild(component());
++ getComponent().then(component => {
++   document.body.appendChild(component);
++ })
+```
+
+Note the use of `webpackChunkName` in the comment. This will cause our separate bundle to be named `lodash.bundle.js` instead of just `[id].bundle.js`. For more information on `webpackChunkName` and the other available options, see the [`import()` documentation](/api/module-methods#import-). Let's run webpack to see lodash separated out to a separate bundle:
```
TheDutchCoder: So this isn't happening. I use webpack 2.6.1, which is the latest 2.x release and my bundles are named `0.bundle.js` and `index.bundle.js` respectively. I will see if I can find out more about why this is.

TheDutchCoder: I think I understand what's going on.

The **chunk** is named correctly, but the output isn't, because we removed it from the webpack config. It'll just generate an index-based name as far as I can tell.

So the output is correct, just this part of the guide isn't ;)

skipjack: I think we need to include `chunkFilename` in the `output` configuration. I had a feeling that might be needed...

skipjack: See the [output documentation](https://webpack.js.org/configuration/output/#output-chunkfilename). I think that will fix the issue (e.g. `chunkFilename: '[name].bundle.js'`).

TheDutchCoder: Yes correct, I just tried it and it now works as expected, so we should add this to the guide.

skipjack: Will do! Thanks!

content/guides/code-splitting.md
skipjack: Yep, good point. @TheDutchCoder what do you think, should I just manually insert it into the `index.html` file or include the `HTMLWebpackPlugin` as shown in _Output Management_ seeing as we want this to stem from that guide anyway. I'm thinking the latter is probably best even if `output-management` isn't fully ready yet.

Also, on a related note, what do you guys think about showing the `/dist` directories contents in the __project__ file diffs? I think once past _Output Management_, every guide should show the `HTMLWebpackPlugin` maybe with a comment in the file diffs saying everything in there is auto-generated and should not be edited manually.

content/guides/code-splitting.md
skipjack: Updated the diffs and added that option.

content/guides/code-splitting.md
TheDutchCoder: @skipjack yes I would include the HtmlWebpackPlugin for this so it flows better from the previous guide. 👍 

Diffs in the dist dir would be useful, but only if we add something that cleans up the `/dist` directory (it gets really cluttered early on). Maybe I can put that in `Output Management` as it makes sense to do it there as we're also moving the `index.html` into the `/src` folder.

Then from that point on we can include the diffs on the `/dist` folder.

skipjack: > @skipjack yes I would include the HtmlWebpackPlugin for this so it flows better from the previous guide.

Ok I'm going to add that to the examples.

> Diffs in the dist dir would be useful, but only if we add something that cleans up the `/dist` directory (it gets really cluttered early on). Maybe I can put that in `Output Management` as it makes sense to do it there as we're also moving the `index.html` into the `/src` folder.

I actually meant I don't think we should show the contents of `/dist` and instead just show add a note like so:

``` diff
webpack-demo
|- package.json
|- webpack.config.js
|- /dist 
  |- auto-generated files
|- /src
  |- index.js
+ |- another-module.js
|- /node_modules
```

Or even just leave out the `|- auto-generated files` bit. Seeing as with the `HTMLWebpackPlugin`, users shouldn't ever have to modify the contents of `/dist`.

skipjack:
Ok I just cleaned up this branch and added some final commits. Once the build passes I'll squash and merge. @TheDutchCoder once it's merged feel free to give these guides one last pass and pr any updates you think might be needed. 

I'll start reviewing your other PRs once this sucker is merged.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,290 @@
...
+module.exports = {
+  entry: {
+    index: './src/index.js',
+    another: './src/another-module.js'
+  },
+  output: {
+    filename: '[name].bundle.js',
+    path: path.resolve(__dirname, 'dist')
+  },
+  plugins: [
```
TheDutchCoder: Minor: In other examples we put `plugins` before `output`. I don't care much about the order of things, but it might be nice to keep it consistent.

content/guides/code-splitting.md
```diff
@@ -1,41 +1,290 @@
...
+            Asset    Size  Chunks                    Chunk Names
+  index.bundle.js  544 kB       0  [emitted]  [big]  index
+another.bundle.js  544 kB       1  [emitted]  [big]  another
+   [0] ./~/lodash/lodash.js 540 kB {0} {1} [built]
+   [1] (webpack)/buildin/global.js 509 bytes {0} {1} [built]
+   [2] (webpack)/buildin/module.js 517 bytes {0} {1} [built]
+   [3] ./src/another-module.js 87 bytes {1} [built]
+   [4] ./src/index.js 216 bytes {0} [built]
+```
+
```
TheDutchCoder: Make mention of a possible `index.html` as output as well?
It depends on how we tie this guide to the previous one, where a template is provided to the HtmlWebpackPlugin.

Let's wait with finalizing the outputs until `Output Management` is merged, so we know for sure.

content/guides/lazy-loading.md
```diff
@@ -0,0 +1,109 @@
...
++   element.appendChild(br);
++   element.appendChild(button);
++
++   // Note that because a network request is involved, some indication
++   // of loading would need to be shown in a production-level site/app.
++   button.onclick = e => import(/* webpackChunkName: "print" */ './print').then(module => {
++     var print = module.default;
++
++     print();
++   })
```
TheDutchCoder: `<nitpick>`Add semi colon :)`</nitpick>`

TheDutchCoder:
💯 

Some minor things we can add in a later PR, but this is looking fantastic!

content/guides/code-splitting.md
skipjack: I agree, I'll change in a follow up PR. I actually like the order of `entry` => `module` => `plugins` => (anything else) => `output` as it kind of reflects the order of operations.

content/guides/lazy-loading.md
skipjack: 😆  will do -- I'm in the ASI boat myself but I'll fix for consistency.

Krinkle:
This seems to have broken links to <https://webpack.js.org/guides/code-splitting-async/>. Could a redirect be put in place? E.g. linked from <https://eng.uber.com/m-uber/> and various other places.

skipjack:
@Krinkle good catch, will do. I'll add some for the other removed removed pages as well while I'm at it.

PR #1339, opened 2017-06-25
Miscellaneous Edits

### LINKED ISSUE #569: General - Clarify Loader Configuration
Recently (during a discussion of https://github.com/webpack/webpack/issues/2797) I went to see how loaders are configured in Webpack 2. I found it in the end, but the journey was a long one:

- [Concepts -> Loaders](https://webpack.js.org/concepts/loaders/) only mentions they can accept query parameters without describing what that is
- [Guides](https://webpack.js.org/guides/)... May be there are loaders, but I don't know in which one
- [Documentation](https://webpack.js.org/configuration/). There's a spec for the documentation. And you can see the `options` param somewhere in `rules`
- Then I went like this: Documentation -> Module -> Rule.loader -> [optional: Rule.use] -> [UseEntry.loader](https://webpack.js.org/configuration/module/#useentry). 

Note: the last step of the journey:
- contradicts concepts. Concepts only mention the query. Both `UseEntry.loader` and config spec only mention `options`
- is a bit confusing (there's now `loader` and `loaders`, _and_ `Rule.use`? and all of them do the same thing?)

skipjack:
> In resolving #569, I had to grep through much of the docs to look for loader usage inconsistencies. In doing this I noticed a variety of other minor (and some not-so-minor) things that needed fixing so I'm addressing them here as well. This should probably have been split into multiple PRs but if the reviewers just view the commits instead of the full diff it shouldn't be so bad (sorry it's late and I'm tired 😞).

A bunch of cleanup including...

- Resolving some formatting inconsistencies
- De-duplicating content between `/concepts/output` and `/configuration/output`
- Clarifying loader usage and configuration (see #569)

Resolves #569

cc @dmitriid (see [this commit][1] in particular)

[1]: https://github.com/webpack/webpack.js.org/pull/1339/commits/9171a0a9fceaeaea6bdced3f8ada50215317d587

### DIFF
--- a/content/api/loaders.md
+++ b/content/api/loaders.md
@@ -139,10 +139,10 @@ In the example: `"/abc/loader1.js?xyz!/abc/node_modules/loader2/index.js!/abc/re
 
 ### `this.query`
 
-1. In case the loader was configured with an [`options`](/configuration/module/#useentry) object, this will be a reference to the object.
+1. If the loader was configured with an [`options`](/configuration/module/#useentry) object, this will point to that object.
 2. If the loader has no `options`, but was invoked with a query string, this will be a string starting with `?`.
 
-T> Use the [`getOptions` method from the `loader-utils`](https://github.com/webpack/loader-utils#getoptions) to extract the given loader options.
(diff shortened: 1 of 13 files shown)


skipjack:
@bebraw @dmitriid merged this... but please still feel free to review and I'll do my best to make any corrections or additions in a separate pr.

PR #1349, opened 2017-06-28
fix definition of `webpack -p`

JonnieCache:
fix definition of `webpack -p` command line option expansion, at the top of this page its missing the needed single quotes which are present down below. this caused me some grief ;)

incidentally `--optimize-minimize --define process.env.NODE_ENV="'production'"` *doesnt* produce the same result as `-p` for me, only the first correctly causes react to build in production mode without the attedent warnings. this is with webpack 1.15.0 and react 15.6.1. I assume this is a doc issue rather than something i need to create as an issue on either react or webpack but let me know if im wrong here.

### DIFF
--- a/content/guides/production.md
+++ b/content/guides/production.md
@@ -20,7 +20,7 @@ The following article describes the best practices and tools to use when using w
 
 ## The Automatic Way
 
-Running `webpack -p` (or equivalently `webpack --optimize-minimize --define process.env.NODE_ENV="production"`). This performs the following steps:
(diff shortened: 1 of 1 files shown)


skipjack:
Hmmm well I think you are correct that the quotes are needed but will let @bebraw have the final say. Also, just as a general note, this repo is geared towards the latest major releases of webpack, i.e. v2 and v3. While I believe you're correction still stands, you should migrate to a newer version as soon as you can. There are a variety of other things documented on this site that will not work in `1.x.x` versions.

bebraw:
Yup, better go with quotes I think. That said, I don't use the CLI myself but the fix looks valid.

skipjack:
@JonnieCache merged, but please see my comment above as well.

PR #1356, opened 2017-07-01
Correcting output log

vsmahesh:
Output log has been updated to match what is been displayed in the console when followed the steps mentioned.

Results I am seeing when followed the article is:

```
~/Projects/js/webpack-demo
$ webpack
Hash: ea8bef37f3fc13f92afb
Version: webpack 3.0.0
Time: 658ms
    Asset    Size  Chunks                    Chunk Names
bundle.js  544 kB       0  [emitted]  [big]  main
   [0] ./src/index.js 219 bytes {0} [built]
   [2] (webpack)/buildin/global.js 509 bytes {0} [built]
    + 1 hidden module
~/Projects/js/webpack-demo
```


content/guides/getting-started.md
```diff
@@ -246,7 +246,7 @@ Version: webpack 2.2.0
 Time: 390ms
     Asset    Size  Chunks                    Chunk Names
 bundle.js  544 kB       0  [emitted]  [big]  main
-   [0] ./~/lodash/lodash.js 540 kB {0} [built]
+   [0] ./src/index.js 219 bytes {0} [built]
    [1] (webpack)/buildin/global.js 509 bytes {0} [built]
    [2] (webpack)/buildin/module.js 517 bytes {0} [built]
    [3] ./src/index.js 278 bytes {0} [built]
```
skipjack: `./src/index.js` is already mentioned at the bottom of these three examples so either that should be replaced with the `+ 1 hidden module` or we should re-evaluate this change. What version of webpack are you on? I think maybe we should just add a warning at the top of this guide or in `/guides/index.md` to indicate that output might vary between versions of webpack -- trying to always keep these up to date will not be easy.

cc @TheDutchCoder

vsmahesh: I am using webpack 3.0.0. I assumed that the output log will always start with the entry file. It is a logical assumption; isn't it? If that is not the case or it can't be guaranteed, a generic guide line as you have mentioned will do.

skipjack: Yeah, and I think that's ok -- but we'd need to update the whole diff in this case not just the first module (see my comment above).

vsmahesh: See the index of files in output; it is 0, 2, 3! I believe we should go with the generic guiding statement that would cover any such discrepancy. 

```
Hash: ea8bef37f3fc13f92afb
Version: webpack 3.0.0
Time: 658ms
    Asset    Size  Chunks                    Chunk Names
bundle.js  544 kB       0  [emitted]  [big]  main
   [0] ./src/index.js 219 bytes {0} [built]
   [2] (webpack)/buildin/global.js 509 bytes {0} [built]
    + 1 hidden module
```

skipjack: Yeah maybe `[1]` is the hidden module.

> I believe we should go with the generic guiding statement that would cover any such discrepancy.

Sure would you like to add this statement or should I close this PR and open a separate one to update `/guides/index` (and maybe `/guides/getting-started`).

vsmahesh: Please go ahead.

Thank you

PR #1359, opened 2017-07-01
Custom File Systems

### LINKED ISSUE #141: API - Using Preloaded Assets with the Compiler
_From @danschumann on March 11, 2015 14:0_

I have all my files in memory, for various reasons I won't get into ( maybe one is because I wrote a loading module `loaddir` and tend to use it ), but part of it is re-using code on the front end and back end.

My loaded files look like this (`asObject: true`):

```
require('loaddir')({path: './scripts', asObject: true }).then(function(directory) {
  /* directory == {
     'file1': 'console.log("file1");',
     'subDirectory': {
         'file2': 'console.log("subDir/file2");',
     },
  } */
});
```

or, my loaded files look like this (`asObject: false`)

```
require('loaddir')({path: './scripts' }).then(function(directory) {
  /* directory == {
    'file1': 'console.log("file1");',
    'subDirectory/file2': 'console.log("subDir/file2");',
  } */
});
```

How do I give a manifest like this to webpack, telling it "Hey webpack, I already loaded everything, can you just compile using this data plz?"

_Copied from original issue: webpack/webpack#874_

skipjack:
Elaborate on how one passes a custom file system and mention that this is do-able for the `inputFileSystem` as well.

Resolves #141

### DIFF
--- a/content/api/node.md
+++ b/content/api/node.md
@@ -259,9 +259,9 @@ webpack({
 ```
 
 
-## Compiling to Memory
+## Custom File Systems
 
-webpack writes the output to the specified files on disk. If you want webpack to output them to a different kind of file system (memory, webDAV, etc), you can set the `outputFileSystem` option on the compiler:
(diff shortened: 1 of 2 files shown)


skipjack:
@sokra merging, but please let me know if you see anything that needs changing and I'll resolve in a separate PR.

PR #1361, opened 2017-07-03
Replace `use` with `loader`

morajabi:
keyword `use` throws error with webpack 3.

### DIFF
--- a/content/guides/typescript.md
+++ b/content/guides/typescript.md
@@ -57,7 +57,7 @@ module.exports = {
    rules: [
      {
        test: /\.tsx?$/,
-       use: 'ts-loader',
+       loader: 'ts-loader',
        exclude: /node_modules/,
      }
    ]


skipjack:
Hmmm, are you sure you're on the right version of webpack? Can you share a screenshot of your configuration and output? I'm pretty sure we wanted to stick with `use` going forward...

cc @webpack/documentation-team @sokra

skipjack:
That would be a big breaking change and I don't see it mentioned in the [changelog](https://github.com/webpack/webpack/releases). If it was changed this may actually be a bug in the v3 release.

skipjack:
Closing... I tested this out and was not able to replicate. Please share more details if this is still an issue for you.

morajabi:
Look at here https://webpack.js.org/configuration/module/#useentry It's using `loader` for single, non-array loaders. So why it is here?

skipjack:
@morajabi I agree it is a bit confusing, however we are trying to at least be consistent with the recommended usage in the docs (see #569). The basic premise is that for the top level `rule` objects, `use` is recommended. Within nested `UseEntry` objects, e.g. when you need to specify `options`, `loader` is used. I think the old keys like `loader`, which currently still exist for compatibility, will be phased out in upcoming releases.

cc @webpack/core-team -- we've heard about this a few times now in this repo alone... just a heads up as maybe it's a sign we should start phasing these old properties out.

PR #1368, opened 2017-07-05
docs(config): Document how to use built-in Node.js modules

Rob--W:
The documentation about using built-in Node.js modules was very poor (i.e. non-existent). This PR expands the documentation of the "node" configuration option, and shows how one can require built-in modules if desired.

Furthermore, all possible effects of the options are explicitly documented. Instead of being vague of what happens when `false` is used, it is explicitly spelled out what happens.

References:

- https://github.com/webpack/webpack/blob/a589a6c9789a9d342fc630e36ab81827dd20289b/lib/WebpackOptionsApply.js
  shows when the NodeStuffPlugin and NodeSourcePlugin plugins are used.
- https://github.com/webpack/webpack/blob/a589a6c9789a9d342fc630e36ab81827dd20289b/lib/NodeStuffPlugin.js
  is the plugin that is used for every target.
- https://github.com/webpack/webpack/blob/a589a6c9789a9d342fc630e36ab81827dd20289b/lib/node/NodeSourcePlugin.js
  is the plugin that is only used for "browser" and "webworker" targets.
- https://github.com/webpack/webpack/blob/a589a6c9789a9d342fc630e36ab81827dd20289b/lib/MultiModule.js#L65-L67
  is the generated "webpackMissingModule" code for unresolved modules. (when the NodeSourcePlugin is not used, the environment's "require" is used. In Node.js, this also throws a "Cannot find module 'modulename'" error (with single quotes instead of double quotes).


content/configuration/node.md
```diff
@@ -109,9 +120,18 @@ Default: `true`
...
+Polyfills for Node.js core libraries from
+[`node-libs-browser`](https://github.com/webpack/node-libs-browser) are used if
+available, when the `NodeSourcePlugin` plugin is enabled.
+See the list of [Node.js core libraries and their polyfills](https://github.com/webpack/node-libs-browser#readme).
+
+By default, webpack will polyfill each library if there is a known polyfill or do nothing if there is not one.
+In the latter case, webpack will behave as if the module name was configured with the `false` value.
 
-By default, Webpack will polyfill each library if there is a known polyfill or do nothing if there is not one.
+T> To import a built-in module, use [`__non_webpack_require__`](/api/module-variables/#__non_webpack_require__-webpack-specific-),
```
wearhere: > to import a built-in module

Instead of a polyfill, you mean?

content/configuration/node.md
```diff
@@ -109,9 +120,18 @@ Default: `true`
 
 `boolean | "mock" | "empty"`
 
-Many other Node.js core libraries can be configured as well. See the list of [Node.js core libraries and their polyfills](https://github.com/webpack/node-libs-browser).
+W> This option is only activated (via `NodeSourcePlugin`) when the target is unspecified, "browser" or "webworker".
```
wearhere: `W>`?

content/configuration/node.md
Rob--W: Also. Importing using the default require instead of a polyfill, mock, or erroring.

content/configuration/node.md
Rob--W: This puts the text in a warning box. `T>` is a tip box, `?>` is a TODO box.

content/configuration/node.md
```diff
@@ -5,9 +5,16 @@ contributors:
...
 ---
 
-These options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html) and modules. This allows code originally written for the Node.js environment to run in other environments like the browser. This feature is provided by webpack's internal [`NodeStuffPlugin`](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js).
+These options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html) and modules.
+This allows code originally written for the Node.js environment to run in other environments like the browser.
+This feature is provided by webpack's internal
+[`NodeStuffPlugin`](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js) plugin.
+If the target is "web" (default) or "webworker", the
+[`NodeSourcePlugin`](https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js)
+plugin is also activated.
```
skipjack: Can you leave the formatting as is? We may automatically enforce a max line length at some point (webpack-contrib/webpack-defaults#73) but for now we typically keep paragraphs all on one line and encourage contributors to word-wrap while editing these files. That said, it probably makes sense to split it into 2 paragraphs as such:

``` md
These options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html) and modules. This allows code originally written for the Node.js environment to run in other environments like the browser. 

This feature is provided by webpack's internal [`NodeStuffPlugin`](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js) plugin. If the target is "web" (default) or "webworker", the [`NodeSourcePlugin`](https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js) plugin is also activated.
```

content/configuration/node.md
```diff
@@ -19,7 +26,9 @@ This is an object where each property is the name of a Node global or module and
 - `true`: Provide a polyfill.
 - `"mock"`: Provide a mock that implements the expected interface but has little or no functionality.
 - `"empty"`: Provide an empty object.
-- `false`: Provide nothing. Code that expects this object to be defined may crash.
+- `false`: Provide nothing. Code that expects this object may crash with a `ReferenceError`.
+  Code that attempts to import the module using `require('modulename')` may
+  trigger a `Cannot find module "modulename"` error.
```
skipjack: I'm not sure this extra description is necessary. I see where you're going with it, but there could be any number of errors that occur so just specifying that it `may crash` should be enough.

content/configuration/node.md
```diff
@@ -109,9 +120,18 @@ Default: `true`
 
 `boolean | "mock" | "empty"`
 
-Many other Node.js core libraries can be configured as well. See the list of [Node.js core libraries and their polyfills](https://github.com/webpack/node-libs-browser).
+W> This option is only activated (via `NodeSourcePlugin`) when the target is unspecified, "web" or "webworker".
+
+Polyfills for Node.js core libraries from
+[`node-libs-browser`](https://github.com/webpack/node-libs-browser) are used if
+available, when the `NodeSourcePlugin` plugin is enabled.
+See the list of [Node.js core libraries and their polyfills](https://github.com/webpack/node-libs-browser#readme).
```
skipjack: Again, the changes here look 👍  but please drop the line breaks as it's all one paragraph.

content/configuration/node.md
```diff
@@ -109,9 +120,18 @@ Default: `true`
...
+
+By default, webpack will polyfill each library if there is a known polyfill or do nothing if there is not one.
+In the latter case, webpack will behave as if the module name was configured with the `false` value.
```
skipjack: Same here.

content/configuration/node.md
```diff
@@ -109,9 +120,18 @@ Default: `true`
...
 
-By default, Webpack will polyfill each library if there is a known polyfill or do nothing if there is not one.
+T> To import a built-in module, use [`__non_webpack_require__`](/api/module-variables/#__non_webpack_require__-webpack-specific-),
+   i.e. `__non_webpack_require__('modulename')` instead of `require('modulename')`.
```
skipjack: Same here -- i.e. the whole `T>` should all fall on line `133`.

content/configuration/node.md
Rob--W: Done. Note that I did not just split the paragraph, I also added "plugin" after "NodeStuffPlugin", so that it has become "NodeStuffPlugin plugin".

content/configuration/node.md
Rob--W: I added explicit error messages so that search engines can find the documentation more easily.

content/configuration/node.md
Rob--W: I combined the first three lines that formed one sentence in one line. The last sentence is still separate because the message that it conveys is not depending on the previous line.

content/configuration/node.md
Rob--W: Done.

content/configuration/node.md
Rob--W: Done.

PR #1345, opened 2017-06-28
Add name/names requirement for async commons chunk

andy-j-d:
To use the `async` option with the `CommonsChunkPlugin` in the current version of webpack (3.0.0, going back to 2.5.0), a `name` or `names` option which matches entry name/s is also required. This PR changes the code example for async commons chunk to show that.

PR #1370, opened 2017-07-05
Clarify CommonChunksPlugin async and name options

jdbevan:
The aim was to incorporate into the documentation the changes/workaround documented here: https://github.com/webpack/webpack/issues/5109#issuecomment-311154920


skipjack:
I think this overlaps with #1345 slightly.

skipjack:
Also #1428.

Issue #1258, opened 2017-05-31
Guides - Review and Simplify


TheDutchCoder:
I'm currently following the guides as I'm new to webpack and even though the start (setup and first example) is very clear, the guides quickly become vague and (worse) they don't seem to yield working results.

To give you a concrete example: Code Splitting - CSS:
The first example on this page will result in an error, as no input or output is specified. Nor does it contain any information on how to load the result in a page. This might seem arbitrary to experienced users, but I think it's important.

Another one is: Code Splitting - Async:
The `import` example will actually result in webpackJSONp errors and create a `0.bundle.js` file that can not be resolved when loaded in a page. It's also a very unexpected output (why is this file here and what is it supposed to do?)

I believe this can be incredibly discouraging to beginners as they might want to see "fully working" examples so they can grasp what is going on.

Would it be appreciated if I fork this part of the docs and try to improve them a bit? Not sure how contributions are normally done ;)

Since I'm a beginner myself, I might need some help on it, as I'll probably won't be able to get everything working without good docs 😄

TheLarkInn:
I wonder if we should be adding html-webpack-plugin as a easy bootstrap to getting people started, even if they don't understand right away. It will handle some of the nuances like putting async and main chunks in the right order on the html, etc.

TheDutchCoder:
Yeah that might be a good idea. It would certainly help testing out the examples without the frustration of having to figure out how files should be called in a page.

Currently following all guides and proposing some changes (if that's alright) ,so they're consistent and basically build off of the first main example.

rafaeleyng:
> Another one is: Code Splitting - Async:
> The import example will actually result in webpackJSONp errors and create a 0.bundle.js file that > can not be resolved when loaded in a page. It's also a very unexpected output (why is this file here > and what is it supposed to do?)

I will give my experience at this: some months ago, even already being a Webpack user for 1 year and having worked with RequireJS, I tried to setup code splitting with Webpack following the guides, and I was unable to get it working.

I'm planning to check it again the next few days, so I can give a more up-to-date feedback then.

skipjack:
> I'm currently following the guides as I'm new to webpack and even though the start (setup and first example) is very clear, the guides quickly become vague and (worse) they don't seem to yield working results.

[...]

PR #1371, opened 2017-07-05
修改文件名

189:


skipjack:
@189 thanks for this but I think we'd actually prefer to leave that filename as is because it stays consistent with the rest of the guides (which we're currently in the process of reviewing and cleaning up -- #1258). The output library will still be named properly with the configuration used:

image

Also, and this is more of a sidenote, none of the current maintainers speak Chinese so if you know english it would be much appreciated if you could translate titles/descriptions (e.g. "修改文件名" => "Modify the file name").

PR #1373, opened 2017-07-05
Mention that asynchronous chunk loading depends on Promises

jessehattabaugh:


content/guides/code-splitting.md
```diff
@@ -284,7 +286,6 @@ __src/index.js__
   });
 ```
 
-
```
skipjack: Can you leave this new line in -- leaving two before headers helps readability when editing these documents.

content/guides/code-splitting.md
```diff
@@ -165,6 +165,8 @@ Here are some other useful plugins and loaders provide by the community for spli
 
 Two similar techniques are supported by webpack when it comes to dynamic code splitting. The first and more preferable approach is use to the [`import()` syntax](/api/module-methods#import-) that conforms to the [ECMAScript proposal](https://github.com/tc39/proposal-dynamic-import) for dynamic imports. The legacy, webpack-specific approach is to use [`require.ensure`](/api/module-methods#require-ensure). Let's try using the first of these two approaches...
 
+T> Webpack uses Promises to asynchronously load chunks. If you want to support browsers that don't have Promises like Internet Explorer, you'll need to polyfill Promises *before* loading your bundled code.
```
skipjack: This is definitely good to mention -- I might rephrase slightly to:

``` md
T> `import()` calls use [promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). If you want to support older browsers that lack `Promise` support (e.g. Internet Explorer), you'll need to include a `Promise` polyfill _before_ your primary bundles.
```

BTW this is mentioned on the [Module Methods]() page but at least for now I agree it's probably good to mention it here as well.

simon04:
Related PR: #1374

PR #1377, opened 2017-07-06
remove repeated underline in getting-started.md guide 😀

iamkun:
# remove repeated underline in getting-started.md guide

Might be a repeated underline in CLI command.

cheers~


simon04:
The need for repeated two dashes is explained in the same sentence. We cannot invalidate the example for UI issues.

PR #1383, opened 2017-07-07
fix how to be stable module ids e.g

shaodahong:
This should be introduced here how to be stable module ids e.g

### DIFF
--- a/content/plugins/commons-chunk-plugin.md
+++ b/content/plugins/commons-chunk-plugin.md
@@ -111,7 +111,7 @@ new webpack.optimize.CommonsChunkPlugin({
 <script src="app.js" charset="utf-8"></script>
 ```
 
-Hint: In combination with long term caching you may need to use the [`ChunkManifestWebpackPlugin`](https://github.com/diurnalist/chunk-manifest-webpack-plugin) to avoid that the vendor chunk changes. You should also use records to ensure stable module ids.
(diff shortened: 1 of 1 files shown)


content/plugins/commons-chunk-plugin.md
```diff
@@ -111,7 +111,7 @@ new webpack.optimize.CommonsChunkPlugin({
 <script src="app.js" charset="utf-8"></script>
 ```
 
-Hint: In combination with long term caching you may need to use the [`ChunkManifestWebpackPlugin`](https://github.com/diurnalist/chunk-manifest-webpack-plugin) to avoid that the vendor chunk changes. You should also use records to ensure stable module ids.
+Hint: In combination with long term caching you may need to use the [`ChunkManifestWebpackPlugin`](https://github.com/diurnalist/chunk-manifest-webpack-plugin) to avoid that the vendor chunk changes. You should also use records to ensure stable module ids. e.g. `NamedModulesPlugin` or `HashedModuleIdsPlugin`
```
simon04: You should also use records to ensure stable module ids. e.g. `NamedModulesPlugin` or `HashedModuleIdsPlugin`
→
You should also use records to ensure stable module ids, e.g., using `NamedModulesPlugin` or `HashedModuleIdsPlugin`.

shaodahong: Right, and can be combined with [this](https://github.com/webpack/webpack.js.org/issues/1384). What do you think?

PR #1399, opened 2017-07-08
update: resolving src path in entry point

ali-master:


skipjack:
@ali-master does this change fix something for you? Maybe broken file paths on windows or something? I think we we're trying to keep it simple on purpose.

content/guides/getting-started.md
```diff
@@ -189,7 +189,7 @@ __webpack.config.js__
 var path = require('path');
 
 module.exports = {
-  entry: './src/index.js',
+  entry: path.resolve(__dirname, 'src', 'index.js'),
```
skipjack: Another option here is to simply use `context`:

``` js
module.exports = {
  context: path.resolve(__dirname, 'src'),
  entry: 'index.js',
  // ...
```

ali-master:
Hi @skipjack,

Ok, I agree with you for try to keep it simple on purpose.

PR #1402, opened 2017-07-09
docs(guides): add diff part and src/index.html codes

Hangeer:
docs(guides): add diff part and src/index.html codes

### DIFF
--- a/content/guides/output-management.md
+++ b/content/guides/output-management.md
@@ -1,9 +1,10 @@
 ---
 title: Output Management
-sort: 4
+sort: 5
 contributors:
   - skipjack
   - TheDutchCoder
+  - Hangeer
 ---
 
 >T This guide extends on code examples found in the [`Asset Management`](/guides/asset-management) guide.
@@ -199,7 +200,7 @@ __webpack.config.js__
 ``` diff
   const path = require('path');
   const HtmlWebpackPlugin = require('html-webpack-plugin');
-  const CleanWebpackPlugin = require('clean-webpack-plugin');
++ const CleanWebpackPlugin = require('clean-webpack-plugin');
 
   module.exports = {
     entry: {
@@ -210,8 +211,8 @@ __webpack.config.js__
 +     new CleanWebpackPlugin(['dist']),
       new HtmlWebpackPlugin({
         title: 'Output Management',
-        filename: 'index.html',
-        template: 'src/index.html'
++       filename: 'index.html',
++       template: 'src/index.html'
       })
     ],
     output: {
@@ -219,6 +220,20 @@ __webpack.config.js__
       path: path.resolve(__dirname, 'dist')
     }
   };
+```
+
+__src/index.html__
+
+``` html
+<html>
+  <head>
+    <title>Output Management</title>
+    <script src="./print.bundle.js"></script>
+  </head>
+  <body>
+    <script src="./app.bundle.js"></script>
+  </body>
+</html>
 ```
 
 Now run an `npm run build` and inspect the `/dist` folder. If everything went well you should now only see the files generated from the build and no more old files!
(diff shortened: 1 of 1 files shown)


skipjack:
@TheDutchCoder if you decide to merge this one, let's close #1401 as it adds just one part of the changes added here. I approved that one as it was a pretty simple change but would rather let you have a chance to dig into this one seeing as there's more going on.

content/guides/output-management.md
```diff
@@ -1,9 +1,10 @@
 ---
 title: Output Management
-sort: 4
```
TheDutchCoder: What's the reason for moving this guide down in the list?

content/guides/output-management.md
```diff
@@ -199,7 +200,7 @@ __webpack.config.js__
 ``` diff
   const path = require('path');
   const HtmlWebpackPlugin = require('html-webpack-plugin');
-  const CleanWebpackPlugin = require('clean-webpack-plugin');
++ const CleanWebpackPlugin = require('clean-webpack-plugin');
```
TheDutchCoder: 👍

content/guides/output-management.md
```diff
@@ -210,8 +211,8 @@ __webpack.config.js__
 +     new CleanWebpackPlugin(['dist']),
       new HtmlWebpackPlugin({
         title: 'Output Management',
```
TheDutchCoder: Technically this should be a diff line too, as a comma is added to the end.
No biggie, just for future ref.

content/guides/output-management.md
```diff
@@ -221,6 +222,20 @@ __webpack.config.js__
   };
 ```
 
+__src/index.html__
```
TheDutchCoder: I'm okay with this. It touches on using templates, which we removed, but I think it's a good change for this guide.

👍

content/guides/output-management.md
Hangeer: Sorry for that, I read the guide so I followed, may I just remove these lines and commit?

TheDutchCoder: Yeah go for it!

Hangeer: I changed my commit in a patch, is that ok?

content/guides/output-management.md
TheDutchCoder: I'm just wondering if this is actually used, because of the `HtmlWebpackPlugin`.

Let me run a build with it and see if it's used or not.

TheDutchCoder: Yeah this doesn't work correctly, as the `HtmlWebpackPlugin` dynamically adds the assets, but they're also hard-coded in the index.html file.

We can do two things:

1. Drop this altogether, just not use a template
2. Use a template, but make it dynamic.

We removed the 2nd option because the guide was getting quite long and people can read up on how to use the `HtmlWebpackPlugin` templates in the respective docs.

I would recommend removing it for now, is that okay?

Hangeer: Yeah, but if this is not used, update this guide and add some explanation for it will be good too.

TheDutchCoder: It's referenced here: "You can also take a look at html-webpack-template which provides a couple of extra features in addition to the default template." 😉

Hangeer: All right, you've done a good job bro.

TheDutchCoder: Just go ahead and take this piece out then and commit it, then I can approve this and merge it in ;)

Thanks for helping!

TheDutchCoder:
Don't forget to adjust this diff as well: 

```
+       filename: 'index.html',
+       template: 'src/index.html'
```

😉

Hangeer:
Sorry for that, do you mean just remove src/index.html
or keep that, add some comments.
I didn't get that so well.

content/guides/output-management.md
```diff
@@ -210,8 +210,8 @@ __webpack.config.js__
 +     new CleanWebpackPlugin(['dist']),
       new HtmlWebpackPlugin({
         title: 'Output Management',
-        filename: 'index.html',
-        template: 'src/index.html'
++       filename: 'index.html',
```
TheDutchCoder: Just remove those lines, they're not used in the diff before and they're not related to the clean plugin.

PR #1403, opened 2017-07-09
docs(guides): Typo in Output-Management

wendelb:
Hi there,

there was a typo in one of the diffs in the output-management.md guide. This PR fixes that.

Regards.

PS: CLA is signed


skipjack:
@wendelb thanks for your contribution -- I think the same change will go in as a part of #1402 so let's close in favor of that one.

PR #1410, opened 2017-07-11
docs(module): update Rule.enforce typo

hisland:
docs(module): update Rule.enforce typo
according to my test, it might be like this

### DIFF
--- a/content/configuration/module.md
+++ b/content/configuration/module.md
@@ -91,13 +91,13 @@ Specifies the category of the loader. No value means normal loader.
 
 There is also an additional category "inlined loader" which are loaders applied inline of the import/require.
 
-All loaders are sorted in the order `post, inline, normal, pre` and used in this order.
+All loaders are sorted in the order `pre, normal, inline, post` and used in this order.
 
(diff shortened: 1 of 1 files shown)


skipjack:
This change looks valid to me.. I'm pinging the rest of the team though just to be sure.

cc @webpack/documentation-team

hisland:
my test is here: https://github.com/hisland/my-learn/tree/master/webpack2/write-loader/07.prefixing-omit-some

simon04:
I tried to verify this change with the [source code of `NormalModuleFactory`](https://github.com/webpack/webpack/blob/cd688f8/lib/NormalModuleFactory.js#L185-L202) and I'm confused due to [`asyncLib.parallel`](https://caolan.github.io/async/docs.html#parallel) there.

skipjack:
@sokra can you clarify if this change is valid?

sokra:
It's not valid. The original was correct.

Regarding the order: It's expected that the log statements are emitted in reverse order, because the normal phase of loaders are executed in right to left order.

PR #1411, opened 2017-07-11
doesn't remove lodash import given it's still used in the example

olistik:


TheDutchCoder:
Good catch, but maybe I'm missing something, but I don't see the change to lodash?

If you could add the import in, then we should be good I think.

olistik:
@TheDutchCoder you're right. A Diff's Diff is not a joy to handle :-D
Let me know if the import has been correctly restored by my latest commit.

TheDutchCoder:
Fixed the whitespace, rest looks good ;)

Now hoping Travis works (there's been an issue with React components linting it seems)

PR #1412, opened 2017-07-11
docs(guides): Add minimum Node version

TheDutchCoder:
Adds a warning to use Node 6+ to run the site locally (`Running the Site`), as otherwise `npm start` will fail.

### DIFF
--- a/content/writers-guide.md
+++ b/content/writers-guide.md
@@ -48,7 +48,7 @@ Note that `related` will generate a __Further Reading__ section at the bottom of
 
 ## Running the Site
 
-W> Node 6 is required to run the site locally, make sure you have the right version by running `node -v` in the command line.
+W> Node 6 (or higher) is required to run the site locally, make sure you have the right version by running `node -v` in the command line.
 
 1. `npm install`
 2. `npm start`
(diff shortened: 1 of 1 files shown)


simon04:
We have some duplication on how to run the project:
* https://github.com/webpack/webpack.js.org#contributing
* https://github.com/webpack/webpack.js.org/blob/master/.github/CONTRIBUTING.md#setup
* https://github.com/webpack/webpack.js.org/blob/master/content/writers-guide.md#running-the-site

Can we reduce the number of occurrences? To my understanding, writers-guide.md should focus on the content and the technical part on how to build/dev-start the page should be somewhere else.

TheDutchCoder:
@simon04 I think that's a good idea, we could reference it from this doc anyway (just a link).

Do you want me to take this section out instead?

simon04:
@TheDutchCoder I'd remove the "Running the Site" section from writers-guide.md, but I'm open to differing suggestions as well. 😄 

Can you mention the required Node version in https://github.com/webpack/webpack.js.org/blob/master/.github/CONTRIBUTING.md?

TheDutchCoder:
@simon04 Alright I removed it :)

It seems the Node version is already mentioned in Contributing, so I think we're good there?

PR #1413, opened 2017-07-11
Fix Build Issues

skipjack:
fix(dropdown): convert all tabs to spaces

### DIFF
--- a/components/dropdown/dropdown.jsx
+++ b/components/dropdown/dropdown.jsx
@@ -2,46 +2,46 @@ import React from 'react';
 import LanguageIcon from '../../assets/language-icon.svg';
 
 export default class Dropdown extends React.Component {
-	state = {
-		active: false
+  state = {
+    active: false
   };
 
-	render() {
-		let { className = '', items = [] } = this.props;
-		let activeMod = this.state.active ? "dropdown__list--active" : "";
+  render() {
+    let { className = '', items = [] } = this.props;
+    let activeMod = this.state.active ? "dropdown__list--active" : "";
 
-		return (
-			<div
-				tabIndex="0"
-				className={ `dropdown ${className}` }
-				onMouseOver={ this._toggle.bind(this, true) }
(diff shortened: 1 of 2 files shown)


skipjack:
@Munter this is a bit bizarre -- I'm getting all 200s when I go to the actual site or `curl` it but for some reason on Travis it gets all 502s... I've rerun the build a few times already all with the same result.

Munter:
```
$ curl https://jspm.io -I
HTTP/1.1 502 Bad Gateway
Date: Wed, 12 Jul 2017 07:46:38 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Set-Cookie: __cfduid=d65c6b17eead63406cc24a080503b3b8d1499845598; expires=Thu, 12-Jul-18 07:46:38 GMT; path=/; domain=.jspm.io; HttpOnly
Expires: Thu, 01 Jan 1970 00:00:01 GMT
Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
Pragma: no-cache
X-Frame-Options: SAMEORIGIN
Server: cloudflare-nginx
CF-RAY: 37d263cbf89e3cef-CPH
```

This seems to happen on `HTTP HEAD` only. I have special casing for retrying with the slower and more resource intensive `HTTP GET` here: https://github.com/Munter/hyperlink/blob/master/lib/index.js#L172-L175

It seems another case needs to be added here

skipjack:
@Munter ah man that's a pain... what would you even flag on from that response? I may just rethink this link for now as it's not really crucial to our docs -- keep us posted though if you find a good solution and re-release `hyperlink`.

Munter:
It's an easy fix. I already have a retry mechanism for when a server hasn't implemented `HTTP HEAD`. It just doesn't activate on `5xx` responses. I'll get this out pretty soon.

Munter:
I released a new major version that includes this fix and also comes with a new `canonicalroot` option that should be helpful if you ever need canonical urls in your build output and want to test things locally

PR #1417, opened 2017-07-12
in next section, lodash shouldn't be remove.

yoyoys:


content/guides/asset-management.md
```diff
@@ -591,7 +591,7 @@ __webpack.config.js__
 __src/index.js__
 
 ``` diff
-- import _ from 'lodash';
+import _ from 'lodash';
```
simon04: Good catch! Please replace `- ` with `  ` to keep it consistent with other unchanged lines in the file.

simon04:
Ah, I've just found that we already have a (slightly older) PR on the same issue: #1411.

We're happy to review/merge further contributions 😄

PR #1419, opened 2017-07-12
Fixed build error due to whitespace

siddharthkp:


TheDutchCoder:
Is already being worked on in [#1413](https://github.com/webpack/webpack.js.org/pull/1413)

PR #1425, opened 2017-07-15
Update migration guide title to reflect that it applies to both v2 and v3

eteeselink:


skipjack:
I'm actually not sure it does apply to v3 -- I was about to submit a PR that makes this clear with an intro paragraph. I may change the title as well.

skipjack:
@eteeselink thanks for your contribution, I'm going to close in favor of #1426 which should be merged soon. Feel free to continue the discussion here, or on that PR though.

eteeselink:
@skipjack hmm are you sure? I just upgraded from v1 straight to v3 by doing the steps in this guide. Shouldn't v2 to v3 be a noop for virtually all cases? Anyway great work and feel free to ignore this PR indeed! :-)

skipjack:
Yeah, the migration from v2 to v3 is pretty painless which is why you probably didn't notice any issues after following the steps for v2. I did a quick scan and definitely didn't see any mention of v3 in the guide (aside from some "this may be deprecated in v3" statements). 

If someone wanted to update the guide to include v2 to v3 and then rename it _Migrating Versions_ or something like that I wouldn't be opposed to it. Not sure that's necessary though.

PR #1434, opened 2017-07-16
Docs - remove extra grid 

styopdev:
Check this page https://webpack.js.org/plugins/


skipjack:
@styopdev thanks for your contribution! This was actually fixed in #1429 which I just merged so let's close this one out. Should be deployed within the next few hours.

PR #1435, opened 2017-07-16
Update to Hyperlink 3.x and use --canonicalroot

Munter:
Fixes issue with jspm.io responding 502 to hyperlinks checks.
Also adds usage of the new `--canonicalroot` option, which allows hyperlink to resolve links to canonical resources as though they were local

Note that hyperlink no longer checks source maps in this configuration. If you are deploying source maps and want to verify their correct linking, add `--source-maps` to the hyperlink options

@skipjack


skipjack:
@Munter awesome, thank you!

PR #1436, opened 2017-07-17
Rework Caching Guide

### LINKED ISSUE #652: Guides - Explain Hash Changes in Caching Guide

skipjack:
Trying to combine the current [caching guide][1] with @timse's [post][2] and make it a bit easier to read with clear, fully-fleshed out examples. I'll be submitting a PR to @TheDutchCoder's [examples repo][3] soon as well for the small test project I created in relation to this.

Resolves #652 (hopefully -- and hopefully a bunch of the related open issues on the main repo)
Part of #1258 (will update that now)

[1]: https://webpack.js.org/guides/caching/
[2]: https://medium.com/webpack/predictable-long-term-caching-with-webpack-d3eee1d3fa31
[3]: https://github.com/TheDutchCoder/webpack-guides-code-examples


skipjack:
@TheDutchCoder this should be ready for an initial review.

@timse I used your post in combination with the content that @okonet and @jouni-kantola already had here. Using the `HashedModuleIdsPlugin` allowed me to avoid (I think) certain steps in the post you wrote. I still need to test using `import()`, `externals` and a few other things, so another section or two may be warranted. It seems you have a lot of experience on the subject, so if you could help review that would be great.

@okonet @jouni-kantola I can't assign you, but if either of you can review that would be much appreciated.

Everyone should note that I also have a PR open at TheDutchCoder/webpack-guides-code-examples#17 that directly reflects the final version of the files/code in the guide. This should be a good testing ground for us to figure out exactly what does and doesn't mess up hashes. I want to keep things fairly straightforward so the guide doesn't go on forever, so maybe leaving some rare edge cases out is ok, but we should definitely try to cover all common use cases.

skipjack:
I think once we get this to a good state, we should locate all related issues on the main repo (here I think it's only #652), see if they are fully or almost fully answered by the guide updates and close them with a note saying that people should read through the guide and, if any issues arise create a targeted issue with what needs to change.

As per @jouni-kantola's point in #652, it would be great if we could get to one source of truth for this documentation.

content/guides/caching.md
```diff
@@ -22,7 +22,7 @@ related:
 
 T> This examples in this guide stem from [getting started](/guides/getting-started), [output management](/guides/output-management) and [code splitting](/guides/code-splitting).
```
TheDutchCoder: S/This examples/The examples

content/guides/caching.md
```diff
@@ -22,7 +22,7 @@ related:
...
 
-So we're using webpack to bundle our modular application, deploying our `/dist` directory to the server, and clients, typically browsers, are hitting that server to grab the site and its assets. The last step can be time consuming, which is why browsers use a technique called [caching](). This allows sites to load faster with less unnecessary network traffic, however it can also cause headaches when you need new code to be picked up.
+So we're using webpack to bundle our modular application, deploying our `/dist` directory to the server, and clients, typically browsers, are hitting that server to grab the site and its assets. The last step can be time consuming, which is why browsers use a technique called [caching](https://en.wikipedia.org/wiki/Cache_(computing)). This allows sites to load faster with less unnecessary network traffic, however it can also cause headaches when you need new code to be picked up.
```
TheDutchCoder: Would rephrase this "...directory to the server, and clients, typically browsers, are hitting that server..." so it flows a little better.

content/guides/caching.md
```diff
@@ -241,13 +241,13 @@ runtime.1400d5af64fc1b7b3a45.js    5.85 kB       2  [emitted]         runtime
     + 1 hidden module
 ```
 
-... we can see that all three have. This is because the [`module.id`]() of each is incremental by default, so...
+... we can see that all three have. This is because the [`module.id`](/api/module-variables#module-id-commonjs-) of each is incremental by default, so...
 
 - The `main` bundle changed because of it's new content.
```
TheDutchCoder: S/it's/its

content/guides/caching.md
```diff
@@ -241,13 +241,13 @@ runtime.1400d5af64fc1b7b3a45.js    5.85 kB       2  [emitted]         runtime
...
 - The `vendor` bundle changed because it's `module.id` was changed.
```
TheDutchCoder: S/it's/its

content/guides/caching.md
```diff
@@ -241,13 +241,13 @@ runtime.1400d5af64fc1b7b3a45.js    5.85 kB       2  [emitted]         runtime
...
 - And, the `runtime` bundle changed because it now contains a reference to a new module.
 
-The first and last are expected -- it's the `vendor` hash we want to fix. Luckily, there are two plugins that can help us out with this dilemma. First, the [`NamedModulesPlugin`]() which the path to the module rather than a numerical ID. While this plugin is useful during development for easier to read output, it does take a bit longer to run. The second option is the `HashedModuleIdsPlugin`, which is what we'll use as these examples are more targeted toward production builds:
+The first and last are expected -- it's the `vendor` hash we want to fix. Luckily, there are two plugins that can help us out with this dilemma. First, the [`NamedModulesPlugin`](/plugins/named-modules-plugin) which the path to the module rather than a numerical ID. While this plugin is useful during development for easier to read output, it does take a bit longer to run. The second option is the `HashedModuleIdsPlugin`, which is what we'll use as these examples are more targeted toward production builds:
```
TheDutchCoder: "...which the path..." seems like a word is missing here.

TheDutchCoder: Link to `HashedModuleIdsPlugin`?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
-For an even more optimized setup:
+This guide focuses on the configuration changes needed to ensure that your `output` files are cached when appropriate, but re-requested when changed.
 
-1. Use compiler stats to get the file names when requiring resources in HTML.
-2. Generate the chunk manifest JSON and inline it into the HTML page before loading resources.
 
+## Output Filenames
 
-## The problem
+A simple way to ensure the browser picks up changed files is by using `output.filename` [substitutions](/configuration/output#output-filename). The `[hash]` substitution can be used to include a build-specific hash in the filename, however it's even better to use the `[chunkhash]` subsitution which include a bundle-specific hash in the filename.
```
TheDutchCoder: S/include/includes

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
++ import Print from './print';
 
-To fix that, we should use [`ChunkManifestWebpackPlugin`](https://github.com/diurnalist/chunk-manifest-webpack-plugin), which will extract the manifest to a separate JSON file. This replaces the chunk manifest with a variable in the webpack runtime. But we can do even better; we can extract the runtime into a separate entry by using `CommonsChunkPlugin`. Here is an updated `webpack.config.js` which will produce the manifest and runtime files in our build directory:
+  function component() {
+    var element = document.createElement('div');
 
-```js
-// webpack.config.js
-var ChunkManifestPlugin = require("chunk-manifest-webpack-plugin");
+    // Lodash, currently included via a script, is required for this line to work
```
TheDutchCoder: Remove this comment (I think)

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
+  import _ from 'lodash';
+- import Print from './print';
++ // import Print from './print';
 
-## Manifest inlining
+  function component() {
+    var element = document.createElement('div');
 
-Inlining the chunk manifest and webpack runtime (to prevent extra HTTP requests), depends on your server setup. There is a nice [walkthrough for Rails-based projects](https://brigade.engineering/setting-up-webpack-with-rails-c62aea149679). For server-side rendering in Node.js you can use [webpack-isomorphic-tools](https://github.com/halt-hammerzeit/webpack-isomorphic-tools).
+    // Lodash, currently included via a script, is required for this line to work
```
TheDutchCoder: Same here.

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
 
-## References
+## Conclusion
 
-* https://medium.com/@okonetchnikov/long-term-caching-of-static-assets-with-webpack-1ecb139adb95#.vtwnssps4
-* https://gist.github.com/sokra/ff1b0290282bfa2c037bdb6dcca1a7aa
-* https://github.com/webpack/webpack/issues/1315
-* https://github.com/webpack/webpack.js.org/issues/652
-* https://presentations.survivejs.com/advanced-webpack/
+Caching gets messy. Plain and simple. However the walk-through above should give you a running start to deploying consistent, cache-able assets. See the _Further Reading_ section below to learn more.
```
TheDutchCoder: S/cache-able/cachable

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
+  - title: Advanced Webpack Presentation
+    url: https://presentations.survivejs.com/advanced-webpack/
+  - title: Issue 1315
+    url: https://github.com/webpack/webpack/issues/1315
+  - title: Issue 652
+    url: https://github.com/webpack/webpack.js.org/issues/652
 ---
 
-To enable long-term caching of static resources produced by webpack:
+T> This examples in this guide stem from [getting started](/guides/getting-started), [output management](/guides/output-management) and [code splitting](/guides/code-splitting).
```
okonet: `This` and `examples` don't fit together grammatically

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
 
-1. Use `[chunkhash]` to add a content-dependent cache-buster to each file.
-2. Extract the webpack manifest into a separate file.
-3. Ensure that the entry point chunk containing the bootstrapping code doesn’t change hash over time for the same set of dependencies.
+So we're using webpack to bundle our modular application, deploying our `/dist` directory to the server, and clients, typically browsers, are hitting that server to grab the site and its assets. The last step can be time consuming, which is why browsers use a technique called [caching](https://en.wikipedia.org/wiki/Cache_(computing)). This allows sites to load faster with less unnecessary network traffic, however it can also cause headaches when you need new code to be picked up.
```
okonet: I'm not sure this guide needs this intro. It is good for a blog post but I'd expect this guide to get directly to the point. Probably replace it with a link to [my post](https://medium.com/@okonetchnikov/long-term-caching-of-static-assets-with-webpack-1ecb139adb95) and I could cross-link to this guide in the beginning to guide the traffic here?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
 
-For an even more optimized setup:
+This guide focuses on the configuration changes needed to ensure that your `output` files are cached when appropriate, but re-requested when changed.
```
okonet: `This guide focuses on the configuration needed to ensure files produced by webpack compilation can remain cached unless their contents has changed.` ?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
-But the problem here is, builds after *any file update* will update all filenames and clients will have to re-download all application code. So how can we guarantee that clients always get the latest versions of assets without re-downloading all of them?
+?> Add bash output
 
+Now our bundle's name should change to reflect its content, meaning if `src/index.js` is changed, the filename should change. However, if we run another build, we'll see that this is not the case:
 
-## Generating unique hashes for each file
+?> Add bash output
 
-What if we could produce the same filename, if the contents of the file did not change between builds? For example, it would be unnecessary to re-download a vendor file, when no dependencies have been updated, only application code.
+This is because webpack includes certain boilerplate, specifically the runtime and manifest, in the entry chunk.
```
okonet: In the previous sentence you're saying the file contents should have been changed but I think you meant to say "should not change"?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
-module.exports = {
-  /*...*/
-  output: {
-    /*...*/
--   filename: "[name].[hash].js"
-+   filename: "[name].[chunkhash].js"
-  }
-};
-```
+## Extracting Boilerplate
```
okonet: `Extracting webpack boilerplate`?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
+    ],
+    output: {
+      filename: '[name].[contenthash].js',
+      path: path.resolve(__dirname, 'dist')
+    }
+  };
 ```
 
-Although, each time we build for production, we’ll get different file names. Something, that looks like this:
+Now let's run a build to see if it worked:
```
okonet: By reading this sentence it's not clear if it worked or not and you're not clarifying it after the code block. I'd rephrase by making it less ambiguous.

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
-module.exports = {
-  /*...*/
-  plugins: [
-    function() {
-      this.plugin("done", function(stats) {
-        require("fs").writeFileSync(
-          path.join(__dirname, "build", "stats.json"),
-          JSON.stringify(stats.toJson()));
-      });
+Another thing we may want to extract is our core third-party libraries, in our case `lodash`, as they are less likely to change than our source code. This step will allow clients to request even less from the server to stay up to date. Let's do this using a combination of a new `entry` point along with another `CommonsChunkPlugin` instance:
```
okonet: Nitpicking but `thing` and `libraries` don't match to me. What about

> It's a good practice to extract third-party libraries, such as `lodash` or `react`, to a separate `vendor` chunk as they are less likely to change than our source code.

okonet: I'd replace `Let's do this ` with `This can be done by`

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
+      filename: '[name].[contenthash].js',
+      path: path.resolve(__dirname, 'dist')
     }
-  ]
-};
+  };
 ```
 
-Alternatively, just use one of these plugins to export JSON files:
+W> Note that order matters here. The `'vendor'` instance must be included prior to the `'runtime'` instance.
```
okonet: `instance`? Did you mean `chunk`?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
-    new webpack.HashedModuleIdsPlugin(),
-    new WebpackChunkHash(),
-    new ChunkManifestPlugin({
-      filename: "chunk-manifest.json",
-      manifestVariable: "webpackManifest",
-      inlineManifest: true
-    })
-  ]
-};
+... we can see that all three have. This is because the [`module.id`](/api/module-variables#module-id-commonjs-) of each is incremental by default, so...
```
okonet: `is incremental by default` => `is not consistent by default`?

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
+
+- The `main` bundle changed because of it's new content.
```
okonet: Since this list is a part of the sentence, I'd make `The` lowercase and remove `.` in the end of each but last line

content/guides/caching.md
```diff
@@ -4,284 +4,358 @@ sort: 11
...
+- The `vendor` bundle changed because it's `module.id` was changed.
+- And, the `runtime` bundle changed because it now contains a reference to a new module.
+
+The first and last are expected -- it's the `vendor` hash we want to fix. Luckily, there are two plugins that can help us out with this dilemma. First, the [`NamedModulesPlugin`](/plugins/named-modules-plugin) which the path to the module rather than a numerical ID. While this plugin is useful during development for easier to read output, it does take a bit longer to run. The second option is the [`HashedModuleIdsPlugin`](/plugins/hashed-module-ids-plugin), which is what we'll use as these examples are more targeted toward production builds:
```
okonet: I'd remove `dilemma`

okonet: `which the path to the module` is missing the verb

okonet: `ID` => `identifier`?

okonet: `for easier to read output` => `for simpler output` or `for the output that is easier to read`?

content/guides/caching.md
skipjack: 👍

content/guides/caching.md
skipjack: 👍

content/guides/caching.md
skipjack: 👍

content/guides/caching.md
skipjack: Yeah, I guess it's ok to append `able` to words even if you it's not an actual variation of that word. Using the `-` prevents spell checkers from throwing an error but I guess it does read better without it.

content/guides/caching.md
skipjack: Hmmm I see your point but I think it is helpful to give a little context before jumping into things. I did use the link so as not to explain caching in too much detail -- also updated per the last review to make it flow a little better. Maybe we can still tighten it up a bit more? @TheDutchCoder what do you think?

content/guides/caching.md
skipjack: 👍

content/guides/caching.md
skipjack: 👍

content/guides/caching.md
skipjack: 👍

content/guides/caching.md
skipjack: Makes sense on both counts -- will update.

content/guides/caching.md
skipjack: I meant `CommonsChunkPlugin` instance but maybe I should specify that.

content/guides/caching.md
skipjack: No I did mean "incremental" in the sense that modules ids are simply incremented up in the order they are resolved. However I do see how this could be a bit confusing... I'll rephrase.

content/guides/caching.md
skipjack: I see what you mean but the way I updated the sentence above makes this change unnecessary (I think).

content/guides/caching.md
skipjack: Idk I think I'd rather leave this title as is for now. Partly because the mix of lowercase and uppercase (I know this is defined in our style guide/media repo) feels a bit odd but also I think it's pretty clear based on context.

skipjack:
@okonet @TheDutchCoder awesome feedback, thank you both! Just pushed the updates...

What did you think of the article in general? My feeling is that it's a solid base that's now in line with our guides, but we may have to add a few of more of the oddities that can break caching, like those discussed in the various other guides we're linking to.

content/guides/caching.md
jouni-kantola: Personally I'd prefer `Extract webpack runtime`. I interpret `boilerplate` like something that isn't variable. Also, later in the PR the asset is named `runtime`.

content/guides/caching.md
```diff
@@ -4,284 +4,356 @@ sort: 11
...
-Hash: cfba4af36e2b11ef15db
-Version: webpack 2.2.0
-Time: 66ms
-                         Asset     Size  Chunks             Chunk Names
-vendor.50cfb8f89ce2262e5325.js  2.58 kB       0  [emitted]  vendor
-  main.70b594fe8b07bcedaa98.js  2.57 kB       1  [emitted]  main
-   [0] ./src/index.js 63 bytes {1} [built]
-   [1] ./src/vendor.js 63 bytes {0} [built]
-```
+As we learned in [code splitting](/guides/code-splitting), the [`CommonsChunkPlugin`](/plugins/commons-chunk-plugin) can be used to split modules out into separate bundles. A lesser-known feature of the `CommonsChunkPlugin` is extracting webpack's boilerplate and manifest which can change with every build. By specifying a name not mentioned in the `entry` configuration, the plugin will automatically extract what we want into a separate bundle:
```
jouni-kantola: Isn't the runtime added to the last specified commons chunk? Maybe that is only when specified as array.

content/guides/caching.md
```diff
@@ -4,284 +4,356 @@ sort: 11
...
++       'lodash'
++     ]
++   },
+    plugins: [
+      new CleanWebpackPlugin(['dist']),
+      new HtmlWebpackPlugin({
+        title: 'Caching'
+      }),
++     new webpack.optimize.CommonsChunkPlugin({
++       name: 'vendor'
```
jouni-kantola: Don't we need `minChunks: Infinity` as property, or is it default?

content/guides/caching.md
```diff
@@ -4,284 +4,356 @@ sort: 11
...
+``` diff
+  const path = require('path');
+  const CleanWebpackPlugin = require('clean-webpack-plugin');
+  const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+  module.exports = {
+    entry: './src/index.js',
+    plugins: [
+      new CleanWebpackPlugin(['dist']),
+      new HtmlWebpackPlugin({
```
jouni-kantola: Maybe we should leave out `HtmlWebpackPlugin` from the caching docs?

jouni-kantola:
One thing I'd like to point out is that this PR mostly describes how to setup a proper config, not what to be aware of during development, I.e. adding extra entries. What I've always found hardest is to keep hashes intact during development and refactoring. What seem like small changes can have bigger impact than expected (https://gist.github.com/jouni-kantola/1c1e2bfaebf30de50d1b6a71b869da13).

timse:
Heya :) not sure about `HashedModuleIdsPlugin` might be working some magic that makes things I used unnecessary. I personally like the `NamedModulesPlugin`/`NamedModulesChunks` better for a couple of reasons, that dont necessarily count as a general guideline :)

content/guides/caching.md
skipjack: I actually think the `HtmlWebpackPlugin` is key for this doc. If not, users have to track bundle names themselves using the manifest and I think that complicates the guide significantly. Plus, it's usage is already covered in `output-management` which comes before this guide.

content/guides/caching.md
skipjack: How about `Extract Runtime & Libraries`? I think that pretty much covers the whole section. Could also be `Extract Runtime, Manifest, and Libraries` but the runtime and manifest are being extracted into the same bundle so I think it's ok to just use `Runtime`.

content/guides/caching.md
skipjack: It's extracted to a separate chunk when you use a name that doesn't already exist (i.e. no entry chunk exists by that name). I played around with the array syntax as well but ran into some issues trying to do it all with one instance.

content/guides/caching.md
skipjack: See the example at TheDutchCoder/webpack-guides-code-examples#17... I don't think it's necessary but I also didn't test chunks too much yet (though I do have a bit more code to push that uses `import()` to generate a child chunk).

skipjack:
> One thing I'd like to point out is that this PR mostly describes how to setup a proper config, not what to be aware of during development, I.e. adding extra entries. What I've always found hardest is to keep hashes intact during development and refactoring. What seem like small changes can have bigger impact than expected.

@jouni-kantola I see your point, but I mean ideally we want to provide the configuration changes needed so that unexpected hash changes don't occur, right? Maybe some discussion in the conclusion of what to watch out for would help though?

> Heya :) not sure about `HashedModuleIdsPlugin` might be working some magic that makes things I used unnecessary. I personally like the `NamedModulesPlugin` / `NamedModulesChunks` better for a couple of reasons, that dont necessarily count as a general guideline :)

@timse I used the `HashedModuleIdsPlugin` because I found it simpler when testing (TheDutchCoder/webpack-guides-code-examples#17) and it seemed I didn't need to include any of the `Named***` plugins. That said, I did see what you said in your post about the `Named***` plugins yielding slightly smaller bundles.

> That being said, I would use the named chunks plugin - But probably with a saner fallback function than specified in the guide a wrote :)

I will take a look at that plugin tomorrow. I'm going to merge this for now but I definitely would like to keep the conversation going and do some follow up PRs. Maybe we can all test the example I put together in TheDutchCoder/webpack-guides-code-examples#17 and see if we can poke holes any holes in it? I don't think we have to cover extreme edge cases but if there are some common scenarios that break it we should definitely address them in the guide.

Issue #1388, opened 2017-07-07
Release - Document Version 3.1.0


simon04:
Document changes according to https://github.com/webpack/webpack/releases/tag/v3.1.0

- [x] Allow different library names for UMD
- [x] Support for passing a defined identifier as this in a IIFE
- [x] Use the new resolve performance option cacheWithContext: false by default when it's safe
- [x] Support array of functions as configuration
- [ ] add sortModules to Chunk which is required in extract-text-plugin to support webpack 3

simon04:
> add sortModules to Chunk which is required in extract-text-plugin to support webpack 3

See https://github.com/webpack/webpack/pull/5225. We do not have an API doc for the `Chunk` class. The immediate effect of this is a bugfix. Nothing to be done for this item?

simon04:
> Support for passing a defined identifier as this in a IIFE

The PR https://github.com/webpack/webpack/pull/5076 classifies this a bugfix. So we can skip a documentation update?

skipjack:
@simon04 yep I agree on webpack/webpack#5076 (just looked over that pr) and the other. It seems that they're both just classified as bug fixes. However, I'll keep the missing `Chunk` class doc in mind. The __API__ section definitely needs the most help/attention/review of this whole site. Still working my way through the backlog but I'm hopeful start cleaning that up soon. I think we'll have some other help coming on soon as well.

I'll close but, if you feel there's still something that needs to be addressed, feel free to re-open.

Issue #1318, opened 2017-06-20
Release - Document Version 3.0.0


simon04:
Document changes according to https://github.com/webpack/webpack/releases/tag/v3.0.0

- [x] #1309 
- [x] #1313
- [x] #1314 
- [x] #1315 
- [x] #1316 
- [x] #1317 
- [x] #1320 
- [ ] Scope Hoisting (webpack/webpack#4873)
- [x] node_modules no longer mangle to ~ in stats, https://github.com/webpack/webpack/pull/4144 (if needed?)

halfzebra:
There's also #4979 for `noParse`

simon04:
That's already covered via #1316.

skipjack:
@simon04 awesome! Thank you for tackling this with the rapid fire PRs! Starting to review/merge them now...

skipjack:
Down to two... was scope hoisting tackled by #1309 or is it more of an overarching concept/feature we need to document? I'm happy to help push this forward if there's anything I can do.

skipjack:
> node_modules no longer mangle to ~ in stats, webpack/webpack#4144 (if needed?)

Is the only change just to the CLI output? Judging from the PR it looks like it. If so, I'd say this doesn't really require any changes. If anything we can go through the docs once things have stabilized a bit and updated all webpack bash output examples to the latest version.

simon04:
> was scope hoisting tackled by #1309 or is it more of an overarching concept/feature we need to document?
#1309 addressed the CLI change only. The changes in https://github.com/webpack/webpack/pull/4873 seem to be larger than that, but I haven't investigated in detail yet. This example is relevant: https://github.com/webpack/webpack/tree/master/examples/scope-hoisting

> > node_modules no longer mangle to ~ in stats, webpack/webpack#4144 (if needed?)
> Is the only change just to the CLI output?
I think so. It is considered a "breaking change" and thus listed in the release notes. I'm fine with not documenting this one.

skipjack:
@sokra, or anyone else familiar with this new feature (cc @webpack/documentation-team), could you outline how this changes things for the end user so we can get it documented? I read through [this example](https://github.com/webpack/webpack/tree/master/examples/scope-hoisting) and webpack/webpack#4873, but still don't understand exactly what's changed in terms of usage.

kentcdodds:
[...]

PR #1445, opened 2017-07-18
docs(configuration): Change chunkhash to hash

ematipico:
The `filename` options supports `hash` and not `chunkhash`. 
`chunkhash` is only for `chunkFilename` option.

### DIFF
--- a/content/configuration/index.md
+++ b/content/configuration/index.md
@@ -45,7 +45,7 @@ module.exports = {
 
     <details><summary>[filename](/configuration/output#output-filename): "bundle.js", // string</summary>
     [filename](/configuration/output#output-filename): "[name].js", // for multiple entry points
-    [filename](/configuration/output#output-filename): "[chunkhash].js", // for [long term caching](/guides/caching)
+    [filename](/configuration/output#output-filename): "[hash].js", // for [long term caching](/guides/caching)
     </details>
     // the filename template for entry chunks
 


skipjack:
Thanks for your contribution but `chunkhash` is [available for `filename`](https://webpack.js.org/configuration/output/#output-filename) (see the table) and should be the default based on the practices recommended in the [caching guide](https://webpack.js.org/guides/caching). Could you document why this change is necessary or provide an example using the latest version of webpack where this doesn't work?

skipjack:
Related: #1436

ematipico:
That's weird. Webpack 3.3.0 was complaining about that saying that I needed to use `hash` instead of `chunkhash` for `filename`

skipjack:
Hmmm I just used in TheDutchCoder/webpack-guides-code-examples#17 with webpack `3.3.0` and didn't get any errors. Can you maybe pull that example and see if you are able to reproduce the issue? It may be a bug with webpack 3 but as far as I know `chunkhash` should be available.

skipjack:
Maybe it's some combination of things in your config that's causing that error to be thrown, even if that's the case though it may be a bug or maybe the error isn't detailed enough. Or, maybe a breaking change was introduced that I'm not aware of but this wasn't mentioned in either of the release docs issues ( #1318 or #1388 ) or in the main repo's changelog as far as I know.

cc @webpack/documentation-team

ematipico:
OK probably I narrowed the issue.

My output config

```javascript
    output: {
        path: path.resolve(path.join(__dirname, '/')),
        filename: '[name].[chunkhash].js',
        chunkFilename: '[name].[chunkhash].js',
        publicPath: '/'
    }
```
```bash
[name].[chunkhash].js
Cannot use [chunkhash] for chunk in '[name].[chunkhash].js' (use [hash] instead)
```

Probably it's not possible to use `[chunkhash]` inside `filename` when you have `chunkFilename` defined. I didn't find anything inside the documentation though

skipjack:
I think the maybe that error is actually flagging the use of `[chunkhash]` within `chunkFilename`. Maybe for `chunkFilename` only `[hash]` is supported. I agree that this error is a bit confusing. I'll re-open so we can get some clarification.

cc @bebraw @sokra @TheLarkInn

sokra:
With HMR enabled you can't use `[chunkhash]` for technical reasons.

For dev builds with HMR -> use `[hash]`.

ematipico:
The configuration that I am using doesn't use the HMR. The local web server is launched using another technology. Is it because there's the `--watch` flag?

skipjack:
> Is it because there's the --watch flag?

It seems that's likely the case, e.g. if webpack is in "Watch Mode" -- `[chunkhash]` is forbidden. I'll let @sokra clarify but if this is the case maybe we should just add a warning under the substitutions table like so:

``` md
W> Note that `[chunkhash]` cannot be used in conjunction with "Watch Mode". For example, if you're using `webpack-dev-server` or `webpack-dev-middleware` during development, you will get an error using `[chunkhash]` within your development configuration.
```

ematipico:
I get this error even on production build. I don't run any watcher or HMR thing. 
I run the command 

```bash
 webpack --config webpack.config.prod.js --optimize-minimize --colors --profile --progress
```

The `output` is the same of the one that I pasted before. I tried to reproduce it on another smaller repo but no joy. I guess there's a combination of settings that fires that error...

ematipico:
Found the cause of that error, which is now filed here https://github.com/webpack/webpack/issues/5337

I close the PR

skipjack:
Awesome -- thanks @ematipico for following up on this!

PR #1450, opened 2017-07-21
docs(getting-started): refine module section

simon04:
* bring together related sentences
* add links to further documentation


skipjack:
I have a fix for the build errors in #1436. It was caused by a change to the uglifyjs plugin's readme. I'll try to get that one and a few others, including this, in tomorrow.

skipjack:
There's a minor markdown linting issue we need to fix and then this should be good to go. We should either extend the allowed line length in that rule or break up that paragraph. This is another instance where reference links would be nice as I think those characters get counted.

PR #1439, opened 2017-07-17
HMR guide rewrite

sbaidon:
This PR contains some changes to the `HMR` guide:

-  Update examples to be consistent with `Development` guide

PR #1454, opened 2017-07-26

evolve2k:
Progressing through the tutorial so far, each tute lead on nicely to the next.
This one however breaks the flow and provided no clues as to which lines to add like previous files do.

Basically the convention of the green lines beginning with + signs is not present in this file, breaking the flow of the tutorial at this point.

I've marked up a couple of key ones that would be useful to highlight.
This is helpful as is it is but could probably do with more work also.

### DIFF
--- a/content/guides/hot-module-replacement.md
+++ b/content/guides/hot-module-replacement.md
@@ -11,6 +11,7 @@ contributors:
   - drpicox
   - skipjack
   - gdi2290
+  - evolve2k
 related:
   - title: Concepts - Hot Module Replacement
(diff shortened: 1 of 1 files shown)


TheDutchCoder:
Thanks for your contribution!

We're actually in the process of rewriting this guide, as you can see in this [PR](https://github.com/webpack/webpack.js.org/pull/1439)

I would suggest giving your input on that one if you like.

I'll close this PR in the mean time ;)

evolve2k:
Ah excellent, I was hoping that was the case. Thanks for your ongoing work on this great project!

PR #1456, opened 2017-07-26

sbaidon:
Just noticed the lodash import was missing in a  __index.js__ diff block

### DIFF
--- a/content/guides/hot-module-replacement.md
+++ b/content/guides/hot-module-replacement.md
@@ -246,6 +246,7 @@ body {
 __index.js__
 
 ``` diff
+  import _ from 'lodash';
   import printMe from './print.js';
 + import './styles.css';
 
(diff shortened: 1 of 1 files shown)


content/guides/hot-module-replacement.md
```diff
@@ -213,7 +213,7 @@ __webpack.config.js__
       new webpack.HotModuleReplacementPlugin()
     ],
     output: {
-      filename: 'bundle.js',
+      filename: '[name].build.js',
```
skipjack: I think `[name].bundle.js` is more commonly used than `[name].build.js` but I'm not sure if this is what is mentioned in the other guides. @TheDutchCoder do you know?

Personally I would prefer `[name].bundle.js` but I don't feel too strongly about it.

sbaidon: You are right it should be `[name].bundle.js` that is what's on the other guides, will update

Issue #1463, opened 2017-07-29
Guides - Hot Module Replacement


maheshpec:
Hello,

While going through the [hot module replacement guide](https://webpack.js.org/guides/hot-module-replacement/#gotchas), I noticed a snippet line did not have + sign indicating that its a new statement.

**What is the current behavior?**
[HMR guide page's _Gotcha_ section](https://webpack.js.org/guides/hot-module-replacement/#gotchas) has a code snippet. This code snippet introduces a `let element = component();` statement. Its not present in the _Enabling HMR_ section. However, it does not have a + sign to indicate that its a new statement.

**What is the expected behavior?**
The `let element = component();` statement should be treated as a new statement by having a + in front of it and highlighting it in green.

Thank you

skipjack:
@maheshpec I'm going to close and quote your comment in #1461 as that one mentions this error as well. That way we can keep the discussion in one place. I think we may be able to update your PR with one or two other fixes to resolve these issues for good 👍

PR #1464, opened 2017-07-30
docs(guides): fix HMR Gotcha section snippet (#1463)

### LINKED ISSUE #1461: Guides - Warnings/Errors in Hot Module Replacement
Hello! I am new to Webpack, and I'm trying to learn it by the guide, as I advanced to HMR, I got some errors...

I've written the code showed in that chapter until the Stylesheets part (which I haven't read yet), and I got this console message while using `npm start`:

```
ERROR in chunk print [entry]
bundle.js
Conflict: Multiple assets emit to the same filename bundle.js
```

I'm not sure, but I think this error is about having multiple entries, so I guess it needs to have a output name for each one of it.

Then I noticed that the author of the guide changed the `filename` property of `output` from `[name].bundle.js` to `bundle.js`, so I typed it back to the previous value to see what happens, then I get no error from the console. But also no hot edits work also...

When I edit(and save) the `print.js` file, I get the following browser's console:

image

The `ReferenceError` is understandable(and they need to be solved on the guide), but even getting through them we have the warnings, that tell us the modules can't be hot updated, and that's the focus of this section of the guide...

Does anybody who had read the article about HMR know how to proceed? I really don't have a clue of how to solve and improve it...

maheshpec:
Fix for snippet in the _Gotcha_ section of [_Hot Module Replacement Guide_](https://webpack.js.org/guides/hot-module-replacement/#gotchas).

Resolves #1461

### DIFF
--- a/content/guides/hot-module-replacement.md
+++ b/content/guides/hot-module-replacement.md
@@ -30,6 +30,8 @@ W> __HMR__ is not intended for use in production, meaning it should only be used
 
 This feature is great for productivity. All we need to do is update our [webpack-dev-server](https://github.com/webpack/webpack-dev-server) configuration, and use webpack's built in HMR plugin.
 
+W> You cannot update the entry point for a chunk. It will not be accepted.
+
 __webpack.config.js__
 
 ``` diff
@@ -39,8 +41,9 @@ __webpack.config.js__
 
   module.exports = {
     entry: {
-      app: './src/index.js',
-      print: './src/print.js'
+-      app: './src/index.js',
+-      print: './src/print.js'
++      app: './src/index.js'
     },
     devtool: 'inline-source-map',
     devServer: {
@@ -72,13 +75,6 @@ __index.js__
   import _ from 'lodash';
   import printMe from './print.js';
 
-+ if (module.hot) {
-+   module.hot.accept('./print.js', function() {
-+     console.log('Accepting the updated printMe module!');
-+     printMe();
-+   })
-+ }
-
   function component() {
     var element = document.createElement('div');
     var btn = document.createElement('button');
@@ -94,6 +90,14 @@ __index.js__
   }
 
   document.body.appendChild(component());
++
++ if (module.hot) {
++   module.hot.accept('./print.js', function() {
++     console.log('Accepting the updated printMe module!');
++     printMe();
++   })
++ }
+
 ```
 
 Start changing the `console.log` statement in `print.js`, and you should see the following output in the browser console.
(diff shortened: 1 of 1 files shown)


skipjack:
@maheshpec does this resolve the issue you opened? Also, I think your issue may overlap with #1461 so I may close yours in favor of the first but I will double check before doing so.

skipjack:
See my comment above... I think we can make a few changes to put the confusion to bed and make the code snippets flow a bit better.

@sbaidon can you explain more what those other warnings mentioned in #1461 were about? You mentioned some change to the `entry` configuration fixed them?

sbaidon:
Ok, so I re-organized the block as you mentioned. I am not getting any kind of error, but the warning is still showing up:

screen shot 2017-07-30 at 12 05 00 pm

As you can see the `print.js` module is being updated, but then the warning is telling us it wasn't ?

My guess was that somehow webpack was trying to update the module twice, and it is only being accepted the first time. 

I thought it had something to do with having both `print.js` and `index.js` listed as entry points in our configuration.

So when I removed `print.js` as an entry this is the output I got:

screen shot 2017-07-30 at 12 27 00 pm

By the way I added the `new webpack.NamedModulesPlugin()` to the config so we could see names instead of ids.

Still not sure if I am doing anything else wrong...

maheshpec:
I think @sbaidon is right. It is trying to load it twice - once for the `index` and other for the `print` entry point. When its trying to load the `print` entry point, it is getting `unaccepted` in `getAffectedStuff` function within `hotApply` function of `HotModuleReplacement.runtime.js` file within webpack (I'm using 3.4.1). The exact condition is 

```javascript
if(module.hot._main) {                                                                                  
        return {                                                                                                           
          type: "unaccepted",                                                                                                        
          chain: chain,                                                                                                    
          moduleId: moduleId                                                                                               
        };                                                                                                                 
} 
```

IMHO the approach to fix would be to add a _Preparation_ section which removes the print entry point. Please let me know what you guys think

maheshpec:
I made the changes that @sbaidon mentioned and the ones about ordering that @skipjack mentioned. I've tested this with 3.4.1 version of webpack and it is working without any console warnings for me.

skipjack:
@maheshpec @sbaidon yes good catch and @maheshpec thank you for updating! 

I should've caught that earlier, you shouldn't have an entry file that's consumed by another module without using the `CommonsChunkPlugin` and even there it usually just makes sense for vendor code splitting -- I completely missed that that was going on in these examples. I do understand how it happened when trying to synchronize this with the previous guides though. If we wanted to split `print.js` out to a separate chunk in the current setup, then `import()` would be a better solution.

As soon as the build passes, I'll get this merged so we can close out #1461.

content/guides/hot-module-replacement.md
```diff
@@ -94,6 +88,13 @@ __index.js__
   }
 
   document.body.appendChild(component());
++
```
sbaidon: There's just this extra plus sign left

skipjack: I think that's ok, it's just indicating that a new line was added in between the last line of the script and new content. That's how we've done it in the other guides as well.

skipjack: Ah or maybe you meant there's two plus signs? That's actually just a looking at a `diff` within a `diff` issue 😆  (there's really only one in the document).

maheshpec: It is a little confusing to view these diffs :)

skipjack: yeah like some kind diff inception lol

sbaidon: oh I did mean that the new line was added, but if that's how it has been done on the other guides is fine 👍

PR #1466, opened 2017-07-31
Display sponsors in categories

sokra:
Categories: Platin, Gold, Silber, Bronze
Logos have different sizes

image

### DIFF
--- a/components/splash/splash.jsx
+++ b/components/splash/splash.jsx
@@ -31,11 +31,21 @@ export default props => {
 
           <p>Through contributions, donations, and sponsorship, you allow webpack to thrive. Your donations directly support office hours, continued enhancements, and most importantly, great documentation and learning material!</p>
 
-          <h2>Sponsors</h2>
-          <Support number={ 100 } type="sponsors" />
+          <h2>Platinum Sponsors</h2>
+          <Support type="sponsors" rank="platinum" />
+
+          <h2>Gold Sponsors</h2>
+          <Support type="sponsors" rank="gold" />
+
+          <h2>Silver Sponsors</h2>
+          <Support type="sponsors" rank="silver" />
+
+          <h2>Bronze Sponsors</h2>
+          <Support type="sponsors" rank="bronze" />
 
           <h2>Backers</h2>
-          <Support number={ 130 } type="backers" />
+          <Support type="backers" />
+
         </Container>
       </div>
     </div>
--- a/components/support/support-style.scss
+++ b/components/support/support-style.scss
@@ -8,29 +8,40 @@
 
   &__item {
     position: relative;
-    margin: 0 5px 5px 5px;
+    margin: 0 2px 2px 2px;
   }
 
   &__sponsors-avatar {
-    height: 64px;
+    &-bronze, &-normal {
+      height: 32px;
+    }
+    &-silver {
+      height: 64px;
+    }
+    &-gold {
+      height: 96px;
+    }
+    &-platinum {
+      height: 128px;
+    }
   }
 
-  &__backers-avatar {
-    width: 62px;
-    height: 62px;
+  &__backers-avatar-normal {
+    width: 31px;
+    height: 31px;
     border-radius: 50%;
-    border: 2px solid white;
+    border: 1px solid white;
     overflow: hidden;
   }
 
   &__outline {
     position: absolute;
     top: 0;
-    margin: -2px;
-    width: calc(100% + 4px);
-    height: calc(100% - 2px);
+    margin: -1px;
+    width: calc(100% + 2px);
+    height: calc(100% - 4px);
     border-radius: 50%;
-    border: 2px solid rgb(112, 202, 10);
+    border: 1px solid rgb(112, 202, 10);
   }
 
   &__bottom {
--- a/components/support/support.jsx
+++ b/components/support/support.jsx
@@ -2,27 +2,60 @@ import React from 'react';
 import Additional from './support-additional.json';
 import './support-style';
 
+const ranks = {
+  bronze: {
+    maximum: 2000
+  },
+  silver: {
+    minimum: 2000,
+    maximum: 10000
+  },
+  gold: {
+    minimum: 10000,
+    maximum: 50000
+  },
+  platinum: {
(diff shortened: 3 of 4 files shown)


components/splash/splash.jsx
```diff
@@ -31,11 +31,21 @@ export default props => {
 
           <p>Through contributions, donations, and sponsorship, you allow webpack to thrive. Your donations directly support office hours, continued enhancements, and most importantly, great documentation and learning material!</p>
 
-          <h2>Sponsors</h2>
-          <Support number={ 100 } type="sponsors" />
+          <h2>Platin Sponsors</h2>
+          <Support size="platin" minimum={ 50000 } type="sponsors" />
```
skipjack: Should this be "Platinum Sponsors"?

components/splash/splash.jsx
```diff
@@ -31,11 +31,21 @@ export default props => {
...
+
+          <h2>Gold Sponsors</h2>
+          <Support size="gold" minimum={ 10000 } maximum={ 50000 } type="sponsors" />
+
+          <h2>Silber Sponsors</h2>
+          <Support size="silber" minimum={ 2000 } maximum={ 10000 } type="sponsors" />
```
skipjack: Should this be "Silver Sponsors"?

skipjack: If the answer is yes to both of these then we should update the semantics throughout the whole PR (e.g. props, css classes, etc.).

components/support/support-style.scss
```diff
@@ -8,29 +8,40 @@
...
     top: 0;
-    margin: -2px;
-    width: calc(100% + 4px);
-    height: calc(100% - 2px);
+    margin: -1px;
+    width: calc(100% + 2px);
+    height: calc(100% - 4px);
     border-radius: 50%;
-    border: 2px solid rgb(112, 202, 10);
+    border: 1px solid rgb(112, 202, 10);
```
skipjack: Just gauging from the screenshot, it seems some of these style changes might have caused the normal backers display to be thrown off. I'll check locally as well but is there a reason you changed the "normal" backers style? They look pretty good to me as is in Chrome.

components/support/support.jsx
```diff
@@ -4,25 +4,34 @@ import './support-style';
 
 export default class Support extends React.Component {
   render() {
-    let { number, type } = this.props;
+    let { minimum, maximum, size, type } = this.props;
```
skipjack: I might rename the `size` prop to `level` or `rank` or something similar. "Size" sounds more like count to display or image size to me than their rank based on donation amount. Also, maybe instead of `minimum` and `maximum` being props along with `level`, we could have that data within the component file so only `level` needs to be passed.

components/support/support.jsx
```diff
@@ -4,25 +4,34 @@ import './support-style';
...
+    }
+
+    if (typeof maximum === 'number') {
+      supporters = supporters.filter(item => item.totalDonations < maximum * 100);
+    }
+
     return (
       <div className="support">
         {
-          supporters.slice(0, number).map((supporter, index) => (
```
skipjack: Just want to make sure this is intentional -- we don't want to limit the backers or supporters displayed at all?

components/support/support-style.scss
skipjack: By the way -- this is what I mean re the backer design:

image

There's some noticeable extra whitespace at the bottom of each backer that makes things look a bit off.

skipjack:
It seems the build error was a 404'd backer image... not sure if that was just a temporary outage or caused by these changes. I restarted the build to see if it's reproducible.

components/support/support-style.scss
sokra: The screenshot is on 50% browser zoom. On 100% it's looks normal.

I reduced the size of the backers by 50% to match the size of the bronze sponsors.

components/support/support.jsx
sokra: Yep it was intentional.

sokra:
Updated version: (now in 100% browser zoom)

image

sokra:
Updated version:
image


CI seem to be correct with some URLs being broken:

Adam Howard
https://d1ts43dypk8bqh.cloudfront.net/v1/avatars/5e2529e2-ca5e-495c-8e2f-facb62961cdc 404

Dana Woodman
https://opencollective-production.s3-us-west-1.amazonaws.com/danawoodmancom_7bada460-bd8b-11e6-9ca1-bb6c7a9a84e0.png 200 but length = 0

That's the data we receive from opencollective API. cc @xdamman

components/support/support-style.scss
skipjack: 👍  I'll just double check this locally then when testing to make sure everything looks as it should.

components/support/support.jsx
skipjack: 👍

skipjack:
> That's the data we receive from opencollective API.

Yeah just weird it's only being thrown on this PR -- ah I guess because more backers are now being shown.

skipjack:
Test locally, looks good. Working on the build failures, it looks like the change in class name might have thrown off our `scripts/check-links.js` ignore pattern.

xdamman:
Love this!

Yeah some backers don't always provide a valid avatar :-/ 
Not sure what we can do about it. Ideas welcome.

The only thing that I'd add to this is an explanation of what makes a platinum/gold/silver sponsor. One of the thing you can do is to redirect the button become a sponsor to a custom URL e.g. https://opencollective.com/webpack/donate/500/month/silver%20sponsor

See also our new buttons on https://opencollective.com/widgets

skipjack:
> Yeah some backers don't always provide a valid avatar :-/ Not sure what we can do about it. Ideas welcome.

Yeah not sure there's much you can do. I updated our link testing to ignore them which should be fine for now.

> The only thing that I'd add to this is an explanation of what makes a platinum/gold/silver sponsor. 

Yes I agree, I think a small description under each section is probably warranted. @sokra we can probably just add a `description` field to each object in the `rank` data in `support.jsx` and display those dynamically.

> One of the thing you can do is to redirect the button become a sponsor to a custom URL

Awesome, thanks!

PR #1468, opened 2017-07-31
Add documentation for devServer cli options

orteth01:
Originally was just adding documentation for the allowedHosts cli option https://github.com/webpack/webpack-dev-server/pull/1012.

However, because there was no documentation for other CLI options I went ahead and added those too.

### DIFF
--- a/content/configuration/dev-server.md
+++ b/content/configuration/dev-server.md
@@ -74,6 +74,11 @@ allowedHosts: [
 ]
 ```
 
+To use this option with the CLI pass the `--allowed-hosts` option a comma-delimited string.
+
+```
+webpack-dev-server --entry /entry/file --output-path /output/path --allowed-hosts .host.com,host2.com
+```
 
 ## `devServer.clientLogLevel`
 
--- a/content/configuration/dev-server.md
+++ b/content/configuration/dev-server.md
@@ -421,6 +421,20 @@ Usage via the CLI
 webpack-dev-server --pfx /path/to/file.pfx
 ```
 
+## `devServer.pfxPassphrase`
+
+`string`
+
+The passphrase to a SSL PFX file.
+```js
+pfxPassphrase: 'passphrase'
+```
+
+Usage via the CLI
+```
+webpack-dev-server --pfx-passphrase passphrase
+```
+
 ## `devServer.port` - CLI only
 
 `number`
--- a/content/configuration/dev-server.md
+++ b/content/configuration/dev-server.md
@@ -621,6 +621,21 @@ setup(app){
 }
 ```
 
+## `devServer.socket`
+
+`string`
+
+The Unix socket to listen to (instead of a host).
+
+```js
+socket: 'socket'
+```
+
+Usage via the CLI
+```
+webpack-dev-server --socket socket
+```
+
 
 ## `devServer.staticOptions`
 
--- a/content/configuration/dev-server.md
+++ b/content/configuration/dev-server.md
@@ -549,7 +549,7 @@ webpack-dev-server --progress
 ```
 
 
-## `devServer.public` - CLI only
+## `devServer.public`
 
 `string`
 
@@ -559,6 +559,11 @@ For example, the dev-server is proxied by nginx, and available on `myapp.test`:
 
 ```js
 public: "myapp.test:80"
+```
+
+Usage via the CLI
+```
+webpack-dev-server --public myapp.test:80
 ```
 
 
--- a/content/configuration/dev-server.md
+++ b/content/configuration/dev-server.md
@@ -237,7 +237,7 @@ hot: true
 T> Note that you must also include a `new webpack.HotModuleReplacementPlugin()` to fully enable HMR. See the [HMR concepts page](/concepts/hot-module-replacement) for more information.
 
 
-## `devServer.hotOnly` - CLI only
+## `devServer.hotOnly`
 
 `boolean`
 
@@ -247,6 +247,10 @@ Enables Hot Module Replacement (see [`devServer.hot`](#devserver-hot)) without p
 hotOnly: true
 ```
 
+Usage via the CLI
+```
+webpack-dev-server --hot-only
+```
 
 ## `devServer.https`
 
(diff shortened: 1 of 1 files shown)


simon04:
This would be the first CLI option explicitly documented for webpack-dev-server. We might want to do the same for [all other options](https://github.com/webpack/webpack-dev-server/blob/1cf4359aa4b9aef5fedeeb89de9e9a7e4ea52687/bin/webpack-dev-server.js#L66-L213) in a consistent way?

"Flag" typically relates to a Boolean option (it's there or it isn't); I find "option" more adequate.

orteth01:
> "Flag" typically relates to a Boolean option (it's there or it isn't); I find "option" more adequate.

i'll update it. I can also take a shot at documenting the other CLI options.

content/configuration/dev-server.md
```diff
@@ -129,6 +165,25 @@ To disable `contentBase`:
 contentBase: false
 ```
 
+Usage via the CLI
+```
+webpack-dev-server --content-base /path/to/content/dir
+```
+
+## `devServer.disableHostCheck`
```
orteth01: was this option intentionally left out of the docs to prevent people from using it? If so, I can remove this.

skipjack:
This overlaps with #1196... there was some discussion there on where this documentation should go. I'm actually ok with small sections below each config example as both as you've done although I think there are some formatting issues (which may be why the build is failing).

content/configuration/dev-server.md
```diff
@@ -74,6 +74,24 @@ allowedHosts: [
...
+## `devServer.bonjour`
+
+This option broadcasts the server via ZeroConf networking on start
+
+```js
+bonjour: true
+```
+
+Usage via the CLI
+```
```
skipjack: You need to add a line break between each fenced code block and the paragraph above it.

skipjack: This is why the build is currently failing.

content/configuration/dev-server.md
```diff
@@ -312,7 +439,36 @@ overlay: {
...
+The passphrase to a SSL PFX file.
+```js
+pfxPassphrase: 'passphrase'
+```
+
+Usage via the CLI
+```
+webpack-dev-server --pfx-passphrase passphrase
+```
+
```
skipjack: It should be two line breaks between each section -- it seems on some you did add two but at least a few of these only have one.

content/configuration/dev-server.md
```diff
@@ -403,6 +566,11 @@ For example, the dev-server is proxied by nginx, and available on `myapp.test`:
 public: "myapp.test:80"
 ```
 
+Usage via the CLI
+```
+webpack-dev-server --public myapp.test:80
+```
+
 
 ## `devServer.publicPath` 🔑
```
skipjack: @simon04 (or @orteth01) do either of you know the purpose of the key icons? Can we drop them?

skipjack: Or @SpaceK33z? I'm not sure what they're supposed to indicate. If they were meaningful maybe we can think of a more intuitive display.

orteth01: > Options that are compatible with webpack-dev-middleware have 🔑 next to them.

skipjack: Ah I see, thank you.

orteth01:
> there was some discussion there on where this documentation should go. 

@skipjack yea I was originally thinking about creating a separate readme for the CLI but it seemed like that would result in a lot of duplicated explanations of options - that's why I opted to just include CLI usage in the devServer readme

skipjack:
> yea I was originally thinking about creating a separate readme for the CLI but it seemed like that would result in a lot of duplicated explanations of options - that's why I opted to just include CLI usage in the devServer readme

Yeah I agree re duplication -- I think at some point, once we've finished reviewing and structuring the documentation on this site we should really start thinking about how more of the documentation can be auto-generated from code comments or other data that provides a single source of truth. This is a perfect example of where this would work well, i.e. that same [data](https://github.com/webpack/webpack-dev-server/blob/1cf4359aa4b9aef5fedeeb89de9e9a7e4ea52687/bin/webpack-dev-server.js#L66-L213) being passed to `yargs.options` could really be the source of truth for this whole page. However this definitely gets tricky with the need for examples and other more detailed guides/content.

content/configuration/dev-server.md
```diff
@@ -129,6 +173,29 @@ To disable `contentBase`:
...
 ```
 
+Usage via the CLI
+
+```bash
+webpack-dev-server --content-base /path/to/content/dir
+```
+
+
+## `devServer.disableHostCheck`
```
orteth01: > was this option intentionally left out of the docs to prevent people from using it? If so, I can remove this.

commented on an earlier, now-out-of-date commit. just want to make sure this is alright

skipjack: I think if it's an available option and it's noted by the CLI when running `webpack-dev-server -h` then it should be documented here as well. Your "THIS IS NOT RECOMMENDED..." is prominent enough that people should notice this, and if they choose to proceed anyway, so be it.

I'm going to do one last skim and then merge this. @SpaceK33z if you have a chance to review this later on and come across any issues -- please ping us so we can resolve in a separate PR.

content/configuration/dev-server.md
```diff
@@ -252,6 +337,29 @@ https: {
...
+`boolean`
+
+Output cli information. It is enabled by default.
+
+```bash
+webpack-dev-server --info=false
+```
+
 
 ## `devServer.inline` - CLI only
```
skipjack: I think we should drop the "CLI only" flag here as well unless it was removed in a recent version. I've definitely used `inline` via the configuration before.

skipjack: @orteth01 I'm just going to make this change so we can get this merged. Ping me if I'm wrong on this though and we can fix in a separate PR.

orteth01: @skipjack yea you're totally right. I must have just missed this one. my bad!

skipjack: No worries, thanks again for this update! If you don't mind I may ping you in the future if there's other questions re this page since it seems like you have some familiarity with the webpack-dev-server repo.

orteth01: > If you don't mind I may ping you in the future if there's other questions re this page

yea no problem! happy to help!

Issue #1331, opened 2017-06-23
Guides - Issues with Tree Shaking & Production


jakearchibald:
I was trying to get tree shaking working. Here are a few notes about the [tree shaking](https://webpack.js.org/guides/tree-shaking/) and [production](https://webpack.js.org/guides/production/) guides:

The article explains what tree shaking is (really well), but it doesn't really cover how to achieve it.

`--optimize-minimize` will do it, but that also throws in uglify, which can make it difficult to confirm if treeshaking has worked, and will throw errors if you're using ES2016+.

One of the guides should cover how to enable tree shaking without enabling other optimisations. This will help users confirm that code is being removed, without having to read minified code.

The production guide should cover how to avoid errors when using ES2016+, either by using the uglify harmony branch, or BabiliWebpackPlugin. None of the docs I've seen cover how to use BabiliWebpackPlugin instead of uglify while retaining tree shaking.

Tree shaking will break if using babel + the es2015 preset, as it transpiles the imports. I've filed https://github.com/babel/babel-loader/issues/477, but it may be worth mentioning in the guide too.

jakearchibald:
The production guide suggests:

```js
new webpack.LoaderOptionsPlugin({
  minimize: true,
  debug: false
})
```

…will perform tree-shaking, but that plugin appears to be deprecated, and doesn't appear to be handling tree-shaking.

insin:
After playing about with this, it seems like Uglify is still responsible for actually removing the unused code - Webpack's role in tree shaking appears to be avoiding generating `__webpack_exports__[id] = whatever` statements for unused ES6 exports so Uglify doesn't count them as "used".

jakearchibald:
Yeah, it seems to behave that way. Babili does the same.

I guess the docs led me to think the code removal was performed by webpack.

skipjack:
> After playing about with this, it seems like Uglify is responsible for actually removing the unused code - Webpack's role in tree shaking appears to be avoiding generating __webpack_exports__[id] = whatever statements for unused exports so Uglify doesn't count them as "used" for the purposes of dead code elimination.

@insin @jakearchibald I've seen a few other discussions recently that all came to the same conclusion as well. Essentially it's the combination of ES6 static `import` / `export` statements and the `UglifyJsPlugin` that enable tree shaking from my understanding. A warning was added recently in #1474 that starts to clarify this but I'll elaborate on this and address the other things mentioned above.

PR #1474, opened 2017-08-01
Add note about --optimize-minimize is not required for tree shaking

### LINKED ISSUE #1473: Inform --optimize-minimize is not required for tree shaking
There seems to be some confusion as to when tree shaking occurs (ie what configuration is necessary). Ref #1331 https://github.com/webpack/webpack/issues/4713#issuecomment-317601066 

Correct me if I am wrong but it seems multiple separate parts of webpack is responsible for different parts of tree shaking? Anywho, following https://github.com/webpack/webpack/issues/4713#issuecomment-317601066 I thought it might be a good idea to note in the documentation that `--optimize-minimize` is not required for tree shaking to occur.

I have added an initial PR #1474 . This is not meant as a final solution but more of an opener for a discussion of how to explain that `--optimize-minimize` is not only way to tree shaking but a production config is.

esbenp:
Idea for #1473 (also see #1331). Opener for discussion - feedback welcome.

### DIFF
--- a/content/guides/tree-shaking.md
+++ b/content/guides/tree-shaking.md
@@ -84,3 +84,5 @@ function(e,t,n){"use strict";function r(e){return e*e*e}t.a=r}
 /* ... */
 function(e,t,n){"use strict";Object.defineProperty(t,"__esModule",{value:!0});var r=n(0);console.log(n.i(r.a)(5))}
 ```
+
+T> Note, using the `--optimize-minimize` flag is NOT required for tree shaking to occur. Tree shaking will occur when using a production configuration. This can also be configured directly in a `webpack.config.js` file. Read [production build](/guides/production) for an in-depth guide on how to create a production config file.
--- a/content/guides/tree-shaking.md
+++ b/content/guides/tree-shaking.md
@@ -85,4 +85,4 @@ function(e,t,n){"use strict";function r(e){return e*e*e}t.a=r}
 function(e,t,n){"use strict";Object.defineProperty(t,"__esModule",{value:!0});var r=n(0);console.log(n.i(r.a)(5))}
 ```
 
-T> Note, using the `--optimize-minimize` flag is NOT required for tree shaking to occur. Tree shaking will occur when using a production configuration. This can also be configured directly in a `webpack.config.js` file. Read [production build](/guides/production) for an in-depth guide on how to create a production config file.
+T> Note that the `--optimize-minimize` flag enables tree shaking by including the `UglifyJsPlugin` behind the scenes. Alternatively, the `UglifyJsPlugin` can be included manually in the `plugins` section of your configuration file. The plugin, combined with webpack's resolving of `import` and `export` statements, is what makes tree shaking possible. See the [production build](/guides/production) guide for more information.


content/guides/tree-shaking.md
```diff
@@ -84,3 +84,5 @@ function(e,t,n){"use strict";function r(e){return e*e*e}t.a=r}
 /* ... */
 function(e,t,n){"use strict";Object.defineProperty(t,"__esModule",{value:!0});var r=n(0);console.log(n.i(r.a)(5))}
 ```
+
+T> Note, using the `--optimize-minimize` flag is NOT required for tree shaking to occur. Tree shaking will occur when using a production configuration. This can also be configured directly in a `webpack.config.js` file. Read [production build](/guides/production) for an in-depth guide on how to create a production config file.
```
skipjack: Instead of saying:

> Tree shaking will occur when using a production configuration.

I think we should be more specific on exactly what's necessary which I think, as I mentioned in #1473, is allowing webpack to actually resolve `import` / `export` statements (i.e. disabling any transpilers like babel from resolving them). And, including the `UglifyJsPlugin`. The `import` / `export` is already mentioned at the top of this guide so maybe just noting that either `--optimize-minimize` should be used (which includes the `UglifyJsPlugin` under the hood) or the plugin can be configured directly (likely in a "production" configuration).

esbenp: Okay, I think that makes sense. However, I am unsure about the first part. Do you mean using something like [babel-plugin-resolver](https://github.com/tleunen/babel-plugin-module-resolver) when you say resolving using transpilers? I am not sure how transpilers work, but I am assuming loading your JS files with `babel-loader` does not resolve any imports, right?

What do you think about something along the lines of

> Note, using the `--optimize-minimize` flag is NOT required for tree shaking to occur. Tree shaking will occur when Webpack resolves your import/export statements and when you are including the `UglifyJsPlugin`. You can either include the `UglifyJsPlugin` using the `--optimize-minimize` flag or by adding the plugin manually to your `plugins` key of your configuration file. Read [production build](/guides/production) for an in-depth guide on how to create a production config file.

skipjack: > However, I am unsure about the first part. Do you mean using something like babel-plugin-resolver when you say resolving using transpilers? I am not sure how transpilers work, but I am assuming loading your JS files with babel-loader does not resolve any imports, right?

Depends on how you have the `babel` and the `babel-loader` configured. With certain plugins, you'll need to set `{ modules: false }` in order to let webpack resolve the `import` and `export` syntax.

> What do you think about something along the lines of...

Definitely sounds better... I might reword/simplify as:

> Note that the `--optimize-minimize` flag enables tree shaking by including the `UglifyJsPlugin` behind the scenes. The `UglifyJsPlugin` can also be included manually in the `plugins` section of your configuration file. The plugin, combined with webpack's resolving of `import` and `export` statements, is what makes tree shaking possible. See the [production build guide](/guides/production) for more information.

What do you think?

esbenp: Yes, this is a better wording, I agree. I have updated the diff to your suggestion.

skipjack:
Great, thanks @esbenp! Re-running the build due to some timeouts (which we should really be ignoring -- will address this finally in a separate PR).

Once the build passes we should be able to get this merged -- at which point I'll close out #1473 but make a note in #1331 that this should be one of the things we address in the rewrite.

PR #1485, opened 2017-08-04
Update getting-started.md

firatkucuk:
without dist it doesn't work.


firatkucuk:
I didn't notice the index.html file in dist directory. But I thought dist directory must be temporary there shouldn't be any persistent file there. As a result my mistake sorry.

PR #1486, opened 2017-08-04
Fix display errors: <li><p><code>

pierreneter:
Fix display error for the list in page `/development/plugin-patterns/`, and any list that uses ` `` `
https://webpack.js.org/development/plugin-patterns/


pierreneter:
webpack-site20170508

skipjack:
Weird -- there's a list [here](https://webpack.js.org/plugins/prefetch-plugin/) that doesn't have the same issues. I wonder if there's something else that's causing it on that page.

Anyway thanks for your contribution, I'll review this more this weekend.

pierreneter:
Oh, I checked the difference between 2 md files.
Because `li`s is interrupted by an enter in `plugin-patterns` file, so it creates a `p` tag that wrapper the `li`, and cause the other `li` to be wrappered by the tag `p`.

pierreneter:
Oh, I checked the difference between 2 md files.
Because `li`s is interrupted by an enter in `plugin-patterns` file, so it creates a `p` tag that wrapper the `li`, and cause the other `li` to be wrappered by  `p` tag.

pierreneter:
I think we should keep delete `display: inline-block;` as in my previous commit.
I have checked and it does not affect anything if it was removed.
Keep it will spoil the list UI have `li`s are followed by enter written by someone in the future.

skipjack:
> I think we should keep delete display: inline-block; as in my previous commit.

@pierreneter yeah I'm slightly worried that this was an intentional for some other reason when @pastelsky worked on the markdown styling updates so I'm hoping it doesn't break anything. That said `display: inline-block;` also causes inline code spanning more than one line (e.g. code at the end of a line) to be bumped down to the next line instead of naturally wrapping. I also checked how GitHub and some other style `code` tags within paragraphs and the default `display: inline;` seems to be fine.

> I have checked and it does not affect anything if it was removed.

Thanks for testing, but there is huge amount of content on this site so it's a bit hard to be sure that we haven't missed an edge case. However, if we have, I'm sure someone will bug us about it 😆  so it's not too big a deal.

> Keep it will spoil the list UI have lis are followed by enter written by someone in the future.

True, hopefully we'll enforce more in our markdown linting soon so formatting quirks like that won't slip past our CI process.

PR #1441, opened 2017-07-17
docs(guides): add using-webpack-dev-middleware section

navarroaxel:

PR #1488, opened 2017-08-05
Add missing comma in development guides 

reyesyang:


skipjack:
@reyesyang thanks, however I think this overlaps with #1441 as @TheDutchCoder made the same correction there. That one should be merged soon so I think we can close this...

PR #1492, opened 2017-08-06
Update Index Page

### LINKED ISSUE #1300: Index - Link to Concepts/Loader Docs
I recently have made the move to Webpack from the Rails asset pipeline and found myself incredibly confused when getting familiar with the concepts. The first experience I had with Webpack was actually with Rails 5.1 via the Webpacker gem and fell in love. Since I enjoyed the experience so much I thought I would go ahead and try out setting up things on my own and found myself incredibly confused when getting started. I'm not sure if this is up for debate or not, but I'd like to suggest a slightly different approach to getting started with Webpack.

I imagine most people do what I do and do a DuckDuckGo or Google search for Webpack, visit the homepage and click the getting started page. I spent hours trying to figure out how to use a loader because the getting started guide doesn't really explain. It wasn't until I heard from Sean Larkin that the concepts page existed to explain loaders and plugins. From that moment on I've been so happy because if I wouldn't have found the concepts page I would have gone back to using the Webpacker gem and never learning how to do this all on my own.

I'd like to suggest that concepts are highlighted more on the homepage to help people out that are coming from different backgrounds like myself.

Thanks!

skipjack:
This is both a follow up to #1466 and resolves #1300. The index page now contains short descriptions for each sponsor/backer section:

image

cc @sokra @xdamman @joshmosh

### DIFF
--- a/components/splash/splash-style.scss
+++ b/components/splash/splash-style.scss
@@ -10,6 +10,11 @@
       background-color:#f3f3f3;
     }
 
+    p {
+      margin: 0 auto;
+      max-width: 800px;
+    }
+
     .container {
       padding:5em 1em;
 
--- a/components/support/support-style.scss
+++ b/components/support/support-style.scss
@@ -4,7 +4,18 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: center;
-  padding: 1em 0.5em;
+  padding: 0 0.5em 1em;
+
+  &__description {
+    flex: 0 0 100%;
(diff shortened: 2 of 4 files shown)


sokra:
> who have pledged to $2000 to webpack

I'm not an english expert, but this sounds a bit weird to me. Shouldn't it be `pledged up to $2000 to webpack`?

skipjack:
> I'm not an english expert...

@sokra haha apparently neither am I! Good catch, I'll update.

PR #1514, opened 2017-08-13
Loading Images

srsandy:
Respected Sir,

import _ from 'lodash';
  import './style.css';
  import Icon from './icon.png';

  function component() {
    // Lodash, now imported by this script
    element.classList.add('hello');
+   // Add the image to our existing div.
+   var myIcon = new Image();
+   myIcon.src = Icon;
+   element.appendChild(myIcon);

    return element;
  }

Here in the above code. We are appending a new child  which adds the image after 'Hello webpack' not behind it

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -257,7 +257,7 @@ Time: 895ms
    [9] ./src/index.js 503 bytes {0} [built]
 ```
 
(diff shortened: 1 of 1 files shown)


srsandy:
Sir this is a documentation error

skipjack:
> Respected Sir,
> ...
> Sir this is a documentation error

😆 thanks for the polite greeting with your PR! We're a bit more informal here so no need for "sir". I requested a review from @TheDutchCoder whose done a lot of work on the guides and should be able to help. Re the error, you are correct -- this was caused by a plugin name change in `webpack-contrib` -- please ignore and I will resolve in a separate PR.

PR #1524, opened 2017-08-14
Mention Python dependency in contribution guide

ChrisChinchilla:
docs(contributing): Mention Python dependency


ChrisChinchilla:
@skipjack I got an error when I ran `npm install', as I had an unsupported Python version. But agree with the ordering.

skipjack:
Ah ok nvm mentioning 'npm run lint' then.

ChrisChinchilla:
OK @skipjack updated the order.

PR #1490, opened 2017-08-06
[PWA] rename components and move app specific code to /src

pierreneter:
Related to #1400 
Finished the first two lines in this checklist.

PR #1533, opened 2017-08-22
Improve documentation on devtool

### LINKED ISSUE #273: Config - Elaborate on Source Map Options

sokra:
fixes webpack/webpack#2725
fixes webpack/webpack#4936
fixes webpack/webpack#2766
fixes webpack/webpack#2145
fixes webpack/webpack#1689
fixes #273


content/configuration/devtool.md
```diff
@@ -26,53 +26,84 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
-eval-source-map               | --    | +       | no         | original source
-source-map                    | --    | --      | yes        | original source
-inline-source-map             | --    | --      | no         | original source
-hidden-source-map             | --    | --      | yes        | original source
-nosources-source-map          | --    | --      | yes        | without source content
-
-T> `+` means faster, `-` slower and `o` about the same time
-
-Some of these values are suited for development and some for production. For development you typically want fast Source Maps at the cost of bundle size, but for production you want separate Source Maps that are accurate.
+T> Instead of using the `devtool` option you can also use the `SourceMapDevToolPlugin`/`EvalSourceMapDevToolPlugin` directly. It has more options. You must never use both `devtool` option and plugin!
```
ChrisChinchilla: T> Instead of using the `devtool` option you can also use `SourceMapDevToolPlugin`/`EvalSourceMapDevToolPlugin` directly as it has more options. Never use both the `devtool` option and plugin together.

ChrisChinchilla: @sokra Can we say why someone shouldn't use them together?

content/configuration/devtool.md
```diff
@@ -26,53 +26,84 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
+cheap-source-map               | +     | o       | no         | transformed code (lines only)
+cheap-module-source-map        | o     | -       | no         | original source (lines only)
+inline-cheap-source-map        | +     | o       | no         | transformed code (lines only)
+inline-cheap-module-source-map | o     | -       | no         | original source (lines only)
+source-map                     | --    | --      | yes        | original source
+inline-source-map              | --    | --      | no         | original source
+hidden-source-map              | --    | --      | yes        | original source
+nosources-source-map           | --    | --      | yes        | without source content
+
+T> `+++` super fast, `++` pretty fast, `+` fast, `o` medium, `-` slow, `--` pretty slow
```
ChrisChinchilla: @sokra You had a previous commit that used `very fast`, I think that's actually better, as 'pretty' would typically be one before 'fast'.

content/configuration/devtool.md
```diff
@@ -26,53 +26,84 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
+
+Some of these values are suited for development and some for production. For development you typically want fast Source Maps at the cost of bundle size, but for production you want separate Source Maps that are accurate and support minimizing.
 
 W> There are some issues with Source Maps in Chrome. [We need your help!](https://github.com/webpack/webpack/issues/3165).
 
 T> See [`output.sourceMapFilename`](/configuration/output#output-sourcemapfilename) to customize the filenames of generated Source Maps.
 
+### Qualities
```
ChrisChinchilla: @sokra Can we add a line break between the explanation and the examples?

content/configuration/devtool.md
```diff
@@ -26,53 +26,84 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
 
 ### Production
 
 These options are typically used in production:
 
-`source-map` - A full SourceMap is emitted as a separate file. It adds a reference comment to the bundle so development tools know where to find it.
+`(none)` (Omit the `devtool` option) - No SourceMap is emitted. This is a good option to start with.
 
-`hidden-source-map` - Same as `source-map`, but doesn't add a reference comment to the bundle. Useful if you only want SourceMaps to map error stack traces from error reports, but don't want to expose your SourceMap for the browser development tools.
+`source-map` - A full SourceMap is emitted as a separate file. It adds a reference comment to the bundle so development tools know where to find it. Warning: You should configure your server to disallow access to the Source Map file for normal users!
```
ChrisChinchilla: @sokra Could we add a line break between the explanation and the warning? Ideally this would be formatted differently, but we'll worry about that later.

content/configuration/devtool.md
```diff
@@ -26,53 +26,84 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
 
-`cheap-source-map` - A SourceMap without column-mappings ignoring loaded Source Maps.
+`hidden-source-map` - Same as `source-map`, but doesn't add a reference comment to the bundle. Useful if you only want SourceMaps to map error stack traces from error reports, but don't want to expose your SourceMap for the browser development tools. Warning: You should not deploy the Source Map file to the webserver. Instead only use it for error report tooling.
```
ChrisChinchilla: @sokra Could we add a line break between the explanation and the warning? Ideally this would be formatted differently, but we'll worry about that later.

content/configuration/devtool.md
```diff
@@ -26,53 +26,84 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
 
-`cheap-module-source-map` - A SourceMap without column-mappings that simplifies loaded Source Maps to a single mapping per line.
+`nosources-source-map` - A SourceMap is created without the `sourcesContent` in it. It can be used to map stack traces on the client without exposing all of the source code. You can deploy the Source Map file to the webserver. Warning: It still exposes filenames and structure for decompiling, but it doesn't expose the original code.
```
ChrisChinchilla: @sokra Could we add a line break between the explanation and the warning? Ideally this would be formatted differently, but we'll worry about that later.

content/configuration/devtool.md
sokra: The CI didn't allowed me to use `very`.

sokra:
@ChrisChinchilla Thanks for your feedback.

content/configuration/devtool.md
ChrisChinchilla: Interesting, I guess it labels it as a 'weasel word', which is kind of true, but an indication there are times you might want it 😆 .

Thanks!

content/configuration/devtool.md
```diff
@@ -26,53 +26,91 @@ Choose a style of [source mapping](http://blog.teamtreehouse.com/introduction-so
...
 
 ### Production
 
 These options are typically used in production:
 
+`(none)` (Omit the `devtool` option) - No SourceMap is emitted. This is a good option to start with.
+
 `source-map` - A full SourceMap is emitted as a separate file. It adds a reference comment to the bundle so development tools know where to find it.
 
+W> You should configure your server to disallow access to the Source Map file for normal users!
```
skipjack: @ChrisChinchilla we actually can format them differently (as @sokra did here). We have custom markdown syntax enabled so that you can use `W>` for a warning, `T>` for a tip, and `?>` for a TODO. I think this is mentioned in the [writer's guide](https://webpack.js.org/writers-guide) but if it's not feel free to submit a PR.

skipjack:
I'm going to work on resolving the conflicts here and in the other approved PRs as soon as I get some time so we can get these things merged. They're due to the major restructuring from #1490.

skipjack:
@sokra merged, you should be able to close those issues on the main repo now. (it doesn't seem like the "fixes" keyword works across repos unfortunately)

PR #1539, opened 2017-08-23
add HMR with node.js API example to guides

bdwain:
Fixing issue filed with webpack-dev-server (https://github.com/webpack/webpack-dev-server/issues/1051#issuecomment-324189203).

There was no example of doing HMR with the Node API in the docs, and it turns out an undocumented function (addDevServerEntrypoints) was required to make it work. 

This adds a working example to the docs for people setting up the dev server with node.


src/content/guides/hot-module-replacement.md
```diff
@@ -125,6 +126,64 @@ main.js:4395 [WDS] Hot Module Replacement enabled.
 + main.js:4330 [HMR] Consider using the NamedModulesPlugin for module names.
 ```
 
+### HMR With the Node.js API
+
+When using Webpack Dev Server with the Node.js API, you won't put the dev server options on the webpack config object. Instead, you'll pass them as a second parameter to `new WebpackDevServer(compiler, options)`.
```
ChrisChinchilla: @bdwain Replace `you won't` with `don't`.

src/content/guides/hot-module-replacement.md
```diff
@@ -125,6 +126,64 @@ main.js:4395 [WDS] Hot Module Replacement enabled.
...
+
+To enable HMR, you'll also need to modify your webpack configuration object to have the HMR entry points. Webpack Dev Server exports a method called `addDevServerEntrypoints` which you can use to do this.
```
ChrisChinchilla: @bdwain Change `you'll` with `you`.

bdwain:
fixed! @ChrisChinchilla Thanks for the suggestions.

skipjack:
To elaborate on my previous comment. I think just including the section without the `diff`s would be enough, for example:

> When using `webpack-dev-server` with the Node.js API, don't put the dev server options in the webpack config object. Instead, pass them as a second parameter upon creation, for example:
> 
> ```
> new WebpackDevServer(compiler, options)
> ```
>
> And remember that to enable HMR with the Node.js API, you will need to modify your webpack configuration object to include the HMR entry points. The `webpack-dev-server` package includes a handy method called `addDevServerEntrypoints` which you can use to do this.

src/content/guides/hot-module-replacement.md
```diff
@@ -125,6 +126,64 @@ main.js:4395 [WDS] Hot Module Replacement enabled.
...
+``` diff
+  const path = require('path');
+  const HtmlWebpackPlugin = require('html-webpack-plugin');
+  const webpack = require('webpack');
+
+  module.exports = {
+    entry: {
+       app: './src/index.js',
+       print: './src/print.js'
+       app: './src/index.js'
```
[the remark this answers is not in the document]
bdwain: oops. i'm gonna remove this part of the diff anyway, but good catch

bdwain:
sure i can remove some of the diffs. I like your example, though I'd like to show how to use addDevServerEntrypoints because there are no examples of it anywhere else and it's not immediately clear.

bdwain:
is there anything left that is needed for this?

skipjack:
@bdwain yeah sorry for the delay. Ideally I'd like to push that content to an `api/webpack-dev-server` page at some point but since that doesn't exist at the moment, I think this is ok. I'm going to make some small tweaks and merge, thanks again for your work on this.

bdwain:
sweet thanks for merging it!

wesleymostien:
after every changes, server reloads the page in the browser with this setup ? Clues ?

skipjack:
@wesleymostien this isn't really a support channel but I think you may be running into [this issue](https://github.com/webpack/webpack-dev-server/issues/1021). There was a buggy release of the dev server a while back so you may have to upgrade.

Or, are you saying you tried the __dev-server.js__ example in the guide is throwing that error?

wesleymostien:
no, this happens when you forget the options 'host' and 'port'. So fixed it, by adding these to the config options. Although I'm still trying to find a way to mute console logs in the browser (without success)...

PR #1540, opened 2017-08-23
`inline: true` is `required` rather than `recommended` for HMR

jasonslyvia:
CMIW, but when `inline` is `false`, HMR will not work at all.

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -377,7 +377,7 @@ Usage via the CLI
 webpack-dev-server --inline=false
 ```
 
-T> Inline mode is recommended when using Hot Module Replacement.
+T> Inline mode is **required** when using Hot Module Replacement.
 
 
 ## `devServer.lazy` 🔑
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -377,7 +377,7 @@ Usage via the CLI
 webpack-dev-server --inline=false
 ```
 
-T> Inline mode is **required** when using Hot Module Replacement.
+T> Inline mode is recommended for Hot Module Replacement as it includes an HMR trigger from the websocket. Polling mode can be used as an alternative, but requires an additional entry point, `'webpack/hot/poll?1000'`.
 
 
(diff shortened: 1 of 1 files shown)


skipjack:
Hmmm, not sure on this one. Pinged @sokra and @SpaceK33z for review as they might know better. I haven't used the dev server without `inline: true` in a _very long time_.

sokra:
Inline mode includes the HMR trigger from WebSocket. You'll lose this when not using inline mode.

You could use HMR polling mode when using HMR without inline mode.

So technically it's not required.

jasonslyvia:
@sokra I fail to see how to enable `HMR polling mode` in [here](https://webpack.js.org/guides/hot-module-replacement/), [here](https://webpack.js.org/api/hot-module-replacement/) and [here](https://webpack.js.org/concepts/hot-module-replacement/).

sokra:
I guess it's not documented:

```
entry: [
  "webpack/hot/poll?1000",
  ...
]
```

skipjack:
@jasonslyvia I'd be happy to work with you on this. I think there are a few pages (i.e. the ones you linked to) that need to be updated.

jasonslyvia:
@sokra Thanks for the clarification.  I just wonder if everyone is fine with `inline` mode, is there any necessity to introduce `polling` mode? Given that webpack is known for plenty of configuration.

@skipjack Do you mean adding `polling` mode or just marking `inline` as required for HMR?

skipjack:
@jasonslyvia I mean note the `polling` mode via the entry point but also update the guide to include `inline: true` as that's a more sensible default.

jasonslyvia:
@skipjack I'm good with it, how can I help here?

skipjack:
If you have time and could make those changes that would be great. Just avoid saying "required" and instead maybe word it as suggested. I'll review and help get it merged once you finish.

jasonslyvia:
Sure, I'll try it soon.

skipjack:
@jasonslyvia updated this tip based on our discussion. Will merge later today as long as no one has issues or suggestions regarding this change.

PR #1541, opened 2017-08-23
Clarify interactions between output.libraryTarget and output.library

metatoaster:
@TheLarkInn as promised, I expanded on the entire `output.libraryTarget` section and checked a few things that kind of related to this.  However, I really don't like how when I had sectioned `output.libraryTarget` behaviors (i.e. the variable, object assignment or the formalized module definition systems) in a sub heading still ended up being on the sidebar which probably is a bad thing in terms of consistency. Though this is the first pass that attempted to correct the various major misgivings I had about the entire thing about that.  Please review carefully because I am by no means experienced at dealing with all the various intricacies of the various JavaScript module systems.


src/content/configuration/output.md
```diff
@@ -293,13 +293,13 @@ If using the [`output.library`](#output-library) option, the library name is aut
 
 `string` or `object` (since webpack 3.1.0; for `libraryTarget: "umd"`)
 
-Use `library`, and `libraryTarget` below, when writing a JavaScript library that should export values, which can be used by other code depending on it. Pass a string with the name of the library:
+Use `output.library`, and `output.libraryTarget` below, when writing a JavaScript library that should export values, which can be used by other code depending on it. Pass a string with the name of the library:
```
ChrisChinchilla: @metatoaster Here are you trying to say that you pass a string with the name of the library, and you add or replace that name if the library should export values? Or something else?

src/content/configuration/output.md
```diff
@@ -356,23 +356,50 @@ MySubModule.doSomething();
...
 
-Configure how the library will be exposed. Any one of the following options can be used.
+Configure how the library will be exposed. Any one of the following options can be used. Please note that this option works in conjunction with the value assigned to [`output.library`](#output-library). For the following examples, it is assumed that this value is configured as `MyLibrary`.
 
-The following options are supported:
+Note that `_entry_return_` in the example code below is the value returned by the entry point. In the bundle itself, it is the output of the function that is generated by webpack from the entry point.
+
+### Exposing the library via variable
+
+These options simply assign the return value of the entry point (e.g. whatever the entry point exported) to the name provided by `output.library` at whatever scope the bundle was included at.
```
ChrisChinchilla: @metatoaster Remove 'simply', these sorts of words should generally be avoided.

src/content/configuration/output.md
```diff
@@ -356,23 +356,50 @@ MySubModule.doSomething();
...
 
 `libraryTarget: "var"` - (default) When your library is loaded, the **return value of your entry point** will be assigned to a variable:
 
 ```javascript
 var MyLibrary = _entry_return_;
 
-// your users will use your library like:
+// your users, in a separate script, will use your library like:
```
ChrisChinchilla: @metatoaster I know it's not you're original wording, but whilst we're here :)

I would add 'like the following' to 'like'.

src/content/configuration/output.md
```diff
@@ -356,23 +356,50 @@ MySubModule.doSomething();
...
 MyLibrary.doSomething();
 ```
 
-W> Not specifying a `output.library` will cancel this `"var"` configuration.
+W> an empty `output.library` with this libraryType will result in no assignment being done.
```
ChrisChinchilla: @metatoaster I think you can remove 'being done'.

src/content/configuration/output.md
```diff
@@ -356,23 +356,50 @@ MySubModule.doSomething();
...
+
+
+`libraryTarget: "assign"` - Here webpack will blindly generate an implied global.
```
ChrisChinchilla: @metatoaster 'blindly' sounds a bit irresponsible, maybe 'automatically'? Or even remove the word completely if it's not really needed?

src/content/configuration/output.md
```diff
@@ -356,23 +356,50 @@ MySubModule.doSomething();
...
+
+W> an empty `output.library` with this libraryType will result in a broken output bundle.
+
 
+### Exposing the library via object assignment
 
-`libraryTarget: "this"` - When your library is loaded, the **return value of your entry point** will be assigned to this, the meaning of `this` is up to you:
+These options simply assign the return value of the entry point (e.g. whatever the entry point exported) to a specific object under the name defined by `output.library`.
+
+If `output.library` is not assigned a non-empty string, the default behavior is that all properties returned by the entry point be assigned to the object as defined for the particular `output.libraryTarget`, via the following code fragment:
```
ChrisChinchilla: @metatoaster Instead of 'be assigned', 'are assigned' might be better.

src/content/configuration/output.md
```diff
@@ -356,23 +356,50 @@ MySubModule.doSomething();
...
+
+```javascript
+(function(e, a) { for(var i in a) e[i] = a[i]; }(${output.libraryTarget}, _entry_return_)
+```
+
+W> Note that not setting a `output.library` will result in all properties returned by the entry point be blindly assigned to whatever object that was passed in.
```
ChrisChinchilla: @metatoaster Again the 'blindly' comment as before.

src/content/configuration/output.md
```diff
@@ -422,20 +453,14 @@ module.exports = _entry_return_;
 require("MyLibrary").doSomething();
 ```
 
-T> Wondering the difference between CommonJS and CommonJS2? Check [this](https://github.com/webpack/webpack/issues/1114) out (they are pretty much the same).
+Note that `output.library` is omitted, thus it is not required for this particular `output.libraryType`.
 
+T> Wondering the difference between CommonJS and CommonJS2? Check [this](https://github.com/webpack/webpack/issues/1114) out (while they are similar, there are some subtle differences that are not usually relevant in the context of webpack).
```
ChrisChinchilla: @metatoaster Better would be "Wondering what the difference between CommonJS and CommonJS2 is?

ChrisChinchilla: @metatoaster And maybe "Read [this issue](https://github.com/webpack/webpack/issues/1114)" instead of "Check [this](https://github.com/webpack/webpack/issues/1114) out"

src/content/configuration/output.md
```diff
@@ -446,14 +471,33 @@ output: {
...
-users will be able to call your library like so:
+The generated output will be defined with the name "MyLibrary", i.e.
+
+```javascript
+define("MyLibrary", [], function() {
+  // This module return value is what your entry chunk returns
+});
+```
+
+The bundle can simply be included as part of a script tag, and the bundle can be invoked like so:
```
ChrisChinchilla: @metatoaster Again, I recommend removing 'simply'.

src/content/configuration/output.md
metatoaster: This really could be elaborate more - I did not change much from the original text (see that the change was adding the `output.` prefix.  The context is that the explicit details as to what the values do are listed below that, so this fact should be made more clear.

metatoaster: Actually, I scrapped that entire sentence and example, it contributed nothing.

src/content/configuration/output.md
metatoaster: Again, copied verbatim from existing text, but while I am at it, as you said...

ChrisChinchilla:
@metatoaster Cool! From a docs side this looks good to me :)

metatoaster:
What about the new sections defined under the `output.libraryTarget` showing up on the same level as the navigation tree when rendered?  I am a bit worried that doesn't look too good, and I am not sure how to go about it while keeping these headings to group them by these categories to help with comprehension.  If this is fine then I supposed no harm done.

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
  window["MyLibrary"] = _entry_return_;
 
-//your users will use your library like:
+//your users will use your library like the following:
 window.MyLibrary.doSomething();
  ```
 
 
-`libraryTarget: "global"` - When your library is loaded, **the return value of your entry point** will be part `global` object.
+`libraryTarget: "global"` - When your library is loaded, **the return value of your entry point** will be part `global` object under the property named by `output.library`.
```
skipjack: I would rephrase this the same way:

> `libraryTarget: "global"` - The return value of your entry point will be assigned to the `global` object using the `output.library` value.

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
  global["MyLibrary"] = _entry_return_;
 
-//your users will use your library like:
+//your users will use your library like the following:
 global.MyLibrary.doSomething();
  ```
 
 
-`libraryTarget: "commonjs"` - When your library is loaded, **the return value of your entry point** will be part of the exports object. As the name implies, this is used in CommonJS environments:
+`libraryTarget: "commonjs"` - When your library is loaded, **the return value of your entry point** will be part of the `exports` object under the property named by `output.library`. As the name implies, this is used in CommonJS environments:
```
skipjack: Again (sorry I know this is going on forever):

> `libraryTarget: "commonjs"` - The __return value of your entry point__ will be assigned to the `exports` object using the `output.library` value. As the name implies, this is used in CommonJS environments.

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
 exports["MyLibrary"] = _entry_return_;
 
-//your users will use your library like:
+//your users will use your library like the following:
 require("MyLibrary").doSomething();
 ```
 
+### Exposing the library under formalized module definition systems
+
+These options will result in a bundle that comes with a more complete header to ensures compatibility with various module systems.  The `output.library` option will take on a different meaning under the following `output.libraryTarget` options.
```
skipjack: "to ensures" => "to ensure"

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
+
 
 `libraryTarget: "commonjs2"` - When your library is loaded, **the return value of your entry point** will be part of the exports object. As the name implies, this is used in CommonJS environments:
```
skipjack: I know this one wasn't even changed but we should update for consistency:

> `libraryTarget: "commonjs2"` - The **return value of your entry point** will be assigned to the `module.exports`. As the name implies, this is used in CommonJS environments.

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
 module.exports = _entry_return_;
 
-//your users will use your library like:
+//your users will use your library like the following:
 require("MyLibrary").doSomething();
 ```
 
-T> Wondering the difference between CommonJS and CommonJS2? Check [this](https://github.com/webpack/webpack/issues/1114) out (they are pretty much the same).
-
+Note that `output.library` is omitted, thus it is not required for this particular `output.libraryType`.
```
skipjack: Again unless I've missed something and `libraryType` is a configuration option, I think we should rephrase the last bit as "... this particular library type."

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
+T> Wondering the difference between CommonJS and CommonJS2 is? While they are similar, there are some subtle differences between them that are not usually relevant in the context of webpack. (For further details, please [read this issue](https://github.com/webpack/webpack/issues/1114).)
 
-Note that your entry chunk must be defined with the `define` property, if not, webpack will create the AMD module, but without dependencies. The output will be something like this:
 
-```javascript
-define([], function() {
-  // This module returns is what your entry chunk returns
-});
-```
+`libraryTarget: "amd"` - In this case webpack will make your library an AMD module.
```
skipjack: "In this case webpack will make your library an AMD module." => "This will expose your library as an AMD module."

src/content/configuration/output.md
```diff
@@ -446,14 +465,33 @@ output: {
...
 
+If `output.library` is undefined, the following is generated instead.
+
+```javascript
+define([], function() {
+  // This module returns is what your entry chunk returns
+});
+```
+
+This bundle will not work as expected, or not work at all (in the case of the almond loader) if loaded directly with a `<script>` tag. It will only work through RequireJS compatible asynchronous module loader through the actual path to that file, so in this case, the `output.path` and `output.filename` may become important for this particular setup if these are exposed directly on the server.
```
skipjack: "... through RequireJS compatible ..." => "... through a RequireJS compatible ..."

src/content/configuration/output.md
```diff
@@ -446,14 +465,33 @@ output: {
...
+
+
 `libraryTarget: "umd"` - This is a way for your library to work with all the module definitions (and where aren't modules at all). It will work with CommonJS, AMD and as global variable. Take a look at the [UMD Repository](https://github.com/umdjs/umd) to learn more.
```
skipjack: Maybe we can simplify to:

> `libraryTarget: "umd"` - This exposes your library under all the module definitions allowing it to work with CommonJS, AMD and as global variable. Take a look at the [UMD Repository](https://github.com/umdjs/umd) to learn more.

src/content/configuration/output.md
```diff
@@ -482,6 +520,31 @@ And finally the output is:
 });
 ```
 
+Note that omitting `library` will result in the assignment of all properties returned by the entry point be assigned directly to the root object, as documented under the [object assignment section](#exposing-the-library-via-object-assignment).  Example:
```
skipjack: This is pretty minor but note that we typically only use one space at the start of a new sentence. So `... assignment).  Example:` would be come `... assignment). Example:`.

src/content/configuration/output.md
```diff
@@ -498,13 +561,7 @@ output: {
...
 
 
-`libraryTarget: "assign"` - Here webpack will blindly generate an implied global.
-
-``` javascript
-MyLibrary = _entry_return_;
-```
-
-Be aware that if `MyLibrary` isn't defined earlier your library will be set in global scope.
+### Other libraryTargets
```
skipjack: I would simplify this title to just `Other Targets` the whole section is already in the context of `libraryTarget`. Also re your headers/sidebar question -- I think the other titles above should be simplified as well. This is a general problem we have across the site and while I do think we should try to keep titles more concise, we can also consider only showing `h2` level anchors in the sidebar.

src/content/configuration/output.md
```diff
@@ -293,13 +293,7 @@ If using the [`output.library`](#output-library) option, the library name is aut
...
 `string` or `object` (since webpack 3.1.0; for `libraryTarget: "umd"`)
 
-Use `library`, and `libraryTarget` below, when writing a JavaScript library that should export values, which can be used by other code depending on it. Pass a string with the name of the library:
-
-``` js
-library: "MyLibrary"
-```
-
-The name is used depending on the value of the [`output.libraryTarget`](#output-librarytarget) options.
+How the value of the `output.library` is used depends on the value of the [`output.libraryTarget`](#output-librarytarget) options; please refer to that section for the details.
```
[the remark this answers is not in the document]
metatoaster: Good point.  I still think it is worthwhile to rephrase this to make it more clear.

src/content/configuration/output.md
```diff
@@ -356,86 +350,111 @@ MySubModule.doSomething();
...
+
+`libraryTarget: "assign"` - Here webpack will generate an implied global.  This has the potential to reassign an existing value, use with caution.
+
+``` javascript
+MyLibrary = _entry_return_;
+```
+
+Be aware that if `MyLibrary` isn't defined earlier your library will be set in global scope.
+
+W> an empty `output.library` with this libraryType will result in a broken output bundle.
```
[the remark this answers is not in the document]
metatoaster: I think I meant `libraryTarget`.  Good catch, thank you.

Issue #1525, opened 2017-08-14
Dev - General Updates and Fixes


skipjack:
> Moving dev stuff from #1380 for more targeted work...

We either need to finish #980 -- which still needs some work, possibly on the [antwar][1] level -- or port the site to a more stable build tool like [phenomic][2] or [gatsby][3]. Porting the site would also offer a few things, like full HMR, that #980 doesn't address but will require some serious thought. Either way these are the infrastructure issues we need to solve:

- [x] Resolve `<title>` issues (e.g. #720). 
- [x] Display index pages correctly (#1085).
- [x] Bail if YAML/Markdown parsing fails (#1070).
- [x] Easier specification of custom layouts per guide (e.g. #1367).
- [ ] Consider a better hosting/deployment setup like Netlify (e.g. #845).
- [x] Mark releases (synced with main package) and archive them (#176, #1017).
- [x] Consider landing page image load improvements (#1548).

__UPDATE: Fixed on The `rebuild` Branch__

- [x] Interactive bundles contain duplicate code (#877).
- [x] Migrate away from _marked_ and clean up process (webpack-contrib/webpack-defaults#73).
- [x] Better HMR, faster builds, and quicker build tests. 
- [x] Continue component/utility modularization and simplification.

The last one refers to outsourcing much of the custom dev-related stuff to more modular packages as we've done with the [voting app][5] and [extracted banner][6] (though neither are actually being consumed yet).

__Other Dev-Related Issues__

- #1519 - Synchronized routing and `/content` structure with less configuration.
- #1400 - Turn the site into a [PWA][4] (sync w @TheLarkInn).
- #469 - Allow illustrations via Mermaid.

cc @montogeek @probablyup @geoffdavis92 @MoOx @bebraw @hulkish. If I've missed anyone please let me know, but I've tried to include everyone who has shown interest in site dev or has previously worked on site dev regularly.

[1]: https://github.com/antwarjs/antwar
[2]: https://phenomic.io/
[3]: https://github.com/gatsbyjs/gatsby
[4]: https://developers.google.com/web/progressive-web-apps/
[5]: https://github.com/webpack-contrib/voting-app
[6]: https://github.com/skipjack/react-banner

skipjack:
> From #1526...

[...]

Issue #1400, opened 2017-07-08
Dev - Convert to PWA (checklist)


hulkish:
- [x] Refactor naming convention for react components. ( #1490, #1538 )
- [x] Refactor structure of app specific code to live in `/src` dir. ( #1490, #1538 )
- [x] Eliminate FOIT (_Flash Of Invisible Text_). ( #1536 )
- [x] Consider inlining [svg icons][1] instead of including them as a font. ( #4192 )
- [ ] Add Jest & create [snapshot tests][2] for react components.
- [x] Add service worker.  ( #3000 )
- [x] Score at least a 92 on [lighthouse][4].
- [x] Create Web App Manifest (using [webpack-pwa-manifest][5] maybe?).
- [ ] Site appropriately informs the user when they're offline.
- [x] Site use [cache first networking][6].  ( #3000 )
- [x] Full offline support for all docs pages.  ( #3000 )
- [x] Upgrade to use latest webpack.
- [ ] Discuss: Inferno instead of Preact?

## Score at least a 92 on lighthouse

- Test on [lighthouse][4] or now in the `Audits` section of chrome devtools.
- Should follow [these guidelines][7].

## Create Web App Manifest

As described here https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/

## Site appropriately informs the user when they're offline
- Use the [Network Information API][8] to show the user an indication when they're offline.
- I'm thinking [offline.js][9] for this.

[1]: https://github.com/webpack/webpack.js.org/tree/master/src/styles/icons
[2]: https://facebook.github.io/jest/docs/snapshot-testing.html
[3]: https://workboxjs.org/reference-docs/latest/module-workbox-webpack-plugin.html
[4]: https://developers.google.com/web/tools/lighthouse/
[5]: https://www.npmjs.com/package/webpack-pwa-manifest
[6]: https://developers.google.com/web/progressive-web-apps/checklist#site-uses-cache-first-networking
[7]: https://developers.google.com/web/progressive-web-apps/checklist
[8]: https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API
[9]: https://github.com/hubspot/offline

Dean-Coakley:
Hoping to have a PR ready for Components naming convention refactor within 24 hours. (Depending on how my testing goes)

Should WebP be used for image optimization? (Whenever that is to be done) How would that work when new images are added when a new sponsor is added, etc?
[...]

PR #1557, opened 2017-08-29
Observe fonts to prioritize content. Handle repeated visits with sessionStorage

jeremenichelli:
_A note before starting with what's inside this PR. I'm not completely happy with the code, basically because **antwar** doesn't help a lot with above the fold styles and javascript and font strategy technics, so if this PR doesn't go further I'm totally fine and I will consider doing this again if we migrate to other content managment generator or strategy._

What's happening here? I'm replicating a technic I've used in other static sites, you can read more about it here: https://jeremenichelli.github.io/2016/05/font-loading-strategy-static-generated-sites/

I'm observing the font change when the site _mounts_ and in case sessionStorage indicates the font is already cached the className is changed in advanced overhead.

One of my concerns is that I don't know how **antwar** is gonna handle this _global code_ related to the site, because my choice would have been to create a new entry with font face observing code only being loaded when needed, something I couldn't do.

Locally, **look at the screenshot on bottom left from dev tools**, it really works since instead of this text-less page on first paint:
captura de pantalla 2017-08-29 a la s 12 29 38

User might be able to get this system font text screen:
captura de pantalla 2017-08-29 a la s 16 36 12

And when fonts get available they will switch to this:
captura de pantalla 2017-08-29 a la s 16 36 26

I will wait to see what build says, but as I mentioned before I don't feel 100% confident about this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "yaml-frontmatter-loader": "0.0.3"
   },
   "dependencies": {
+    "fontfaceobserver": "^2.0.13",
     "preact": "^6.2.1",
     "preact-compat": "3.11.0",
     "prop-types": "^15.5.8",
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { Component } from 'react';
 import Interactive from 'antwar-interactive';
 import { GoogleAnalytics } from 'antwar-helpers';
+import FontFaceObserver from 'fontfaceobserver';
 import NotificationBar from '../NotificationBar/NotificationBar';
 import Navigation from '../Navigation/Navigation';
 import Footer from '../Footer/Footer';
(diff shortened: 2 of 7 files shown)


skipjack:
@jeremenichelli yeah I see where you're going with this and I like that we want to do more in terms of optimizing the site performance/experience. However there's also something to be said about complicating the codebase (which already pretty convoluted imo). This is kind of been my feeling in general re the dev work discussed in #1525 and #1400 -- meaning I'd much rather do things to clean up the codebase like finish #980 and hit any cleanup-like tasks mentioned in those issues (e.g. the changes in #1490) before we start adding a new things. 

The blocker for me so far has been that at the end of the day the site's content should take priority over everything else imo. No matter how performant we get this site, if the content isn't well thought out and better organized than we still are left with a bad site (all be it a fast bad site 😆 ). For this reason I've spent the most of this year trying to address the large backlog of content-related issues we had open.

The good news is that this work is nearing completion and starting to be much more manageable with the recent wave of help that we've gotten. Now we're down to ~20 issues and I'm hoping we can keep it that way. With the content more under control and people continuing to review and improve it, we should be able to get the infrastructure cleaned up and then start making some improvements.

By the way, I'm not saying we have to hold off on all infrastructure improvements, just ones like this where it feels a bit dirty and we could probably come up with something cleaner once the repos in a better place.

jeremenichelli:
All understood and I have the same opinion about the priorities and original mission of this site.

I will close this pull request since I myself don't feel confident about getting this code to master. But it might be a good place to start thinking on a change for the future. I love React, I think is a powerful tool, but using it for _everything_ like static sites doesn't feel super right in my opinion. For example here, when you are trying to accomplish some static content first approach.

It might be good to take a look at Next if we want to stick with React for example, or take a look at some static site generators, there are a lot out there working well on production. Also to simplify the DX, that for a static site is really complex in my opinion.

That way we might be able to make _content_ first citizen in this repo.

skipjack:
> I will close this pull request since I myself don't feel confident about getting this code to master. But it might be a good place to start thinking on a change for the future.

Definitely, feel free to leave the branch around.

> I love React, I think is a powerful tool, but using it for everything like static sites doesn't feel super right in my opinion. For example here, when you are trying to accomplish some static content first approach.
>
> It might be good to take a look at Next if we want to stick with React for example, or take a look at some static site generators, there are a lot out there working well on production. Also to simplify the DX, that for a static site is really complex in my opinion.

So first of all, the infrastructure of this site __definitely__ needs work and _developer experience_ (I'm assuming that's what you mean by DX) really needs to be simplified.

I think React can be a good static site generator but the setup that we have now is definitely not ideal imo. I think the biggest point of confusion has been the `Interactive` components, which are used because antwar doesn't do true SSR (server-side rendering) while other generators like Phenomic and Gatsby do. This makes it easier to bind events, import css, and prevents some of the duplicate bundle issues we've been running into.

To antwar's credit, @bebraw has given us a lot of freedom in terms of updating it to cater to a nice solution on this site. The problem has been that @bebraw (nor I) has had the time to make certain updates to the tool and even just managing content and development issues on this site is tough. There's also certain changes (like full on SSR) that @bebraw doesn't want to add to antwar as it will complicate what is now a fairly simple codebase.

I, and others I think, have also played around with the idea of a port. I do think _Next.js_ might be a good option and potentially _Phenomic_ or _Gatsby_ could be as well. I did do a bit of testing on right after the v1 releases of _Gatsby_ and _Phenomic_ and had a lot of trouble with the provided examples for both.

I think we're close to getting #980 merged which will at least fix some things and put us on the latest version of webpack, antwar, and a bunch of other packages. From there, I think we can more fairly evaluate a port and talk about the pros and cons of each. There's just a lot of moving pieces with this site so I just want to get a good plan together before we dive into anything. Rushing a port could leave us in an even worse position than we are now.

> That way we might be able to make content first citizen in this repo.

This is really what @bebraw and I have always discussed trying to get to. If we keep our build process and functionality fairly simple and externalize as much as possible to other packages then we can keep the focus on well organized and well written content. Which, at the end of the day, is what the users of the site need from us.

jeremenichelli:
@skipjack yes, I agree with basically all stated. Just to be clear I'm not against using **antwar**, I just thing there are better options for this specific type of site.

Of course any port would a major change which should be widely discussed by all members. I'll be around to collaborate with team on both topics, content and infrastructure 🎉

skipjack:
> Just to be clear I'm not against using antwar, I just think there are better options for this specific type of site.

Yeah I understand. We just need to weigh our options, list out the pros and cons, and pick something that makes the most sense for this site.

> I'll be around to collaborate with team on both topics, content and infrastructure.

👍

PR #1561, opened 2017-08-31
Updating table spacing

rishantagarwal:
Currently, table spacing is not proper and needs to be scrolled to view whole content. I removed some extra space to make it completely visible.

### DIFF
--- a/src/content/api/plugins/compiler.md
+++ b/src/content/api/plugins/compiler.md
@@ -92,32 +92,32 @@ webpack([config1, config2], (err, stats) => {
 
 This a reference guide to all the event hooks exposed by the `Compiler`.
 
-Event name                    | Reason                                  | Params                    | Type
------------------------------ | --------------------------------------- | ------------------------- | ----------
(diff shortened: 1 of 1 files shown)


pierreneter:
No relation. Tabs will be removed during HTML rendering.

capture

skipjack:
@rishantagarwal actually I think @pierreneter makes a good point. I'm not 100% sure your change actually makes a difference in the table... could you share some screenshots?

@pierreneter did you test this, are you sure it the spacing change doesn't make a difference? I approved as it doesn't seem like it'd hurt anything but I was wondering the same thing. I feel like whitespace is typically trimmed and the column's width is just determined by the longest trimmed cell.

skipjack:
@rishantagarwal closing, but please feel free to re-open and share some screenshots if @pierreneter and I are mistaken.

rishantagarwal:
@pierreneter @skipjack makes sense, my bad ! I am so sorry .

PR #1562, opened 2017-08-31
fix(hot-module-replacement): Fix and improve HMR management API

sokra:
In webpack 2 `check` and `apply` return Promises.

There are more options.

PR #1564, opened 2017-08-31

AlenQi:
add the '`',abide by the grammar


pierreneter:
It will be fixed with d0791ac of #1562. And this is a PR after the PR, is this a coincidence?

pierreneter:
It will be fixed with d0791ac of #1562. And here is a PR after that PR, is this a coincidence?

skipjack:
@pierreneter good catch, I just merged the other so let's close this. Thanks anyway for your contribution @AlenQi!

PR #1565, opened 2017-08-31
Add more information about optimization bailouts

pspeter3:
Attempt to transfer content from the blog post to the official docs. I want to follow up with a PR for better bailout reasons. The hope is that the reasons will match those listed in the table.


src/content/plugins/module-concatenation-plugin.md
```diff
@@ -16,3 +16,78 @@ new webpack.optimize.ModuleConcatenationPlugin()
 ```
 
 > Scope Hoisting is specifically a feature made possible by ECMAScript Module syntax. Because of this webpack may fallback to normal bundling based on what kind of modules you are using, and [other conditions](https://medium.com/webpack/webpack-freelancing-log-book-week-5-7-4764be3266f5).
+
+# Optimization Bailouts
```
skipjack: Can you change this to an `h2` (i.e. `##`)? The beginning of every page starts with an `h1` with the page title and, aside from keeping the design consistent, it's not good to switch back and forth for SEO reasons. Also we typically drop two new lines between `h2` sections.

src/content/plugins/module-concatenation-plugin.md
```diff
@@ -16,3 +16,78 @@ new webpack.optimize.ModuleConcatenationPlugin()
...
+| Imported From Other Chunk                     | Root    |
+| Imported By Multiple Other Module Groups      | Root    |
+| Imported With `import()`                      | Root    |
+| Affected By `ProvidePlugin` Or Using `module` | Prevent |
+| HMR Accepted                                  | Root    |
+| Using `eval()`                                | Prevent |
+| In Multiple Chunks                            | Prevent |
+| `export * from "cjs-module"`                  | Prevent |
+
+## Module Grouping Algorithm
```
skipjack: Same here, just add another new line before this header.

src/content/plugins/module-concatenation-plugin.md
```diff
@@ -16,3 +16,78 @@ new webpack.optimize.ModuleConcatenationPlugin()
 ```
 
 > Scope Hoisting is specifically a feature made possible by ECMAScript Module syntax. Because of this webpack may fallback to normal bundling based on what kind of modules you are using, and [other conditions](https://medium.com/webpack/webpack-freelancing-log-book-week-5-7-4764be3266f5).
+
+# Optimization Bailouts
+
+As the article explains, there Webpack attempts to achieve partial scope hoisting. Webpack attempts to merge modules into a single scope but cannot do so in every case. If webpack cannot merge a module, the two alternatives are Prevent and Root. Prevent means the module must be in its own scope. Root means the module begins a new module group. The following table explains the conditions:
```
skipjack: As per the [branding guidelines](https://github.com/webpack/media), "Webpack" should be always be written in lowercase.

src/content/plugins/module-concatenation-plugin.md
pspeter3: Sure!

src/content/plugins/module-concatenation-plugin.md
pspeter3: Thanks!

src/content/plugins/module-concatenation-plugin.md
pspeter3: Thanks!

skipjack:
@pspeter3 made some minor formatting, grammar and spelling corrections. Will merge as soon as the build passes. Thanks again for your work on this!

PR #1567, opened 2017-09-02
Wrong devtool type at Production>Source Mapping guide

diegofaria:
### Problem

At the [Production - Source Mapping guide](https://webpack.js.org/guides/production/#source-mapping), there's a suggestion to use a *devtool style* not supposed for production use, as the [Devtool guide explains](https://webpack.js.org/configuration/devtool/#devtool).

### Solution

Suggest a *devtool style* suited for production, based on the [Devtool guide](https://webpack.js.org/configuration/devtool/#devtool).

### DIFF
--- a/src/content/guides/production.md
+++ b/src/content/guides/production.md
@@ -155,7 +155,7 @@ If you decide to try another, just make sure your new choice also drops dead cod
 
 ## Source Mapping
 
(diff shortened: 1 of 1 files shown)


skipjack:
Good catch, @sokra does `source-map` look like a good recommendation to you?

PR #1570, opened 2017-09-04
Make sure images are responsive

mitogh:
Prevent stretch images with defined sizes in smaller breakpoints.

**Before**
screen shot 2017-09-03 at 10 55 15 pm

**After**
screen shot 2017-09-03 at 10 55 11 pm

### DIFF
--- a/src/styles/markdown.scss
+++ b/src/styles/markdown.scss
@@ -242,6 +242,7 @@
 
   img {
     max-width: 100%;
+    height: auto;
   }
 
   b, strong {


skipjack:
Ah @pierreneter we both missed something here. That `height: auto` caused the supporter images to be enlarged. Just pushed a fix, should be deployed shortly.

pierreneter:
@skipjack hey, I don't see any problem with the splash page. Can you explain it to me? I think the height fixed on the home page is entirely reasonable and it does not cause any problems with `responsive`.

skipjack:
Yeah by the time you saw it the fix was probably deployed (unless you tested locally and checked out to before my fix). The problem was that `height: auto` had a higher specificity than the support image `height` declarations. Most of the support images (particularly the sponsors I think) we're rendering to the actual height of the image. Give me a sec, I can check out and share a screenshot.

skipjack:
The images:

image

The issue:

image

mitogh:
ha! Sorry guys :|

skipjack:
@mitogh no worries we should've caught this. As soon as I saw it, I was like "Woah! Those are some big sponsors!"

PR #1573, opened 2017-09-04
Update resolve.md

luankevinferreira:
Add an example of alias usage with images and HTML files.


skipjack:
@luankevinferreira it is true that you can do that but not by default. The block of code you added also means the reader would also have to include the `html-loader` in their `module` configuration. The `resolve` options, including `alias`, will work for anything webpack resolves, e.g. relative `url(...)`s with the `css-loader` as well. But, by default, only `import` and `require` statements are handled.

I'm going to close but if you wanted to add something more generic about `resolve.alias` so people know it can be used across the board when you are loading and resolving other module types, then I'd be happy to re-open.

PR #1588, opened 2017-09-12
fix sponsors

sokra:
* fix sponsors for opencollective change
* add additional sponsors
* allow to merge additional sponsors to existing sponsors
* limit width of logos to 3 x height to disallow very wide logos
* format money values with commas


src/components/Support/Support.jsx
```diff
@@ -19,17 +27,37 @@ const ranks = {
...
   render() {
-    let { rank, type } = this.props;
-    let supporters = require(`./support-${type}.json`);
+    let { rank } = this.props;
+    let supporters = [
+      ...GoldSponsors,
+      ...SilverSponsors,
+      ...Sponsors,
+      ...Backers,
+    ];
```
skipjack: Since we end up combining anything anyway, maybe we can simplify the fetch script to put everything in one file in the first place, making some of the code here unnecessary. Anyway, I was in a bit of rush right now so I'll review the code more in depth later on and see if there's anything we can simplify. Let me know what you think about the one file idea.

PR #1589, opened 2017-09-12
Update tree-shaking.md

MijaelWatts:
As a conclusion, I think it's a good idea to show case a high-level explanation of why it's called tree shaking. Once the user reads the guide and ends up with a high-level explanation the word tree shaking will stick to the user's head.


src/content/guides/tree-shaking.md
```diff
@@ -146,4 +147,6 @@ So, what we've learned is that in order to take advantage of _tree shaking_, you
 - Use ES2015 module syntax (i.e. `import` and `export`).
 - Include a minifier that supports dead code removal (e.g. the `UglifyJSPlugin`).
 
+You can imagine your app as a tree. The code you actually use are the green and alive leafs of your tree. The dead-code are the brown and dead leafs of your tree consumed by the autumn. In order to get rid of the dead leafs, you have to shake your tree, so that all the dead leafs fall. Tha's why it's called  _tree shaking_
```
skipjack: I like this addition but there are some grammar and spelling issues. I would rephrase as such:

> You can imagine your application as a tree. The source code and libraries you actually use represent the green, living leaves of the tree. Dead code represents the brown, dead leaves of the tree that are consumed by autumn. In order to get rid of the dead leaves, you have to _shake the tree_, causing them fall.

MijaelWatts: Cool... I already updated the paragraph, thanks for the rephrasing.

PR #1594, opened 2017-09-13
update pfx and passphrase information

navarrorc:
Spent several hours trying to figure out why it didn't work, realized that the documentation was not accurately showing the correct information. I reviewed https://nodejs.org/api/https.html and figured out the correct way to assign these two values. I hope that saves others hours of frustration. Thanks.


skipjack:
@shellscape can you review? Happy to merge but I'd like to confirm with someone who is more familiar with advanced `webpack-dev-server` usage.

shellscape:
You can specify the pfx and passphrase options like that, but it's not required when `options.https` is set to `true`. You can see this in action here https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L360-L371. So I'm really not sure how we should document this. Should we have the docs say that `options.https` should be true, or should we encourage people to specify an object for that parameter?

navarrorc:
@shellscape Thanks for clarifying, I understand now. However, setting `pfx: '/path/to/file.pfx'` doesn't work in my environment, I have to use `fs` as follows: `pfx: fs.readFileSync('/path/to/file.pfx')`. Can you verify that this is correct? 

My options are:
```
...
https: true,
pfx: fs.readFileSync('/path/to/file.pfx'),
pfxPassphrase: 'passphrase'
```

shellscape:
@navarrorc I'd say that's correct. if we look at the CLI for webpack-dev-server, it's doing the same: https://github.com/webpack/webpack-dev-server/blob/master/bin/webpack-dev-server.js#L315

So for the CLI, just providing the relative path should be enough. For using the API directly, reading the file contents is needed. This is one of many discrepancies between the API and CLI that have crept into the project over the years.

skipjack:
Closing, as I actually think #1595 resolves this issue quite nicely. @navarrorc @shellscape please correct me if I'm wrong or feel free to submit a follow up pr with more to elaborate.

PR #1595, opened 2017-09-15
Update dev-server.md

cyberram:
Documentation updated based on the answer here.
https://stackoverflow.com/questions/37026880/p12-certificate-not-enough-data-error


src/content/configuration/dev-server.md
```diff
@@ -470,7 +470,7 @@ overlay: {
 
 `string`
 
-Path to a SSL pfx file.
+Path to a SSL pfx file while used via CLI. In options, should be the bytestream of the PFX file.
```
ChrisChinchilla: @cyberram Just to make this slightly clearer, I would suggest changing to…

"When used via the CLI, a path to an SSL .pfx file. If used in options, it should be the bytestream of the. pfx file"

Yup, SSL is one of those odd cases where it is 'an', not 'a' :)

skipjack:
Related to #1594

src/content/configuration/dev-server.md
cyberram: Wow. Thanks for enlightening me @ChrisChinchilla

PR #1608, opened 2017-09-27
devserver: add before, after hook info

shellscape:
Added info for newly added `before` and `after` hooks, and deprecation of `setup` in webpack-dev-server


skipjack:
@shellscape it looks like you may have committed from a separate account which is why the CLA isn't signed (unless you just haven't signed it on your @shellscape account yet). One thing you can do to resolve this is rebase the branch using your main account and then force update the remote copy -- this will recommit everything under your main account.

You can also try to add that separate email address under your main account.

shellscape:
Dangit. Split accounts for work and all :) I'll rewrite the info for that commit. Fix should be up in a minute

shellscape:
@skipjack CLA bot caught up and it looks good. this is documentation for a release that was published yesterday, so we'd love to get this up as soon as we can.

PR #1613, opened 2017-10-01
I think here is native event, not react's one

yanlee26:
waiting


src/content/guides/caching.md
```diff
@@ -358,8 +358,8 @@ __src/index.js__
 
     // Lodash, now imported by this script
     element.innerHTML = _.join(['Hello', 'webpack'], ' ');
--   element.onClick = Print.bind(null, 'Hello webpack!');
-+   // element.onClick = Print.bind(null, 'Hello webpack!');
+-   element.onclick = Print.bind(null, 'Hello webpack!');
```
asapach: then you should probably update line 258 as well

PR #1622, opened 2017-10-09
fix(sponsors): opencollective api changed, add validation

sokra:
opencollective api changed again...

I also added a validation this time so build will fail instead of leaving an empty sponsors list.

### DIFF
--- a/src/components/Support/Support.jsx
+++ b/src/components/Support/Support.jsx
@@ -1,10 +1,22 @@
 import React from 'react';
-import GoldSponsors from './support-goldsponsors.json';
-import SilverSponsors from './support-silversponsors.json';
-import Sponsors from './support-sponsors.json';
 import Backers from './support-backers.json';
 import Additional from './support-additional.js';
 import './Support.scss';
+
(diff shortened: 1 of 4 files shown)


skipjack:
@sokra running locally -- for the most part it works well and I like how the code's been simplified. I am seeing some issues with the images though...

Some images are failing because they're interpreted as local paths:

image

One of the failures is their placeholder image which means a lot show up blank:

image

It looks like those all need to be prepended with `https://opencollective.com`. I'll see if I can add a check but this should likely be added on their end, I would guess.

skipjack:
Ok that fix (`.startsWith('/') ...`) would actually break one of our local sponsor images so I think I'll merge and reply to that thread to note that the URLs should be fully qualified. Also, I'm not sure about falling back to a `<span>` in the backers section though I do understand that approach for the core sponsors to make sure their names are displayed no matter what.

Anyway, for now I think I'll merge since the whole section is broken on the site. I'll ping that email thread with a note about the relative URLs and maybe I'll do a follow up PR to clean up the design for missing avatars at some point.

PR #1624, opened 2017-10-10
Update devServer doc —-host to avoid using IP 0.0.0.0

hanxue:
Since [v2.4.3](https://github.com/webpack/webpack-dev-server/releases/tag/v2.4.3)
webpack-dev-server will disallow access from IP address that does not
match `Host` header.

There are many users who are affected, see https://github.com/webpack/webpack-dev-server/issues/882. Update documentation for
clarity

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -300,14 +300,18 @@ For more options and information, see the [connect-history-api-fallback](https:/
 Specify a host to use. By default this is `localhost`. If you want your server to be accessible externally, specify it like this:
 
 ```js
-host: "0.0.0.0"
+host: "192.168.1.2"
 ```
 
 Usage via the CLI
 
 ```bash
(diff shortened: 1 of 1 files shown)


hanxue:
@shellscape Are we on agreement that `--host 0.0.0.0` is a edge case? Isn't the normal case to specify an internal IP address, e.g. `--host 10.10.0.18`?

My concern is a new user reading the documentation and setting the value to `--host 0.0.0.0`  without setting `disableHostCheck` and expecting to be able to access the dev-server output on another machine will encounter an error.

shellscape:
@hanxue based on the discussion in that issue, I'm confident that it is an edge case that appears under only very specific conditions. If you wanted to add a note/caveat that users _may_ have to also include `disableHostCheck` in their config, I think that would be acceptable.

hanxue:
OK. I would rather defer to your opinion as a maintainer 😄
Appreciate your response and explanation.

PR #1612, opened 2017-09-30
Info Architecture

skipjack:
Tighten up page organization and sorting as discussed in #1519. This PR should resolve the following items:

- [x] Sort all existing pages in an intuitive way (already done for guides, see #1258).
- [x] _Support_, _Development_, and maybe sponsors/backers => _Contribute_ (#1365).
- [x] _CLI_ nav link should be removed (it will instead just be found under _API_).
- [x] The articles should then be organized via nested directories or a `group` field.
- [ ] Top level pages should use a wider layout (e.g. #1367).

The last one I may leave out as the stranded pages are a bit more joined together after #980. Hopefully once this is done, the content will be more usable, the structure more intuitive, and new features like `group` available for contributors and package maintainers. One thing we need to remember prior to merging this is:

- [x] Add redirects for all broken routes.

This should put us in a good place to continue pushing some of the other dev cleanup discussed in #1525 forward, however it would be great if some folks could help with the remaining content related issues (most of which are related to improving or fixing issues in the _API_ section).

Resolves #1519

PR #1626, opened 2017-10-10
docs(config): style update to webkit scroll bars

rouzbeh84:
I tried pulling this against `refactor-info-architecture` but couldn't since changes didn't effect the branch. All this does is add a webkit scrollbar style to hide it as mentioned in [this comment](https://github.com/webpack/webpack.js.org/pull/1612#issuecomment-334996066) about the proximity of content to scroll bars. 

**_this should only get looked at after #1612 is merged in._**

> **_NOTE:_**
> this does not fix it for FF/IE as those browsers do not support this behavior. an alternative would be to add padding to both `.sidebar__inner` and `.page__content`


rouzbeh84:
chrome

firefox

PR #1629, opened 2017-10-11
Update production.md

kelset:
Modifying the doc accordingly to [this tweetstorm](https://twitter.com/TheLarkInn/status/917524723579338752) where Dan Abramov & Sean Larkin kindly explained this to me.

### DIFF
--- a/src/content/guides/production.md
+++ b/src/content/guides/production.md
@@ -13,6 +13,7 @@ contributors:
   - redian
   - skipjack
   - xgqfrms
+  - kelset
 ---
 
 In this guide we'll dive into some of the best practices and utilities for building a production site or application.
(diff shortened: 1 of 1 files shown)


src/content/guides/production.md
```diff
@@ -175,6 +176,7 @@ __webpack.prod.js__
   })
 ```
 
+Note: _We have not used `inline-source-map` because it's not optimized for production and would generate a giant bundle, and so reduce the overall performances. Same goes for `devtool`s that contain `eval` in the name._
```
skipjack: Hmmm I would use our custom tip syntax and tweak the wording a bit from what was used in the tweet, e.g.:

``` md
T> Avoid `inline-***` and `eval-***` use in production as they can increase bundle size and reduce the overall performance.
```

kelset: Got it :) I'll update it now

kelset:
Updated

PR #1647, opened 2017-10-16
Add devServer.index

psi-4ward:
Add devServer.index Option to config documentation

See https://github.com/webpack/webpack-dev-server/blob/1af8f0e508e08b4e8f036311ae6c72b77f77600f/lib/optionsSchema.json#L318

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -374,6 +374,16 @@ To pass your own certificate via the CLI use the following options
 webpack-dev-server --https --key /path/to/server.key --cert /path/to/server.crt --cacert /path/to/ca.pem
 ```
 
+## `devServer.index` - API only
+
+`string`
+
+The filename that is considered the index file.
+
+```javascript
+index: 'index.htm'
+```
+
 
 ## `devServer.info` - CLI only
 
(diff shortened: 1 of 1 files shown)


shellscape:
@skipjack the change looks good, but I leave the `* only` flags up to you. At the moment we have a lot of disconnect between what options the CLI will allow you to set versus what options are available via the API. to that end, we could _really_ use a small icon set that we could use to indicate via which route an option is available. it could be argued that it's apparent based on the code block that accompanies an option (bash/console block for CLI, javascript block for API) but users don't always pick up on that. It's worth noting that a few options are marked with "CLI Only" as well.

psi-4ward:
Yes it's only a prop for node-api, ie 
```javascript
// ...
  devServer: {
    index: 'index.htm',
    // ...
```
but there is no corresponding CLI argument (or at least it's not documented) so I added this hint as counterpart to `CLI only`. Let's remove this cause `proxy` and others are API-only commands as well and do not have this.

skipjack:
@psi-4ward thanks for updating.

> At the moment we have a lot of disconnect between what options the CLI will allow you to set versus what options are available via the API. to that end, we could really use a small icon set that we could use to indicate via which route an option is available. it could be argued that it's apparent based on the code block that accompanies an option (bash/console block for CLI, javascript block for API) but users don't always pick up on that. It's worth noting that a few options are marked with "CLI Only" as well.

@shellscape yeah I see your point though instead of an icon set I think it might be better to just separate the content into two or three pages -- CLI, Node API, and Configuration. We'll probably have to give it a bit of thought seeing as there's a lot of overlap. Happy to discuss with you more offline, the _API_ section of the site is really just starting to take shape but it might be a good place for some of this content.

PR #1648, opened 2017-10-17
Fix wrong path to bundled result js file

rruimor:
Correct the path to the bundled js file in the index.html as stated in the guide.


rruimor:
Not need, sorry!

PR #1649, opened 2017-10-17
docs(guides): update hot-module-replacement.md

### LINKED ISSUE #1623: Guides - Document NamedModulesPlugin in HMR
Hi!

If I set up HMR by following the steps here:
https://webpack.js.org/plugins/hot-module-replacement-plugin/
https://webpack.js.org/guides/hot-module-replacement/

The resultant console output includes (seen both locally and in the guide example; [source](https://github.com/webpack/webpack/blob/v3.6.0/hot/log-apply-result.js#L36)):
```
main.js:4330 [HMR] Consider using the NamedModulesPlugin for module names.
```

Currently there doesn't appear to be any other documentation of this other than a brief mention on:
https://webpack.js.org/plugins/named-modules-plugin/
...which isn't linked at all from the HMR pages.

I think it would be  good to:
* mention the suggested use of `NamedModulesPlugin` on both of the HMR plugin and guide pages above
* specify whether the order of the plugins matters (the test-case [here](https://github.com/webpack/webpack/blob/2475c6a4d6c8306c42ee37e20c76e6e33bae19ec/test/HotTestCases.test.js#L66-L67) uses one order, and [this blog guide](http://remarkablemark.org/blog/2017/05/07/hot-module-replacement/) another, so perhaps it doesn't matter, but would be good to confirm)

Many thanks :-)

skipjack:
Add the `NamedModulesPlugin` for easier debugging as the relative paths
to each updated dependency will be shown in the patch messages.

Resolves #1623

### DIFF
--- a/src/content/guides/hot-module-replacement.md
+++ b/src/content/guides/hot-module-replacement.md
@@ -59,6 +59,7 @@ __webpack.config.js__
       new HtmlWebpackPlugin({
         title: 'Hot Module Replacement'
       }),
++     new Webpack.NamedModulesPlugin(),
 +     new webpack.HotModuleReplacementPlugin()
     ],
     output: {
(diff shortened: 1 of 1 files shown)


edmorley:
Many thanks :-)

PR #1663, opened 2017-10-20
Update tapable.md

jakewies:
Fixing some small (teeny tiny) errors in the docs.

### DIFF
--- a/src/content/api/tapable.md
+++ b/src/content/api/tapable.md
@@ -8,13 +8,13 @@ contributors:
   - e-cloud
 ---
 
-[Tapable](https://github.com/webpack/tapable) is small library that allows you to add and apply plugins to a javascript module. It can be inherited or mixed in to other modules. It is similar to NodeJS's `EventEmitter` class, focusing on custom event emission and manipulation. However, in addition to this, `Tapable` allows you to have access to the "emittee" or "producer" of the event through callbacks arguments.
+[Tapable](https://github.com/webpack/tapable) a is small library that allows you to add and apply plugins to a javascript module. It can be inherited or mixed in to other modules. It is similar to NodeJS's `EventEmitter` class, focusing on custom event emission and manipulation. However, in addition to this, `Tapable` allows you to have access to the "emittee" or "producer" of the event through callbacks arguments.
 
 `Tapable` has four groups of member functions:
 
-- `plugin(name:string, handler:function)`: This allows a custom plugin to register into a **Tapable instance**'s event. This acts as the same as `on()` of `EventEmitter`, for registering a handler/listener to do something when the signal/event happens.
+- `plugin(name:string, handler:function)`: This allows a custom plugin to register into a **Tapable instance**'s event. This acts similar to `on()` of `EventEmitter`, for registering a handler/listener to do something when the signal/event happens.
 - `apply(…pluginInstances: (AnyPlugin|function)[])`: `AnyPlugin` should be a class (or, rarely, an object) that has an `apply` method, or just a function with some registration code inside. This method is just to **apply** plugins' definition, so that the real event listeners can be registered into the _Tapable_ instance's registry.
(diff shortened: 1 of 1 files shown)


src/content/api/tapable.md
```diff
@@ -8,13 +8,13 @@ contributors:
   - e-cloud
 ---
 
-[Tapable](https://github.com/webpack/tapable) is small library that allows you to add and apply plugins to a javascript module. It can be inherited or mixed in to other modules. It is similar to NodeJS's `EventEmitter` class, focusing on custom event emission and manipulation. However, in addition to this, `Tapable` allows you to have access to the "emittee" or "producer" of the event through callbacks arguments.
+[Tapable](https://github.com/webpack/tapable) a is small library that allows you to add and apply plugins to a javascript module. It can be inherited or mixed in to other modules. It is similar to NodeJS's `EventEmitter` class, focusing on custom event emission and manipulation. However, in addition to this, `Tapable` allows you to have access to the "emittee" or "producer" of the event through callbacks arguments.
```
skipjack: Hmm, I think you meant `is a` instead of `a is`?

src/content/api/tapable.md
```diff
@@ -8,13 +8,13 @@ contributors:
...
 
 `Tapable` has four groups of member functions:
 
-- `plugin(name:string, handler:function)`: This allows a custom plugin to register into a **Tapable instance**'s event. This acts as the same as `on()` of `EventEmitter`, for registering a handler/listener to do something when the signal/event happens.
+- `plugin(name:string, handler:function)`: This allows a custom plugin to register into a **Tapable instance**'s event. This acts similar to `on()` of `EventEmitter`, for registering a handler/listener to do something when the signal/event happens.
```
skipjack: I like this change but maybe we can improve it even more to `...similar to the `on()` method of the `EventEmitter` which is used for...`?

src/content/api/tapable.md
```diff
@@ -8,13 +8,13 @@ contributors:
...
 - `apply(…pluginInstances: (AnyPlugin|function)[])`: `AnyPlugin` should be a class (or, rarely, an object) that has an `apply` method, or just a function with some registration code inside. This method is just to **apply** plugins' definition, so that the real event listeners can be registered into the _Tapable_ instance's registry.
-- `applyPlugins*(name:string, …)`: The _Tapable_ instance can apply all the plugins under a particular hash using these functions. These group of method act like `emit()` of `EventEmitter`, to control the event emission meticulously with various strategy for various use cases.
+- `applyPlugins*(name:string, …)`: The _Tapable_ instance can apply all the plugins under a particular hash using these functions. This group of methods act like `emit()` of `EventEmitter`, to control the event emission meticulously with various strategies and for various use cases.
```
skipjack: Maybe the end of the sentence could be improved even more?

> This group of methods act like `emit()` of `EventEmitter` and are used to meticulously control the event emission, accommodating various strategies and use cases.

src/content/api/tapable.md
jakewies: @skipjack whoops you're absolutely right, thanks for the correction!

jakewies: i will take a look at the rest of these later when I'm home, fix them up so we can move forward. Quick question though, I made these corrections through the GitHub UI. If I want to update this PR to handle your feedback, should that be done on a local branch on my computer? I don't see anywhere on GitHub to revise this specific PR.

Newbie here :)

Thanks again Greg!

skipjack: > i will take a look at the rest of these later when I'm home, fix them up so we can move forward.

Great, no rush.

> If I want to update this PR to handle your feedback, should that be done on a local branch on my computer? I don't see anywhere on GitHub to revise this specific PR.

If you go to the `Files changed` tab,  you should see an "Edit" icon on the file that will allow you to edit it again through the UI. You could also `git clone` your fork and push to that `patch-1` branch but you should be able to just make the changes through the interface.

image

You should see that same pencil icon on the right.

PR #1664, opened 2017-10-20
fix vote page

sokra:
`page.url` was different, probably due to refactoring


skipjack:
@sokra tested and verified. Soon I'm going to finish the `port-voting-app` branch and maybe try to tackle the voting redesign in its [new home](). In the PR I submit I'll also include it in the top navigation since we now have room.

PR #1667, opened 2017-10-22
Fixed typo in environment-variables.md

aidant:
Arrow functions can not return Objects without enclosing parentheses.


asapach:
That's not an object - that's the arrow function body.

aidant:
Huh you are right. Idk why I thought it was an object.

PR #1671, opened 2017-10-24
Adding windows command for build

Vanguard90:
`./node_modules/.bin/webpack --config webpack.config.js` command does not work at windows 7 with webpack 3.8.1. 

Correct way to run it is  `./node_modules/.bin/webpack.cmd --config webpack.config.js`. 

This page is the first step of the guide for newbies and this omission can be discouraging for windows users that wants to learn Webpack. I want to add this information so that other windows users can proceed without experimenting.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -217,7 +217,7 @@ bundle.js  544 kB       0  [emitted]  [big]  main
    [3] ./src/index.js 278 bytes {0} [built]
 ```
 
-W> Note that when calling `webpack` via its path on windows, you must use forward slashes instead, e.g. `node_modules\.bin\webpack --config webpack.config.js`.
+W> Note that when calling `webpack` via its path on windows, you must use backslashes instead, e.g. `node_modules\.bin\webpack --config webpack.config.js`.
 
 T> If a `webpack.config.js` is present, the `webpack` command picks it up by default. We use the `--config` option here only to show that you can pass a config of any name. This will be useful for more complex configurations that need to be split into multiple files.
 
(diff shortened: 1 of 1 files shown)


skipjack:
@bebraw @sokra can either of you verify this? I don't have a windows machine to test on but I'm surprised that the `webpack.js` file in `node_modules/webpack/.bin` would have a different extension on windows. Should it be resolved by modifying the [`PATHEXT`](http://windowsitpro.com/systems-management/how-do-i-execute-exe-files-without-typing-extension) environment variable?

Vanguard90:
@skipjack @bebraw @sokra To make your job easier, here's a screenshot of the `.bin` folder. You can see that windows sees a `'webpack.cmd'` file there.

webpack-cmd

asapach:
The correct way to run it would be `node_modules\.bin\webpack --config webpack.config.js`. The `.cmd` extension has nothing to do with this, it's automatically resolved by the shell.

skipjack:
@Vanguard90 thanks for sharing that screenshot. @asapach so it seems the issue is the leading `./`? @Vanguard90 could you try what @asapach suggested to see if that works? In that case we cane just update the original to remove the `./`.

asapach:
I think the issue is with `/` being used. `.\node_modules\.bin\webpack` works just as well, but less commonly used on Windows.

Vanguard90:
I just tried running  `node_modules\.bin\webpack --config webpack.config.js` at the basic training folder I set up and it indeed does work, whereas its `/` counterpart `node_modules/.bin/webpack --config webpack.config.js` gives an error. So @asapach identified the issue. 

Changing forward slash to backward slash sounds like a better update rather than adding an another line just for windows. @skipjack

asapach:
> you must use forward slashes instead

It's [backslash](https://en.wikipedia.org/wiki/Backslash), not forward slash.

Vanguard90:
@asapach Thanks for the catch, I need to go to sleep I guess :)
@skipjack Made the changes and updated the commit.

PR #1673, opened 2017-10-26
docs(concepts): beginner friendly introduction (WIP)

### LINKED ISSUE #1416: Concepts - Introduction page should use less technical language
The [introduction page](https://webpack.js.org/concepts/) uses language which is likely unfamiliar to someone new to modules and bundling. It also could do with some front-loading of important info in each section.

Some examples of technical language:

- "dependencies" should be initially referred to as "input files and libraries"
- "bundle" should be referred to as "output file"
- the concept of a "dependency graph" could possibly be addressed in less technical terms

Examples of front-loading important info:

- "The _entry_ describes the file that webpack should process first. There may be more than one entry file."
- "The _output_ describes the file which should be produced by webpack. There is generally one output file produced for each entry."
- "A _loader_ describes how files of a certain type should be processed."
- "A _plugin_ adds custom functionality to the webpack bundling process."

TheDutchCoder:
Attempt to make the Introduction a bit more beginner friendly by explaining concepts a little more.

Wait with a review until I give the heads-up ;)

Resolves #1416

### DIFF
--- a/src/content/concepts/index.md
+++ b/src/content/concepts/index.md
@@ -14,10 +14,10 @@ At its core, *webpack* is a _module bundler_ for modern JavaScript applications.
 
 It is [incredibly configurable](/configuration), but to get started you only need to understand four **Core Concepts**:
 
- * Entry
- * Output
- * Loaders
- * Plugins
+- Entry
+- Output
+- Loaders
+- Plugins
 
 This document is intended to give a **high-level** overview of these concepts, while providing links to detailed concept specific use cases.
 
--- a/src/content/concepts/index.md
+++ b/src/content/concepts/index.md
@@ -10,9 +10,10 @@ contributors:
   - TheDutchCoder
 ---
 
-At its core, *webpack* is a _module bundler_ for modern JavaScript applications. When webpack processes your application, it recursively builds a _dependency graph_ that includes every module your application needs, then packages all of those modules into one or more _bundles_ to be loaded by the browser.
+At its core, *webpack* is a _module bundler_ for modern JavaScript applications. When webpack processes your application, it recursively builds a _dependency graph_ that includes every module your application needs, then packages all of those modules into one or more _bundles_.
 
 It is [incredibly configurable](/configuration), but to get started you only need to understand four **Core Concepts**:
+
  * Entry
  * Output
  * Loaders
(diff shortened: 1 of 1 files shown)


TheDutchCoder:
@skipjack Okay a first run through the changes should be a good idea for now.

Maybe @tanyagray can provide a bit more constructive feedback as well to see if we can improve this even more.

src/content/concepts/index.md
```diff
@@ -59,20 +62,18 @@ In the example above, we use the `output.filename` and the `output.path` propert
...
+The `output` property has [many more configurable features](/configuration/output) and if you like to know more about the concepts behind the `output` property, you can [read more in the concepts section](/concepts/output).
 
 
 ## Loaders
 
-The goal is to have all of the assets in your project be **webpack's** concern and not the browser's (though, to be clear, this doesn't mean that they all have to be bundled together). webpack treats [every file (.css, .html, .scss, .jpg, etc.) as a module](/concepts/modules). However, webpack itself **only understands JavaScript**.
+*Loaders* enable webpack to process more than just JavaScript files (webpack itself only understands JavaScript). They make sure that you can leverage webpack's bundling abilities for files that aren't JavaScript by treating them as [webpack modules](/concepts/modules).
 
-**Loaders in webpack _transform these files into modules_ as they are added to your dependency graph.**
+In short: **Loaders in webpack _transform non JavaScript files into modules_ as they are added to your dependency graph.**
```
skipjack: Not sure if we should keep this whole sentence bold. Also maybe we could rephrase to something like:

> Essentially, webpack loaders transform all types of files into modules that included in your application's dependency graph.

skipjack: Also, as you've already started doing, I think we can lowercase "Loaders" unless it's used at the beginning of a sentence.

src/content/concepts/index.md
```diff
@@ -109,9 +110,9 @@ There are other, more specific properties to define on Loaders that we haven't y
 
 ## Plugins
 
-While Loaders only execute transforms on a per-file basis, `plugins` are most commonly used to perform actions and custom functionality on "compilations" or "chunks" of your bundled modules [(and so much more!)](/concepts/plugins). The webpack Plugin system is [extremely powerful and customizable](/api/plugins).
+While Loaders only execute transforms on a per-file basis, `plugins` are most commonly used to perform actions and custom functionality on pieces (e.g. chunks) of your bundled modules [(and so much more!)](/concepts/plugins). The webpack Plugin system is [extremely powerful and customizable](/api/plugins).
```
skipjack: I would lowercase "Loaders" and maybe simplify the sentence even more...

> While loaders are used to transform certain types of modules, `plugins` can be leveraged to perform a wider range of tasks. Plugins range from bundle optimization and minification all the way to defining environment-like variables. The [plugin interface](/api/plugins) is extremely powerful and can be used to tackle a wide variety of tasks.

TheDutchCoder:
:+1: On it!

TheDutchCoder:
@skipjack I processed your feedback, please take a look and see if this is good enough to merge!

Thanks again for your review!

PR #1676, opened 2017-10-27
External Source Map Example in SourceMapDevToolPlugin

hkrutzer:


src/content/plugins/source-map-dev-tool-plugin.md
```diff
@@ -45,3 +47,14 @@ new webpack.SourceMapDevToolPlugin({
   exclude: ['vendor.js']
 })
 ```
+
+### Host sourcemaps externally
```
skipjack: I think we should captilize the title, split "sourcemaps" into "Source Maps", and maybe simplify slightly to "External Source Maps" (though I'm also ok with "Host Source Maps Externally").

src/content/plugins/source-map-dev-tool-plugin.md
```diff
@@ -45,3 +47,14 @@ new webpack.SourceMapDevToolPlugin({
...
+
+Set a URL for sourcemaps. Useful for hosting them on a host that requires authorization.
```
skipjack: Split "sourcemaps" into two words ("source maps") here as well. Just for consistency with the rest of our docs.

hkrutzer:
No problem. I've addressed your comments.

PR #1686, opened 2017-11-03
A new --concatenate-modules flag

iamakulov:
This adds documentation for the new `--concatenate-modules` CLI flag (see https://github.com/webpack/webpack/pull/5914).

Additionally, this documents `ModuleConcatenationPlugin` in the Production guide. I believe it’s reasonable to do this because the `--concatenate-modules` flag is included under the `-p` umbrella.

### DIFF
--- a/src/content/guides/production.md
+++ b/src/content/guides/production.md
@@ -239,7 +239,7 @@ __src/index.js__
 
 ## Module Concatenation
 
-By default, webpack wraps each module into a function. This helps helps to isolate them, but brings additional overhead. You can try reducing this overhead by enabling [`ModuleConcatenationPlugin`](/plugins/module-concatenation-plugin) which, where possible, merges multiple modules together in a single module wrapper:
+By default, webpack wraps each module into a function. Wrapping helps to isolate them but brings additional overhead. You can try reducing this overhead by enabling the [`ModuleConcatenationPlugin`](/plugins/module-concatenation-plugin), which, where possible, merges multiple modules in a single module wrapper:
 
 __webpack.prod.js__
 
--- a/src/content/guides/production.md
+++ b/src/content/guides/production.md
@@ -205,7 +205,7 @@ __webpack.prod.js__
 +       'process.env': {
 +         'NODE_ENV': JSON.stringify('production')
 +       }
-+     })
++     }),
     ]
   })
 ```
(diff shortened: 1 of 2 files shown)


src/content/guides/production.md
```diff
@@ -236,9 +237,33 @@ __src/index.js__
...
+
+Read more about module concatenation [in the plugin docs](/plugins/module-concatenation-plugin).
 
 ## CLI Alternatives
 
-Some of what has been described above is also achievable via the command line. For example, the `--optimize-minimize` flag will include the `UglifyJSPlugin` behind the scenes. The `--define process.env.NODE_ENV="'production'"` will do the same for the `DefinePlugin` instance described above. And, `webpack -p` will automatically invoke both those flags and thus the plugins to be included.
+Some of what has been described above is also achievable via the command line:
+* the `--optimize-minimize` flag will include the `UglifyJSPlugin` behind the scenes;
+* the `--define process.env.NODE_ENV="'production'"` will do the same for the `DefinePlugin` instance described above;
+* the `--concatenate-modules` flag will enable the `ModuleConcatenationPlugin`.
```
skipjack: To fix the build errors add a line break between 262/263 and change the list style to `-`. I also think you can drop the leading `the` in these and simplify like so:

> - `--optimize-minimize` will include the `UglifyJSPlugin` behind the scenes;
> - `--define process.env.NODE_ENV="'production'"` will allow dependencies to key off `process.env.NODE_ENV`;
> - `--concatenate-modules` will include the `ModuleConcatenationPlugin`.

src/content/guides/production.md
```diff
@@ -236,9 +237,33 @@ __src/index.js__
...
++ const webpack = require('webpack');
+  const merge = require('webpack-merge');
+  const common = require('./webpack.common.js');
+
+  module.exports = merge(common, {
+    plugins: [
++     new webpack.optimize.ModuleConcatenationPlugin()
+    ]
+  })
+```
```
skipjack: This example should be a true diff by including the full content of the previous __webpack.prod.js__ example, e.g.

``` diff
  const merge = require('webpack-merge');
  const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
  const common = require('./webpack.common.js');

  module.exports = merge(common, {
    devtool: 'cheap-module-source-map',
    plugins: [
      new UglifyJSPlugin({
        sourceMap: true
      }),
      new webpack.DefinePlugin({
        'process.env': {
          'NODE_ENV': JSON.stringify('production')
        }
-     })
+     }),
+     new webpack.optimize.ModuleConcatenationPlugin()
    ]
  })
```

src/content/guides/production.md
```diff
@@ -236,9 +237,33 @@ __src/index.js__
...
+
+Read more about module concatenation [in the plugin docs](/plugins/module-concatenation-plugin).
```
skipjack: Minor nitpick: we typically include two line breaks before/after `h2` headers.

iamakulov:
Thanks for the feedback! Updated :–)

src/content/guides/production.md
iamakulov: Done!

Instead of doing

```diff
-     })
+     }),
```

I added a comma in the `DefinePlugin` section.

src/content/guides/production.md
```diff
@@ -236,9 +237,46 @@ __src/index.js__
   document.body.appendChild(component());
 ```
 
+## Module Concatenation
+
+By default, webpack wraps each module into a function. This helps helps to isolate them, but brings additional overhead. You can try reducing this overhead by enabling [`ModuleConcatenationPlugin`](/plugins/module-concatenation-plugin) which, where possible, merges multiple modules together in a single module wrapper:
```
asapach: >This _**helps helps**_ to isolate them, ...

src/content/guides/production.md
```diff
@@ -204,7 +205,7 @@ __webpack.prod.js__
 +       'process.env': {
 +         'NODE_ENV': JSON.stringify('production')
 +       }
-+     })
++     }),
```
asapach: why add a trailing comma? this is inconsistent with the existing code style

src/content/guides/production.md
iamakulov: Good catch, thank you!

src/content/guides/production.md
iamakulov: OK, I’ve looked at `UglifyJSPlugin` here:

image

and thought that if the selected line has a comma in this example, it also has a trailing comma in the example before that – but turns out it doesn’t.

Will update this.

skipjack:
@iamakulov thanks for making the updates so promptly! Will merge shortly.

iamakulov:
Whoops, wait, the feature is not released yet – https://github.com/webpack/webpack/pull/5914 :–)

Or is that OK, and this wouldn’t be published until the feature is released?

PR #1690, opened 2017-11-05
Update html-webpack-plugin.md

Jocs:
_describe your changes..._

Update output.path to absolute path in html-webpack-plugin.md

### DIFF
--- a/src/content/plugins/html-webpack-plugin.md
+++ b/src/content/plugins/html-webpack-plugin.md
@@ -27,7 +27,7 @@ var HtmlWebpackPlugin = require('html-webpack-plugin');
 var webpackConfig = {
   entry: 'index.js',
   output: {
-    path: 'dist',
+    path: __dirname + '/dist',
     filename: 'index_bundle.js'
   },
   plugins: [new HtmlWebpackPlugin()]


src/content/plugins/html-webpack-plugin.md
```diff
@@ -27,7 +27,7 @@ var HtmlWebpackPlugin = require('html-webpack-plugin');
 var webpackConfig = {
   entry: 'index.js',
   output: {
-    path: 'dist',
+    path: __dirname + '/dist',
```
skipjack: Even better would be to use Node's built-in `path.resolve` method:

``` js
path: path.resolve(__dirname, './dist')
```

This would allow the same configuration to work on windows as well.

Jocs: ok, I will change it ,and use path.resolve.

Jocs: Hi, I created another PR, based on my forked repository's master branch.
[1698](https://github.com/webpack/webpack.js.org/pull/1698)
You can close the PR now.

Sorry for the trouble which I caused.
🙃

skipjack: No worries, will do.

PR #1691, opened 2017-11-05
Use normal function to fix scope in loader

ThaNarie:
When using an arrow function, `this` is undefined.
When using a normal function, `this` is the loaderContext.

(tested using the tutorial on this page)

### DIFF
--- a/src/content/contribute/writing-a-loader.md
+++ b/src/content/contribute/writing-a-loader.md
@@ -234,13 +234,15 @@ __src/loader.js__
 ``` js
 import { getOptions } from 'loader-utils';
 
-export default source => {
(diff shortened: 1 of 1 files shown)


src/content/contribute/writing-a-loader.md
```diff
@@ -234,13 +234,15 @@ __src/loader.js__
 ``` js
 import { getOptions } from 'loader-utils';
 
-export default source => {
+function loader(source) {
```
asapach: you could do `export default function loader(...) {` to keep the code consistent

ThaNarie: You're right, updated!

PR #1696, opened 2017-11-11
Use shorten form for `process.env.NODE_ENV`

ogonkov:
Shorten form for `DefinePlugin` example.

### DIFF
--- a/src/content/guides/production.md
+++ b/src/content/guides/production.md
@@ -201,9 +201,7 @@ __webpack.prod.js__
         sourceMap: true
       }),
 +     new webpack.DefinePlugin({
-+       'process.env': {
-+         'NODE_ENV': JSON.stringify('production')
-+       }
++       'process.env.NODE_ENV': JSON.stringify('production')
 +     })
     ]
   })


pierreneter:
Hey, please check: https://github.com/webpack/webpack/issues/5392 and https://github.com/webpack/webpack/pull/5450
I think the old way is better.

skipjack:
Yeah that’s a fair point, let’s close this out. @ogonkov your version is slightly simpler from a line count perspective but nesting it adds very little overhead.

ogonkov:
@skipjack it's a webpack version, not mine
https://webpack.js.org/plugins/define-plugin/#feature-flags

>When defining values for process prefer 'process.env.NODE_ENV': JSON.stringify('production') over process: { env: { NODE_ENV: JSON.stringify('production') } }. Using the latter will overwrite the process object which can break compatibility with some modules that expect other values on the process object to be defined.

skipjack:
> When defining values for process prefer 'process.env.NODE_ENV': JSON.stringify('production') over process: { env: { NODE_ENV: JSON.stringify('production') } }. Using the latter will overwrite the process object which can break compatibility with some modules that expect other values on the process object to be defined.

Ah thanks for the citation @ogonkov, I see why you changed it now. @pierreneter I'd say that's a somewhat bigger problem than the destructuring problem.

@ogonkov what do you mean by `it's a webpack version` though? Does the overwriting issue only occur in an older webpack version or do you just mean that because of that warning we should lean towards not nesting?

ogonkov:
@skipjack i mean that documentation have a single recommendation for using `process.env` override, so it's better to have all examples to be consistent

skipjack:
@ogonkov yep I agree... @pierreneter based on that point I think I'm going to merge this. Hopefully both of those issues (overwriting and destructuring) will be resolved at some point so it won't matter what we recommend. This is a case where I don't think approach matters too much if the issues you both brought up didn't exist.

PR #1697, opened 2017-11-11
Loader API Revisions

### LINKED ISSUE #1607: API - Document Adding Metadata via Loaders
Feature

Loader chains are pretty much stateless, if a loader mutates content significantly but next loader needs some critical information (which is now tampered) to do it's thing, it'll have hard time.

NA

Take `css-loader` and config `use: ExtractTextPlugin.extract({use: 'css-loader'})` for example.
 If `css-loader` could add some metadata when it's running, something like `meta: {rawContent: "body {margin: 0;}", sourceMap: {...}}` and is accessible by subsequent loaders, ExtractTextPlugin in this example, they won't have to run mini-compilers to extract the content, since it's now directly accessible.
 
See *What is the expected behavior?*

**Please mention other relevant information such as the browser version, Node.js version, webpack version and Operating System.**
NA

---

This issue was moved from webpack/webpack#5715 by @sokra. Orginal issue was by @NeekSandhu.

> Great. It'd beautiful if documentation is fixed a bit.
> 
> As on this page: https://webpack.js.org/api/loaders it doesn't say anything about third argument.
> 
> And second, at https://webpack.js.org/api/loaders/#this-callback it says third argument is AST although it can be anything really.

Yep you should send a PR to the documentation @NeekSandhu.

skipjack:
Add some lead in descriptions to the `Examples` section and clarify that meta
data can be passed along via the fourth parameter to `this.callback` and accepted
as the third parameter to a loader function.

Resolves #1607

cc @NeekSandhu

### DIFF
--- a/src/content/api/loaders.md
+++ b/src/content/api/loaders.md
@@ -42,7 +42,7 @@ module.exports = function(content, map, meta) {
 
 ### Asynchronous Loaders
 
-For asynchronous loaders, [`this.async`]() is used to retrieve the `callback` function:
+For asynchronous loaders, [`this.async`](/api/loaders#this-async) is used to retrieve the `callback` function:
 
 __async-loader.js__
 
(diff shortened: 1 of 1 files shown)


NeekSandhu:
Thank you very much!

PR #1700, opened 2017-11-13
Linking compiler docs on plugins page

chrmod:
While authoring plugins the list of hooks is essential knowledge, so link to hooks documentation page should be accessible in the top of plugins authoring guide.
The exact location is not very important, but it should be easy to find.

(for newcomer like myself it was not easy to realize the link between compiler api and plugins at first)

### DIFF
--- a/src/content/contribute/writing-a-plugin.md
+++ b/src/content/contribute/writing-a-plugin.md
@@ -11,7 +11,7 @@ A plugin for `webpack` consists of
 
 - A named JavaScript function.
 - Defines `apply` method in it's prototype.
-- Specifies webpack's event hook to attach itself.
+- Specifies an [event hook](/api/compiler/#event-hooks) on which to bind itself.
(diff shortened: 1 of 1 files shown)


src/content/contribute/writing-a-plugin.md
```diff
@@ -11,7 +11,7 @@ A plugin for `webpack` consists of
 
 - A named JavaScript function.
 - Defines `apply` method in it's prototype.
-- Specifies webpack's event hook to attach itself.
+- Specifies webpack's event hook to attach itself. (all hooks are listed at [compiler](https://webpack.js.org/api/compiler/#event-hooks) docs)
```
skipjack: I think this would be better phrased as:

``` md
- Specifies an [event hook](/api/compiler/#event-hooks) on which to bind itself.
```

chrmod:
@skipjack make sense! thanks for review

PR #1707, opened 2017-11-19
SourceMapDevToolPlugin: Enable path stripping from source map filenam…

EugeneHlushko:
…e path when using custom append.

Documentation update according to: https://github.com/webpack/webpack/pull/5986


EugeneHlushko:
@skipjack thank you, looks good!

PR #1708, opened 2017-11-20
Use npx

MajorBreakfast:
This PR changes `./node_modules/.bin/webpack` to `npx webpack`. `npx` has been shipping with node for a few months now. And I'd say that using `npx` is the more ideomatic way to execute a binary of a locally installed package.


ematipico:
For people that don't have the new version yet, I think the documentation should explain both ways.

MajorBreakfast:
@ematipico I've modified the text, so that it's clear were to find the binary, so that it can be called directly if npx isn't there for some reason.

MajorBreakfast:
@skipjack I totally agree with that motto :D 

BTW my hopes for getting this actually merged weren't that high. Not all projects think like that. IMO the docs should always show the most idiomatic way. That way it's nice and straightforward for newbies. I also learned about npx only a few months after it started shipping with node because I saw it used somewhere.

PR #1717, opened 2017-11-26
Port/Redesign Voting App

skipjack:
As the "Vote" section is basically another web app/site in and of itself, we ported it to a [separate repo](https://github.com/webpack-contrib/voting-app), and I finally found the time to implement the redesign we'd been discussing. This pull request...

- [x] Picks the redesigned app up as a package (only linked right now).
- [x] Removes all obsolete code from this repo.
- [x] Adds a prominent link to the mini-application in the header.
- [x] Updates the issue template to make it clear what's external.

Related to webpack-contrib/voting-app#7


skipjack:
@bebraw I think `antwar-interactive` may have issues with resolving `node_modules`. See the [build errors](https://travis-ci.org/webpack/webpack.js.org/builds/319014508#L839-L850). I'm testing locally and the error seems related to the `id` prop. I've been poking around but I can't quite figure out how that's used within `antwar`.

bebraw:
@skipjack It's maybe better to replicate that in the Antwar repo itself. Easier to debug and eventually fix.

skipjack:
Ok confirmed it's an issue with `id`. Changing the `id` to `./node_modules/webpack.vote` fixes the issue (at least when running the build locally). I'll push here momentarily and we'll see if it passes on Travis.

> It's maybe better to replicate that in the Antwar repo itself. Easier to debug and eventually fix.

Ok I'll see if I can find time to create a minimal example. Want me to create an issue?

bebraw:
> Ok I'll see if I can find time to create a minimal example. Want me to create an issue?

Yup, otherwise I'll lose sight of it and forget.

skipjack:
Ok @TheLarkInn @sokra, merging but I'll keep an eye on the site and once it gets deployed I'll check the login functionality (and the rest) to make sure it's working properly.

skipjack:
@sokra also by one of your comments it sounded like there is a way to test oauth login locally using a hosts file or something? Can you elaborate or point me in the direction of an article that explains how to do that?

sokra:
It's only possible to set one callback/redirect url. It's would be a security problem anyway.

You can edit your HOSTS file to point webpack.js.org to your localhost. You can also copy the token from the production page to the localhost localStorage to test the page after login.

skipjack:
Will do. This still hasn't deployed yet for some reason but I'll try to test login locally to avoid any surprises.

PR #1720, opened 2017-11-29
Update shimming doc with build output example

agudulin:
Add build output example over `TODO`.

### DIFF
--- a/src/content/guides/shimming.md
+++ b/src/content/guides/shimming.md
@@ -90,8 +90,6 @@ What we've essentially done here is tell webpack...
 If we run a build, we should still see the same output:
 
 ``` bash
-./node_modules/.bin/webpack --config webpack.conf.js
-
 Hash: f450fa59fa951c68c416
 Version: webpack 2.2.0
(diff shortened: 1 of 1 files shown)


src/content/guides/shimming.md
```diff
@@ -90,7 +90,17 @@ What we've essentially done here is tell webpack...
 If we run a build, we should still see the same output:
 
 ``` bash
-TODO: Include output
+./node_modules/.bin/webpack --config webpack.conf.js
+
```
skipjack: Can you drop these two lines (`93` and `94`)? Most of the guides stem from _Getting Started_ where setting up npm scripts (i.e. `npm run build`) is [recommended](https://webpack.js.org/guides/getting-started/#npm-scripts).

skipjack: Also, in most of the other output examples we don't include the actual command.

agudulin: Thank you for review, I copied it from another page mentioned in this guide :) Removing..

PR #1739, opened 2017-12-11
Update getting-started.md

wohuifude123:
At the beginning, should include the description such as 'don't include via any script'.


skipjack:
Lodash is being included via a script though, the [basic setup](https://webpack.js.org/guides/getting-started/#basic-setup) section is meant to be written all at once. There shouldn't be any intermediate steps, and if it wasn't included via a `<script>` tag, none of the code in that script would run.

PR #1742, opened 2017-12-13
Just an uppercase fix

jampow:
Just a simple uppercase fix


skipjack:
@jampow yeah so it does feel a bit funny but that's intentional according to our [branding guidelines](https://github.com/webpack/media#name).

PR #1747, opened 2017-12-15
docs(HMR): Fix code example

RobbertWolfs:
Updating code example in guides.
Guides told to remove the `print.js` but it deleted two lines and putted back one line (same one as one of the deleted). So I just deleted the one line that was necessary

### DIFF
--- a/src/content/guides/hot-module-replacement.md
+++ b/src/content/guides/hot-module-replacement.md
@@ -46,8 +46,8 @@ __webpack.config.js__
 
   module.exports = {
     entry: {
-       app: './src/index.js'
--      print: './src/print.js'
+      app: './src/index.js'
+-     print: './src/print.js'
     },
     devtool: 'inline-source-map',
     devServer: {
--- a/src/content/guides/hot-module-replacement.md
+++ b/src/content/guides/hot-module-replacement.md
@@ -2,6 +2,7 @@
 title: Hot Module Replacement
 sort: 6
 contributors:
+  - RobbertWolfs
   - jmreidy
   - jhnns
   - sararubin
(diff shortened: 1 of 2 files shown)


src/content/guides/hot-module-replacement.md
```diff
@@ -45,9 +46,8 @@ __webpack.config.js__
 
   module.exports = {
     entry: {
--      app: './src/index.js',
+       app: './src/index.js',
```
asapach: you have a trailing comma now

src/content/guides/hot-module-replacement.md
```diff
@@ -45,9 +46,8 @@ __webpack.config.js__
 
   module.exports = {
     entry: {
--      app: './src/index.js',
+       app: './src/index.js'
```
asapach: now compare this line: you've changed it, but it's not reflected in the diff
does this PR make sense?

RobbertWolfs: The following sentence is written in the markdown file : "We'll also remove the entry point for print.js..."

So I guess this should be the correct example 
screen shot 2017-12-15 at 22 36 48


And this is the current example
screen shot 2017-12-15 at 22 38 29

Which removes the `app.js` line and then re-adds it.. Which is a step to much I guess

src/content/guides/hot-module-replacement.md
RobbertWolfs: fixed

asapach:
That's how diff works: when you change a line it shows the old text as "removal" and the new text as an "addition". By removing the trailing comma in the first line you've change it, which needs to be reflected in the diff.

RobbertWolfs:
Ok, thought it was weird but.. ignore this PR then

PR #1750, opened 2017-12-16
Make intro more realistic

sexualmaniac:
Current message gives the wrong feeling of being an easy tool for the job. In reality, with all the setup, options, plugins, etc that you need to know it's not even close to be simple.

### DIFF
--- a/src/content/index.md
+++ b/src/content/index.md
@@ -59,7 +59,7 @@ __page.html__
 
 Then run `webpack` on the command-line to create `bundle.js`.
 
-## It's that simple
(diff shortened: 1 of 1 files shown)


skipjack:
Unfortunate you feel that way... webpack is definitely a low-level tool and there is some complexity in some of the more advanced features but the __Guides__ section has pretty thorough step-by-step walkthroughs of the core features and getting started really isn't that bad. If you'd like to submit a more serious PR to help improve one of the articles, please feel free to.

Issue #1706, opened 2017-11-17
webpack v4 migration guide and feature additions


TheLarkInn:
Hey docs team and all of our beloved contributors!!!! 

We are getting close to merge webpack/webpack#next (v4) into master soon and we have drafted a WIP change log and this means we can start laying the ground work and documentation for our migration guide and other changes!

Listed is the link to the changelog:

https://github.com/webpack/webpack/wiki/Changelog-WIP

I'd give you ownership to break up the work however you see fit! Let me know if the core team can help in any way!!!

🔥🔥Lots of exciting changes!🔥🔥

reergymerej:
I've got a [PR in core] that modifies a new feature (`sideEffects` in package.json) and I'd like to make the sister change in the docs.  Can anyone point me in the right direction?

[PR in core]: https://github.com/webpack/webpack/pull/6314

skipjack:
Hmmm, maybe in the __API__ section somewhere? Also target your PR to the `next` branch.

reergymerej:
Thanks, @skipjack.  I searched the `next` branch and didn't see anything.  I may be the first to add it.

skipjack:
> From #1795...

Create an upgrade path section with clearly defined, individual pages:

- Upgrade from v1 to v2
- Upgrade from v2 to v3
- Upgrade from v3 to v4

Each page should include any script helpers, and examples particular to it's version of the upgrade.

skipjack:
I like the note from from #1795 but I think we should create a "Migrating" `group` under __Guides__ and title the articles simply `v1 - v2`, `v2 - v3`, etc.

TheLarkInn:
Here is the first issue with a set of migration diffs. I think this is something we can run with? 

https://github.com/webpack/webpack.js.org/issues/1706

TheLarkInn:
@skipjack, I've been starting to make a bunch of videos that are going to hopefully be embeddable into our documentation. I'm not sure where I can start to organize them. Would it work best to dump into a standalone issue?

TheLarkInn:
> I like the note from from #1795 though in terms of actual wording/implementation, I think we should create a "Migrating" group under Guides and title the articles simply v1 - v2, v2 - v3, etc. (or 1.x.x to 2.x.x).

Has this already landed in next branch? If not, I will create an issue so that first time contributor can tackle just this work (creating the group).

skipjack:
[...]

PR #1754, opened 2017-12-18
(Plugin) API Rewrite

### LINKED ISSUE #940: API - Clarify Plugins Structure and Flow
The page especially missing:
1. Description of all class instance properties
2. A flow of the plugins - a) where you can hook and where it is in the flow b) when webpack do what internally. I found this chinese website which go the direction I mean: https://github.com/liangklfangl/webpack-compiler-and-compilation

3. This is difficult to understand:
```
Plugin Interfaces
There are two types of plugin interfaces.

Timing based

sync (default): As seen above. Use return.
async: Last parameter is a callback. Signature: function(err, result)
parallel: The handlers are invoked parallel (async).

Return value

not bailing (default): No return value.
bailing: The handlers are invoked in order until one handler returns something.
parallel bailing: The handlers are invoked in parallel (async). The first returned value (by order) is significant.
waterfall: Each handler gets the result value of the last handler as an argument.
```

4. Show ES6 class variant of the plugin architecture beside the prototyped one (encourage plugin developer to use classes)

5. Explain entirely what **seal** means in webpack

6. compilation(c: Compilation, params: Object)
> A Compilation is created. A plugin can use this to obtain a reference to the Compilation object. The params object contains useful references

I **LOVE** useful references. But what kind of?

### LINKED ISSUE #1458: API - Document When a Plugin Should Call `doResolve()`
Documentation request

No documentation about how plugins are processed

Every webpack plugin follows the following structure:

```
resolver.plugin(this.source, function(request, callback) {
    if (something) {
         resolver.doResolve(target, obj, "aliased with mapping '" + name + "': '" + ...)
    } else {
         callback(...);
    }
```
Can anyone please explain when I should call `doResolve(...)` and when `callback(...)`. I've found this phrase in the webpack docs:

> To pass the request to other resolving plugins, use the this.doResolve(types: String|String[], request: Request, callback) method

However, I don't know what to make of it. 

It seems that `doResolve` starts the process from the start - resolve types are repeated. Inside `doResolve` the `callback.stack` has the following entries:
```
[
    "resolve: (/d/...",    <------------- resolve here
    "new-resolve: (/d/...",
    "parsed-resolve: (/d/...",
    "described-resolve: (/d/...",
    "resolve: (/d/...",    <--------------- and it's repeated here
    "new-resolve: (/d/...",
    "parsed-resolve: (/d/...",
    ...
]
```

[Here is](http://imgur.com/a/OEUjr) the screenshot. Why does that happen?

Here is how the stack looks like in `doResolve`.

And also all plugins that hook into `resolve` events use the function `getInnerRequest(resolver, request);`. What is inner request?

It's hard to write a plugin without understanding the processing flow. I'm not asking for a full-blown detailed documentation, maybe just a small passage explaining the matter in general terms so that I get correct mindset when exploring the code through the debugger.

---

This issue was moved from webpack/webpack#5380 by @sokra. Orginal issue was by @maximusk.

### LINKED ISSUE #1505: API - Resolver documentation confusing
Regarding [this page](https://webpack.js.org/api/plugins/resolver/).

I noticed webpack allows plugins in the `resolve` config ([`resolve.plugins`](https://webpack.js.org/configuration/resolve/#resolve-plugins)). So I assumed these kind of plugins are defined as per the documentation found [here](https://webpack.js.org/api/plugins/resolver/).

However, this page gives absolutely no information at all about creating a "resolver" or how to go about implementing one. It mentions what methods a resolver has access to and should use, but not how to create one.

It would be of great use if there was at least a linked example, or the boilerplate of such a plugin.

Edit: Removed second issue, as it was a scoping problem on my end.

skipjack:
> Work in progress...

This point of this pr is to:

- Resolve some longstanding __API__ issues (see below).
- Standardize, simplify, and clarify various parts of the plugin API.
- Do some reformatting of the __API__ section in general.
- Update and dedupe `tapable` docs.
- Bring the documentation up to date with the latest hooks available in v4.

Resolves #1505 
Resolves #1458 
Resolves #940 

This also begins the process of resolving #1706. Notice that I've targeted this to a `next` branch where we can begin merging `v4` related PRs. Once webpack v4 is fully released (around february it sounds like) we can... 

- Tag master with the latest v3 version (e.g. `3.10.x`) to make it easy to archive and make available later on.
- Rebase the `next` to make sure it's fully up to date with anything on `master`.
- Merge the `next` branch to master and continue improving the docs.


skipjack:
@TheLarkInn @sokra I could really use some help with finishing the documentation on this branch soon (mostly just need to know the missing hook descriptions). I also think we need to update the following utility READMEs as such:

### `tapable`

The updated docs were actually pretty solid and easy to follow. I think all that's missing is a description for each available hook.

### `enhanced-resolve`

It seems the README for this package is totally out of date with the current code (though I could be wrong). I think the following needs to be documented:

- Each available hook (preferably using the same format we use here)
  - Type of underlying `tapable` hook
  - Brief description and optionally a small example
  - Any parameters passed
- A description of each available class method and their parameters (e.g. `doResolve`)

@mlcohen, seeing as you dug into this a bit in #1458, would you be willing to help out with updating the `enhanced-resolve` README? You may want to collaborate with @sokra and @TheLarkInn as I think the API changed a bit recently.

As for the documentation of both these utilities (`tapable` and `enhanced-resolve`) in this site, I'm tightening up the existing documentation in this pull request and linking out to each package for details. IMHO the public-facing APIs of those packages should be documented in only one place: their respective READMEs. The only thing that should be noted on this site is how the core of webpack uses their APIs internally. As I've mentioned elsewhere, once those READMEs are updated, we can pull them into a new "Utilities" section in this site (similar to __Loaders__ and __Plugins__).

Trying to have this PR in a good place for review within the next 24 - 48 hours. Hoping we can then work together to knock this stuff out as this should solidify the __API__ section of our site significantly and hopefully clarify a lot for our readers/users.

TheLarkInn:
> The updated docs were actually pretty solid and easy to follow. I think all that's missing is a description for each available hook.

So you mean this would be for authors looking to write their own hooks for their own plugin systems etc? @mlcohen could definitely help here.

skipjack:
> So you mean this would be for authors looking to write their own hooks for their own plugin systems etc?

That too, but also for webpack plugin authors who want to know how the `SyncHook`, `SyncWaterfallHook`, etc work behind the scenes. Something like...

> ### SyncHook
> 
> Creates a hook that is executed synchronously... (etc.)
>
> Methods: `tap`

...for each would be great. I know they may seem obvious to people who are very familiar with `tapable` or other similar systems but for new plugin authors (e.g. me) it's helpful to have some background. The "Methods" section could be left out I guess because I think it's made pretty clear below that `Sync***` hooks only allow `tap` and `Async***` hooks allow all three.

I can move some of this to an issue in the `tapable` repo if that would help. Re @mlcohen, I was thinking they'd be able to help more with `enhanced-resolve` per the discussion in #1458 but if they have a good understanding of `tapable` as well that would be great. (this is all of course if @mlcohen has time 😜 ... no pressure)

skipjack:
@TheLarkInn merging before this gets too far behind. I highly recommend that either you, @sokra, or someone else who is familiar with the webpack v4 hooks go through the following pages and fill out the hook descriptions:

- `compiler-hooks.md`
- `compilation-hooks.md`
- `parser.md`
- `resolvers.md`

I standardized the format of each hook section and I think I got all new/renamed hooks added. I didn't feel qualified enough to fill in the descriptions for the rest (i.e. every one with `...`). I'm going to create issues on the `tapable` and `enhanced-resolve` repos to note what I mentioned above.

PR #1763, opened 2017-12-26
docs(config): clarify use of wildcard with proxy url matching

ramast:
Clarify the possibility of using micromatch expressions inside the proxy's context array

_describe your changes..._

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -629,6 +629,16 @@ proxy: [{
 }]
 ```
 
+[Micromatch](https://www.npmjs.com/package/micromatch#matching-features) wildcard expressions are also supported:
+
+```js
+proxy: [{
+  context: ["/admin/**", "**/api/**", "(@/index.html/)"],
+  target: "http://localhost:3000",
+}]
+```
+But you may not mix plain text and wildcard expressions in the same context array
+
 ## `devServer.progress` - CLI only
 
 `boolean`


src/content/configuration/dev-server.md
```diff
@@ -629,6 +629,16 @@ proxy: [{
...
 ```
 
+[Micromatch](https://www.npmjs.com/package/micromatch#matching-features) wildcard expressions are also supported:
+
+```js
+proxy: [{
+  context: ["/admin/**", "**/api/**", "(@/index.html/)"],
+  target: "http://localhost:3000",
+}]
+```
```
skipjack: You need to add another line break after the code block, before the next paragraph.

src/content/configuration/dev-server.md
```diff
@@ -629,6 +629,16 @@ proxy: [{
...
+But you may not mix plain text and wildcard expressions in the same context array
+
```
skipjack: Here as well (we usually keep two lines between each `##` (h2) section.

shellscape:
I have to give a 👎 on this one. Not because of the content - but because all the proxy feature does is use https://github.com/chimurai/http-proxy-middleware and pass-through options to that middleware. We really need to avoid adding a ton of docs around that, as it should be documented on that project page - and linked to from the webpack docs.

ramast:
I can see your point but as a new comer I didn't realize that wild card matching is an advanced topic which I'd find it in the http-proxy-middleware documentation.

Maybe my edit can be replaced with something like

wildcard matching is also supported, see https://github.com/chimurai/http-proxy-middleware#context-matching for more information

shellscape:
@ramast yeah my point was that the docs need to be clearer that we're leveraging that module, and they need to point the user to that module's docs in an obvious manner. I'd like to see this PR pivot to providing that.

ramast:
Would my proposal to replace current PR with something like

Wildcard matching is also supported, see https://github.com/chimurai/http-proxy-middleware#context-matching for more information

Is acceptable to you? or you want a more generic text?
I am asking because there is already one generic text that says 

> The dev-server makes use of the powerful http-proxy-middleware package. Checkout its documentation for more advanced usages.

skipjack:
> I am asking because there is already one generic text that says

@ramast closing as per @shellscape's comment and the existing link. I think the existing text says enough, if users want to utilize `http-proxy-middleware` features they can go dig into the documentation there.

Mizziw:
I still haven't spoken with my lawyer yet and will get back to you after the new year

Mizziw:
I still haven't spoken with my lawyer yet and will get back to you after
the new year

skipjack:
@Mizziw huh? Was this sent by accident? I'm going to delete these comments unless there's something I'm missing...

Mizziw:
Just wanted to make sure that all the basis are covered

Mizziw:
And transition is smooth

PR #1764, opened 2017-12-27
docs(guides): remove useless code

Dream4ever:
The reference of `./print.bundle.js` is unnecessary, because it is already imported in ap.bundle.js.
Remove the reference will not change the result.


skipjack:
@Dream4ever this is a good point but removing it kind of changes the whole guide. If anything the `CommonsChunkPlugin` should be included which would wipe out the duplicated code in the main bundle. This is one naive method of _Code Splitting_ but was an easy way to generate multiple bundles to demonstrate use of the `HTMLWebpackPlugin`.

cc @TheDutchCoder

TheDutchCoder:
@Dream4ever You're technically correct (the best kind of correct!), but I agree that this would actually change the whole guide a bit.

I'm actually a proponent of doing that, as I now realize the example is a bit poor (importing print.js twice), so could you maybe to a more thorough review of this guide where the whole guide is "fixed" from that point forth?

if not I can probably take care of it one of these days, though I'm a bit preoccupied with our newborn ;)

skipjack:
I'm gonna close due to inactivity but if someone wants to revive this based on the comments above I'm all for it.

PR #1765, opened 2017-12-27
docs: Fix formatting

yangshun:
`&nbsp;` shouldn't be used as they are wider than normal spaces. In some cases, the `&nbsp;`s aren't necessary. The rest can be replaced with `{' '}` instead.

### Before

screen shot 2017-12-26 at 9 29 29 pm

Note the spacing between (`can` and `be`) and (`learn` and `different`) is wider than usual.

### After

screen shot 2017-12-26 at 9 30 42 pm

### DIFF
--- a/src/content/guides/migrating.md
+++ b/src/content/guides/migrating.md
@@ -544,7 +544,7 @@ webpack now supports template strings in expressions. This means you can start u
 
 ### Configuration Promise
 
-webpack now supports returning a `Promise` from the configuration file. This allows to do async processing in you configuration file.
+webpack now supports returning a `Promise` from the configuration file. This allows to do async processing in your configuration file.
 
 **webpack.config.js**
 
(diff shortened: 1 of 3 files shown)


src/content/guides/migrating.md
```diff
@@ -544,7 +544,7 @@ webpack now supports template strings in expressions. This means you can start u
 
 ### Configuration Promise
 
-webpack now supports returning a `Promise` from the configuration file. This allows to do async processing in you configuration file.
+webpack now supports returning a `Promise` from the configuration file. This allows to do async processing in your configuration file.
```
skipjack: We should probably clean up that last sentence even more:

> This allows async processing in your configuration file.

Or even combine the whole thing to one:

> webpack now supports returning a `Promise` from the configuration file which allows async processing in your configuration file.

PR #1773, opened 2018-01-06
docs(concepts): Direct new-comers to read about JS modules

tamirSwag:
Added a direction to learn about JS modules when reading the concepts page, meant mostly for new-comers. Because the modules doc page does not discuss the JS module systems, I also added a further reading link on this subject.

This is after a colleague of mine had an issue understanding what and why about webpack when reading the docs. What he was missing is an understanding of JS modules. You can't understand a *module* bundler without knowing modules.

### DIFF
--- a/src/content/concepts/index.md
+++ b/src/content/concepts/index.md
@@ -12,7 +12,7 @@ contributors:
 
 At its core, *webpack* is a _static module bundler_ for modern JavaScript applications. When webpack processes your application, it recursively builds a _dependency graph_ that includes every module your application needs, then packages all of those modules into one or more _bundles_.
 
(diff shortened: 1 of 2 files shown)


src/content/concepts/index.md
```diff
@@ -12,6 +12,8 @@ contributors:
 
 At its core, *webpack* is a _static module bundler_ for modern JavaScript applications. When webpack processes your application, it recursively builds a _dependency graph_ that includes every module your application needs, then packages all of those modules into one or more _bundles_.
 
+T> Learn more about JavaScript modules and webpack modules [here](/modules).
```
skipjack: Shouldn't the link be to `/concepts/modules` -- e.g. [here](https://webpack.js.org/concepts/modules).

PR #1734, opened 2017-12-08
 docs(config) update how to write the webpack configuration with typescript

youta1119:
issue #1735 
Update how to write the webpack configuration with typescript. Current configuration is wrong.
ts-node dosenot  support the module syntax of other than commonjs module. 
So, if module in compilerOptions in tsconfig.json is not commonjs, webpack getting error. like this.

```bash
(function (exports, require, module, __filename, __dirname) { import * as path from "path";
                                                              ^^^^^^

SyntaxError: Unexpected token import
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:588:28)
    at Module.m._compile (/Users/hogefuga/webpack-config-with-typescript/node_modules/ts-node/src/index.ts:392:23)

```
I have a two solution.
1. modify tsconfig.json
1. install tsconfig-paths

### First solution:

edit module in compilerOptions in tsconfig.json. please setting commonjs.

### Secoud solution:

first install tsconfig-paths

```bash
npm install --save-dev tsconfig-paths
```

**ts-node can resolve tsconfig.json using the environment variable provided by tsconfig-path.**

and setting environment variable `process.env.TS_NODE_PROJECT` provided by tsconfig-path like this.

```json
//tsconfig-for-webpack-config.json
{
  "compilerOptions": {
    "module": "commonjs"
  }
}

//package.json
{
  "scripts": {
      "build": "TS_NODE_PROJECT=\"tsconfig-for-webpack-config.json\" webpack"
  }
}
```
I make sample project. please see [this repository](https://github.com/youta1119/webpack-config-with-typescript)

PR #1775, opened 2018-01-06
Update configuration-languages.md

nielsbot:
I could not get this to work as-is... For me (macOS) TS_NODE_PROJECT is not set in the context of webpack. But using `cross-env` works. Found in this SO answer: https://stackoverflow.com/a/37356976/210171

### DIFF
--- a/src/content/configuration/configuration-languages.md
+++ b/src/content/configuration/configuration-languages.md
@@ -77,7 +77,7 @@ __package.json__
 ```json
 {
   "scripts": {
-    "build": "TS_NODE_PROJECT=\"tsconfig-for-webpack-config.json\" webpack"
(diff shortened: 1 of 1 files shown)


skipjack:
Hmmm, @youta1119 made these changes in #1734. @youta1119 does this change look right to you?

cc @dmitriid

@nielsbot if they agree that `cross-env` is necessary then we should also add `cross-env` the `npm install` statement above. (I don't use TypeScript myself, so I rely on our other TypeScript contributors to test and validate).

dmitriid:
I think `TS_NODE_PROJECT=\"tsconfig-for-webpack-config.json\"` should be `TS_NODE_PROJECT=tsconfig-for-webpack-config.json`, and then it should work without `cross-env`

I could be wrong though :)

PR #1777, opened 2018-01-09
Typo: Should this be import not export?

lauriejones:
Tiny thing, but this made more sense to me with `imported` rather than exported. 

I could be wrong though 🙂


lauriejones:
Thanks @skipjack 👍

PR #1785, opened 2018-01-13
Update module-factories.md

dear-lizhihua:
image

### DIFF
--- a/src/content/api/module-factories.md
+++ b/src/content/api/module-factories.md
@@ -1,6 +1,6 @@
 ---
 title: Module Factories
-group: Plugins
+group: Modules
 sort: 5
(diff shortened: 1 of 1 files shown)


skipjack:
@dear-lizhihua that page actually does belong in the "Plugins" section though I see how it can be confusing. Note that I do have a [plugin-api-rewrite]() branch open that refactors and standardizes the whole plugins section which should make things more clear. I took out the pages that mostly contained `TODO`s but will re-add some of those pages like this one later on (as the factories are `tapable`).

I'll make sure that it's clear why this page is in the "Plugins" section and maybe rename it to "Factories" to prevent confusion.

PR #1790, opened 2018-01-14
Remove unneeded file property

sokra:
This causes HTML to be really huge

reported by @lcxfs1991


dear-lizhihua:
image
`npm run build`, show error: 'page.file.attributes', i think you should delete `sort` function

PR #1792, opened 2018-01-15
docs(config): Add caveat about proxying request to root

yetithefoot:
_If `devServer.index` is not specified, requests to root not handled by proxy feature. There is no notes in docs about that, so its can be confusing (was happened with me). 

I added simple note to `devServer.proxy` section about specifying `devServer.index`_.


skipjack:
@shellscape let me know what you think.

@yetithefoot if @shellscape (webpack-dev-server maintainer) is good with this, then we can discuss grammar and fixing the build errors.

yetithefoot:
Uh, sorry. I'm not a native speaker, I rely on Grammarly here. 👨‍🎓 

This was my case: I used `webpack-dev-server` to proxy all requests to backend, but index file that should be served by node backend wasn't proxying by some reasons, I tried to find option in proxy config docs, but then found myself reading `webpack-dev-middleware` sources. Just tried to sum up this experience with this PR.

shellscape:
@skipjack unfortunately I stepped away from wds. probably best to follow up with @TheLarkInn or @sokra now.

skipjack:
@shellscape good to know, thanks. I'll reach out to them.

PR #1803, opened 2018-01-26
Add note on how plugins can intercept stats

nmonti:
I was trying to suppress some stats in my project which used the webpack Node API, so I set my stats object according to that documentation. It wasn't working, and I had no idea why. That's [until I asked on SO](https://stackoverflow.com/questions/48452322/webpack-builds-twice-and-ignores-stats-option/48453063) and was told webpack-dev-middleware intercepts stats. It's not obvious that happens, so I want to add this note for devs to be aware of such things.

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -13,6 +13,8 @@ The `stats` option lets you precisely control what bundle information gets displ
 
 T> For webpack-dev-server, this property needs to be in the `devServer` object.
 
+T> Some plugins, such as `webpack-dev-middleware`, intercept `stats`. Therefore you must set `stats` in that plugin's config.
+
 W> This option does not have any effect when using the Node.js API.
 
 ## `stats`


src/content/configuration/stats.md
```diff
@@ -13,6 +13,8 @@ The `stats` option lets you precisely control what bundle information gets displ
 
 T> For webpack-dev-server, this property needs to be in the `devServer` object.
 
+T> Some plugins, such as `webpack-dev-middleware`, intercept `stats`. Therefore you must set `stats` in that plugin's config.
```
skipjack: I'm a little confused about what this is suggesting. Would this be targeted at plugin authors who want to intercept `stats`? E.g. _why_ would I want to set `stats` in that plugin's config?

nmonti: I set `stats` to hide all warnings in my webpack config, but I was still getting warnings. I then tried to suppress all stats because I thought I was going crazy, but that didn't work either. Finally I discovered that `webpack-dev-middleware` was the plugin that was displaying the stats and overriding what I set in my webpack config. This was annoying to track down and not plainly obvious, especially for devs who are jumping into a complex webpack project

PR #1804, opened 2018-01-27
Docs update for relative output path

chrisdothtml:
Docs update for [webpack#6214](https://github.com/webpack/webpack/pull/6214)


skipjack:
Thanks for the follow through. Please ping us when the main PR is merged.

skipjack:
Ah, actually upon reading through this again and looking at the other PR I'm not sure this change makes sense. First, `path.resolve` is used not only to make the path absolute but also to allow compatibility between windows and mac/unix/linux builds as the `./` only works on the latter set of systems. Secondly, I'm not sure everyone wants their `output` directory to be relative to the `context`. For example, many times `./src` is used as `context` and `./dist` should be located at the same level as `src`, not nested within it.

Not saying nobody would use this new feature but I think for that reason I'd rather leave the example as is. cc @jeremenichelli @montogeek @EugeneHlushko

chrisdothtml:
@skipjack seems like your comment is against the feature (rather than the docs update) and is better suited for the [main PR](https://github.com/webpack/webpack/pull/6214). I've been dying for this kind of feedback on it since it's been sitting silent for months, despite it being approved and ready for merge.

However, If it *does* get merged (I'm doubtful at this point), and the nature of how `context` is used with `output.path` changes, surely the docs should be updated to reflect that, though, right?

skipjack:
> @skipjack seems like your comment is against the feature (rather than the docs update) and is better suited for the main PR.

Actually, I don't think it's a terrible feature. I just probably wouldn't use it too frequently myself and I don't think we should start recommending it. But to your point, if it did get merged, I think we should note it in the appropriate places within the __Configuration__ section.

> I've been dying for this kind of feedback on it since it's been sitting silent for months, despite it being approved and ready for merge.

Yeah, sorry I'm not a maintainer on the main repo (and slowly phasing myself off of here, I just stop by to help out when I can).

> However, If it does get merged (I'm doubtful at this point), and the nature of how context is used with output.path changes, surely the docs should be updated to reflect that, though, right?

Yeah, see my comment above. I just wouldn't recommend updating it everywhere in the docs (e.g. in __Guides__, __Concepts__, etc.). A few small additions in the __Configuration__ section should be adequate. Let's re-open if your main PR is merged.

chrisdothtml:
Fair enough. Will ping you if it's merged

jeremenichelli:
Hi @chrisdothtml, thanks for the PR and sorry for the waiting. I second Greg about not modifying the config paths, I´m pretty sure it's gonna generate confussion or issue reports.

If the main PR to core gets merge I think we should add a section or extend the description on the Configuration page.

nyngwang:
> Ah, actually upon reading through this again and looking at the other PR I'm not sure this change makes sense. First, `path.resolve` is used not only to make the path absolute but also to allow compatibility between windows and mac/unix/linux builds as the `./` only works on the latter set of systems. Secondly, I'm not sure everyone wants their `output` directory to be relative to the `context`. For example, many times `./src` is used as `context` and `./dist` should be located at the same level as `src`, not nested within it.
> 
> Not saying nobody would use this new feature but I think for that reason I'd rather leave the example as is. cc @jeremenichelli @montogeek @EugeneHlushko

This sounds reasonable, but you missed the point that I can set the `context` to the project root. then everything inside `entry` will be relative paths like `src/...`, and those inside `output` will be relative paths like `dist/...`.

Wait for your explanation.

PR #1817, opened 2018-02-02
Update docs with info-verbosity option

EugeneHlushko:
Add the newly added `info-verbosity` to the docs. Reference to addition: webpack/webpack-cli#200

### DIFF
--- a/src/content/configuration/watch.md
+++ b/src/content/configuration/watch.md
@@ -71,6 +71,13 @@ poll: 1000 // Check for changes every second
 T> If watching does not work for you, try out this option. Watching does not work with NFS and machines in VirtualBox.
 
 
+## `info-verbosity`
+
+`string`: `none` `info` `verbose`
+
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Hi man @skipjack 
Can you review this one? We did close a linked issue in webpack repo but the doc is not showing how to use the flag without this pr

EugeneHlushko:
I've rebased this on as it got too old :/

EugeneHlushko:
This landed into master, so its live in latest cli version @skipjack

skipjack:
Yeah, my question is more that it doesn't seem like `webpack-cli` is shipped with `webpack` in `v3`:

https://github.com/webpack/webpack/blob/v3.11.0/bin/webpack.js

In `v4` I think it's required as a separate install. So I might switch this to `next` seeing as `v4` actually requires users to use `webpack-cli`. Let me know if this makes sense or not, if yes then I'll repoint it to `next` and merge.

EugeneHlushko:
very true, i forgot that 3.11 still not sing cli. So `next` it is @skipjack

PR #1818, opened 2018-02-02
npm run start instead of npm start

elkhan:
_describe your changes..._

### DIFF
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -235,7 +235,7 @@ __package.json__
   }
 ```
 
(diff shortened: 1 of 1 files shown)


pierreneter:
What is the problem if we keep `npm start`? I see that `npm run start` is equivalent to` npm start`.

elkhan:
It is definitely not a big deal just caught my eyes because earlier in the Getting Started part it is `npm run build`. I thought of consistency.

skipjack:
@elkhan thanks for your contribution. I see your point re consistency but I think `npm start` is used commonly enough and is slightly more convenient because it's shorter. People can always take a look at NPM's docs to learn the difference between built-in scripts and `run` scripts. Maybe integrating a link to their docs on this in the paragraph would be more appropriate. Closing but feel free to re-open if you add a link.

PR #1827, opened 2018-02-09
 Add reference to UglifyJsPlugin dependency

neilkennedy:
Added a dependency section to mention that a source map will not be generated if the user is also using the `UglifyJsPlugin` and has not set the `sourceMap` property to `true`.

### DIFF
--- a/src/content/plugins/source-map-dev-tool-plugin.md
+++ b/src/content/plugins/source-map-dev-tool-plugin.md
@@ -40,6 +40,7 @@ The `fileContext` option is useful when you want to store source maps in an uppe
 
 T> Setting `module` and/or `columns` to `false` will yield less accurate source maps but will also improve compilation performance significantly.
 
+W> Remember that when using the [`UglifyJSPlugin`](/plugins/uglify-js-plugin), you must utilize the `sourceMap` option.
 
 ## Examples
 
@@ -92,17 +93,4 @@ Will produce the following URL:
 
 ``` js
 https://example.com/project/sourcemaps/bundle-[hash].js.map`
-```
-
-## Dependencies
-
-If you are also using `UglifyJsPlugin` then source maps will only be generated if you set the `sourceMap: true` property.
-
-``` js
-plugins: [
-    new webpack.optimize.UglifyJsPlugin({ comments: false, sourceMap: true }),
-    new webpack.SourceMapDevToolPlugin({
-        filename: '[file].map'
-    })
-]
 ```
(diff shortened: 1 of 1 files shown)


jeremenichelli:
Agree with this being something that needs to be present in the plugin documentation and not here. Would be good to check if this is clarified there and submit a PR with changes if it's not.

Good to have a link to the plugin's documentation in case we don't.

neilkennedy:
Hi. I believe the UglifyJsPlugin is pretty clear that you need to set the boolean to true to generate source maps. The problem that I encountered was that I didn't know to even look at that plugin to solve my problem. It wasn't obvious to me that a plugin for uglifying my JS would cause my sourcemaps to not generate.

montogeek:
@neilkennedy I see your confusion, you expect webpack do handle source maps _natively_ not through a plugin.

The note could be useful

neilkennedy:
Not necessarily natively, which I think it can do, but in my situation I added this plugin to help me generate source maps and I followed all the instructions but I didn't get any source maps generated. I didn't expect an unrelated plugin that I was also using to uglify my code would have an affect on the source map plugin.

It just seemed helpful to me to have something on this page to list any plugins that might have an affect on the output of this plugin.

montogeek:
Which instructions did you followed?

montogeek:
At the end of this https://webpack.js.org/configuration/devtool/ guide it says:
> When using the uglifyjs-webpack-plugin you must provide the sourceMap: true option to enable SourceMap support.

That page is the first Google result when you google "webpack source maps"

neilkennedy:
Yes, it was actually that page that lead me to the solution in the end. But when I first went to that page the first paragraph says

> Use the SourceMapDevToolPlugin for a more fine grained configuration

So I clicked on the link to that plugin and continued from there, skipping the rest of the `devtools` page because I assumed the SourceMapDevToolPlugin would do everything I needed it to.

neilkennedy:
@skipjack I definitely understand your hesitation, it's a slippery slope. But as webpack gets bigger and more and more plugins get added, it's going to be difficult to keep track of all of the inter-dependencies. It would be nice if the UglifyJsPlugin didn't have any affect on source map creation at all, but seen as it does, I do believe it's useful to point that out.

skipjack:
> It would be nice if the UglifyJsPlugin didn't have any affect on source map creation at all, but seen as it does, I do believe it's useful to point that out.

Yeah, I understand where you're coming from. With the update, I'm comfortable merging this. Reading your comment though made me realize that maybe there is a more appropriate fix: Maybe `sourceMap` should just default to `true`? I would recommend reaching out to [UglifyJSPlugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) maintainers by creating an issue over there (and feel free to link back to this discussion for reference).

Maybe there's something we don't know but it does seem like it would be more intuitive for the plugin to just handle them by default are at least do it if it sees that one is specified in the bundle it's minifying. Seems like turning it off is more of a performance thing though I'm not sure what the majority of people would expect when it comes to minification/uglification.

PR #1828, opened 2018-02-13
AngularJS + Webpack

var-bin:
Hi there,

In [this article](https://medium.com/@var_bin/angularjs-webpack-lazyload-bb7977f390dd), I’d like to show you how to realize lazyLoad modules for AngularJS app using Webpack, [UI-Router](https://ui-router.github.io/ng1/), and [ocLazyLoad](https://oclazyload.readme.io/). All behind-the-scene magic will be done with the help of ocLazyLoad tool. Last, but not least in the last part of the article I've shown how to use [Dynamic imports](https://github.com/tc39/proposal-dynamic-import).

P.S. Feedback is appreciated, it should be less than ~7 min reading.

Best regards,
Vitalii


var-bin:
Hi @skipjack,

Thank you for your review and approve. I quite agree with your comment according to the title, but I prefer to not change it. Basically, webpack does lazyLoad for AngularJS modules/services/controllers/etc with errors. These errors appear because of a [DI](https://docs.angularjs.org/guide/di). [oc.lazyLoad](https://oclazyload.readme.io/docs) module helps to do lazy loading in the right way.

Cheers ;)

PR #1833, opened 2018-02-15
Start adding Nonce documentation

### LINKED ISSUE #1802: Document __webpack_nonce__
There's a clever bit of functionality built into webpack for Content Security Policy purposes, enabled via `__webpack_nonce__`. It'd be great to formally document it and give examples for why it's useful.

https://github.com/webpack/webpack/blob/ab2270263e9af5b0dd83e454c20ab3aa1797424a/lib/APIPlugin.js#L26

https://stackoverflow.com/questions/42922784/what-s-the-purpose-of-the-html-nonce-attribute-for-script-and-style-elements

EugeneHlushko:
Add basic nonce documentation.

Fixes #1802

### DIFF
--- a/src/content/guides/csp.md
+++ b/src/content/guides/csp.md
@@ -1,5 +1,5 @@
 ---
-title: Content Security Policy
+title: Content Security Policies
 sort: 17
 contributors:
   - EugeneHlushko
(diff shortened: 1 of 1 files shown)


probablyup:
Looking good! A few notes to add to the docs:

1. The nonce should always be a base64-encoded string.

2. A corresponding header (`Content-Security-Policy`) or meta tag (`<meta http-equiv="Content-Security-Policy" ...>`) needs to be sent with the document to instruct the browser to enable CSP. 
  
This article goes into a lot of great detail on the subject: https://www.troyhunt.com/locking-down-your-website-scripts-with-csp-hashes-nonces-and-report-uri/ and [MDN also has a great article](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).

It'd also be good to include an example of a CSP that allows loading JS from a trusted CDN, since that's a common use case for FE projects:

```
Content-Security-Policy: default-src 'self' trusted.cdn.com
```

EugeneHlushko:
I've added more links and useful information, thank for nice catch @probablyup 

cc @TheLarkInn can anyone review docs PRs they seem to hang in the air :(

devdigital:
How is this intended to be used when a nonce is created *per request*? Is this just a static nonce at build time?

EugeneHlushko:
depends on your set up but a new nonce needed per request

devdigital:
This is a global variable checked at run-time - i.e. you just need to set `window.__webpack_nonce__` before your output webpack bundle executes on the page. This isn't very obvious from the current docs - the term *entry script* and example showing the `React` import could be misread as a build time setting.

probablyup:
CSP is somewhat complicated in that if it’s set up right, an attacker can’t
get a script on the page in the first place to run with the pregenerated
nonce.

It is true that nonce is better generated dynamically per page load, but
that isn’t workable if a CDN is part of your deployment process. That’s why
a combination of default src self + your CDN host + a nonce works decently
in this case.

Injected ads and stuff like that won’t match the two allowed hosts and if
it somehow does, it certainly won’t have the correct nonce since it isn’t
allowed to execute JS on the page to detect it. Therefore, the only
loophole I can think of is if someone malicious injects code through your
build process.

devdigital:
What does webpack do with the nonce value? as it appears this var is also used by some CSS-in-JS libs at runtime when they add inline styles etc.

probablyup:
It is appended to async-loaded chunk script tags and some libraries use it
for dynamically injected style tags, like styled-components.

html-webpack-plugin might (should) also use it if set.

devdigital:
Ok, so would the guidance be to reduce the scope of the static nonce to just webpack and those libraries (e.g. glamor, styled-components) that use it, and use another nonce (created *per request*) for other inline scripts? and therefore use multiple `nonce` sources in the content security policy?

probablyup:
Oh I suppose that could work. I actually wasn’t aware you could do multiple
nonce.

devdigital:
I haven't tried, but a static nonce used for all inline scripts seems to bypass the purpose of a nonce?

frontendphil:
@skipjack @probablyup @EugeneHlushko is someone working on a follow-up? Because currently, the documentation leaves me with more questions than answers. I think @devdigital raised some important question. Where can I put that `__webpack_nonce__` so that `webpack` can pick it up? It looks like a build-time config even though it should be something that I could inject at least on every page load. What are the hooks, mechanisms, etc.?

EugeneHlushko:
I think this can be achieved using an SSR set-up, but probably not out of the box, I would suggest filing an issue in webpack/webpack and i (or someone from the team) will pick it up shortly @frontendphil

steffektif:
@EugeneHlushko might [this](https://github.com/styled-components/styled-components/issues/887) be a suitable workaround?

EugeneHlushko:
@steffektif the last comment seems pretty much what i thought of but in the light of `styled-components`

PR #1837, opened 2018-02-17
Fix bundle.js path, update code snippets

dhruvdutt:
Update snippets
- Update webpack version 
- Fix bundle.js path issue

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -120,7 +120,7 @@ Now run your build command:
 npm run build
 
 Hash: 9a3abfc96300ef87880f
-Version: webpack 2.6.1
+Version: webpack 3.11.0
 Time: 834ms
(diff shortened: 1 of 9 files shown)


skipjack:
@dhruvdutt thanks for this submission but there are some issues with these changes. First, to update the webpack version, you would need to re-run the examples with the latest version of webpack as `3.11.0`s output might be somewhat different from what previous versions' output. The `./dist/bundle.js` change is invalid due to the fact the the `index.html` file is supposed to be moved within `/dist`, see the __project__ diff in the __Creating a Bundle__ section:

https://github.com/webpack/webpack.js.org/blob/master/src/content/guides/getting-started.md#creating-a-bundle

Closing as this would require a fair amount of changes but feel free to re-open if you have the time to update the full output and remove the path change.

Issue #1838, opened 2018-02-18
Include recognition sections for translators


hangpark:
Contributors of each page of documents are listed at the bottom. I'm starting to translate the doc into Korean at [webpack-korea/webpack.js.org](https://github.com/webpack-korea/webpack.js.org) and I have a plan to add recognition sections for translators just as contributors.

This enhancement can be applied in that repo only, but it seems good if this change applied into `translate` branch so that makes each translating project of various languages can show translators as a default functionality. I believe this would make translators more enthusiastic on their works.

Implementation is quite simple since it would be done by using already-implemented `Contributors` component.

PR #1839, opened 2018-02-18
[#1838] Include recognition sections for translators

hangpark:
Please read #1838.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -96,6 +96,8 @@ contributors:
   - bebraw
   - Jhnns
   - SpaceK33z
+translators:
+  - hangpark
 ---
 
 Some documentation...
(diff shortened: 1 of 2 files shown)


dear-lizhihua:
@Hangeer @skipjack @lcxfs1991 @Lutece
Added contributor in the markdown file, easily lead to the problem of line number does not correspond. It may not be convenient to merge branch and diff lines operations.

skipjack:
> Added contributor in the markdown file, easily lead to the problem of line number does not correspond. It may not be convenient to merge branch and diff lines operations.

@dear-lizhihua so do you think we should __not__ merge? I see your point but I also see the need for acknowledging the work, maybe there's a way to ignore frontmatter when diffing? Or at least exclude the `translators` section?

dear-lizhihua:
I agree with your point, just some my ideas, if there is a way to exclude it is better, but you can merge this pull request directly.

PR #1840, opened 2018-02-18
Update cli.md

Legends:
Added parameter `--display-error-details` to section Debug Options:
`--display-error-details	Display details about errors	boolean	false`

It is mentioned already in the "Stats Options", but deserves to be in under debug options, very useful parameter.


TheDutchCoder:
Hi @Legends thanks for the contribution!

I agree that it's a useful parameter to list, so let's get this merged! :)

PR #1841, opened 2018-02-18
Update getting-started.md

Legends:
This addition would have helped me a lot:

Add `--display-error-details` for detailed error messages, .i.e. `npm run build --display-error-details`.
Without this flag development is a black box and painful


skipjack:
I think the addition in `cli.md` from #1840 probably makes sense but I don't think it's necessary to add it to this guide. Also, just fyi, to pass parameters to an npm script you need an additional `--` so it knows it isn't an `npm run` flag, i.e.:

``` bash
npm run build -- --display-error-details
```

PR #1844, opened 2018-02-21
Update code-splitting.md

nicky-lau:
add dynamic import plugin


montogeek:
Hi @nicky-lau, is that plugin necessary?

montogeek:
Also, this would likely change for webpack 4 since `CommonsChunkPlugin` was removed entirely

nicky-lau:
@montogeek yes,this plugin is necessary, if not using it, the following error occurs at webpack 3.11.0:
```
Module build failed: SyntaxError: 'import' and 'export' may only appear at the top level
```

also, `dynamic import` is just for SPA, but `CommonChunkPlugin` is working for mulit entry app,

skipjack:
@nicky-lau have you specified `modules: false` in your `.babelrc` to allow webpack to process ES6 module syntax? I use dynamic `import()`s a lot and haven't needed that plugin.

nicky-lau:
@skipjack thank ur answer.
here is my `.babelrc`:
```
{
	"presets": [
		["env", {
			"targets": {
				"browsers": ["last 2 versions", "> 1%"]
			},
			"modules": false
		}]
	]
}
```
entry file: 
``` javascript
const arr = [1, 6, 3, 4, 5, 2];

if (arr.includes(2)) {
	import(/* webpackChunkName: "lodash" */'lodash')
	.then(_ => {
		console.log(_.join(['hello', 'world', '!'], ' '));
	})
	// require.ensure([], () => {
	// 	const _ = require('lodash');
	// 	console.log(_.join(['Hello', 'World', '!'], ' '));
	// }, 'lodash')
}
```
error info: 
``` 
Module build failed: SyntaxError: 'import' and 'export' may only appear at the top level
```

skipjack:
@nicky-lau could you create a demo repo? @montogeek can you take a look at that with @nicky-lau can see if it's something else that needs fixing and if not we can continue this PR to get it merged? Embarrassed to say most of my projects still use the `es2015` and `stage-*` presets so maybe it has something to do with `babel-preset-env` (which I know is now the recommended practice).

montogeek:
It would say that the plugin is necessary to allow Babel to parse those `import`s.
@skipjack `babel-preset-env` does not really add plugins automatically, it only add polyfills and transforms.

src/content/guides/code-splitting.md
```diff
@@ -126,6 +131,20 @@ __webpack.config.js__
       index: './src/index.js',
       another: './src/another-module.js'
     },
++  module: {
++     rules: [
++       {
++         test: /\.js$/,
++         exclude: /node_modules/,
++         ues: {
```
dhruvdutt: Type ~`ues`~ => `uses`

src/content/guides/code-splitting.md
```diff
@@ -126,6 +131,20 @@ __webpack.config.js__
...
++           loader: 'babel-loader',
```
dhruvdutt: Can you please keep consistency in object keys quotes? Some have double, some have single.

src/content/guides/code-splitting.md
```diff
@@ -126,6 +131,20 @@ __webpack.config.js__
...
++           "plugins": [
++           "syntax-dynamic-import" // important!
```
dhruvdutt: Wrong indentation.

src/content/guides/code-splitting.md
nicky-lau: :+1:

src/content/guides/code-splitting.md
```diff
@@ -126,6 +131,18 @@ __webpack.config.js__
       index: './src/index.js',
       another: './src/another-module.js'
     },
++  module: {
```
dhruvdutt: Wrong indentation.

nicky-lau: thank you for pointing that.

dhruvdutt (dismissed):
LGTM. Please rebase.

src/content/guides/code-splitting.md
```diff
@@ -114,6 +114,11 @@ The first of these two points is definitely an issue for our example, as `lodash
 
 The [`CommonsChunkPlugin`](/plugins/commons-chunk-plugin) allows us to extract common dependencies into an existing entry chunk or an entirely new chunk. Let's use this to de-duplicate the `lodash` dependency from the previous example:
 
+First, let's install this plugin: `syntax-dynamic-import`
+``` npm
+npm install --save-dev babel-plugin-syntax-dynamic-import
+```
```
jeremenichelli: Hi @nicky-lau thanks for the contribution. IT's important to note users that this is only needed if they are transpiling their code with babel which might not be the case.

So I suggest changing this to a `W>` blockquote to inform users that **if** they are using Babel, they will need to include this so the Babel understands the dynamic import syntax, and remove it from the code below. Would you be able to do this?

PR #1850, opened 2018-02-23
Added notification for webpack 4 beta availability w/links to articles

gricard:
Modified the NotificationBar to add an additional note about webpack 4 being in beta and some helpful links. 

Please let me know if the formatting/wording can be improved and I'll make the necessary changes.

Also, is there a possibility that if someone had previously dismissed the notification bar that it will then not show with this new information in it? Should I add a second NotificationBar instead?

### DIFF
--- a/src/components/NotificationBar/NotificationBar.jsx
+++ b/src/components/NotificationBar/NotificationBar.jsx
@@ -15,7 +15,7 @@ export default class NotificationBar extends React.Component {
           <p>
             WEBPACK 4 IS RELEASED, TRY IT TODAY, SEE THE CHANGELOG AND THESE ARTICLES: <br />
             <a target="_blank" href="https://github.com/webpack/webpack/releases/tag/v4.0.0">* Change Log</a><br />
-            <a target="_blank" href="https://medium.com/webpack/webpack-4-beta-try-it-today-6b1d27d7d7e2">* Install and setup Webpack 4</a><br />
+            <a target="_blank" href="https://medium.com/webpack/webpack-4-released-today-6cdb994702d4">* Install and setup Webpack 4</a><br />
             <a target="_blank" href="https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a">* New configuration options in Webpack 4</a><br />
           </p>
           <br />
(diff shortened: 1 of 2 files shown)


gricard:
Oh, and is there a changelog to link to yet? I couldn't find one.

skipjack:
There is a link to the changelog, check the right hand side of the footer. The changes in this PR look great but I would also recommend bumping the `version` variable at the top of the file as well:

``` diff
- const version = 2
+ const version = 3
```

This will force the notification to be displayed again for any users that dismissed previous ones.

skipjack:
Also, and maybe this should be in a separate PR, I think it's time to find an actual home in the site for the old notification (webpack clothing, stickers and such) and remove it from the notification bar. Somewhere in the __Contribute__ section or landing page would probably be best.

cc @jeremenichelli @TheLarkInn

gricard:
@skipjack Thanks. I've updated the version #. I'll see if I can make another PR this weekend to separate the messages.

src/components/NotificationBar/NotificationBar.jsx
```diff
@@ -12,6 +12,12 @@ export default class NotificationBar extends React.Component {
     return (
       <div className={ `notification-bar ${dismissedMod}` }>
         <Container className="notification-bar__inner">
+          <p>
```
TheLarkInn: Heh, since webpack v4 just went release, would you mind linking to this medium article instead, and then keep the other links. [This](https://medium.com/webpack/webpack-4-released-today-6cdb994702d4) is the new article, and you can link this instead.

gricard: Done

gricard: Done

src/components/NotificationBar/NotificationBar.jsx
```diff
@@ -12,6 +12,13 @@ export default class NotificationBar extends React.Component {
...
+            WEBPACK 4 IS RELEASED, TRY IT TODAY, SEE THE CHANGELOG AND THESE ARTICLES: <br />
+            <a target="_blank" href="https://github.com/webpack/webpack/releases/tag/v4.0.0">* Change Log</a><br />
+            <a target="_blank" href="https://medium.com/webpack/webpack-4-released-today-6cdb994702d4">* Install and setup Webpack 4</a><br />
```
dhruvdutt: It is a crime to write ***webpack*** as ~"Webpack"~. :male_detective:

jeremenichelli:
I think this is already out of date, thanks either way @gricard

PR #1860, opened 2018-02-27
Fix typo in the synchronous loader code example

loicplaire:
`this.callback` is called with `sourceMaps` instead of `map`.


montogeek:
Thanks for the correction!

PR #1862, opened 2018-02-27
Updated getting-started.md

TheDutchCoder:
* Changes the guide so it reflects webpack 4 output.
* Changes some wordings here and there and adds some clarifications.
* Adds `webpack-cli` install which is required with v4

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -71,6 +71,32 @@ __index.html__
 </html>
 ```
 
+We also need to adjust our `package.json` file in order to make sure we mark our package as `private`, as well as removing the `main` entry. This is to prevent an accidental publish of your code.
+
+T> If you want to learn more about the inner workings of `package.json`, then we recommend reading [the npm documentation](https://docs.npmjs.com/files/package.json).
+
+__package.json__
+``` diff
+  {
+    "name": "webpack-demo",
+    "version": "1.0.0",
+    "description": "",
++   "private": true,
+-   "main": "index.js",
+    "scripts": {
+      "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
+    "devDependencies": {
+      "webpack": "^4.0.1",
+      "webpack-cli": "^2.0.9"
+    },
+    "dependencies": {}
+  }
+```
+
 In this example, there are implicit dependencies between the `<script>` tags. Our `index.js` file depends on `lodash` being included in the page before it runs. This is because `index.js` never explicitly declared a need for `lodash`; it just assumes that the global variable `_` exists.
 
 There are problems with managing JavaScript projects this way:
@@ -98,13 +124,15 @@ __project__
     |- index.js
 ```
 
-To bundle the `lodash` dependency with `index.js`, we'll need to install the library locally...
+To bundle the `lodash` dependency with `index.js`, we'll need to install the library locally:
 
 ``` bash
-npm install --save-dev lodash
+npm install --save lodash
 ```
 
-and then import it in our script...
+T> As a rule of thumb: if you're installing a package that will be bundled into your rpdocution code, then you should use `npm install --save`. If you're installing a package to help development (e.g. a linter, testing libraries, etc.) then you should use `npm install --save-dev`. More information can be found in the [npm documentation](https://docs.npmjs.com/cli/install).
+
+Now, lets import `lodash` in our script:
 
 __src/index.js__
 
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -144,7 +144,7 @@ In this setup, `index.js` explicitly requires `lodash` to be present, and binds 
 With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `bundle.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
 
 ``` bash
-npx webpack src/index.js dist/bundle.js
+npx webpack src/index.js --output dist/bundle.js
 
 Hash: 857f878815ce63ad5b4f
 Version: webpack 3.9.1
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -20,16 +20,15 @@ Webpack is used to compile JavaScript modules. Once [installed](/guides/installa
 
 ## Basic Setup
 
-First let's create a directory, initialize npm, [install webpack locally](/guides/installation#local-installation), and install the webpack-cli (which we'll need to run webpack from the command line (CLI)):
+First let's create a directory, initialize npm, [install webpack locally](/guides/installation#local-installation), and install the webpack-cli (the tool used to run webpack on the command line):
 
 ``` bash
 mkdir webpack-demo && cd webpack-demo
 npm init -y
-npm install --save-dev webpack
-npm install --save-dev webpack-cli
+npm install webpack webpack-cli --save-dev
 ```
 
-T> Throughout the Guides we will use `diff` blocks to show you what changes we're making to directories, files, code, etc. For example: below we're adding `index.html` (file), `/src` (directory), and `index.js` (file).
+T> Throughout the Guides we will use `diff` blocks to show you what changes we're making to directories, files, and code.
 
 Now we'll create the following directory structure, files and their contents:
 
@@ -147,7 +146,7 @@ In this setup, `index.js` explicitly requires `lodash` to be present, and binds 
 With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `bundle.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
 
 ``` bash
-npx webpack src/index.js --output dist/bundle.js
+npx webpack
 
 Hash: dabab1bac2b940c1462b
 Version: webpack 4.0.1
@@ -181,7 +180,7 @@ Note that webpack will not alter any code other than `import` and `export` state
 
 ## Using a Configuration
 
-As of version 4, webpack doesn't require any configuration, but most projects will need a more complex setup, which is why webpack supports a [configuration file](/concepts/configuration). This is much more efficient than having to manually type in a lot of commands on the CLI, so let's create one to replace the CLI line options used above:
+As of version 4, webpack doesn't require any configuration, but most projects will need a more complex setup, which is why webpack supports a [configuration file](/concepts/configuration). This is much more efficient than having to manually type in a lot of commands in the terminal, so let's create one to replace the CLI line options used above:
 
 __project__
 
@@ -311,6 +310,6 @@ webpack-demo
 |- /node_modules
 ```
 
-T> If you're using npm 5, you'll probably also see a `package-lock.json` file in your directory as well.
+T> If you're using npm 5, you'll probably also see a `package-lock.json` file in your directory.
 
 If you want to learn more about webpack's design, you can check out the [basic concepts](/concepts) and [configuration](/configuration) pages. Furthermore, the [API](/api) section digs into the various interfaces webpack offers.
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -20,15 +20,18 @@ Webpack is used to compile JavaScript modules. Once [installed](/guides/installa
 
 ## Basic Setup
 
-First let's create a directory, initialize npm, and [install webpack locally](/guides/installation#local-installation):
+First let's create a directory, initialize npm, [install webpack locally](/guides/installation#local-installation), and install the webpack-cli (which we'll need to run webpack from the command line (CLI)):
 
 ``` bash
 mkdir webpack-demo && cd webpack-demo
 npm init -y
 npm install --save-dev webpack
+npm install --save-dev webpack-cli
 ```
 
-Now we'll create the following directory structure and contents:
+T> Throughout the Guides we will use `diff` blocks to show you what changes we're making to directories, files, code, etc. For example: below we're adding `index.html` (file), `/src` (directory), and `index.js` (file).
+
+Now we'll create the following directory structure, files and their contents:
 
 __project__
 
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.md
```diff
@@ -20,15 +20,18 @@ Webpack is used to compile JavaScript modules. Once [installed](/guides/installa
 
 ## Basic Setup
 
-First let's create a directory, initialize npm, and [install webpack locally](/guides/installation#local-installation):
+First let's create a directory, initialize npm, [install webpack locally](/guides/installation#local-installation), and install the webpack-cli (which we'll need to run webpack from the command line (CLI)):
```
jeremenichelli: Can we rephrase the sentence inside the parenthesis like _(the tool for running webpack on the command line)_? I feel it lighter and less over explanatory.

src/content/guides/getting-started.md
```diff
@@ -20,15 +20,18 @@ Webpack is used to compile JavaScript modules. Once [installed](/guides/installa
...
 
 ``` bash
 mkdir webpack-demo && cd webpack-demo
 npm init -y
 npm install --save-dev webpack
+npm install --save-dev webpack-cli
 ```
 
-Now we'll create the following directory structure and contents:
+T> Throughout the Guides we will use `diff` blocks to show you what changes we're making to directories, files, code, etc. For example: below we're adding `index.html` (file), `/src` (directory), and `index.js` (file).
```
jeremenichelli: I would erase the _For example..._, reader will continue with the lecture and will get into that alone, no need for anticipation in my opinion.

src/content/guides/getting-started.md
```diff
@@ -20,15 +20,18 @@ Webpack is used to compile JavaScript modules. Once [installed](/guides/installa
...
 ## Basic Setup
 
-First let's create a directory, initialize npm, and [install webpack locally](/guides/installation#local-installation):
+First let's create a directory, initialize npm, [install webpack locally](/guides/installation#local-installation), and install the webpack-cli (which we'll need to run webpack from the command line (CLI)):
 
 ``` bash
 mkdir webpack-demo && cd webpack-demo
 npm init -y
 npm install --save-dev webpack
+npm install --save-dev webpack-cli
```
jeremenichelli: We could join this statement as 

```sh
npm install webpack webpack-cli --save-dev
```

src/content/guides/getting-started.md
```diff
@@ -99,7 +102,7 @@ __project__
 To bundle the `lodash` dependency with `index.js`, we'll need to install the library locally...
 
 ``` bash
-npm install --save lodash
+npm install --save-dev lodash
```
jeremenichelli: This should be `npm install --save lodash` or bundling won't work.

src/content/guides/getting-started.md
```diff
@@ -173,7 +181,7 @@ Note that webpack will not alter any code other than `import` and `export` state
 
 ## Using a Configuration
 
-Most projects will need a more complex setup, which is why webpack supports a [configuration file](/concepts/configuration). This is much more efficient than having to type in a lot of commands in the terminal, so let's create one to replace the CLI options used above:
+As of version 4, webpack doesn't require any configuration, but most projects will need a more complex setup, which is why webpack supports a [configuration file](/concepts/configuration). This is much more efficient than having to manually type in a lot of commands on the CLI, so let's create one to replace the CLI line options used above:
```
jeremenichelli: I wouldn't replace terminal since it's a better represantion of the place where we use CLI statements.

src/content/guides/getting-started.md
```diff
@@ -146,18 +149,23 @@ With that said, let's run `npx webpack` with our script as the [entry point](/co
 ``` bash
 npx webpack src/index.js --output dist/bundle.js
```
jeremenichelli: @TheLarkInn @sokra what is the default input and output directory for webpack v4? I remember src/index.js being the default entry point if I'm not wrong. Would love to highlight the fact that you just need to run `webpack` in the terminal 🌟

src/content/guides/getting-started.md
```diff
@@ -230,14 +243,28 @@ Given it's not particularly fun to run a local copy of webpack from the CLI, we
...
-``` json
+``` diff
 {
-  ...
+  "name": "webpack-demo",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
   "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
```
jeremenichelli: Eliminate this line and description, keywords and maybe author or put something like `webpack support team`, to avoid noise and unnecessary lines in the example.

jeremenichelli:
⚠️  @TheDutchCoder can you change the base branch to `next` instead of `master`? We are merging everything related to v4 there first.

src/content/guides/getting-started.md
montogeek: Maybe we should wait for https://github.com/webpack/webpack/pull/6467? In that case it would be installed along webpack

src/content/guides/getting-started.md
montogeek: Entry https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L20

src/content/guides/getting-started.md
```diff
@@ -279,6 +311,6 @@ webpack-demo
 |- /node_modules
 ```
 
-T> If you're using npm 5, you'll probably also see a `package-lock.json` file in your directory.
+T> If you're using npm 5, you'll probably also see a `package-lock.json` file in your directory as well.
```
montogeek: Isn't this a redundancy: "also" and "as well" at the end?

src/content/guides/getting-started.md
jeremenichelli: I won't honestly, we can't block documentation for a PR that has conflicts and uncertainty around when it could be merged.

If that changes in the future then we can quickly come and change it.

montogeek: Yeah, it seems stalled :/

src/content/guides/getting-started.md
TheDutchCoder: Are you sure about that? It works fine in devDependencies (and it should).

This will only be an issue if you have an app that runs on Node and has a dependency on the server. Client-side bundlers should use devDependencies usually.

I just tested it and it works fine either way.

src/content/guides/getting-started.md
TheDutchCoder: Yeah the current setup should already be the default. I'll remove the options to highlight webpack's ease of use and default configuration ;)

src/content/guides/getting-started.md
TheDutchCoder: This is the default `package.json` file generated by `npm init -y`, I would rather keep it so people feel confident that they're doing the right thing.

I know some of this is a bit verbose to us, but a lot of beginners have expressed their desire for more verbose guides as they feel webpack (and FE dev in general) can be a bit daunting.

I'm okay with making this shorter again, but I don't think it hurts to show the file here.

TheDutchCoder:
@jeremenichelli whoops, I will base against `next`, sorry about that!

src/content/guides/getting-started.md
TheDutchCoder: Nice catch!

src/content/guides/getting-started.md
jeremenichelli: By definition, `dependencies` are needed at run time, doesn't matter if you're bundling for Node or the browser. 

In contrast, `devDependencies` are needed only while developing, again indifferent from where the bundled output is going to run.

Here `lodash` will actually run in the application and not used only during development.

Reference: https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies

TheDutchCoder: I'm fairly sure that in this case lodash gets bundled by webpack, in fact, it would make no sense to have webpack installed as a devDependency but lodash as a regular dependency.

It's a trivial thing, but I'm a proponent of keeping things consistent and I don't see a reason to treat lodash differently in this case.

If you want you can give it a shot yourself and install it as a dev dependency, webpack will still bundle it just fine. It's not being loaded externally.

Can someone else maybe pitch in on this?
It seems to be a growing trend where packages always get installed with `--save` where it's not required (in most cases).

jeremenichelli: @TheLarkInn @sokra word on this?

michael-ciniawsky: Both works for `webpack` since it doesn't care where the contents in `node_modules` are from :). IHMO something like `lodash` is a `dependency` of your `site/spa/...` once it's also included into 'production' code, so `npm i lodash` is the 'current' usage here

ChadKillingsworth: This is a highly confusing topic - but it's all on the perspective of what to expect when another package lists your webpack build as a dependency.

If lodash is bundled by webpack into the dist bundle, it's not needed by the upstream package. So it should be in `devDependencies`.

ljharb:
Commenting on the PR per our discussion in Slack:
 - a top-level app should have `private: true` set, to avoid accidental `npm publish`es, and should omit "main" entirely. Sadly, `npm init` doesn't first ask you if you're making an app or a module.
 - "deps" is for things that are *conceptually* used in production; the existence of a build process doesn't make them devDeps. Obviously it will *work* when they're all in devDeps, but that erases the value of conceptually separating them. One example: you can update all your devDeps at once with virtually zero risk; updating your production dependencies rarely has zero risk.
 - teaching newcomers about package.json is important; any repeated commands they use should go in "scripts", eg, and they definitely need to know about `private: true` and "main" when they create their next npm package.

src/content/guides/getting-started.md
montogeek: In this specific example `lodash` should be in `dependencies`, it is a library needed at runtime, that is bundled my webpack.
In real world examples/applications you never bundle eslint or prettier into your bundles, that's why those are installed as `devDependencies`

montogeek: When you say:
> To bundle the lodash dependency with index.js, we'll need to install the library locally...

It means that it is necessary at runtime, bundled.

Meanwhile
> npm install --save-dev lodash

says: `lodash` is used for development purposes, similar to Flow, Prettier or ESLint, in this is not the case.

And finally:
> and then import it in our script...

Our application script.

TheDutchCoder:
Okay thanks, that reasoning makes sense to me. I'll use these explanations
in the guide as well, it's good for people to understand this.

Thanks guys, much appreciated!

src/content/guides/getting-started.md
TheDutchCoder: Cool, this all makes sense!

Thanks so much for enlightening me on this subject. It has always bothered me and didn't make a lot of sense and I think a lot of beginners feel the same way.

I'll make sure to include a piece of this in the Guides as well!

🎉

TheDutchCoder:
@jeremenichelli ready for another review!

I reverted the lodash change and added more info about npm/package.json.
Let me know if there are other things you want me to tackle as well.

src/content/guides/getting-started.md
```diff
@@ -71,6 +71,32 @@ __index.html__
 </html>
 ```
 
+We also need to adjust our `package.json` file in order to make sure we mark our package as `private`, as well as removing the `main` entry. This is to prevent an accidental publish of your code.
```
jeremenichelli: I would honestly remove this paragraph, it isn't wrong but it feels out of place honestly. This is a quick start guide to webpack and I feel we are bombarding the user with tons of stuff that for a small project are not necessary.

Sorry if it sounds hard or agressive, it's not the intention. It's also hard to me to _garden_ the content that, though useful, doesn't belong to documentation, especially when it is aimed to people new at a tool or library.

jeremenichelli: I would leave the tip quote below though. We are communicating that there's more to know around a topic and pointing the user to the right place to look for it.

src/content/guides/getting-started.md
```diff
@@ -98,13 +124,15 @@ __project__
...
-To bundle the `lodash` dependency with `index.js`, we'll need to install the library locally...
+To bundle the `lodash` dependency with `index.js`, we'll need to install the library locally:
 
 ``` bash
-npm install --save-dev lodash
+npm install --save lodash
 ```
 
-and then import it in our script...
+T> As a rule of thumb: if you're installing a package that will be bundled into your rpdocution code, then you should use `npm install --save`. If you're installing a package to help development (e.g. a linter, testing libraries, etc.) then you should use `npm install --save-dev`. More information can be found in the [npm documentation](https://docs.npmjs.com/cli/install).
```
jeremenichelli: We could shorten this to:

_When installing a package for production or runtime bundles we use the `--save` flag. On the other hand, the `--save-dev` flag is for packages related to development tasks like linters or testing libraries. More information can be found in the [npm documentation](https://docs.npmjs.com/cli/install)._

What do you think?

src/content/guides/getting-started.md
TheDutchCoder: Yeah it was a suggestion by @ljharb . I agree that it should be as simple as we can make it. I can take it out for now (and leave the tip) and see if we get comments on it in the future. We could always add it in later.

src/content/guides/getting-started.md
TheDutchCoder: Hmm, apart from my typo (lol), I kind of like to be a bit explicit about it (for new users) and clearly communicate that libs used in the bundle fall under deps.

I don't think it can hurt in this case to be a bit verbose. I'd rather leave it this way, is that okay with you as well?

jeremenichelli: Yeah, it's ok. It's jsut me trying to keep tips as short as possible to avoid big color blocks and drive away the atention on the main content. How about:

_When installing a package that will be bundled into your production or runtime code, you should use `npm install --save`. If you're installing a package for development purposes (e.g. a linter, testing libraries, etc.) then you should use `npm install --save-dev`. More information can be found in the [npm documentation](https://docs.npmjs.com/cli/install)._

Just some gardening around the words 😊

src/content/guides/getting-started.md
jeremenichelli: 👍

src/content/guides/getting-started.md
ljharb: `s/rpdocution/production`

TheDutchCoder:
@jeremenichelli should be good now! If you can give it one more read through and then approve it if there's nothing more to change, then we can merge this in and move on to the next one ;)

PR #1864, opened 2018-02-28
fix anchor bug (chinese language)

dear-lizhihua:


montogeek:
@dear-lizhihua What was the bug and how this fixes it?

dear-lizhihua:
https://doc.webpack-china.org/guides/asset-management/
image

Ah, here, several chinese is replaced `-`

PR #1865, opened 2018-02-28
update tree shaking guide for webpack 4

probablyup:
updated the treeshaking documentation for webpack 4

### DIFF
--- a/src/content/guides/tree-shaking.md
+++ b/src/content/guides/tree-shaking.md
@@ -10,16 +10,10 @@ contributors:
   - dmitriid
   - probablyup
 related:
-  - title: Tree shaking with webpack 2, TypeScript and Babel
-    url: https://alexjoverm.github.io/2017/03/06/Tree-shaking-with-Webpack-2-TypeScript-and-Babel/
-  - title: Tree-shaking with webpack 2 and Babel 6
-    url: http://www.2ality.com/2015/12/webpack-tree-shaking.html
-  - title: webpack 2 Tree Shaking Configuration
-    url: https://medium.com/modus-create-front-end-development/webpack-2-tree-shaking-configuration-9f1de90f3233#.15tuaw71x
-  - title: Issue 2867
-    url: https://github.com/webpack/webpack/issues/2867
-  - title: Issue 4784
-    url: https://github.com/webpack/webpack/issues/4784
+  - title: "webpack 4 beta — try it today!"
+    url: https://medium.com/webpack/webpack-4-beta-try-it-today-6b1d27d7d7e2#9a67
+  - title: Debugging Optimization Bailouts
+    url: https://webpack.js.org/plugins/module-concatenation-plugin/#debugging-optimization-bailouts
 ---
 
 _Tree shaking_ is a term commonly used in the JavaScript context for dead-code elimination. It relies on the [static structure](http://exploringjs.com/es6/ch_modules.html#static-module-structure) of ES2015 module syntax, i.e. [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export). The name and concept have been popularized by the ES2015 module bundler [rollup](https://github.com/rollup/rollup).
--- a/src/content/guides/tree-shaking.md
+++ b/src/content/guides/tree-shaking.md
@@ -130,6 +130,18 @@ If your code did have some side effects though, an array can be provided instead
 }
 ```
 
+T> Note that any imported file is subject to tree shaking. This means if you use something like `css-loader` in your project and import a CSS file, it needs to be added to the side effect list so it will not be unintentionally dropped in production mode:
+
+```json
+{
+  "name": "your-project",
+  "sideEffects": [
+    "./src/some-side-effectful-file.js",
+    "*.css"
+  ]
+}
+```
+
 ## Minify the Output
 
 So we've cued up our "dead code" to be dropped by using the `import` and `export` syntax, but we still need to drop it from the bundle. To do that, we'll use the `-p` (production) webpack compilation flag to enable the uglifyjs minification plugin.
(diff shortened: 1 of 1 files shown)


probablyup:
cc @sokra 

I think uglify isn't technically needed anymore to do tree shaking right? But seems like a good recommendation to enable it anyway.

jeremenichelli:
Thanks @probablyup, can you change the base from `master`to `next` we are mergin everything there first. Thanks!

probablyup:
@jeremenichelli done

src/content/guides/getting-started.md
```diff
@@ -144,7 +144,7 @@ In this setup, `index.js` explicitly requires `lodash` to be present, and binds
 With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `bundle.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
 
 ``` bash
-npx webpack src/index.js dist/bundle.js
+npx webpack src/index.js --output dist/bundle.js
```
probablyup: not sure where this change came from...

src/content/guides/tree-shaking.md
```diff
@@ -107,23 +108,46 @@ function cube(x) {
 Note the `unused harmony export square` comment above. If you look at the code below it, you'll notice that `square` is not being imported, however, it is still included in the bundle. We'll fix that in the next section.
 
 
-## Minify the Output
+## Mark the file as side-effect-free
+
+In a 100% ESM module world, identifying side effects is straightforward. However, we don't live in that reality just yet, so it's necessary to provide hints to webpack's compiler on the "pureness" of your code.
```
jeremenichelli: Can you rephrase `we don't live in that reality just yet, so` to `we aren't there just yet, in the mean time`

src/content/guides/tree-shaking.md
```diff
@@ -107,23 +108,46 @@ function cube(x) {
...
+  "name": "your-project",
+  "sideEffects": false
+}
+```
 
-So we've cued up our "dead code" to be dropped by using the `import` and `export` syntax, but we still need to drop it from the bundle. To do that, we'll add a minifier that supports dead code removal -- the [`UglifyJSPlugin`](/plugins/uglifyjs-webpack-plugin) -- to our configuration...
+All the code noted above does not contain side effects, so we can simply mark the property as `false` to inform webpack that it can safely prune unused exports.
 
-Let's start by installing it:
+T> A "side effect" is defined as code that performs a special behavior when imported, other than simply exposing one or more exports. For instance: a polyfill which adjusts the global scope but does not provide an export or has one that you are not using in your codebase.
```
jeremenichelli: Can we rephrase this as:

```
A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports like a polyfill, which might affect the global scope and might not provide an export or has one that you are not using in your codebase.
```

src/content/guides/tree-shaking.md
```diff
@@ -107,23 +108,46 @@ function cube(x) {
...
+    "./src/some-side-effectful-file.js"
+  ]
+}
 ```
 
-And then adding it into our config:
+## Minify the Output
+
+So we've cued up our "dead code" to be dropped by using the `import` and `export` syntax, but we still need to drop it from the bundle. To do that, we'll use the `-p` (production) webpack compilation flag to enable the uglifyjs minification plugin.
+
```
jeremenichelli: @TheLarkInn @sokra is the -p flag still recommended in webpack 4? Or should we clarify that is for versions older than v4?

src/content/guides/tree-shaking.md
```diff
@@ -132,57 +156,21 @@ module.exports = {
...
-
-The code used in this guide assumes you perform tree-shaking using UglifyJS plugin. However, there are other tools such as [webpack-rollup-loader](https://github.com/erikdesjardins/webpack-rollup-loader) that may produce different results depending on your setup.
-
-
 ## Conclusion
 
 So, what we've learned is that in order to take advantage of _tree shaking_, you must...
 
 - Use ES2015 module syntax (i.e. `import` and `export`).
+- Add a "sideEffects" entry to your project's `package.json` file.
```
jeremenichelli: This would be _if_ the code is not affecting outside scope, not _always_ so we should add it to here.

probablyup: Actually no, you should always add a sideEffects entry or treeshaking won't work.

jeremenichelli: Well, but if your librari does produce side effects you shouldn't, right? I don't want to give that message to package owners.

probablyup: Perhaps, but in that case set it to `true`, or provide an array with the specific files that include side effects as described in the guide.

It's better to explicitly have the sideEffects known to webpack than to try and infer it.

montogeek: It is required?

jeremenichelli: Maybe is redundant since you already define what `sideEffects` means. I'm OK with it. Sorry for the noise.

src/content/guides/getting-started.md
jeremenichelli: From the `next` rebase maybe.

jeremenichelli:
There are some rephrasing that I think is necessary, the rest looks good @probablyup

probablyup:
@jeremenichelli updated

edmorley:
Do you think it's worth also linking to the [debugging optimization bailouts](https://webpack.js.org/plugins/module-concatenation-plugin/#debugging-optimization-bailouts) section from this page?

edmorley:
I think this should also include:
* the new `sideEffects` selectors added in webpack/webpack/pull/6536
* setting `sideEffects` via module rules (ie allowing downstream apps to override the package.json entry) - webpack/webpack/pull/6129

probablyup:
@edmorley updated

probablyup:
@jeremenichelli could this be merged? I think many people would find it helpful

edmorley:
@probablyup this is great - thank you for the changes :-)

Issue #1866, opened 2018-02-28
Add `<!doctype html>` to html snippets


rikschennink:
Yesterday I blindly copied the Webpack "Getting Started" index.html page to build a quick prototype. The next 50 minutes I'm debugging weird layout bugs. It seems the example snippet does not contain a DOCTYPE and I forgot to add it myself.

Maybe have it there by default?

montogeek:
webpack website shows in its examples the minimal requirements to wort with it, specify details about other technologies are no essential.

rikschennink:
@montogeek I understand that examples need to be kept as concise as possible. It worries me that a DOCTYPE is considered a "specific detail". Following that reasoning, the `<title>` attribute could also be removed.

If other devs are building their products/prototypes based on these examples they could find themselves in unnecessary trouble.

montogeek:
`title` is part of the guide. HTML knowledge is out of the scope of webpack documentation

timseverien:
I'm with @rikschennink here. If you're providing an entire HTML file, I would those to be complete.

bramsmulders:
+1

montogeek:
We are not providing an entire HTML file, just the minimal HTML to show how webpack works

rikschennink:
@montogeek A minimal HTML file (that is still valid) really should have a `<!doctype html>` at the top.

I'm [reading here](https://gist.github.com/cmalven/1885287) that it should also have a `<title>` tag, my earlier statement about removing it seems to be incorrect.

montogeek:
Can you please submit a PR to add it in all necessary examples? Not those generated by plugins.

rikschennink:
@montogeek happy to do that somewhere this week or the next.

TheDutchCoder:
@rikschennink this is a really good point. I'm probably responsible for this mishap, so I've taken the liberty to include it for our v4 Guides (https://github.com/webpack/webpack.js.org/pull/1862) and I will open a quick PR for the current docs as well.

You're totally right: the examples should be working and correct, even if they're minimal examples.

Thanks for the good catch!

rikschennink:
@TheDutchCoder Fantastic! Thanks for taking the time to fix it Reinier. I'll remove it from my todo list 🙂

jeremenichelli:
It would b good to check the rest of the site and try to spot places where this was also happening and fix it.

TheDutchCoder:
[...]

PR #1868, opened 2018-03-01
docs(all): add doctype for valid HTML in all documentation

TheDutchCoder:
As pointed out in Issue #1866, this should be added for valid HTML.

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -21,6 +21,7 @@ Let's make a minor change to our project before we get started:
 __dist/index.html__
 
 ``` diff
+  <!doctype html>
   <html>
     <head>
 -    <title>Getting Started</title>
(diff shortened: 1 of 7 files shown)


montogeek:
There are other placer where this happen, It is ok if I add them into this branch?

TheDutchCoder:
@montogeek good point, I can skim through them as well, but feel free to do so of course.

jeremenichelli:
Can we organize this a little better? Because if not someone will look at this PR and the issues and think this wasn't done. It's important to make easy to track what has been changed for long tasks like this.

That said, are there more places where this is happening? Or we are sure these are all the doctypes that were missing?

TheDutchCoder:
@jeremenichelli I've scanned all Guides and updated where required. Do you want to expand this to all sections of the website instead? I'm okay with that, it'll just widen the scope a bit more.

jeremenichelli:
I would love to have each section as separate PRs, if that's already not possible, then list in the title of this PR all the sections you covered.

montogeek:
I only found another place missing, it seems there is not more places.

PR #1871, opened 2018-03-01
docs(concepts): Update plugins.md

arjendeblok:
Made sample code for plugin working for webpack 4.
Also changed to ES2015.


montogeek:
Could you please add a note regarding that the plugin name should be a camelized version of the name

arjendeblok:
@montogeek Do you mean the name of the class or the string constant. Both are camilized in the example.

montogeek:
Plugin name, something like:
'First parameter of the tap method (or hook?) should be a camelized version of the plugin name'

TheDutchCoder:
Before I merge this in @jeremenichelli should this be PR'ed against the `next` branch instead?

jeremenichelli:
Yes @TheDutchCoder

Issue #1854, opened 2018-02-25
Access to v3 doc?


estaub:
While we wait for all the V4 plugins and issues and docs to settle down, folks still need to get work done with V3.  It would be nice to have access to the docs.

montogeek:
@estaub It is something missing in the official website?

estaub:
@montogeek I don't know what the current state is.  AFAICT, most of the changes needed for 4 aren't there yet, which is one of the things I'm waiting for before switching.  

It will be very difficult to maintain doc that works well for both, though; everything different will need to be labeled 3-only or 4-only somehow, e.g. CommonsChunkPlugin, UglifyjsWebpackPlugin, the new optimization config, etc.  People will forget to preserve 3-only content, and it will make for awkward and error-prone reading.  I'd think it would be easier to just serve two versions of the doc as a whole than to worry about any of this.

jderijke:
missing v3 doc too. Can a snapshot be made accessible in something like https://webpack.js.org/3?

montogeek:
Everything in the website is for webpack 3, AFAIK nothing has been deleted. 
CommonsChunkPlugin does not longer exists in webpack 4, but the page is available (https://webpack.js.org/plugins/commons-chunk-plugin/), the plan is to add a deprecation warning linking to its replacement, never remove it.

Making a snapshot of the docs would be good, but I am afraid it would require a lot of work since it does not exist.

estaub:
@montogeek The black-and-blue chip in the upper corner of the doc saying "webpack v4.0.1" leads readers (like me) to think that it at least attempts to doc v4.

montogeek:
That shows the latest version, but I see your point

wldcordeiro:
@estaub yeah the version pill is problematic and misleading because the docs are definitely 3.x but users will end up getting 4.x if they just run `npm i webpack`. I'd say we should have a version selector similar to something like [Django](https://docs.djangoproject.com/en/2.0/) or [Ember](https://guides.emberjs.com/v3.0.0/)

dbkaplun:
I second having a version selector. I frequently look for nonexistent old docs and am forced to do `git blame` to find the right version, and read source code itself.

montogeek:
Thinking more about this I am not sure if it is worth to have an entire version of the website for each version, although it is a major release, not a lot of things changed, there is 113 pages of documentation from which approximately only 25 (22%) would need an update for webpack 4

estaub:
[...]

PR #1877, opened 2018-03-03
docs(guide): The CLI moved into a separate package: webpack-cli

KumarBhot:
Executing webpack is now possible with `webpack-cli` instead of just `webpack`. Added that to `npm installation` step as well as the final execution instruction. A little unsure about how to add mention to the `--mode` flag.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -25,7 +25,7 @@ First let's create a directory, initialize npm, and [install webpack locally](/g
 ``` bash
 mkdir webpack-demo && cd webpack-demo
 npm init -y
-npm install --save-dev webpack
+npm install --save-dev webpack webpack-cli
 ```
 
 Now we'll create the following directory structure and contents:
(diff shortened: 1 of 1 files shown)


skipjack:
@jeremenichelli maybe this should be re-targeted at `next` if you're going to merge soon?

Always displaying the latest version number in the badge in the sidebar has definitely caused some confusion (e.g. #1854). We should start tagging the repo and versioning the `package.json` with the currently documented version using something like `standard-version`. This would allow us to easily identify where documentation for each version lies.

cc @jeremenichelli @montogeek @TheDutchCoder

TheDutchCoder:
Hi @KumarBhot thanks for your contribution!

We're actually in the process of updating the docs for v4 and this particular change was already merged into the `next` branch: https://github.com/webpack/webpack.js.org/commit/650f69efa3240018272270932f3d734009d05307

We're also planning on making versioned docs.

PR #1878, opened 2018-03-03
Add the `builtAt` option

osenvosem:
According to the source code there must be the `builtAt` option specified which is defaulted to `true`.
https://github.com/webpack/webpack/blob/f6e366b4be1cfe2770251a890d93081824789209/lib/Stats.js#L125

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16295 @@
+{
+  "name": "webpack.js.org",
+  "version": "0.1.2",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.4",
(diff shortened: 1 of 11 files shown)


TheDutchCoder:
Hi @osenvosem thanks for your contribution!

This addition seems to have been introduced in v4 of webpack, so I've changed the base of this PR to the `next` branch.

TheDutchCoder:
@dhruvdutt How would it cause sync issues exactly?

I personally don't use yarn, but npm.

dhruvdutt:
When we add/update a dependency, only one lock file will change based on which command we use.

For instance, if we do `yarn add lodash`, `yarn.lock` will update and if we do `npm i lodash` only `package-lock.json` would update. It's best to maintain only one lock file.

PR #1888, opened 2018-03-08
Installation guide: Add note to install webpack-cli on v4 or higher

kingdaro:
[As suggested here](https://github.com/webpack/webpack-cli/issues/318#issuecomment-371333504)


montogeek:
Thanks!

PR #1889, opened 2018-03-08
Clearer wording about multiple module styles.

ksze:
The original wording is a bit ambiguous. At the first reading, it sounds like there are modules A, B, and C, where A supports AMD, B supports CommonJS, and C is legacy. This makes no sense in the context and requires a second reading to understand.


montogeek:
Thanks!

PR #1892, opened 2018-03-08
style(sidebar): fix vertical spacing when option has an icon

Simek:
Ensure that sidebar items has proper height even when emoticon is used.

screen shot 2018-03-08 at 07 56 54


Simek:
@montogeek Change was tested on Chrome, Firefox, Opera, Vivaldi and Safari on latest Mac OSX. Also worth mentioning that "line-height" is an old (CSS 2.1) property and shouldn't cause any browser specific issues.

montogeek:
Thank you!

PR #1893, opened 2018-03-08
links open collective image button in production

### LINKED ISSUE #1843: Contribute - Donate Link Broken in Production
The link around the image on the `/contribute/` is broken.
It seems to work fine when you run `npm start` but not when you run `npm run build`.

I was going to try and help but the custom lexer implementation for finding markdown in html was a little tough for me to understand.
FYI multi-line html blocks work as expected in the marked lexer.

Anyway, the bit of code that is leading to the bug `webpack.js.org/src/utilities/markdown.js@handleHTML`
Each html tag gets split and parsed by `parseContent` which means that each individual start/end tag is wrapped in a `<p>`.
I believe the regression was introduced in 5d7534e661d

I didn't dig into why the issue doesn't crop up on development builds.
But I did hit a github rate limit trying to work this issue, it seems fetch.sh is doing some work ?

rouzbeh84:
very small issue but thought it would be a good idea to have the contrib button working proper 😸 

fixes  #1843

### DIFF
--- a/src/content/contribute/index.md
+++ b/src/content/contribute/index.md
@@ -12,9 +12,7 @@ contributors:
 The people who contribute to webpack do so for the love of open source, our users and ecosystem, and most importantly, pushing the web forward together. Because of our [Open Collective](http://opencollective.com/webpack) model for funding and transparency, we are able to funnel support and funds through contributors, dependent projects, and the contributor and core teams. To make a donation, simply click the button below...
 
 <div align="center">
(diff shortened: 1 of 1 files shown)


rouzbeh84:
to size the image right since i couldn't do the inline width trick. i think it broke something in the build somehow anyways so i'll probably just close this out in the morning ha

rouzbeh84:
Very odd that this keeps breaking the build. will close out and try another way i guess ¯\_(ツ)_/¯

PR #1856, opened 2018-02-26
Let's ship the webpack 4 documentation

sokra:
Merge the next branch.

PR #1896, opened 2018-03-08
mention defaults in config example

pravdomil:
see diff

### DIFF
--- a/src/content/configuration/index.md
+++ b/src/content/configuration/index.md
@@ -36,6 +36,7 @@ module.exports = {
     b: ["./app/entry-b1", "./app/entry-b2"]
   },
   </details>
+  // defaults to ./src
   // Here the application starts executing
   // and webpack starts bundling
 


montogeek:
Correct default is `src/index.js`. Could you please update it? Thanks!

pravdomil:
https://github.com/webpack/webpack/pull/6140/files#diff-7473dd12403a310ad8cf18d0ab9f10b9R20

montogeek:
Yes, you are right. Also, could you please change base branch to `next`? It is used to track all webpack 4 changes

pravdomil:
@montogeek @webpack it's not `next`; is version 4 is released so I'm merging it to `master`, and guys in webpack do not release new version if you don't update the documentation, that's totally irresponsible in my opinion, if you access webpack.js.org that docs is marked to be for version 4 and it's not.

and you mark it has enhancement? I cannot found the new default values anywhere, so please update the docs as soon as possible

montogeek:
This is a different repo from webpack, `next` is used to track all webpack 4 changes (#1856 ), once everything is ready it will be merged to `master`.

That bagde in the website refers to the latest version available, not the version of the documentation.

pravdomil:
well that should be changed, it's misleading if that's not a version of documentation so what is? look at the screenshot, everybody will think that that's a documentation for version 4

screen

montogeek:
Yes, definitely, we already talked about it in https://github.com/webpack/webpack.js.org/issues/1854

PR #1901, opened 2018-03-09
Show version of webpack.js.org not webpack in sidebar

pravdomil:
https://github.com/webpack/webpack.js.org/issues/1854 https://github.com/webpack/webpack.js.org/pull/1896

**actual**
screen

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "webpack.js.org",
-  "version": "0.1.2",
+  "version": "3.0.0",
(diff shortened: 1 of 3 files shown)


pravdomil:
Fell free to change it package.json

jeremenichelli:
We are actually considering on removing that badge at all since it's really confusing for users. Thanks for rising this, but I think we are gonna remove it at all for sure and see if we can indicate the webapck package version somehwere else in the UI. I don't see the value of showing the verison of the site itself.

PR #1905, opened 2018-03-11
docs(configuration) Update snippets to comply with webpack v4+

### LINKED ISSUE #1869: Update the Configuration section with webpack v4 configuration
In the Configuration section we still have old config snippets and we also are not reflecting the fact that webopack is now cero config out of the box.

Some content might also need to be change so it's compatible with the new webpack default behavior.

EugeneHlushko:
1. dont say that its the simplest configuration because simplest is 0cjs now
2. update snippets

Fixes #1869


EugeneHlushko:
thank you @sokra 👍

montogeek:
Thanks

PR #1909, opened 2018-03-13
Fix typo. niche => nice

mc-zone:
Considering I'm a non-native speaker, feel free to close it if I was wrong.


montogeek:
In this case it is correct: http://www.dictionary.com/browse/niche
> a distinct segment of a market. 

In this context it means that by submitting issues it helps us to cover very specific or distinct uses cases.

Thanks!

mc-zone:
Thank you for the explanation!

PR #1912, opened 2018-03-14
fixed line 62 spelling 'cosnole' -> 'console'

j-d-b:
just a typo fix~

### DIFF
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -59,7 +59,7 @@ __src/print.js__
 ``` diff
   export default function printMe() {
 -   console.log('I get called from print.js!');
-+   cosnole.log('I get called from print.js!');
++   console.log('I get called from print.js!');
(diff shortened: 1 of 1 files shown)


montogeek:
Thank you!

src/content/guides/development.md
```diff
@@ -59,7 +59,7 @@ __src/print.js__
 ``` diff
   export default function printMe() {
 -   console.log('I get called from print.js!');
-+   cosnole.log('I get called from print.js!');
++   console.log('I get called from print.js!');
```
dear-lizhihua: it's not a spelling error.

line 55:
```
Now let's make sure we have something to debug, so let's create an error in our `print.js` file:
```

montogeek: Thanks, completely miss it, D'oh!

PR #1913, opened 2018-03-15
Includes installing webpack-cli

deletosh:
This is now needed before running the `npx webpack` command.


montogeek:
Thanks for your contribution, unfortunately it was already done in `next` branch https://github.com/webpack/webpack.js.org/pull/1856, which is used to track all webpack 4 documentation and will be merged/released soon.

PR #1927, opened 2018-03-18
Inline loaders concept: replace 'overwrite' with 'override'

elliottsj:
'overwrite' implies that the webpack config will be overwritten
with the inline loader config. 'override' is more accurate.


montogeek:
Thanks!

PR #1936, opened 2018-03-21
docs (guide) fixed TODO and deprecated UglifyJsPlugin

MisterDev:
1. Substituted `webpack.optimize.UglifyJsPlugin()` with `webpack.ProgressPlugin()` since `UglifyJsPlugin` is deprecated
> Error: webpack.optimize.UglifyJsPlugin has been removed, please use config.optimization.minimize instead.

2. Edited the Node API example removing the `compiler.apply` call when setting plugins

3. Added my nickname to the contributors


montogeek:
Is `ProgressPlugin` its substitute?

MisterDev:
@montogeek From my understanding the `UglifyJsPlugin` is used by default when setting the `optimization.minimize` configuration option in Webpack 4 ([code](https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L266))([source](https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a))

I have replaced it with `ProgressPlugin` since that part of the documentation just shows examples of different syntax to configure Webpack to use plugins

src/content/concepts/plugins.md
```diff
@@ -75,17 +76,22 @@ module.exports = config;
...
+When using the Node API, you can also pass plugins via the `plugins` property in the configuration.
 
 **some-node-script.js**
 
 ```javascript
   const webpack = require('webpack'); //to access webpack runtime
   const configuration = require('./webpack.config.js');
 
-  let compiler = webpack(configuration);
-  compiler.apply(new webpack.ProgressPlugin());
```
ooflorent: You could use `new webpack.ProgressPlugin().apply(compiler)` instead

MisterDev: Updated, thanks for the suggestion!

MisterDev:
Ok, @montogeek I included the suggested edit

montogeek:
Thanks!

PR #1937, opened 2018-03-21
fix(mode): correct mode type of filename

Weffe:
in **Mode: development** example

There was a small issue with the filename not matching the mode type in the **Mode: development** example.


montogeek:
Thanks!

PR #1940, opened 2018-03-21
Document "none" values for mode

Tvrqvoise:
The schema validator supports a "none" value for `mode`, but it is not documented.  It is helpful if you want to control your own optimization, but not get the warning message or the default optimization behavior when `mode` is not set.

I was also confused by the following:
```
T> Please remember that setting `NODE_ENV` doesn't automatically set `mode`.
```

When I remove "mode", but still set `NODE_ENV=production`, my bundle is getting minified.  Is this a bug, something wrong in my config, or is the above line unclear / incorrect?


nveenjain:
It is not a bug, webpack by default uses mode:production, so when you remove mode, it uses production mode, it doesn't depend on NODE_ENV. Hence your code is minified.

montogeek:
@nveenjain https://github.com/webpack/webpack/issues/2537#issuecomment-370389124
`NODE_ENV=production` affects how your libraries are bundled.

> it doesn't depend on NODE_ENV 
If you don't set mode nor `NODE_ENV`, webpack will set `NODE_ENV=production`https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L286

Tvrqvoise:
@montogeek I updated the description to document the defaulting case as best I could.  Take a look and let me know.

nveenjain:
@Tvrqvoise , could you also update this section? https://webpack.js.org/concepts/#mode. It would be great.

src/content/concepts/mode.md
```diff
@@ -26,12 +26,13 @@ or pass it as a [CLI](/api/cli/) argument:
 webpack --mode=production
 ```
 
-The following string values are supported:
+This value is expected to always be set, and if not set webpack will warn and fallback to `production` mode.  The following string values are supported:
```
jeremenichelli: I think we should rephrase this, webpack does not expect you to change it if you don't want to, but works for production bundles as default.

Myabe we can express this like: _webpack sets `"production"` as a default for `mode` and warn about to let you know about this behavior. The supported values for `mode` are:_

Tvrqvoise: I was unsure about this text too.  Here's the warning message:

```
WARNING in configuration
The 'mode' option has not been set. Set 'mode' option to 'development' or 'production' to enable defaults for this environment.
```

My reading of this is that it implies that the default behavior is `none`, not `production`, since it implies that defaults are not set.  I can update to your copy, but that's currently inaccurate, as it *doesn't* let you know about that behavior.  However, the solve to this is pretty simple -- we should update the warning itself to be clearer.

montogeek: Message says that is not been set, default behavior is `production`, no `none`.
Yes, that message should say that it uses `production` by default

src/content/concepts/mode.md
```diff
@@ -26,12 +26,13 @@ or pass it as a [CLI](/api/cli/) argument:
 webpack --mode=production
 ```
 
-The following string values are supported:
+If unset, webpack sets `production` as a default for mode and warns to let you know about this behavior. The supported values for mode are:
```
nveenjain: Just to be grammatically correct and more reader friendly, shouldn't we say `If not set/provided,  webpack sets/uses production as default value for mode`. Otherwise LGTM

src/content/concepts/index.md
```diff
@@ -151,7 +151,7 @@ Using plugins in your webpack config is straightforward - however, there are man
 
 ## Mode
 
-By setting the `mode` parameter to either `development` or `production`, you can enable webpack's built-in optimizations that correspond with the selected mode.
+By setting the `mode` parameter to either `development` or `production`, you can enable webpack's built-in optimizations that correspond with the selected mode. If you would like to opt out of these optimizations, you may pass `none` instead.
```
jeremenichelli: @Tvrqvoise could you move this line beloe the *Mode: none* heading? I think it makes more sense there, if not, having just a title with a code snippet doens't add enough valuable content in my opinion.

Tvrqvoise: Sure man

montogeek:
Thanks!

PR #1944, opened 2018-03-22
chore: remove package-lock, sync yarn.lock

dhruvdutt:


montogeek:
It is the other way around, delete package-lock

ooflorent:
Why not doing the opposite? Yarn works way better...

montogeek:
Thanks

PR #1946, opened 2018-03-23
misc(gitignore) Add .idea (IDE) and .DS_Store to .gi

EugeneHlushko:
It was annoying always seeing these as `Untracked files` in `git status`


montogeek:
Thanks

PR #1951, opened 2018-03-24
Changing the output file to `main.js`

sakshi27:
`npx webpack` creates an output file as`main.js` instead of 'bundle.js`. We need to point to `main.js` in the `index.html'

_describe your changes..._

- [ ] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [ ] Make sure your PR complies with the [writer's guide][2].
- [ ] Review the diff carefully as sometimes this can reveal issues.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -167,14 +167,14 @@ __dist/index.html__
    </head>
    <body>
 -    <script src="./src/index.js"></script>
-+    <script src="dist/bundle.js"></script>
++    <script src="main.js"></script>
    </body>
   </html>
 ```
 
 In this setup, `index.js` explicitly requires `lodash` to be present, and binds it as `_` (no global scope pollution). By stating what dependencies a module needs, webpack can use this information to build a dependency graph. It then uses the graph to generate an optimized bundle where scripts will be executed in the correct order.
 
-With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `bundle.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.md
```diff
@@ -167,14 +167,14 @@ __dist/index.html__
...
-+    <script src="dist/bundle.js"></script>
++    <script src="main.js"></script>
    </body>
   </html>
 ```
 
+With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `main.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
```
EugeneHlushko: I would suggest to say:
... and `main.js` as the default [output]...

EugeneHlushko:
Please add into description that it: Fixes #1949

src/content/guides/getting-started.md
jeremenichelli: Thanks for the fix @sakshi27, do you think we can rephrase this paragraph? this is my proposal:

```
We can run `npx webpack` with `./src/index.js` as the [entry point](/concepts/entry-points) and `main.js` as the default [output](/concepts/output).

>T The `npx` command ships with Node 8.2 or higher and let us run the webpack binary (`./node_modules/.bin/webpack`) we installed above.
```

xx10t4:
The **Conclusion** section also references the non-existent `dist/bundle.js` file.

montogeek:
@EugeneHlushko @jeremenichelli I had closed 3 PRs with these same changes, your suggestions are welcome but are out of the scope of this PR, I really want to merge this one ASAP since it is a page directly linked from the homepage.

jeremenichelli:
No problem @montogeek, merge this and create an issue with the feedback here. I did that for the split chunks plugin. The important thing is to keep track of what's left to be done so it doesn't get lost in the conversations.

montogeek:
Thanks

PR #1954, opened 2018-03-25
Fix Path in index.html after move to /dist

cscouras:
If the ```<script>```  tag in```index.html``` has the  ```src``` path```"dist/bundle.js"``` 
 then when previewing ```index.html``` in the browser the follwing error is produced after the build: ... __/dist/dist/bundle.js net::ERR_FILE_NOT_FOUND__

Changing the ```src``` path to ```"./bundle.js"```  in the ```<script>``` tag resolves the error and produces the desired result after the build.


jeremenichelli:
@montogeek we should close this one, right?

montogeek:
Thanks!

PR #1863, opened 2018-02-27
Deprecation warning on CommonChunksPlugin. Add SplitChunksPlugin page

jeremenichelli:
Adding Tobias as a reviewer since I used his gist as a base for the SplitChunksPlugin page. I've also added a deprecation warning on CommonChunksPlugin one (not deleting it for now since developers and teams who can immediately migrate might need it).

Issue #1812, opened 2018-02-02
Config - Document `splitChunks` and `minimize`


kingdaro:
For https://github.com/webpack/webpack/pull/6416: include links to `config.optimize.splitChunks` and `config.optimize.minimize` when the relevant docs are merged.

skipjack:
Can you submit a PR directed at the `next` branch to resolve this? BTW, the other v4 changes are tracked in #1706 so I may add this as a comment there. We won't merge `next` into `master` until v4 is actually released, otherwise we'd get a bunch of reports of missing features. But, if you submit the PR to `next`, these docs would be there as soon as it is merged. You could even link directly to the `next` branch in the meantime.

cc @TheLarkInn

jeremenichelli:
@kingdaro are you working on this? If not I can take care of it :)

kingdaro:
@jeremenichelli Go for it, I'm busy.

jeremenichelli:
Here's the PR for it: https://github.com/webpack/webpack.js.org/pull/1863

jeremenichelli:
Closing since we have an issue for optimization section and we already merged splitChunks docs.

PR #1955, opened 2018-03-26
feat: add documentation for SplitChunksPlugin and deprecation warning…

jeremenichelli:
Sorry for the noise, the content for this branch was already approved here: https://github.com/webpack/webpack.js.org/pull/1863 but because of some issues on rebasing I needed to create a new PR to make sure I'm not making regression on content after the `next` branch merge.

Solves https://github.com/webpack/webpack.js.org/issues/1812

### DIFF
--- a/src/content/plugins/commons-chunk-plugin.md
+++ b/src/content/plugins/commons-chunk-plugin.md
@@ -6,9 +6,14 @@ contributors:
   - christopher4lis
   - kevinzwhuang
   - jdbevan
+  - jeremenichelli
 ---
 
-The `CommonsChunkPlugin` is an opt-in feature that creates a separate file (known as a chunk), consisting of common modules shared between multiple entry points. By separating common modules from bundles, the resulting chunked file can be loaded once initially, and stored in cache for later use. This results in pagespeed optimizations as the browser can quickly serve the shared code from cache, rather than being forced to load a larger bundle whenever a new page is visited.
+The `CommonsChunkPlugin` is an opt-in feature that creates a separate file (known as a chunk), consisting of common modules shared between multiple entry points.
+
+W> The CommonsChunkPlugin has been deprecated in webpack v4 legato. To learn how chunks are treated in the latest version, check out the [SplitChunksPlugin](/plugins/split-chunks-plugin/).
+
+By separating common modules from bundles, the resulting chunked file can be loaded once initially, and stored in cache for later use. This results in pagespeed optimizations as the browser can quickly serve the shared code from cache, rather than being forced to load a larger bundle whenever a new page is visited.
 
 ```javascript
 new webpack.optimize.CommonsChunkPlugin(options)
--- a/src/content/plugins/commons-chunk-plugin.md
+++ b/src/content/plugins/commons-chunk-plugin.md
@@ -11,7 +11,7 @@ contributors:
 
 The `CommonsChunkPlugin` is an opt-in feature that creates a separate file (known as a chunk), consisting of common modules shared between multiple entry points.
 
-W> The CommonsChunkPlugin has been deprecated in webpack v4 legato. To learn how chunks are treated in the latest version, check out the [SplitChunksPlugin](/plugins/split-chunks-plugin/).
+W> The CommonsChunkPlugin has been removed in webpack v4 legato. To learn how chunks are treated in the latest version, check out the [SplitChunksPlugin](/plugins/split-chunks-plugin/).
 
 By separating common modules from bundles, the resulting chunked file can be loaded once initially, and stored in cache for later use. This results in pagespeed optimizations as the browser can quickly serve the shared code from cache, rather than being forced to load a larger bundle whenever a new page is visited.
 
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -147,7 +147,7 @@ This can be controlled per cache group.
 
 ### `optimization.splitChunks.chunks: all`
 
-As it was mentioned before this plugin will affect dynamic imported modules. Setting the `optimization.splitChunks.chunks` option to `"all"` initial chunks will get affected by it (even the ones not imported dynamically). This way chunks can even be shared between entrypoints and on-demand loading.
+As it was mentioned before this plugin will affect dynamic imported modules. Setting the `optimization.splitChunks.chunks` option to `"all"` initial chunks will get affected by it (even the ones not imported dynamically). This way chunks can even be shared between entry points and on-demand loading.
 
 This is the recommended configuration.
 
@@ -191,7 +191,7 @@ Here are some examples and their effect:
 
 ### Split Chunks: Example 1
 
-Create a `commons` chunk, which includes all code shared between entrypoints.
+Create a `commons` chunk, which includes all code shared between entry points.
 
 ```js
 splitChunks: {
@@ -205,17 +205,17 @@ splitChunks: {
 }
 ```
 
-W> This downloads more code than neccessary.
+W> This configuration can enlarge your initial bundles, it is recommended to use dynamic imports when a module is not immediately needed.
 
 ### Split Chunks: Example 2
 
-Create a `vendors` chunk, which includes all code from node_modules in the whole application.
+Create a `vendors` chunk, which includes all code from `node_modules` in the whole application.
 
 ``` js
 splitChunks: {
 	cacheGroups: {
 		commons: {
-			test: /[\\/]node_modules[\\/],
+			test: /[\\/]node_modules[\\/]/,
 			name: "vendors",
 			chunks: "all"
 		}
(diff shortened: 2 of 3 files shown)


jeremenichelli:
Hey @skipjack, Travis is throwing an error: `API rate limit exceeded for 35.224.112.202`, did we exceeded some quota around the service? 🤔

montogeek:
@jeremenichelli It is exceeded every a few hours, something to fix in Infrastructure

jeremenichelli:
@montogeek should we just trigger a new build for the branch?

montogeek:
@jeremenichelli I am fixing it for once and all https://github.com/webpack/webpack.js.org/pull/1963

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+title: SplitChunksPlugin
+contributors:
+  - sokra
+  - jeremenichelli
+related:
+  - title: "webpack 4: Code Splitting, chunk graph and the splitChunks optimization"
+    url: https://medium.com/webpack/webpack-4-code-splitting-chunk-graph-and-the-splitchunks-optimization-be739a861366
+---
+
+Originally in webpack, chunks (and modules imported inside them) were connected by a parent-child relationship in the internal webpack graph.
```
skipjack: I would rephrase as (merged with the next lines as well):

> By default, chunks (and the modules inside them) are connected by direct parent-child relationships in the dependency graph. However, there are instances where a dependency may be duplicated across chunks. Previously, the `CommonsChunkPlugin` was use to dedupe these dependencies. In v4, the `optimization.splitChunks` and `optimization.runtimeChunk` configuration options provide a more intuitive interface.
>
> Here is how the new flow works...

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+
+This connection didn't allow to further optimizations and resulted into more downloaded code.
+
+webpack v4 removes the `CommonsChunkPlugin` in favor of `optimization.splitChunks` and `optimize.runtimeChunk` options. Here is how the new flow works.
+
```
skipjack: We typically use two line breaks between `h2` sections. We should really automate this via `remark` and/or `prettier` along with various other markdown guidelines.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+Why:
+
+* Condition 1: The chunk is shared between both import calls
+* Condition 2: `helpers` is bigger than 30kb
+* Condition 3: Number of parallel requests at the import calls is 2
+* Condition 4: Doesn't affect request at initial page load
+
+Putting the content of `helpers` into each chunk will result into its code being downloaded twice. By using a separate chunk this will only happen once. We pay the cost of an additional request, which could be considered a tradeoff. That's why there is a minimum size of 30kb.
+
```
skipjack: Same two line breaks nit (just helps with readability when editing).

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+Originally in webpack, chunks (and modules imported inside them) were connected by a parent-child relationship in the internal webpack graph.
+
+This connection didn't allow to further optimizations and resulted into more downloaded code.
+
+webpack v4 removes the `CommonsChunkPlugin` in favor of `optimization.splitChunks` and `optimize.runtimeChunk` options. Here is how the new flow works.
+
+## Defaults
+
+Out of the box `SplitChunksPlugin` should work great for most users.
```
skipjack: Is it included out of the box or only when certain flags/config are present? Or do you just mean when included with no options, like so:

``` js
plugins: [
  new SplitChunksPlugin()
]
```

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+
+By default it only affects on-demand chunks because changing initial chunks would affect the script tags the HTML file should include to run the project.
```
skipjack: Maybe explain why changing those initial script tags is a problem?

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+		commons: {
+			test: /[\\/]node_modules[\\/],
+			name: "vendors",
+			chunks: "all"
+		}
+	}
+}
+```
+
+W> This downloads more code than neccessary.
```
skipjack: s/neccessary/necessary

skipjack: Also maybe elaborate this warning a bit...

> This will probably yield an unnecessarily large bundle as many applications use a bunch of `node_modules`. It's typically better to extract and load your core frameworks on initial load, and let other dependencies be chunked out and loaded dynamically later on.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+
+You can set `optimization.splitChunks.cacheGroups.default` to `false` to disable the default cache group, same for `vendors` cache group.
+
+The priority of the default groups are negative to allow any custom cache group to take higher priority (the default value is `0`).
+
+Here are some examples and their effect:
+
+### Split Chunks: Example 1
+
+Create a `commons` chunk, which includes all code shared between entrypoints.
```
skipjack: s/entrypoints/entry points

phun-ky:
Nice work, this is very much needed for all webpack4 users!

But, I can't see that you have included all the options for splitchunkcs, for example:

`automaticNameDelimiter` (defaults to: ~)

All options is in: `schemas\WebpackOptions.json`, and the default options are set here: `lib\WebpackOptionsDefaulter.js`

Specially `automaticNameDelimiter` was important to us, because we have a WAF (NAXSI) that had a ruled triggered for the usage of the tilde(~) character, resulting in a 406 Not Acceptable on the server when the chunk is requested: `vendor~moduleA~moduleB.js`.

src/content/plugins/commons-chunk-plugin.md
```diff
@@ -6,9 +6,14 @@ contributors:
   - christopher4lis
   - kevinzwhuang
   - jdbevan
+  - jeremenichelli
 ---
 
-The `CommonsChunkPlugin` is an opt-in feature that creates a separate file (known as a chunk), consisting of common modules shared between multiple entry points. By separating common modules from bundles, the resulting chunked file can be loaded once initially, and stored in cache for later use. This results in pagespeed optimizations as the browser can quickly serve the shared code from cache, rather than being forced to load a larger bundle whenever a new page is visited.
+The `CommonsChunkPlugin` is an opt-in feature that creates a separate file (known as a chunk), consisting of common modules shared between multiple entry points.
+
+W> The CommonsChunkPlugin has been deprecated in webpack v4 legato. To learn how chunks are treated in the latest version, check out the [SplitChunksPlugin](/plugins/split-chunks-plugin/).
```
ooflorent: It is not deprecated: it has been removed.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+}
+```
+
+W> This downloads more code than neccessary.
+
+## `optimization.runtimeChunk`
+
+Setting `optimization.runtimeChunk` to `true` adds an additonal chunk to each entrypoint containing only the runtime.
```
ooflorent: Maybe you could also talk about `optimization.runtimeChunk="single"` which creates a single runtime chunk shared by all entries.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Split Chunks: Example 2
+
+Create a `vendors` chunk, which includes all code from node_modules in the whole application.
+
+``` js
+splitChunks: {
+	cacheGroups: {
+		commons: {
+			test: /[\\/]node_modules[\\/],
```
DevSide: test: /[\\\\/]node_modules[\\\\/]**/**,

src/content/plugins/split-chunks-plugin.md
jeremenichelli: Do we know what's the default value and behavior when this is not set?

jeremenichelli:
Everyone here, I've updated the branch with your feedback. Feel free to take a look at it, if it is I think we should merge it already, information about this plugin should have been there from day one in my humble opinion.

jeremenichelli:
Some things might not be covered on the first version, I'm tracking the missing stuff here: https://github.com/webpack/webpack.js.org/issues/1956

src/content/plugins/split-chunks-plugin.md
```diff
@@ -0,0 +1,241 @@
...
+This configuration object represents the default behavior of the `SplitChunksPlugin`.
+
+```js
+splitChunks: {
+	chunks: "async",
+	minSize: 30000,
+	minChunks: 1,
+	maxAsyncRequests: 5,
+	maxInitialRequests: 3,
+  automaticNameDelimiter: '~',
```
lencioni: I think the indentation is wrong here. It looks like everything else is using tabs, but this line uses spaces. 
screen shot 2018-03-30 at 5 44 50 pm

jeremenichelli: Thanks for the catch!

src/content/plugins/split-chunks-plugin.md
jeremenichelli: I think that stating that they are (present tense) linked by parent-child relationship is confusing/wrong since this has completely changed on the core of the package.

src/content/plugins/split-chunks-plugin.md
jeremenichelli: Honestly, I think that the sentence is pretty clear by itself.

jeremenichelli:
Yes I agree, that is not clear enough. I will add that to the issue. I will
take another look during this month and do sone gardening.
-- 

*Jeremias Menichelli*


*web developer*
skype | jeremias.menichelli
github | jeremenichelli

PR #1958, opened 2018-03-26
Fix a bug in getting-started.md

dpyzo0o:
The path for including bundle.js is not correct. index.html is already inside /dist folder, the correct path should be 'bundle.js' or '../dist/bundle.js'

_describe your changes..._


montogeek:
Thanks

PR #1959, opened 2018-03-26
Update getting-started

wietsevenema:
When working through this tutorial, we noticed that the default is main.js, not bundle.js

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -167,14 +167,14 @@ __dist/index.html__
    </head>
    <body>
 -    <script src="./src/index.js"></script>
-+    <script src="bundle.js"></script>
++    <script src="main.js"></script>
    </body>
   </html>
 ```
 
 In this setup, `index.js` explicitly requires `lodash` to be present, and binds it as `_` (no global scope pollution). By stating what dependencies a module needs, webpack can use this information to build a dependency graph. It then uses the graph to generate an optimized bundle where scripts will be executed in the correct order.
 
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks for your contribution, unfortunately this was already fixed in https://github.com/webpack/webpack.js.org/pull/1951. I suggest checking existing PRs before submitting yours.

wietsevenema:
Thanks for your quick and kind response, good to hear that is was fixed already.

src/content/guides/getting-started.md
```diff
@@ -167,14 +167,14 @@ __dist/index.html__
...
-+    <script src="bundle.js"></script>
++    <script src="main.js"></script>
    </body>
   </html>
 ```
 
+With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `main.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
```
jeremenichelli: Can we rephrase this a little bit? Like:

```
We can run `npx webpack` with our script as the [entry point](/concepts/entry-points) and it will [output](/concepts/output) a `main.js` file as the default.

>T The `npx` command ships with Node 8.2 or higher and let us run the webpack binary (`./node_modules/.bin/webpack`) we installed in the beginning._
```

jeremenichelli: @wietsevenema do you think it's possible to rephrase this? Here's my proposal:

```
We can run `npx webpack` with our script as the [entry point](/concepts/entry-points) and it will [output](/concepts/output) a `main.js` file as the default.

>T The `npx` command ships with Node 8.2 or higher and let us run the webpack binary (`./node_modules/.bin/webpack`) we installed in the beginning._
```

PR #1963, opened 2018-03-27
Added GitHub token for authenticated GitHub API calls

montogeek:
We are hitting GH Rate limit (60 request/hr) very quickly, this will increase it to 5000.

### DIFF
--- a/src/scripts/fetch_package_names.js
+++ b/src/scripts/fetch_package_names.js
@@ -35,6 +35,11 @@ function main() {
 function fetchPackageNames(options, cb) {
   const github = new GitHubApi();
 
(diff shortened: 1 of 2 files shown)


jeremenichelli:
So the problem was that we were fetching even when no token was present? We'll still gonna fetch on Travis I guess, I don't know if this will actually solve our quota issue.

montogeek:
No, we weren't using auth at all, this add authentication if the token exists (on Travis CI), a contributor won't have it locally, so I added a check before using it

jeremenichelli:
Go for it!

EugeneHlushko:
great addition!

PR #1970, opened 2018-03-29
Add forward slash to the regex in Example 2

seckin92:
Added a forward slash to complete the regular expression definition in Example 2.


montogeek:
Thanks!

PR #1973, opened 2018-03-30
Update getting-started.md

xx10t4:
The initial example does not work. Running `npx webpack` creates the file `dist/main.js` so the `<script src="">` attribute needs to point to main.js


montogeek:
Thanks for your contribution, unfortunately it was already fixed in https://github.com/webpack/webpack.js.org/pull/1951, I advice check previous PRs before submitting yours.

PR #1976, opened 2018-04-02
Fixes reference to output bundle in Getting Started guide

gingi:
With no arguments, the bundle that `npx webpack` outputs in `dist/` is `main.js`, not `bundle.js`. Confirmed with both webpack@4.0.1 (the version referenced in the guide) and webpack@4.4.1 (latest stable). This commit renames all references (`bundle.js` → `main.js`) to avoid requiring new users from specifying unnecessary parameters.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -6,6 +6,7 @@ contributors:
   - varunjayaraman
   - cntanglijun
   - chrisVillanueva
+  - gingi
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks for your contribution, unfortunately it was already fixed in https://github.com/webpack/webpack.js.org/pull/1951, I advice check previous PRs before submitting yours.

PR #1582, opened 2017-09-09
Fix Hyperlink/Travis Issues

skipjack:
Reverted 8281949b400 so we can test and try to resolve the `hyperlink` issue on Travis' network. I think there still may be one link failure due to a package's README that we may have to resolve before we can fix the core issue.

Related to Munter/hyperlink#121

PR #1978, opened 2018-04-03
Evaluate all http redirects of outgoing links

Munter:
Automatically replaced all external urls that had redirects pointing them to a different location.

Automated replacements based on output from hyperlink in #1582

### DIFF
--- a/src/components/Footer/Footer.jsx
+++ b/src/components/Footer/Footer.jsx
@@ -25,7 +25,7 @@ const Footer = () => (
       <section className="footer__right">
         <Link className="footer__link" to="/glossary/">Glossary</Link>
         <Link className="footer__link" to="/branding/">Branding</Link>
-        <Link className="footer__link" to="//gitter.im/webpack/webpack">Gitter</Link>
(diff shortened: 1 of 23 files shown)


src/components/Footer/Footer.jsx
```diff
@@ -25,7 +25,7 @@ const Footer = () => (
       <section className="footer__right">
         <Link className="footer__link" to="/glossary/">Glossary</Link>
         <Link className="footer__link" to="/branding/">Branding</Link>
-        <Link className="footer__link" to="//gitter.im/webpack/webpack">Gitter</Link>
+        <Link className="footer__link" to="https://gitter.im/webpack/webpack">Gitter</Link>
```
montogeek: I was going to commend about this https://www.paulirish.com/2010/the-protocol-relative-url/, but then realize is from 8 years ago and see that is now considered an antipattern, also realize how old I am 📟

Munter: I very much know that feeling :D

Munter:
Failing on profanity :/

https://github.com/webpack-contrib/stylelint-webpack-plugin/pull/150

PR #1988, opened 2018-04-05
chore: sort split chunks cache groups by priority

lencioni:
I'm sorting these visually by priority to help clarify the actual priority that they use.


montogeek:
Like a `switch`?

lencioni:
Yeah I suppose it is kinda like a switch, but I think that is maybe just an artifact of how the `default` cacheGroup is named. :) I was merely looking at the value of the `priority`.

montogeek:
Thanks

PR #1989, opened 2018-04-06
fix: Fix variable name

ptrkcsk:
Given this example code:

```js
new webpack.ProvidePlugin({
  _: 'lodash'
})
```

this sentence

> If you encounter at least one instance of the variable `lodash`, include the `lodash` package and provide it to the modules that need it.

should say 

> If you encounter at least one instance of the variable ~`lodash`~ `_`, include the `lodash` package and provide it to the modules that need it.


montogeek:
Thanks!

PR #1990, opened 2018-04-06
Fixed an a/an error in `/configuration/module`

Pimm:
> It can have a `options` property

↓

> It can have an `options` property


montogeek:
Thanks!

PR #1993, opened 2018-04-06
Merge the next branch

dhruvdutt:

PR #1996, opened 2018-04-07
change bundle.js to main.js

yeonhoyoon:
In recent versions of webpack, the default name for output file is 'main.js'.
Users might be confused as they wouldn't be able to locate 'bundle.js'.


montogeek:
Thanks for your contribution, unfortunately this was already updated on https://github.com/webpack/webpack.js.org/pull/1993.

PR #2006, opened 2018-04-09
Fix broken links

Munter:
Yet more broken link fixes based on local hyperlink runs

### DIFF
--- a/src/scripts/fetch_package_files.js
+++ b/src/scripts/fetch_package_files.js
@@ -2,6 +2,7 @@
 // ./fetch_package_files <file> <output> < input
 // ./fetch_package_files "README.md" "./output" < input
 const fs = require('fs');
+const urlModule = require('url');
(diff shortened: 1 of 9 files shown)


src/scripts/fetch_package_files.js
```diff
@@ -88,6 +89,8 @@ function fetchPackageFiles(options, finalCb) {
             // Replace any <h2> with `##`
             .replace(/<h2[^>]*>/g, '## ')
             .replace(/<\/h2>/g, '')
+            // Resolve anchor URL's to avoid broken relative URL's in the docs
+            .replace(/\[([^[\]]*)\]\(([^)]+)\)/g, (match, textContent, href) => `[${textContent}](${urlModule.resolve(url, href)})`)
```
montogeek: Could you please add an example about what this Regex is doing?

PR #2008, opened 2018-04-10
Fix the sequence of loaders

cokencode:
https://github.com/webpack/webpack/blob/master/lib/NormalModuleFactory.js#L300

`inline` loaders are applied after `normal` loaders.


src/content/configuration/module.md
```diff
@@ -92,7 +92,7 @@ Specifies the category of the loader. No value means normal loader.
 
 There is also an additional category "inlined loader" which are loaders applied inline of the import/require.
 
-All loaders are sorted in the order `pre, inline, normal, post` and used in this order.
+All loaders are sorted in the order `pre, normal, inline, post` and used in this order.
```
TheLarkInn: We should be explicit here about the order in someway. Maybe in a tooltop or hint annotation. 

Technically, webpack's loader execution goes from Left to Right to Left. 

So therefore in the first phase (the pitching phase), the order is: `post-inline-normal-pre`

Then for the next loader phase (the normal phase), the order is: `pre-normal-inline-post`

cokencode: What is the purpose of the pitching phase? 

Also, I think the explanation of the two phases needs to be added to [Writing a Loader](https://webpack.js.org/contribute/writing-a-loader/) section.

src/content/configuration/module.md
```diff
@@ -102,6 +102,7 @@ All normal, post and pre loaders can be omitted (overridden) by prefixing `!!` i
 
 Inline loaders and `!` prefixes should not be used as they are non-standard. They may be use by loader generated code.
 
+T> The `pitch` method on loaders is called in the order `post, inline, normal, pre` before loaders are executed in the order `pre, normal, inline, post`. See [this section](/api/loaders/#pitching-loader) for details.
```
EugeneHlushko: I suggest saying `See [Pitching Loader](...) for details.`

EugeneHlushko: `this section` can be misleading

src/content/configuration/module.md
EugeneHlushko: I agree with Sean, maybe mention that there are two phases and then speak about order?

cokencode:
@TheLarkInn @EugeneHlushko Thanks for reviewing.
I have added a list of two phases and mentioned the orders with them instead of adding a tip box.

montogeek:
Thank you!

PR #2012, opened 2018-04-10
Fix references to non-existing "optimize" section

mbudde:
The docs on the `SplitChunksPlugin` refers to `optimize.runtimeChunk`, but the correct section is `optimization`.


montogeek:
Thanks

PR #2026, opened 2018-04-11
fix(config) Fix build time issue using defaut config

EugeneHlushko:
Local build time ~540 seconds. Consequent builds for some reason take much less with a random difference varying from 82 seconds to 110 seconds per build using default `yarn run build` cmd

UPD:
* local build time is unstable, subsequent builds take less time (down to 82 seconds from 540 seconds) for unknown reason, antwar needs more profiling, dropping this work to not waste time
* this branch seems to build 20-25% faster than others due to node.js version bump
* this branch renders contributors only clientside, will finally fix error event not firing for 50X/40X image responses from OC
* redirects fixed, antwar.config.js redirects dont work at all in latest antwar version

### DIFF
--- a/antwar.config.js
+++ b/antwar.config.js
@@ -73,7 +73,7 @@ module.exports = () => ({
       title: 'Migrate',
       url: ({ sectionName, fileName }) => `/${sectionName}/${fileName}/`,
       layout: () => require('./src/components/Page/Page.jsx').default,
-      content: () => require.context('./loaders/page-loader!./src/content/migrate', false, /^\.\/.*\.md$/),
+      content: () => require.context('./src/content/migrate', false, /^\.\/.*\.md$/),
     },
     vote: () => require('./src/components/Vote/Vote.jsx').default,
     organization: () => require('./src/components/Organization/Organization.jsx').default,
--- a/antwar.config.js
+++ b/antwar.config.js
@@ -16,83 +16,67 @@ module.exports = () => ({
       layout: () => require('./src/components/Page/Page.jsx').default,
       content: () => require.context('./loaders/page-loader!./src/content', false, /^\.\/.*\.md$/),
       index: () => require('./src/components/Splash/Splash.jsx').default,
-      redirects: {
-        'support': '/contribute',
-        'writers-guide': '/contribute/writers-guide'
+    },
+    concepts: {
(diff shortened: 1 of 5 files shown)


montogeek:
@EugeneHlushko It is faster?

EugeneHlushko:
@montogeek locally it is stable at 92-106 seconds on my macbook pro 2015, i also noticed that after rebase not all of the redirects were in place and config's alignment went off. just fixed. seems node 8 speed up travis by 20% but not cut in half, trying to think of smth else meanwhile

EugeneHlushko:
@montogeek can you pull this one and try out locally?

webpack.config.js
```diff
@@ -101,12 +101,15 @@ module.exports = (env) => ({
     }),
     new RedirectWebpackPlugin({
       redirects: {
+        'support': '/contribute/',
```
montogeek: Where these new redirects come from?

antwar.config.js
```diff
@@ -16,77 +16,61 @@ module.exports = () => ({
       layout: () => require('./src/components/Page/Page.jsx').default,
       content: () => require.context('./loaders/page-loader!./src/content', false, /^\.\/.*\.md$/),
       index: () => require('./src/components/Splash/Splash.jsx').default,
-      redirects: {
-        'support': '/contribute',
```
EugeneHlushko: from here

montogeek: Right, completely miss them

montogeek:
Just ran `yarn build`:
```
real	16m16.693s
user	65m29.186s
sys	7m44.943s
```

EugeneHlushko:
Ok profiling it down shows that antwar diving into too many operations at once, limiting workers doesnt solve this. for example if you only build "/" and "concepts" pages it takes you 10 seconds, when you build all the pages the overhead is starting to show on my laptop and `maximumWorkers` doesnt solve this for me locally. Suggest to leave as is and focus on removing antwar (`rebuild` branch) @montogeek

EugeneHlushko:
I will rebase this after `migrate` section merge there are conflicts now

EugeneHlushko:
I suggest we merge this, anyone else you want to look at this @montogeek ?

montogeek:
Thanks

PR #2028, opened 2018-04-12
docs(plugins) Add mini-css-extract-plugin

### LINKED ISSUE #2016: mini-css-extract-plugin is missing from plugins
less-loader is linking to https://github.com/webpack-contrib/mini-css-extract-plugin which is rewritten to `/plugins/mini-css-extract-plugin` by `fetch.sh`

I think mini-css-extract-plugin isn't downloaded as part of the documentation because it doesn't match the `webpack-plugin` pattern

EugeneHlushko:
* remove extract-text-webpack-plugin
* add mini-css-extract-plugin
* add redirect from extract-text-webpack-plugin to mini-css-extract-plugin
* fix guides asset management
* fix guides production

fixes #2016

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -7,7 +7,6 @@ contributors:
   - TheDutchCoder
   - sudarsangp
   - chenxsan
-  - EugeneHlushko
 ---
 
 If you've been following the guides from the start, you will now have a small project that shows "Hello webpack". Now let's try to incorporate some other assets, like images, to see how they can be handled.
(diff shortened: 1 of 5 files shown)


src/content/guides/asset-management.md
```diff
@@ -59,16 +61,19 @@ __webpack.config.js__
 +       {
 +         test: /\.css$/,
 +         use: [
-+           'style-loader',
++           MiniCssExtractPlugin.loader,
```
Kovensky: This won't work with hot reloading, at least not yet.

EugeneHlushko: Do you suggest we add a note that hot reloading wont work for now with MiniCssExtractPlugin or is there a known workaround? @Kovensky

Kovensky: The known workaround is to use `style-loader` 😅

EugeneHlushko: Oh, so we probably can skip updating this guide for now (asset-management) and leave it with `style-loader`?

EugeneHlushko: a redirect from `deleted` plugin is added for that, also the text of the page says why it came to replace ETWP: https://github.com/webpack-contrib/mini-css-extract-plugin

Otherwise works fine locally.

@Kovensky am i right we dont need to update asset-management guide for now?

EugeneHlushko:
ok cleaned up `asset-management.md`, its good now from my perspective. Would really appreciate a re-review @jeremenichelli @Kovensky

jeremenichelli:
I just spotted the redirect @EugeneHlushko, yes, we don't delete or do that for deprecated plugins, we just add a deprecation warning similar to the one in here https://webpack.js.org/plugins/commons-chunk-plugin/

Consider teams that are not ready to upgrade and need this documentation for their use case. I think that with that change this would be good to go.

EugeneHlushko:
@jeremenichelli done 👍

montogeek:
Thanks

montogeek:
Thanks

PR #2040, opened 2018-04-14
chore(rebuild): populate page titles

montogeek:


src/server.jsx
```diff
@@ -23,34 +20,35 @@ const bundles = [
...
   let { assets } = locals.webpackStats.compilation;
+  let title = locals.paths[locals.path];
 
   return ReactDOMServer.renderToString(
-    <StaticRouter location={ locals.path } context={{}}>
+    <StaticRouter location={locals.path} context={{}}>
       <html>
         <head>
-          <meta charSet="UTF-8" />
+          <meta charset='utf-8' />
```
EugeneHlushko: these should be doublequotes

EugeneHlushko:
It works locally but i have an issue with title containing empty html comment notation, if you do `yarn run build` and check html files in `./dist/**/*.html` titles will look like this: `<title>webpack<!-- --> | webpack</title>` any idea why @montogeek

montogeek:
@EugeneHlushko Nope, debugged it for 1 hour when I did it, no idea why does it happen. My best guess is that React think there is an _space_ there, but not sure.

webpack.prod.js
```diff
@@ -7,6 +7,17 @@ const SSGPlugin = require('static-site-generator-webpack-plugin');
 
 // Load Common Configuration
 const common = require('./webpack.common.js');
+const Content = require('./src/_content.json');
+
+const paths = Content.children.reduce((paths, page) => {
```
montogeek: This can also be done in the server file, but I think it is better here.

webpack.prod.js
```diff
@@ -27,7 +38,8 @@ module.exports = env => [
     },
     plugins: [
       new SSGPlugin({
-        crawl: true,
+        paths: Object.keys(paths),
```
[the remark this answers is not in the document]
EugeneHlushko: On the readme of static-site-generator-webpack-plugin, they say that you can use a combination of `crawl: true` and providing multiple entry points for crawling in the paths:
```
plugins: [
    new StaticSiteGeneratorPlugin({
      crawl: true,
      paths: [
        '/',
        '/uncrawlable-page/'
      ]
    })
  ]
```
 i didnt try that but maybe you tried and it doesnt work? Just curious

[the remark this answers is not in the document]
skipjack: @montogeek tbh I'm a little confused why the config changes were needed just to add the page titles. Did you try just `import`ing the `Content.json` file and looking for the corresponding page title from within there based on `locals.path`? I thought the `title` addition would be smaller than this (although I might be missing something).

This feedback is just from a glance though, I'll look through your changes a bit more to understand exactly what you did.

src/server.jsx
skipjack: I agree. And in general, we should avoid misc formatting changes like line 26, lines 51 - 53, etc. I understand some of my stylistic choices might be a little different but I'd rather us fix this with something like prettier instead of changing things back and forth due to personal preference. Also, these little things create a lot of noise on PR like this, meaning it's hard to tell what's related to the title population and what are just misc formatting things.

webpack.prod.js
```diff
@@ -7,6 +7,17 @@ const SSGPlugin = require('static-site-generator-webpack-plugin');
 
 // Load Common Configuration
 const common = require('./webpack.common.js');
+const Content = require('./src/_content.json');
```
skipjack: I'm starting to understand the approach now. You can't do it this way because `_content.json` only exists after a build has occurred, meaning if you haven't run a build, the file won't exist and therefore this line will fail. You should take a similar approach, but do it all within the `server.jsx` file.

skipjack: This way you won't have to change any configuration or touch any other parts of code. Then, if you remove the misc formatting things, this should be a pretty isolated, mergable PR (at least in my opinion).

EugeneHlushko: i think we could also .gitignore the _content.json? why is it in repo

src/server.jsx
montogeek: @skipjack Yeah, another prettier change, submit a PR to add it to the repo. Sorry for the noise.

webpack.prod.js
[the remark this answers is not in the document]
montogeek: Yes, that config change shouldn't be part of this PR.
I was a side change when navigating the site testing titles changed between pages.

webpack.prod.js
montogeek: @skipjack I suspected as much, will move to `server.jsx` file

EugeneHlushko: i think moving this part:
```
const paths = Content.children.reduce((paths, page) => {
  if (page.type === 'directory') {
    page.children.forEach(child => (paths[child.url] = child.title));
  } else {
    paths[page.url] = page.title;
  }

  return paths;
}, {});

...
let title = paths[locals.path];
```

To server JSX is actually a good idea, also i just tried

```
new SSGPlugin({
        crawl: true,
        paths: [
          '/',
          '/configuration/'
        ],
        ....
      })
```
and it seems to work, but for some reason it also crawls outgoing urls from backers

montogeek:
@skipjack Addressed your comments

DISCUSSION.md
```diff
@@ -14,7 +14,7 @@ this branch:
 - [ ] Finish re-incorporating mobile sidebar
 - [ ] Re-integrate google-analytics
 - [ ] Re-incorporate `redirects.json`
-- [ ] Populate page title in `server.jsx` (fernando)
+- [x] Populate page title in `server.jsx` (fernando)
```
skipjack: Drop this to resolve the conflict (see comments on other other PRs for context).

src/server.jsx
```diff
@@ -16,22 +16,35 @@ const bundles = [
...
-    <StaticRouter location={ locals.path } context={{}}>
+    <StaticRouter location={locals.path} context={{}}>
       <html>
         <head>
-          <meta charSet="UTF-8" />
+          <meta charset="utf-8" />
           <meta name="theme-color" content="#2B3A42" />
           <meta name="viewport" content="width=device-width, initial-scale=1" />
-          <title>{/* TODO */} | webpack</title>
+          <title>{title} | webpack</title>
```
jeremenichelli: Shouldn't we add a check for `index.md`? If I'm not wrong the result will be: _webpack | webpack_, maybe removing the title there and defaulting to _webpack_ when it's not present.

src/server.jsx
```diff
@@ -16,22 +16,35 @@ const bundles = [
...
+
+const paths = Content.children.reduce((paths, page) => {
+  if (page.type === 'directory') {
+    page.children.forEach(child => (paths[child.url] = child.title));
+  } else {
+    paths[page.url] = page.title;
+  }
+
+  return paths;
+}, {});
```
jeremenichelli: Does it makes sense to generate a `_paths.json` file after `_content.json` is created? We could import it and avoid doing this every time the server file runs, it might improve performance on SSR considering we have lots of pages.

jeremenichelli: If this is the case Fernando, we can leave it for further optimization and I can tackle that once this is merged :)

skipjack: I think better than another file would just be to have a `walk` or `find` utility to find the page node within the main tree (rather than creating a separate array and walking it).

jeremenichelli: Good, to make it clear this is not a blocker for the PR to be merged.

skipjack: Yeah, I think once we've tackled everything I may try to spend a day or two just doing some cleanup and extraction of code to other packages so that it's fairly clean before going into `master`.

montogeek:
@jeremenichelli Addressed your comment, and also fixed weird characters issue: `<title>webpack<!-- --> | webpack</title>`

montogeek:
It is better to do it here now, I will work on it later today

src/server.jsx
skipjack: In #2108, I added a `content-utils.js` file that should make things like this a breeze using the `WalkContent` or `FindInContent` method. I think we should get #2108 and #2074 merged into `rebuild` and then really start a push to finish up the rest of these. From what I've seen, most of the `rebuild` PRs are pretty close and may just need a little refactoring and abstracting to a separate package (e.g. `remark` plugins). Skimming over them again now.

skipjack:
@montogeek I'm gonna wrap this one up. With #2074 in and the content utilities some of the code here can be simplified and adding `react-document-title` should be fairly easy (for populating page titles during SPA routing). Trying to finish this and get it merged tonight.

PR #2041, opened 2018-04-14
chore(rebuild): re-incorporate anchor links

montogeek:


md.js
```diff
@@ -0,0 +1,31 @@
+var markdown = require('remark');
+var slug = require('remark-slug');
+var visit = require('unist-util-visit');
+
+module.exports = function getAnchors(content) {
+  let anchors = [];
```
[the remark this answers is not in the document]
skipjack: I can give you a really simple one I wrote for replacing text within nodes as an example. Docs-wise you should start here though:

https://github.com/unifiedjs/unified#plugin

And I bet @wooorm would answer any questions you have (I'll do my best as well). I think what you have is pretty close, and you should add an option to allow selection of header levels. Then we can use this within the `DirectoryTreePlugin` `enhance` option to enrich each markdown node with the anchor links we need.

We'll need to synchronize this with #2043. Not sure what the best approach is off the top of my head. Might make sense to make these changes in the same PR.

skipjack:
Also, @montogeek let's abstract any newly created `remark` plugins to their own repo and `npm link` them for testing. If you want to keep playing around with it locally for the meantime that's fine, but I'd like to do the extraction before we merge.

src/components/Sidebar/Sidebar.jsx
```diff
@@ -15,6 +15,8 @@ export default class Sidebar extends Component {
     let { fixed, availableHeight, maxWidth } = this.state;
     let group;
 
+    console.log(pages);
```
jeremenichelli: This can be removed, right?

src/components/Page/Page.jsx
```diff
@@ -71,7 +71,7 @@ class Page extends React.Component {
     if ( this.props.content instanceof Promise ) {
       this.props.content
         .then(content => this.setState({
-          content
+          content: content.default
```
skipjack: This should be dropped as well since it was addressed on `rebuild`.

webpack.common.js
```diff
@@ -35,7 +36,9 @@ module.exports = (env = {}) => ({
           options: {
             plugins: [
               // TODO: Add necessary remark plugins
+              require('remark-slug'),
               require('remark-autolink-headings'),
+              require('remark-html'),
```
skipjack: `remark-html` should be dropped as the `remark-loader` already includes it at the end of the chain.

PR #2042, opened 2018-04-14
chore(rebuild): use `sticky` for fixed sidebar

montogeek:


skipjack:
This is perfect, I'm glad you played around with it. We should use `position: sticky` for the `Gitter` component as well.

skipjack:
@montogeek cleaned up the changes here and rebased on `master`. I also tested and...

As far as I can tell this doesn't work as is (at least in Chrome 65.0.3325.181 on OSX) _without_ disabling `flex` positioning (and basically every other style) on the top-level `Site` element. I think the only thing we lose from this is the `min-height: 100vh` combined with page content's `flex: 1 1 auto` style so the page stretches perfectly to the bottom.

We could use a `min-height: calc(100vh - [NavigationHeight] - [FooterHeight]` however the `NavigationHeight` is dynamic, and these would have to be hardcoded unless we add more js to handle resizing. Any other ideas? Does this PR currently work as is for anyone? Afaik `flex` doesn't play well with `sticky` but maybe it's just some abnormality on my machine.

skipjack:
Wow, my diagnosis above was completely off. It wasn't the `flex` positioning that was throwing it off, it was simply the `overflow: hidden` on `.site` (which I can't justify keeping for any reason). I'm going to try the same approach for the `<Sponsors />` and `<Gitter />`  components before merging.

skipjack:
🎉 🎉 :shipit:

PR #2043, opened 2018-04-14
chore(rebuild): re-incorporate header links

montogeek:
Src code patched until
https://github.com/ben-eb/remark-autolink-headings/pull/41 is merged, if
ever, we should probably use a fork.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "yaml-frontmatter-loader": "^0.1.0"
   },
   "dependencies": {
+    "@rigor789/remark-autolink-headings": "^5.1.0",
     "ajv": "^5.5.2",
     "docsearch.js": "^2.5.2",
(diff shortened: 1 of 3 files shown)


wooorm:
Ohh, seems Ben isn’t very active in remark anymore. Let me see and check in with him if he’s interested in moving it to @remarkjs

skipjack:
@montogeek not seeing any links pop up in the `h*` elements... do I need to install the package from that fork in order for it to work?

skipjack:
Nvm used @rigor789's fork, `@rigor789/remark-autolink-headings`, which works. Cleaning up the styling slightly and then pushing/merging and knocking another one down 🎉. Thanks for digging into this @montogeek.

montogeek:
@skipjack Yep, I patched it locally applying same changes as that fork

PR #2044, opened 2018-04-14
chore(rebuild): handle custom blockquotes

montogeek:


tip.js
```diff
@@ -0,0 +1,76 @@
+// var unified = require('unified');
```
montogeek: Should also be moved to a subdirectory, see https://github.com/webpack/webpack.js.org/pull/2046#discussion_r181557862
Need clean up, and tests would be cool!

tip.js
```diff
@@ -0,0 +1,76 @@
...
+// ?> This is a todo.`;
+
+// const header = `## Header`;
+
+// const md = "T> The `output` property has [many more configurable features](/configuration/output) and if you like to know more about the concepts behind the `output` property, you can [read more in the concepts section](/concepts/output)."
+
+const classMap = {
+  'T>': 'tip',
+  'W>': 'warning',
+  '?>': 'todo'
```
dhruvdutt: Great work. :ok_hand:

skipjack:
Rebased and refactored slightly to make this a proper `remark` plugin. @montogeek I'm naming these plugins `remark-***` for now and keeping them in the `/src/utilities` directory but you should think about separating them out to repos in your namespace or in the `remarkjs` organization.

> Need clean up, and tests would be cool!

Cleanup done (at least for simple `mapping` feature) but yeah tests would be great. I'd recommend doing that once it's separated to another repo though. The whole point of creating these is so that we can abstract and maintain elsewhere.

We want to keep the codebase here as light and easy to maintain as possible. This allows the site to remain content-focused and allows others to use the same tools we build for it elsewhere. There's probably even more code (e.g. certain components) that can be factored out into shared open-source packages even after the `rebuild` branch is merged.

cc @jeremenichelli

montogeek:
@skipjack I will try to find some time this week to do it, thanks for cleaning it out

jeremenichelli:
It would be super good to turn our remark plugins to packages. We should
reach out if the remark org is interested in hosting them too.
-- 

*Jeremias Menichelli*


*web developer*
skype | jeremias.menichelli
github | jeremenichelli

PR #2045, opened 2018-04-14
chore(rebuild): re-incorporate google-analytics

montogeek:


src/index.jsx
```diff
@@ -1,15 +1,19 @@
...
+
+if(process.env.NODE_ENV === 'production') {
+  gaID = 'UA-46921629-2';
+}
 
 // Client Side Rendering
 if ( window.document !== undefined ) {
   ReactDOM.render((
-    <BrowserRouter>
+    <BrowserRouter id={gaID}>
```
montogeek: This component triggers an error locally because `id` is required. I think it is fine, haven't tried passing an empty string, or any string for that matter.

skipjack: Yeah, maybe default `gaID` to an empty string and see what happens? If not, we should change it to something like:

``` js
import { BrowserRouter, Route } from 'react-router-dom'
import { BrowserRouter as AnalyticsRouter } from 'react-g-analytics'

if ( process.env.NODE_ENV === 'production' ) {
  BrowserRouter = AnalyticsRouter
}
```

Then in dev, the normal `BrowserRouter` will simply ignore the `id` prop.

webpack.common.js
```diff
@@ -35,7 +35,15 @@ module.exports = (env = {}) => ({
...
-              require('remark-autolink-headings'),
+              require('./tip'),
+              require('remark-slug'),
+              [
+                require('remark-autolink-headings'),
+                {
+                  behaviour: 'append'
+                }
+              ],
+              require('remark-html'),
```
skipjack: Can you separate these changes to another PR? They seem unrelated to the google-analytics fix.

webpack.dev.js
```diff
@@ -5,31 +5,35 @@ const HTMLPlugin = require('html-webpack-plugin');
...
+    ],
+    devServer: {
+      contentBase: path.resolve(__dirname, './dist'),
+      port: 3000,
+      hot: true,
+      inline: true,
+      compress: true,
+      historyApiFallback: true
+    }
+  });
```
skipjack: Same here, I'm not exactly sure what even changed. Was this intentional?

skipjack: Ah you have a sublime plugin for it or something, I see now. Yeah if you don't mind, let's wait to make all those changes as a separate commit until after this is done and merged. Then we can integrate it into the build process in a way that hits everyone's builds.

EugeneHlushko:
I would suggest implementing ga for prod before merging
```
import { BrowserRouter, Route } from 'react-router-dom'
import { BrowserRouter as AnalyticsRouter } from 'react-g-analytics'

if ( process.env.NODE_ENV === 'production' ) {
  BrowserRouter = AnalyticsRouter
}
```

src/index.jsx
montogeek: That would be cool, but `import`s are defined as `const`

montogeek: What about:
```
const BrowserRouter = process.env.NODE_ENV === 'production' ? require('react-g-analytics') : require('react-router-dom').BrowserRouter;
```

EugeneHlushko: yes or if we want imports just define another variable:
```
Router = process.env.NODE_ENV === 'production' ? AnalyticsRouter : BrowserRouter;
...
<Router ...>
```

montogeek:
Feedback addressed.

src/index.jsx
```diff
@@ -1,22 +1,22 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { BrowserRouter, Route } from 'react-router-dom';
+import AnalyticsRouter from 'react-g-analytics';
 import Site from './components/Site/Site';
 
-// TODO: Re-integrate <GoogleAnalytics analyticsId="UA-46921629-2" />
-// Consider `react-g-analytics` package
+const Router = process.env.NODE_ENV === 'production' ? AnalyticsRouter : BrowserRouter;
```
jeremenichelli: We could use `require` here inside a conditional and only import the browser required for produciton. DOn't know if with this approach `BrowserRouter` gets dead code eliminated, if it does I'm fine with it.

montogeek: I think it get DCE, will confirm later.

jeremenichelli: One more thing, shouldn't we keep the _analyticsId_ in a dot file so we don't expose it here, though we already did either way I guess 😅

montogeek: GA keys are public any way or do you mean just for configuration purposes? We can put it into package.json file.

EugeneHlushko: Yeah its safe to expose 👍 moving it out to single source of truth from the template is another question, which i suspect shouldn't be tackled in this PR

PR #2046, opened 2018-04-14
chore(rebuild): re-implement prism highlighting

montogeek:


highlight.js
```diff
@@ -0,0 +1,50 @@
+const refractor = require('refractor/core.js');
```
montogeek: This plugin should be moved to another place `src/utilities/remark/plugins` would be a good place, we could also create a package from it since it doesn't exist.

src/content/contribute/release-process.md
```diff
@@ -16,7 +16,7 @@ When merging pull requests into the `master` branch, select the _Create Merge Co
 
 ## Releasing
 
-```sh
+```bash
```
dhruvdutt: :medal_sports:

skipjack:
Updated this one as well before merging but I didn't extract too much to `options` so this one should probably be given more thought before prior to being extracted into an actual `remark` plugin. For example, `langs` could be an option and be used to detect whether the language will be parsed or not.

PR #1992, opened 2018-04-06
feat(site) Add v3 docs to deployment

montogeek:
Added v3 docs to deploy script by git checking `fix/v3-docs` branch which have necessary changes to make site works under a subfolder and also last v3 docs.

PR #2048, opened 2018-04-14
chore(rebuild): finish navigation refactoring

skipjack:
Re-integrate icon links and searching in the navigation bar. Use the `next` version of `react-banner` which supports more generic banner items.

> For those who don't know, `react-banner` is an abstraction of our `Navigation` component that I built a while ago but couldn't integrate due to build process limitations. Using this package significantly simplifies how much we have to manage within this repository.


src/components/Navigation/Navigation.jsx
```diff
@@ -1,67 +1,67 @@
...
       <Banner
+        blockName="navigation"
         logo={ <Logo light={ true } /> }
         url={ pathname }
-        links={ links }
+        items={[
+          ...links,
+          {
+            title: 'GitHub Repository',
+            url: '//github.com/webpack/webpack',
```
EugeneHlushko: this will always result in https as gh pages forcing `https`, maybe type https explicitly here too like for the other urls below?

src/components/Site/Site.jsx
```diff
@@ -62,16 +62,16 @@ class Site extends React.Component {
...
               children: this._strip(
                 sections.filter(item => item.name !== 'contribute')
               )
             },
-            { title: 'Contribute', url: '/contribute' },
-            { title: 'Vote', url: '/vote' },
-            { title: 'Blog', url: '//medium.com/webpack' }
+            { content: 'Contribute', url: '/contribute' },
+            { content: 'Vote', url: '/vote' },
+            { content: 'Blog', url: '//medium.com/webpack' }
```
EugeneHlushko: here too

src/components/Site/Site.jsx
```diff
@@ -62,16 +62,16 @@ class Site extends React.Component {
           toggleSidebar={ this._toggleSidebar }
           links={[
             {
-              title: 'Documentation',
+              content: 'Documentation',
               url: '/concepts',
```
montogeek: We need to rethink how URLs are formed, when trying to build the site to work under a URL I had to do a lot of changes (https://github.com/webpack/webpack.js.org/tree/fix/v3-docs), we need to work on it since now in order to be able to support multiple versions in the future

src/components/Navigation/Navigation.jsx
skipjack: Will do, I think this was an artifact from a while back.

src/components/Site/Site.jsx
skipjack: Yeah that's definitely something we need to tackle _at some point_, but is that something that needs to happen before `rebuild` is merged? Will these routes conflict with the `/v3` route you created?

Personally I think figure out version management as a follow up after finishing `rebuild` (though I understand we did need to get something up quickly before for v3 when deploying v4).

EugeneHlushko: this is indeed a pain point for community, though lets try to keep it to the plan and get `rebuild` into master asap before diving into much detail on doc versioning

montogeek: No, after merging `rebuild` is fine, would be better to do it before so we can launch new website with multiple versions support, that would make headlines! (jk).

I have tried to push that change, without success https://github.com/webpack/webpack.js.org/pull/1992 :/

Munter:
I can help out finding any links that are of the wrong form with assetgraph. There's a bit of extra work bridging the Gap back to the markdown file, but I'm sure we can script that

src/components/Navigation/Navigation.jsx
```diff
@@ -1,67 +1,67 @@
+// Import External Dependencies
 import React from 'react';
 import Banner from 'react-banner';
+import DocSearch from 'docsearch.js';
+
+// Import Utilities/Images
+import GitHubIcon from '../../styles/icons/github.svg';
```
jeremenichelli: Is this dependency used? I can't spot it. I'm seeing we are using the class for the icon font instead.

skipjack: Good catch, I’ll remove it. Played around with it momentarily and the fixed font metrics instead.

PR #2049, opened 2018-04-14
chore(rebuild): exclude external urls from ssg process

EugeneHlushko:
* prevents urls like `website.com` getting fetched and built into ./dist

Filed up a PR against master of SSGWP markdalgleish/static-site-generator-webpack-plugin#128 to remove fork ref from package.json if gets accepted

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "request": "^2.81.0",
     "sass-loader": "^6.0.6",
     "sitemap-static": "^0.4.2",
-    "static-site-generator-webpack-plugin": "^3.4.1",
+    "static-site-generator-webpack-plugin": "git://github.com/EugeneHlushko/static-site-generator-webpack-plugin.git#957984e4fe7592f534b1949736bef7251b6e832c",
     "style-loader": "^0.18.2",
     "tap-min": "^1.2.1",
     "tap-parser": "^6.0.1",
--- a/webpack.prod.js
+++ b/webpack.prod.js
@@ -30,7 +30,8 @@ module.exports = env => [
         crawl: true,
         globals: {
           window: {}
-        }
+        },
+        disallowDomains: true
       })
     ],
     output: {


Munter:
I think this is the wrong fix. Static site generator already excludes urls that are absolute or protocol relative. When we get wrong output here is because our own urls are invalid.

Most of these are caused by opencollective not cleaning out their profile urls. I had a fix for that merged in their API, but they still send them out wrong.

We can fix it locally by prepending the missing http:// to those urls

package.json
```diff
@@ -90,7 +90,7 @@
     "request": "^2.81.0",
     "sass-loader": "^6.0.6",
     "sitemap-static": "^0.4.2",
-    "static-site-generator-webpack-plugin": "^3.4.1",
+    "static-site-generator-webpack-plugin": "git://github.com/EugeneHlushko/static-site-generator-webpack-plugin.git#957984e4fe7592f534b1949736bef7251b6e832c",
```
dhruvdutt: I think using `git://` can lead to new issues. Maybe use `https://` or can we wait for a while if your PR gets merged? :)

Munter:
https://github.com/opencollective/opencollective-api/pull/1174/files for reference

Munter:
OpenCollectives api will start responding with `http://` prepended urls for backers websites now: https://github.com/opencollective/opencollective-api/pull/1174

So this change might not be needed

EugeneHlushko:
just curios why it says http everywhere, is it not https aware or it is only capable to prepend http and then just fingers crossed for a redirect to a safe location from http? :) Im happy to drop this one out but  

On a second note i dont like that SSGWP doesnt detect a dot in a protocolless url, it can always slip through and we get a folder in a dist with unnecessary code in index.html
i know this is not something to worry much about on a documentation website but i can see at least two vectors for an attack xD

Munter:
You can't assume https unless you actually check for it. I'd have preferred to prepend https as well, and the library can of course do it.

SSGWP already handles protocol relative urls: https://github.com/markdalgleish/static-site-generator-webpack-plugin/pull/128/files#diff-168726dbe96b3ce427e7fedce31bb0bcR201

Your PR to them effectively says, if any url at this point contains a dot, return null. Dots are valid characters in paths and file names. So your PR would break their url resolving capabilities

PR #2050, opened 2018-04-15
docs(guides): change bundle filename

ayvarot:
Replace `main.js` by `bundle.js` to make sense with the text used in the second paragraph of section `Modules`.


montogeek:
PR is not longer valid, all references now use main.js

PR #2051, opened 2018-04-16
fix(docs) Document output.pathinfo default values

### LINKED ISSUE #2014: Document pathinfo is on for dev builds.
Current docs [state it defaults to `false`](https://webpack.js.org/configuration/output/#output-pathinfo ) but it is [enabled for dev builds](https://github.com/webpack/webpack/blob/b30de38eb5311051388db05287aa3fd951fa41de/lib/WebpackOptionsDefaulter.js#L147-L150).

---

This issue was moved from webpack/webpack#7008 by @montogeek. Original issue was by @jdalton.

EugeneHlushko:
* mention default values for `output.pathinfo`

Fixes #2014

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -616,7 +616,7 @@ Note that `[hash]` in this parameter will be replaced with an hash of the compil
 
 `boolean`
 
-Tell webpack to include comments in bundles with information about the contained modules. This option defaults to `true` in `development` and `false` in `production` mode respectively. Please note that `output.pathinfo` **should not** be used in production, but it's very useful in development when reading the generated code.
+Tell webpack to include comments in bundles with information about the contained modules. This option defaults to `true` in `development` and `false` in `production` [mode](/concepts/mode/) respectively. Please note that `output.pathinfo` **should not** be used in production, but it's very useful in development when reading the generated code.
 
 ``` js
 pathinfo: true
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.md
```diff
@@ -616,7 +616,7 @@ Note that `[hash]` in this parameter will be replaced with an hash of the compil
 
 `boolean`
 
-Tell webpack to include comments in bundles with information about the contained modules. This option defaults to `false` and **should not** be used in production, but it's very useful in development when reading the generated code.
+Tell webpack to include comments in bundles with information about the contained modules. This option defaults to `true` in `development` and `false` in `production` [mode](/concepts/mode/) respectively. Please note that `output.pathinfo` **should not** be used in production, but it's very useful in development when reading the generated code.
```
jeremenichelli: I think we should change _Tell_ for _Tells_. Also the final part of the sentence is a little confusing about using the property or enabling it.

I would change it to: _While the data this comments can provide is useful during development, it's recommended to be disabled for production._

I would even put it in a warning block `W>`

EugeneHlushko: rephrasing this a big so it explicitly states it shouldnt be used in prod:

```
W> While the data this comments can provide is very useful in development when reading the generated code, it **should not** be used in production.
```

**when reading the generated code** - should give a clue to the reader what exactly this gives

EugeneHlushko:
@jeremenichelli i did apply warning block with a small modification, pls re-review

PR #2054, opened 2018-04-17
Update default value of resolve.extensions

Volune:
According to WebpackOptionsDefaulter.js


src/content/configuration/resolve.md
```diff
@@ -142,7 +142,7 @@ enforceModuleExtension: false
 Automatically resolve certain extensions. This defaults to:
 
 ``` js
-extensions: [".js", ".json"]
+extensions: [".wasm", ".mjs", ".js", ".json"]
```
dhruvdutt: Should we change the ordering?

montogeek: It doesn't matter

montogeek:
Thanks!

PR #2059, opened 2018-04-18
Update code-splitting.md

appeltonen:
Updated per https://github.com/webpack/webpack.js.org/blob/master/src/content/plugins/commons-chunk-plugin.md


montogeek:
Thanks!

PR #2062, opened 2018-04-18
Fix rogue tab in index.md

maggo:
There was a tab hiding in the code example on the index page that messed up the indentation


maggo:
CI failed because of api rate limiting

maggo:
Seems like this was fixed in another commit. Closing, thanks for the quick reaction!

PR #2072, opened 2018-04-21
Rebuild fetch files

jeremenichelli:
This adds all fetch files into the new `rebuild` flow.

- Files gets fetched with an underscore, later the enhacer of the `DirectoryTreePlugin` detects the files and removes them from computed route name.
- I did a lot of gardening on the fetch scripts, though I'm keeping the `.sh` one as entry point since moving the API to JavaScript is a little unnecessary, I gave it a try and it's just more spaghetti code.
- Moved the Node engine verison, this allows us to use `promisify`, all fetch flows are now Promise based, no more callback which resulted on tons of lines removed and more readable code.
- Updated some version and packages names to prevent warning logs and take the project to a more sane state.
- Created `yarn.lock` file to match the rest of the projects in the organization
- A lot of processing functions were moved to new `utils` folder for better file organization.


jeremenichelli:
And build passes 💃

jeremenichelli:
Did we disabled builds on PRS? 🤔

src/components/Markdown/Markdown.scss
```diff
@@ -19,14 +19,14 @@
...
 
-    &:first-child {
-      margin-top: 0;
-      line-height: 1;
-    }
-
     tt, code { font-size: 90%; color: inherit }
   }
 
+  h1:first-child {
```
montogeek: What is the reason of this change?

src/utilities/yamlHeadmatter.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function yamlHeadmatter(fields) {
```
montogeek: What is this doing?

src/components/Markdown/Markdown.scss
jeremenichelli: This is because, when a README has an **h2** heading as the first child, if we remove the margin it gets collapsed with the page title and it looks bad.

src/utilities/yamlHeadmatter.js
jeremenichelli: This build the yaml matter for the fetched README files, I didn't add this I just moved it to an specific file so the `fetchPackagedFiles` script looks cleaner.

PR #2073, opened 2018-04-22
fix: avoid repeating browser checking

jeremenichelli:
Small improvement to avoid repeating browser global checking in code.


jeremenichelli:
Can you guys take a look at this one? It would simplify some checks I'm doing right now for server side rendering.

src/utilities/isBrowser.js
```diff
@@ -0,0 +1,3 @@
+const isBrowser = window !== undefined & window.document !== undefined;
```
EugeneHlushko: bitwise AND operator? i think you meant `&&` here

montogeek: What you think about `isClient`?

EugeneHlushko: i also thought this was a bit too explicit. i would maybe suggest renaming the whole file into 
`browser.js` and then if we need more similar utils export an object by default with keyed utils like that:

```
const isClient = window !== undefined && window.document !== undefined;
export default {
  isClient,
}
```

that way less changes will land on the file as it will be future proof

jeremenichelli: Not sure about the last comment, doesn't make any sense to rename it `isClient` and then call the file browser. If we want to group things in a bigger constant file we can always do it.

EugeneHlushko: just felt that one file per condition is a bit too much of abstraction

montogeek: What about infamous and to be forgotten `utils.js`?

jeremenichelli: I think is a little bit redundant to have a `utils.js` file in the `utilities` folder. About it being a single condition, it's not an abstraction, it's the nature of the util, some utils will be lots of lines, some others will take more.

jeremenichelli:
@EugeneHlushko did some corrections, renaming, plus rebasing. What do you think? The goal with this is to avoid repeating conditional across the project for something that simple 😃

PR #2074, opened 2018-04-24
Rebuild static html fix

jeremenichelli:
Initially this PR solves a problem that the html files were generated with `[object Object]` content.

That said, we have anothr problem, and it's a BIG one because I know how hard it is solve the server side vs. async content loading issue in React.

Let me show you why. Right now we are creating the pages so they are static and already gather the content for it.

When the bundle hits the browser and runs, it detects the Promise and cleans the content state whiping out the server side generated content and remounting it again.

kapture 2018-04-24 at 14 30 57

This is serious because if we don't solve it we might need to move to full SPA which for a documentation site I don't recommend, or move to another static generator like Gatsby, Next or Nuxt.


EugeneHlushko:
Hydrate should do, i will try to take a look tonight at this

jeremenichelli:
it doesn’t because we have a promise setting the state internally. We are
also passing a function to Site which will always trigger a new render on
the client.

I tried dozens of stuff but nothing worked. Feel free to try it, I suggest
you branch from this PR.
-- 

*Jeremias Menichelli*


*web developer*
skype | jeremias.menichelli
github | jeremenichelli

skipjack:
Yeah and using `remark-react` may help as well. The only problem with that is image resolving/loading but we could solve that with another remark plugin. I’m at a conf this week so pretty busy but I’ll get back to reviewing/merging/helping this weekend.

jeremenichelli:
I'm completely burnt out around this one, will see if I can tackle on the weekend and with a cleared mind something pops up.

jeremenichelli:
@skipjack though I should look at remark and see what it doesn't, I have the feeling that is not going to solve our issue here. I will give it one more try between today and tomorrow to see if I can avoid this, I really want to make sure we did everything we can before jumping into other solutions like Next, nuxt.js or Gatsby.

jeremenichelli:
Either way, this fix is necessary for properly render static pages and hydrate the DOM, so if it looks good, approave it so we can merge it and take this as base for future improvements.

skipjack:
@jeremenichelli yeah I wouldn't worry too much about `remark` and `remark-react` yet as that's not the main issue. This PR looks good to me but I haven't tested yet so I'll do that soon.

I think I have an idea of how to fix the first (and bigger) issue though. We should be able to pre-load that first bundle in `src/index.jsx` and pass that down so there's no flash (i.e. "loading...") on the first `render`. Meaning, do one top-level `import()` so routing can remain in the `Site` and be utilized by both `index.jsx` and `server.jsx`. I'll try this out when I pull and test this branch and let you know what happens. If it works, I'll separate the fix as another PR.

EugeneHlushko:
Sorry for late reply, going to pull this and test tomorrow

jeremenichelli:
I'll see if I can avoid hydrate from remounting the whole component too.

Someone form us should have a solution, I suggest branching from this PR.

montogeek:
Great work @skipjack !

PR #2078, opened 2018-04-26
Update mode.md

MarkJamesHoward:
Synced up the sample code and information for 'development' mode. The namedChunks plugin is specified as being added when using 'development' mode but is not shown in the sample code


montogeek:
Thanks!

PR #2085, opened 2018-04-27
Added documentation for [contenthash] #6839

FadySamirSadek:
Added output file name example for the new introduced [contenthash]


src/content/configuration/output.md
```diff
@@ -213,6 +214,12 @@ Using hashes based on each chunks' content:
 filename: "[chunkhash].bundle.js"
 ```
 
+Using hashes generated for extracted content :
```
montogeek: No space before colon

FadySamirSadek:
@dhruvdutt @montogeek I applied the requested changes but there is an unrelated error in the scripts that is breaking the build and I believe it will do the same for all new PRs

src/content/configuration/output.md
```diff
@@ -213,6 +214,12 @@ Using hashes based on each chunks' content:
 filename: "[chunkhash].bundle.js"
 ```
 
+Using hashes generated for extracted content:
```
edmorley: Is extracted content just content generated by eg `mini-css-extract-plugin` or does it include all other content (eg JS bundles) too? How does it differ from `chunkhash`?

FadySamirSadek: chunkhash is calculated for a chunk (an entry file) while content hash is a special hash generated by eg mini-css-extract-plugin and is calculated for extracted content only not the whole chunk content. This is very useful in the case of mini-css-extract-plugin as when an extracted css is changed the css hash only is invalidated not the entire chunk. 
did not want to add all these extra info here but will create a PR and add it with an example in https://webpack.js.org/guides/caching/#output-filenames  as per your issue

montogeek:
@FadySamirSadek Can you add a note regarding https://github.com/webpack/webpack.js.org/issues/2096 ?

montogeek:
Thanks!

PR #2090, opened 2018-04-29
Add temporary warning

buoyantair:
I guess the "Prevent Duplication" section has to be re-written due to the changes :3 But for now maybe a flag that points to the updated plugin should be okay :D 


jeremenichelli:
I think that here, instead of just putting a warning here we should re-write the section to not include the old `CommonChunksPlugin`, it makes no sense to have a whole section based on it if it's not complaint with our latest release.

What do you think @montogeek

buoyantair:
@jeremenichelli Yeah that's definitely what we need to do 👍

montogeek:
We can't remove, there is a lot of people using v3, we know v4 docs are not ready to be finished, so there is hundred of developers waiting for full docs to start migrating.

Also, we haven't solved the docs versioning, if we had versioned docs we could do it, but we don't, so we still need to show both pages.

jeremenichelli:
But if we solve that, then this content will lie in the 3.0 tag, right?

montogeek:
@jeremenichelli Yes

montogeek:
Thanks!

Issue #2096, opened 2018-05-02
Use `[contenthash]` over `[hash]` and `[chunkhash]` for better long term caching


evilebottnawi:
For long term caching better use `[contenthash]` over `[chunkhash]`.

From slack
```
sokra [2:59 PM]
A chunk contains modules i. e. a.js, b.css, and c.wasm. This results in 3 assets being generated JS, CSS and WASM. `[chunkhash]` is the hash of the complete chunk, which is JS + CSS + WASM. `[contenthash]` is the hash of the content of the file, which is different for each asset.

evilebottnawi [3:10 PM]
I.e. better use `[contenthash]` for chunks?

sokra [5:26 PM]
@evilebottnawi better use `contenthash`.

sokra [5:30 PM]
we probably should rename that in webpack 5 to `[hash]`.

evilebottnawi [5:34 PM]
Only one question why we need `chunkhash` after implement `contenthash` :smile: ?

sokra [5:39 PM]
not sure if there is a use case where chunkhash is better, but at least for backward-compat
```

Ref on issue in `webpack` https://github.com/webpack/webpack/issues/7138.
Example of problem https://github.com/webpack-contrib/mini-css-extract-plugin/issues/117.

FadySamirSadek:
@evilebottnawi @montogeek does this mean chunkhash should be removed from the exampled in the guide and replaced with contenthash with only a mention that sometime it existed so people know it is still supported ?

evilebottnawi:
@FadySamirSadek alright, just backward compatibility

evilebottnawi:
@FadySamirSadek friendly ping, can we solve this problem faster? Looks like many people faced with this

FadySamirSadek:
@evilebottnawi will finish it by the end of today

Streemo:
I had this config and was running into the issue where a change in `./src/styles.css` would cause the `chunkhash` for the main output js file to change, even if `./src/index.js` was constant:
```
config = {
  entry: "./src/index.js",
  plugins: [
    new MiniCSSExtractPlugin({filename: "[contenthash].css"}),
    ...
  ],
  output: {filename: "[chunkhash].js",
  ...
}
```
Changing the output to `{filename: [contenthash].js}` fixed the issue, causing the browser to only reload the changed css file, as expected.

FadySamirSadek:
@Streemo  I am working on fixing the documentation of `[contenthash]` but it has not been published yet

EugeneHlushko:
[...]

PR #2098, opened 2018-05-03
Replace [chunkhash] with [contenthash]

FadySamirSadek:
Replaced [chunkhash] with [contenthash] in code examples and defined [contenthash]

### DIFF
--- a/src/content/guides/caching.md
+++ b/src/content/guides/caching.md
@@ -6,6 +6,7 @@ contributors:
   - jouni-kantola
   - skipjack
   - dannycjones
+  - fadysamirsadek
 related:
   - title: Predictable Long Term Caching
     url: https://medium.com/webpack/predictable-long-term-caching-with-webpack-d3eee1d3fa31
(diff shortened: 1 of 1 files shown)


FadySamirSadek:
#6839  #2085 #2096

evilebottnawi:
@FadySamirSadek looks good, need check other places on `[chunkhash]` to avoid misleading

FadySamirSadek:
@evilebottnawi Will compile a list of all places that mention `[chunkhash]` and work on them. Thank you 👍

naruaway:
Hi, it seems like still https://webpack.js.org/configuration/output/ recommends the use of `chunkhash` rather than `contenthash`.
If we should always use `contenthash` rather than `chunkhash`, should we also update this page?

EugeneHlushko:
@naruaway i have looked at the page and all references there look valid, which one exactly did you mean?

naruaway:
@EugeneHlushko In that page, `contenthash` is almost only mentioned in the sentence below.
But now contenthash is not only meant for ExtractTextWebpackPlugin use case, right?
```
When using the ExtractTextWebpackPlugin, use [contenthash] to obtain a hash of the extracted file (neither [hash] nor [chunkhash] work).
```

EugeneHlushko:
The "note" says that when using `ExtractTextWebpackPlugin`, you shouldn't use other hash placeholders other than `contenthash`. So that is still the case. I suggest you add contenthash to the table here and that would be it:
Screenshot_2019-07-30 Output webpack

What do you think @naruaway

naruaway:
Yes, should we even replace `[chunkhash]` with `[contenthash]` in that table? 
The title of this PR is "Replace [chunkhash] with [contenthash]".
Does `[chunkhash]` still have its valid usage?

I was confused when I read the doc.

evilebottnawi:
@naruaway yes, `chunkhash` still valid, but is not good for long cache, but can be useful fur other purpose

EugeneHlushko:
yup so just add `[contenthash]` to the table should be it

EugeneHlushko:
@naruaway are you at it?

EugeneHlushko:
this was stalled, so i fixed it in #3233 to close #3209

PR #2106, opened 2018-05-05
change reference from dist/bundle.js to dist/main.js

ATGardner:
at this stage, the default webpack execution creates a dist/main.js file


ATGardner:
I see that the webpack.config.json section, later, is using `bundle.js` as the output.filename value.
So either the getting started document needs to update the `index.html` script tag to reference `bundle.js`, or the config should create `main.js`, like in the default.
I can add either option to this PR as well. What would be preferable?

montogeek:
Thanks!

PR #2108, opened 2018-05-05
chore(rebuild): improve static html utilization

skipjack:
This is an addition to #2074 that addresses some core issues with how entry content is loaded. The changes contained here accomplish the following...

- Prevent any flash while entry content loads by preloading the initial page prior to `render` call.
- Abstract all content manipulation methods out of the `Site` component for use elsewhere in the site.
- Move sorting and filtering to the top level to prevent repetition and yield consistency through the site.

### DIFF
--- a/src/utilities/content-utils.js
+++ b/src/utilities/content-utils.js
@@ -1,0 +1,45 @@
+/**
+ * Walk the given tree of content
+ *
+ * @param {object}   tree     - ...
+ * @param {function} callback - ...
+ */
+export const WalkContent = (tree, callback) => {
+  callback(tree);
+
+  if ( tree.children ) {
+    tree.children.forEach(child => {
+      WalkContent(child, callback);
+    });
+  }
+};
+
+/**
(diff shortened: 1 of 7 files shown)


src/utilities/content-utils.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+/**
+ * Get top-level sections
+ *
+ * @param  {object} tree - Any node in the content tree
+ * @return {array}       - Immediate children of the given `tree` that are directories
+ */
+export const ExtractSections = tree => {
+  return tree.children.filter(item => (
```
EugeneHlushko: `item => item.type === 'directory'`

EugeneHlushko:
Looks good, would need to clone and check locally though, still trying to get some time to do that this weekend 👍

jeremenichelli:
Just checked locally and it works like a charm, great work @skipjack. I think this is a big step towards the direction we are heading. Are we keeping the TODO up to date?

PR #2112, opened 2018-05-06
Configuration/optimization updates

### LINKED ISSUE #1968: Webpack 4 - optimization section under configuration
- [x] (#2053) Create the section itself
- [x] (#2053) minimize
- [x] (#2053) minimizer (for custom options reference webpack/webpack#6879)
- [x] (#2053) splitChunks
- [x] (#2112) runtimeChunk
- [x] (#2112) noEmitOnErrors
- [x] (#2271) namedModules
- [x] (#2271) namedChunks
- [ ] (#2456 WIP) portableRecords
- [x] (#2139) nodeEnv (also closes #2138)
- [x] (#2273) removeAvailableModules
- [x] (#2295) removeEmptyChunks
- [x] (#2295) mergeDuplicateChunks
- [x] (#2306) flagIncludedChunks
- [x] (#2345) occurrenceOrder
- [x] (#2392) sideEffects
- [x] (#2345) providedExports
- [x] (#2345) usedExports
- [x] (#2353) concatenateModules
- [x] (#2265) mangleWasmImports

### LINKED ISSUE #2103: splitChunks : none of the options is self-explanatory and documentation is cryptic
I know this ticket might not conform with the issue creation policy but I will try to compile a list of questions here hoping that they can be addressed (I am willing to make PRs to improve the documentation)

**1. optimization.runtimeChunk**
I guess this makes sense if you have more than one entry points.  In any other case, the appropriate value should be `'false'` , correct?

**2. optimization.occurrenceOrder**
Does this option still exist? (not found in documentation using search)

**3. optimization.namedChunks**
Same as above.

**4. maxInitialRequests**
I had cases when a cacheGroup was not generating a file until I increased the default maxInitialRequests value to something like "10". I have no idea why that is.

(and most importantly)
**5. test**
I am migrating a setup with which we were using simple strings to target individual node_modules which we wanted in a separate generated file to split the end result of vendors in smaller files. Some of the test rules don't seem to be doing what I think they should - is there a way to debug this? Can I just use a simple string like "moment.js" or should I target the parent directory using a regexp : `/[\\/]moment[\\/]/`. 

**6. chunks**
The documentation of the options "initial", "async" and "all" is just a repetition of the option word itself. Desperately looking for a tangible example. 

---

This issue was moved from webpack/webpack#7203 by @montogeek. Original issue was by @gkatsanos.

EugeneHlushko:
Tackles two from #1968 
Might close #2103 

- add `optimization.noEmitOnErrors` to `optimization` section
- add `optimization.runtimeChunk` to `optimization` section
- remove `NoEmitOnErrors` from plugins
- link to full description of `optimization.runtimeChunk`


src/content/configuration/optimization.md
```diff
@@ -59,3 +60,49 @@ module.exports = {
...
+```js
+module.exports = {
+  //...
+  optimization: {
+    noEmitOnErrors: true
+  }
+}
+```
+
+W> If you are using the [CLI](/api/cli/), the webpack process will not exit with an error code by enabling this plugin. If you want webpack to "fail" when using the CLI, please check out the [`bail` option](/api/cli/#advanced-options).
```
jeremenichelli: _by enabling_ should be _while this plugin is enabled.

EugeneHlushko: true, i will also update this to be `while using webpack CLI`

montogeek:
Thanks!

PR #2116, opened 2018-05-07
output.chunkFilename can no longer be function

lwyj123:
For technical reasons it must be a string.  #6672 https://github.com/webpack/webpack/pull/6672
since the [pr](https://github.com/webpack/webpack/pull/6672), output.chunkFilename can no longer be function

#2070


montogeek:
Thanks!

PR #2119, opened 2018-05-07
docs(plugins) Remove improper uglify plugin usage. Fix incorrect file…

### LINKED ISSUE #2091: Documentation still mentions Uglify
Using Uglify throws an error (in v4.6, and it should do so since 4.0):
`Error: webpack.optimize.UglifyJsPlugin has been removed, please use config.optimization.minimize instead.`
But the docu still shows its usage still as a valid example:
`https://webpack.js.org/concepts/plugins/`
This should be removed AND - since there are many who struggle with webpack - there must be a better explanation about the new code usage, where exactely to place the code. Currently, a newcomer like me simply has no idea what to do with "config...".

Dead docu writers: Usually you are programmers from what I know. But a docu is not for programmers, but for newcomers first and foremost. Therefore you really need to think about, what a newcomer needs as information! What happens here - just writing sth like "use config.... instead" is absolutely NO HELP at all!
Really! I see this problem so often all over the OSS world! Just google and count for a while where people ask "Where do I have to put that code?" It takes away so much time from all of us! Please - optimze! Thanks!

EugeneHlushko:
Fixes #2091

- use standard `__` for filenames
- remove improper uglifyjs plugin usage example


jeremenichelli:
Sorry @EugeneHlushko but I might be out of context, can you point me where the Uglify config is being removed in this PR?

montogeek:
@jeremenichelli It is deprecated https://github.com/webpack/webpack/blob/master/lib/webpack.js#L174

EugeneHlushko:
Yup, and we have minimizer and minimize both documented now under configuration/optimization section

montogeek:
Thanks!

PR #2121, opened 2018-05-07
feat: replace loading message for placeholder

jeremenichelli:
Ok, probably unexpected but I hope you like it, specially @skipjack.

What if instead of the **Loading...** message we have right now:

before-content-loading

We display a more modern placeholder animated like this:

after-2

That last example is serving the actual static + SPA content with CPU and Network throttling, I swear that live it looks really awesome and improves the experience a ton.

I'm also defering related links and contributors until the real markdown content is there.


src/components/Page/Page.jsx
```diff
@@ -10,14 +10,37 @@ import Gitter from '../Gitter/Gitter';
 // Load Styling
 import './Page.scss';
 
+// Placeholder string
```
EugeneHlushko: move to placeholder component

yarn.lock
```diff
@@ -6900,6 +6904,13 @@ react-banner@^1.0.0-rc.0:
   version "1.0.0-rc.0"
   resolved "https://registry.yarnpkg.com/react-banner/-/react-banner-1.0.0-rc.0.tgz#59627abaf240cdea33b81a9438d499af07aeb103"
 
+react-document-title@^2.0.3:
```
EugeneHlushko: it seems like a 1st level dependency, where did it come from?

yarn.lock
```diff
@@ -6957,6 +6968,13 @@ react-router@^4.2.0:
     prop-types "^15.5.4"
     warning "^3.0.0"
 
+react-side-effect@^1.0.2:
```
EugeneHlushko: this one too

EugeneHlushko:
How does "SSR" look, related and contributors still visible in prod build ? If there is no impact here then i think its a good addition

yarn.lock
jeremenichelli: For some reason after the rebase I'm seeing stuff I didn't modify like this one. I will try to clean this up.

src/components/Page/Page.jsx
jeremenichelli: I did that first, but since it just returns a string I thought it would be confusing it's not a React component. I might switch to that.

EugeneHlushko: yeah a component would be perfect imo

montogeek:
There is no way to make render faster to avoid a placeholder completely? Like the current docs, it takes less than 200ms to load

jeremenichelli:
@montogeek when you are on a wifi network you won't even see the placeholder honestly as it happens too fast. But when your wifi is crappy or you're on a mobile network it really improves the user perception and prevents the content _restacking_ effect.

@EugeneHlushko server side code remains untouch as this only affects Promise based content.

montogeek:
What happens if the user have JS deactivated? I guess old fashion links should work, right?

montogeek:
Once new React API Suspense is released we could improve this to only show it after 150ms.

jeremenichelli:
> What happens if the user have JS deactivated? I guess old fashion links should work, right?

I guess so, though that is out of the scope of this PR and it's more on the Rebuild PR plate.

src/components/Placeholder/Placeholder.js
```diff
@@ -0,0 +1,25 @@
+// Import Styling
+import './Placeholder.scss';
+
+// Placeholder string
+const Placeholder = () => (`
```
skipjack: Any reason we're not making this a real react component rather than using a template string?

src/components/Placeholder/Placeholder.js
```diff
@@ -0,0 +1,25 @@
...
+  <div class="placeholder">
+    <p class="placeholder--medium">&nbsp;</p>
```
skipjack: I think these would actually be elements not modifiers so `placeholder__medium` would make more sense to stay consistent with BEM.

src/components/Placeholder/Placeholder.js
skipjack: Oh I see, because it's passed in `dangerouslySetInnerHTML`. We could key off the `contentLoaded` state and use a proper component. Once we use `remark-react` though we can avoid `dangerouslySetInnerHTML` entirely and this can definitely become a proper component.

jeremenichelli: Yeah, that's the only reason, I actually crafted it initially as a component and had to roll back when I realized that. We can definitely refactor it when we introduce remark to the game.

src/components/Placeholder/Placeholder.js
jeremenichelli: Yeah totally.

jeremenichelli:
@skipjack I've updated the classnames, do you ahve any more concerns for now?

skipjack:
Nope, looks great.. merging.

PR #2123, opened 2018-05-08
Correct suitable for production information in devtool

sokra:
Correct information about devtools:

cheap-source-map devtools are suitable for production usage as they emit separate files


montogeek:
Thanks!

PR #2125, opened 2018-05-08
docs(plugins) Add forward slash to the regex in Example 2 (#1970)

montogeek:
Original PR was merged against the wrong branch.


EugeneHlushko:
zero changes?

montogeek:
It is not longer relevant :(, but a contribution is a contribution

PR #2128, opened 2018-05-08
Update getting-started.md

FlyingDad:
Change line 170 to 'main.js' as web pack.config.js is building filename: 'main.js', not 'bundle.js


FlyingDad:
Issue previously addressed.

PR #2130, opened 2018-05-09
Try to clarify how ignore-plugin works

DullReferenceException:
The way `ignore-plugin` works wasn't very intuitive to me, so I'd like to try to clarify exactly what the regexes mean in case anyone else is similarly confused.


src/content/plugins/ignore-plugin.md
```diff
@@ -18,8 +19,28 @@ The following examples demonstrate a few ways this plugin can be used.
 
 ## Ignore Moment Locales
 
-As of [moment](https://momentjs.com/) 2.18, all locales are bundled together with the core library (see [this GitHub issue](https://github.com/moment/moment/issues/2373)). You can use the `IgnorePlugin` to stop any locale being bundled with moment:
+As of [moment](https://momentjs.com/) 2.18, all locales are bundled together with the core library (see [this GitHub issue](https://github.com/moment/moment/issues/2373)). 
+
+Contrary to what you may guess, The `requestRegExp` parameter is not tested against the resolved file names or absolute module
```
EugeneHlushko: these line breaks within a paragraph should be removed imo

EugeneHlushko: Dont jump into assumption on what the reader might guess or not, this needs rephrasing

src/content/plugins/ignore-plugin.md
```diff
@@ -18,8 +19,28 @@ The following examples demonstrate a few ways this plugin can be used.
...
+names being imported or required, but rather against the _string_ passed to `require` or `import` _within the source code
+where the import is taking place_. For example, if you're trying to exclude `node_modules/moment/locale/*.js`, this won't
+work:
+
+```js
+new webpack.IgnorePlugin(/moment\/locale\//)
+```
+
+Rather, you must know how `moment` itself is importing those files. Inside of the `moment` module, it has this code:
```
EugeneHlushko: Because of the way how `moment` itself...

src/content/plugins/ignore-plugin.md
DullReferenceException: @EugeneHlushko: removed line breaks and did some rewording.

src/content/plugins/ignore-plugin.md
```diff
@@ -19,8 +20,24 @@ The following examples demonstrate a few ways this plugin can be used.
 
 ## Ignore Moment Locales
 
-As of [moment](https://momentjs.com/) 2.18, all locales are bundled together with the core library (see [this GitHub issue](https://github.com/moment/moment/issues/2373)). You can use the `IgnorePlugin` to stop any locale being bundled with moment:
+As of [moment](https://momentjs.com/) 2.18, all locales are bundled together with the core library (see [this GitHub issue](https://github.com/moment/moment/issues/2373)). 
+
+The `requestRegExp` parameter passed to `IgnorePlugin` is not tested against the resolved file names or absolute module names being imported or required, but rather against the _string_ passed to `require` or `import` _within the source code where the import is taking place_. For example, if you're trying to exclude `node_modules/moment/locale/*.js`, this won't work:
+
+```js
```
EugeneHlushko: Maybe make this a ```diff and put a minus before non working solution to make it red, we had similar complaints in the past from doc users

montogeek:
Thanks!

PR #2132, opened 2018-05-09
docs(module) document webpackInclude and webpackExclude otpions

### LINKED ISSUE #2129: Document webpackInclude and webpackExclude
https://github.com/webpack/webpack/pull/5809

EugeneHlushko:
Fixes #2129

### DIFF
--- a/src/content/api/module-methods.md
+++ b/src/content/api/module-methods.md
@@ -82,8 +82,8 @@ import(
 
 // multiple possible targets
 import(
-  /* webpackInclude: "\\.json$" */
-  /* webpackExclude: "\\.noimport\\.json$" */
+  /* webpackInclude: /\.json$/ */
+  /* webpackExclude: /\.noimport\.json$/ */
   /* webpackChunkName: "my-chunk-name" */
   /* webpackMode: "lazy" */
   `./locale/${language}`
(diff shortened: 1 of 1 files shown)


src/content/api/module-methods.md
```diff
@@ -73,11 +73,21 @@ W> This feature relies on [`Promise`](https://developer.mozilla.org/en-US/docs/W
...
+// single target
 import(
   /* webpackChunkName: "my-chunk-name" */
   /* webpackMode: "lazy" */
   'module'
 );
+
+// multiple possible targets
+import(
+  /* webpackInclude: "\\.json$" */
```
ljcrapo: the `webpackInclude` and `webpackExclude` values use javascript regex syntax.

EG:
``` js
/* webpackInclude: /\.json$/ */
```

src/content/api/module-methods.md
```diff
@@ -91,9 +101,15 @@ import(
 
 T> Note that both options can be combined like so `/* webpackMode: "lazy-once", webpackChunkName: "all-i18n-data" */`. This is parsed as a JSON5 object without curly brackets.
 
+`webpackInclude`: A regular expression that will be matched against during import resolution and point onto modules that __should be bundled__.
```
ljcrapo: this documentation could be a little more clear. 

**only** modules that match `webpackInclude` will be bundled.
any module that matches `webpackExclude` **will not** be bundled.

it may also be helpful (but not necessary) to use a diagram such as this:
image

src/content/api/module-methods.md
ljcrapo: the original PR used string literals but later changed to support this syntax.

EugeneHlushko: great catch, changing...

src/content/api/module-methods.md
EugeneHlushko: To be honest **i like the diagram**, but doesnt feel right to add an image there as maintaining images as features change is a pain in the docs :)

EugeneHlushko:
Thank you for feedback @ljcrapo 
To anyone concerned, updates applied

PR #2133, opened 2018-05-09
chore(rebuild): don't load gitter twice

### LINKED ISSUE #2131: Gitter widget
image
We are mounting our gitter widget on every page load, it should be mounted once.

EugeneHlushko:
Fixes #2131 

Open for discussion:

We could move gitter to `Site.jsx` to have cleaner implementation and stop worrying about hiding and showing the chat toggle button but that would result in gitter widget on all pages of the website vs what we currently have only on documentation pages (`Page.jsx`)

### DIFF
--- a/src/components/Gitter/Gitter.jsx
+++ b/src/components/Gitter/Gitter.jsx
@@ -7,32 +7,34 @@ import isClient from '../../utilities/is-client';
 // Load Styling
 import '../Gitter/Gitter.scss';
 
+let sidecar = false;
+
 // Create and export component
 export default class Gitter extends React.Component {
-  _sidecar = null
-
-  render() {
-    return (
-      <div className="gitter">
-        <div
-          className="gitter__button"
-          onClick={() => window.gitterSidecar && window.gitterSidecar.toggleChat(true)}>
-          <i className="gitter__icon icon-gitter" />
-        </div>
-      </div>
-    );
-  }
-
   componentDidMount() {
     if (isClient) {
       import('gitter-sidecar').then(Sidecar => {
-        if (!window.gitterSidecar) {
-          window.gitterSidecar = new Sidecar({
+        if (!sidecar) {
+          sidecar = new Sidecar({
             room: 'webpack/webpack',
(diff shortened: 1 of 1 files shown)


skipjack:
@EugeneHlushko two things...

1. Just moved `Gitter` to `Site` in #2042 to implement sticky behavior (and i don't think it's bad having it on all pages.
2. I _think_ `import()` is smart enough not to reload the bundle. Pretty sure it detects whether a child bundle has already been loaded before fetching it from server.

In regards to the first, sorry I hadn't read through this when I made that change, I would've left a comment first. With those things in mind though, I think we can close this out (re-open if you feel differently though).

EugeneHlushko:
Just pushed another commit to fix rebase issues

skipjack:
Yeah, I misunderstood the loading issue, I thought you meant loading the gitter bundle multiple times not duplicating widget.

src/components/Gitter/Gitter.jsx
```diff
@@ -14,7 +14,9 @@ export default class Gitter extends React.Component {
   render() {
     return (
       <div className="gitter">
-        <div className="gitter__button js-gitter-toggle-chat-button">
+        <div
+          className="gitter__button"
+          onClick={() => window.gitterSidecar && window.gitterSidecar.toggleChat(true)}>
```
skipjack: You think the imperative call is more intuitive than the class binding?

skipjack: I think I'm ok with this change but the module-level variable makes more sense to me and would shorten this call:

``` js
onClick={() => sidecar && sidecar.toggleChat(true))
```

EugeneHlushko: Yeah i think we can give it a `let` in the top of `Gitter.jsx`

EugeneHlushko:
@skipjack i did move methods above render and used a module level variable (component level variable gets lost as a new copy of component is being instantiated by page)

Also as you suggested did a component level click handler for better look

PR #2134, opened 2018-05-09
changed service worker default name

richtmat:
I changed the filename in the docs from `sw.js` to the workbox default name `service-worker.js`.


src/content/guides/progressive-web-application.md
```diff
@@ -135,7 +135,7 @@ __index.js__
 
 + if ('serviceWorker' in navigator) {
 +   window.addEventListener('load', () => {
-+     navigator.serviceWorker.register('/sw.js').then(registration => {
++     navigator.serviceWorker.register('/service-worker.js').then(registration => {
```
EugeneHlushko: did you run the guide from scratch?

As i see there is an output on the page which says that `sw.js` with a size of 1kb was emited.

montogeek:
Thanks!

PR #2136, opened 2018-05-10
replace 'bundle.js' with 'main.js' in examples

royair:
'npx webpack' default output file 'main.js',
while in examples 'bundle.js' is being used.

corrected examples accordingly.


EugeneHlushko:
wasnt this already collected and merged?

EugeneHlushko:
i think i found the root cause of this one @montogeek 

As soon as you have `webpack.config.js` file in the directory where you run `npx webpack`, even **without explicitly specifying** `--config webpack.config.js`, the config is being used by webpack.

This is how we got different results while testing this behaviour by running the exact same command. Which, if a user follows the guide from scratch, makes this pr completely valid.

royair:
@EugeneHlushko i got 'main.js' as output file **even without having a webpack.config.js file**. 
(this can be easily tested and verified)
btw, i'm using a windows OS, if that makes any different.

EugeneHlushko:
Exactly, thats what i'm also trying to say. Without config you get `main.js`. But then as soon as you get a config named `webpack.config.js` regardless of pointing webpack onto it using `--config %path%` it gets picked up, thats how we previously made this mistake by substituting `main.js` into `bundle.js` for configless run.

montogeek:
Thanks

PR #2142, opened 2018-05-10
Update html-webpack-plugin.md

jeremyodell:
Changed the basic usage webpack.config.js file to use  'module.exports = {' as opposed to the var webpackConfig.

Using the latter created an error when trying to run a minimal example.  It would complain that the ./src was not found.


montogeek:
Thanks!

PR #2144, opened 2018-05-11
fix: indentation in code sample of split-chunks-plugin

Priestch:


Priestch:
It would be better if the markdown linter can detect such error, @EugeneHlushko you know why my CLA not signed, I had signed already

byzyk:
Agree with @EugeneHlushko, I guess linter scripts have to be revised since it's not the first time formatting is broken in the examples.

I've created an issue for this one #2151

jeremenichelli:
Good point, yeah we should try to actually update the linter and run it on all the content.

byzyk:
@jeremenichelli does it mean we can start on #2151? Please check my comment there with a suggested approach.

montogeek:
Thanks!

PR #2147, opened 2018-05-12
docs(cli): update cli docs with execution warning

### LINKED ISSUE #2141: webpack.js.org/src/content/concepts/mode.md does not define precendence
Hey maybe it's defined elsewhere but if so I think it bears repeating in this section of the documentation.

It says that we can define the `mode` entry to dictate the mode to webpack, and that we can also define it in the command line using `--mode=` flag.

However it doesn't say which will be in effect if both the command line and webpack.config.js entry are defined.

Also does anyone know which wins? I assume command line will override config as it is more immediate, but I don't want to assume.

ev1stensberg:
Fixes #2141 . PTAL as I couldn't compile the site on my machine ( prism console err).


montogeek:
@ev1stensberg Prism error is just a warning, Did you see a different error?

src/content/api/cli.md
```diff
@@ -95,6 +95,8 @@ This will form the bundle with both the files as separate entry points.
 
 ### Common Options
 
+W> Note that Command Line Interface has a higher precendence for the arguments you use it with than than your configuration file. Examplewise if you pass [`--mode="production"`](https://webpack.js.org/concepts/mode/#usage) to the CLI and your configuration file uses `development`, `production` will be used.
```
montogeek: - Repeated word `than`
- Space between `Example` and `wise`
- Use relative links (`/concepts/mode/#usage`)

PR #2148, opened 2018-05-13
feat(rebuild): improve font loading strategy

jeremenichelli:
This is my first try on trying to improve loading times on the rebuild branch. What I'm doing is applying what is consider best practice on font loading.

The new observer module doesn't affect the initial bundle since it's dynamically loaded when the site component mounts. I also removed the code web font since is pretty much the same as the system one in Mac and saves as tons of KiBs.

This should eliminate the flash of invisible content that we have on the site right now.

An example here on how it looks when the fonts kicks off:

kapture 2018-05-13 at 14 35 21

I've introduced some small css changes to prevent the stacking effect to be huge, but the loading times on the site should improve drastically. Also, once the SPA bundle kicks off, there's no stacking anymore since we are not reloading the site.

TODO:
 - Fix recalculation on TextRotater component for homepage.
 - Fix banner styles on mobile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "@rigor789/remark-autolink-headings": "^5.1.0",
     "ajv": "^5.5.2",
     "docsearch.js": "^2.5.2",
+    "fontfaceobserver": "^2.0.13",
     "gitter-sidecar": "^1.2.3",
     "prop-types": "^15.5.10",
     "react": "^16.2.0",
--- a/src/components/Markdown/Markdown.scss
+++ b/src/components/Markdown/Markdown.scss
@@ -3,7 +3,7 @@
 @import 'prism-theme';
 
 .markdown {
-  line-height:1.5em;
+  line-height: 1.5em;
 
   h1 { font-size: getFontSize(4); }
   h2 { font-size: getFontSize(3); }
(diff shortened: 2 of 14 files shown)


Munter:
You'd get a simpler setup by serving the Google fonts from the docs server directly and using font-display: swap;

Further massive improvements can be gained by preloading the woff2 versions of those fonts.

If you want to improve the limbo state of unloaded web fonts, using Monica Dinculescus font style matcher will also get you a much smaller jump in text styles from limbo to loaded fonts

jeremenichelli:
@Munter yeah I thought about all of that. I've used Monica's font swapper to better match this, and as tested in my machine both are pretty close.

About `font-display`, though it's still missing Edge and support is not that bad, it partially solves the problem I'm mentioning. Serving directly from Google Fonts still presents a render blocking resource, lazy loading the styles with the bundle prevents that and `fontfaceobserver` gives us all browser wide support for font switching.

Munter:
I might have been unclear. I'm saying don't serve fonts it their CSS directly from Google fonts. Instead download the fonts to the docs repository and have the font face declaration be a part of the base CSS. Then preload the fonts, which now have a known URL. This will cut away the font loading limbo almost entirely

jeremenichelli:
Oh yeah, sorry maybe I got it wrong @Munter. Yeah I agree though for now it sounds overengineering to me. My intention is for the users to have the content as fast as possible without having to wait for the web font to be loaded. We can later apply preconnect and preload strategies for sure.

jeremenichelli:
I will close this since it was impossible to make the home apge animation work with font changing involved, but will work on it later with more time or maybe if we decide to change the design of the site.

PR #2150, opened 2018-05-14
Replace whitespaces with tabs in split-chunks-plugin.md

ybiquitous:
This fixes incorrect indentation in code example in <https://webpack.js.org/plugins/split-chunks-plugin/>.

image

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -181,7 +181,7 @@ splitChunks: {
 			test: /[\\/]node_modules[\\/]/,
 			priority: -10
 		},
-    default: {
+		default: {
 			minChunks: 2,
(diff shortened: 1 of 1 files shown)


ybiquitous:
I recommend using whitespaces instead of tabs for code example, because tab's indentation is invisible and fragile.

And if possible, I think better to use tool like [Prettier](https://github.com/prettier/prettier) and format markdown file automatically.

In this PR, avoiding large diffs, I fixes only corresponding place.

byzyk:
I guess linter scripts have to be revised since it's not the first time formatting is broken in the examples.

byzyk:
Duplicate of #2144

ybiquitous:
> Duplicate of #2144

Oh, sorry. I will close this. 🙏

PR #2152, opened 2018-05-15
Fix documented default for resolve.extensions

taion:
Per https://github.com/webpack/webpack/blob/v4.8.3/lib/WebpackOptionsDefaulter.js#L305, the currently documented default extensions are no longer correct.


montogeek:
Thanks!

PR #2153, opened 2018-05-15
Update code-splitting.md

### LINKED ISSUE #2154: Need update docs: Dynamic import doesn't work as shown in the docs
# Bug report

As shown in the [documentation](https://webpack.js.org/guides/code-splitting/#dynamic-imports) about dynamic imports, we should get lodash from `import(/* webpackChunkName: "lodash" */ 'lodash').then(_ => doSomething());` but instead the lodash value is wrapped inside a namespace with default property.

The docs should show that we need to access _ via `namespace.default`

webpack version:  4.8.3
Node.js version: 8.9.4
Operating System: Mac OS X
Additional tools:

---

This issue was moved from webpack/webpack#7302 by @sokra. Original issue was by @TienDo1011.

TienDo1011:
Since webpack 4, when using `dynamic import` to resolve commonJS module, it no longer resolve directly to the value of `module.exports`, but create an artificial namespace object to wrap it instead.

Fixes #2154


EugeneHlushko:
@montogeek @jeremenichelli this is exactly why we had issue with page's module content, if you remember we did `module.default` check

montogeek:
This was done in another PR

montogeek:
This PR have same changes but also some others, we can merge both.

EugeneHlushko:
Yeap but this one is complete thanks to the explanation paragraph and also uses modern assignment approach e.g. `({ default: _ }) => ...` i would suggest merging both (the older first) or only this one as it is in a complete state for both examples on line 246 and 295

montogeek:
Yeah, still not sure if we want to push ES6 on our code examples, need to check other examples style.

I would say that it is safe these days, but still ¯\_(ツ)_/¯

EugeneHlushko:
mm in that case we might ask @TienDo1011 to update this with es5 ?

TienDo1011:
I think using ES6 is the way to go, as we already use it in that example (with arrow function). Beside, our examples look a lot more clean.

src/content/guides/code-splitting.md
```diff
@@ -242,7 +243,7 @@ __src/index.js__
 -
 -   // Lodash, now imported by this script
 -   element.innerHTML = _.join(['Hello', 'webpack'], ' ');
-+   return import(/* webpackChunkName: "lodash" */ 'lodash').then(_ => {
++   return import(/* webpackChunkName: "lodash" */ 'lodash').then(({ default: _ }) => {
```
asapach: There's a similar line of code on line 286, which should also be updated.

src/content/guides/code-splitting.md
```diff
@@ -281,7 +284,7 @@ __src/index.js__
 -
 -   }).catch(error => 'An error occurred while loading the component');
 +   var element = document.createElement('div');
-+   const _ = await import(/* webpackChunkName: "lodash" */ 'lodash');
++   const { default: _ } = await import(/* webpackChunkName: "lodash" */ 'lodash');
```
asapach: ...and a couple of lines above - see `.then(_ => { ...}`

montogeek:
@TienDo1011 Friendly ping

montogeek:
Thanks!

PR #2158, opened 2018-05-16
fix(fetch) done add two underscores for supporters.json

EugeneHlushko:
Currently fetch command creates `__supporters.json` instead of `_supporters.json` because an underscore was added in both filename and file path string literal.

This patch fixes it, if you do a git clean and fetch currently dev server errors on trying to read non existent `_supporters.json` because it was fetched with two underscores.

### DIFF
--- a/src/utilities/fetchSupporters.js
+++ b/src/utilities/fetchSupporters.js
@@ -26,7 +26,7 @@ request(url)
     }
 
     // Write the file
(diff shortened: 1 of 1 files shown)


jeremenichelli:
OMG, that's my bad, thanks for fixing it @EugeneHlushko

montogeek:
Thanks

PR #2162, opened 2018-05-16
Update externals.md to include an example of more than one library

chanoch:
Give an example of where multiple libraries can be defined. I've used react because it typically requires multiple libraries to be included and there are no examples of multiple libraries being defined as external. Externals is a plural after all so implies multiple libraries can be excluded from the build.

I've capitalised react as the global object in the UMD build from unpkg is usually capitalised.

### DIFF
--- a/src/content/configuration/externals.md
+++ b/src/content/configuration/externals.md
@@ -76,7 +76,8 @@ externals: {
 
 ``` js
 externals : {
-  react: 'react'
+  react: 'React',
+  react-dom: 'ReactDOM'
 }
 
 // or


src/content/configuration/externals.md
```diff
@@ -76,7 +76,8 @@ externals: {
 
 ``` js
 externals : {
-  react: 'react'
+  react: 'React',
+  react-dom: 'ReactDOM'
```
asapach: the property syntax is invalid - should be `'react-dom'`

dhruvdutt: @chanoch Any update on this? :)

montogeek:
Friendly ping @chanoch

montogeek:
Friendly ping @chanoch

EugeneHlushko:
Closing due to inactivity, please re-open after actions applied @chanoch

PR #2163, opened 2018-05-17
Fix `done` compiler hook type

jamiekowalski:
Code reference: https://github.com/webpack/webpack/blob/v4.8.3/lib/Compiler.js#L34


montogeek:
Thanks!

PR #2164, opened 2018-05-17
Fix link to html-webpack-plugin options section

alexeyraspopov:
Setting correct anchor so the link shows exactly what it should.


montogeek:
Thanks!

PR #2167, opened 2018-05-17
Do not need to modify entry

helmutgranda:
Do not need to remove and re-add app: './src/index.js'. We just need to remove the print declaration.

Updated the way the entry declaration is being displayed.


asapach:
It's a duplicate of #1747

helmutgranda:
You are 100% correct, I completely forgot about the comma. 

Thanks for the super quick verification.

PR #2174, opened 2018-05-20
CommonsChunkPlugin page - detailed infos removed

Legends:
Removed webpack 3 infos about CommonsChunkPlugin and left only:
The CommonsChunkPlugin has been removed in webpack v4 legato. To learn how chunks are treated in the latest version, check out the [SplitChunksPlugin](/plugins/split-chunks-plugin/).

I would leave this piece of information for wp3 users and not delete the page entirely.

TBD: Add a reference to webpack.js.org v3 for CommonsChunkPlugin!


Legends:
It's not clear to me, do we have separate docs for v3 and v4?
I don't like this mix, it will get messier with any further version...

montogeek:
We can't remove it until v4 is fully documented.

Legends:
Alright, closing this then....

PR #2183, opened 2018-05-23
SplitChunksPlugin: cleanup and organization

chrisdothtml:
## Summary

Updating the docs based on discussed in [webpack#7217](https://github.com/webpack/webpack/issues/7217)

- adding info about providing function for `chunks`
- also fixing wrong indents in a code snippet

## Potential improvements

Ideally, where it says

> (your function is given the chunk object as a parameter)

it would link to the documentation for the schema of that chunk object. If someone could link me to that (if it exists), I can add it here as well

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -11,7 +11,7 @@ related:
 
 Originally, chunks (and modules imported inside them) were connected by a parent-child relationship in the internal webpack graph. The `CommonsChunkPlugin` was used to avoid duplicated dependencies across them, but further optimizations where not possible
 
-Since version 4 the `CommonsChunkPlugin` was removed in favor of `optimization.splitChunks` and `optimization.runtimeChunk` options. Here is how the new flow works.
+Since webpack v4, the `CommonsChunkPlugin` was removed in favor of `optimization.splitChunks` and `optimization.runtimeChunk` options. Here is how the new flow works.
 
 
 ## Defaults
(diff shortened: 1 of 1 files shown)


src/content/plugins/split-chunks-plugin.md
```diff
@@ -181,7 +182,7 @@ splitChunks: {
 			test: /[\\/]node_modules[\\/]/,
 			priority: -10
 		},
-    default: {
```
EugeneHlushko: i think this was fixed somewhere in an open PR before.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -147,7 +148,7 @@ It can match the absolute module resource path or chunk names. When a chunk name
 
 With the `chunks` option the selected chunks can be configured.
 
-There are 3 values possible `"initial"`, `"async"` and `"all"`. When configured the optimization only selects initial chunks, on-demand chunks or all chunks.
+There are 3 values possible `"initial"`, `"async"` and `"all"`. When configured the optimization only selects initial chunks, on-demand chunks or all chunks. Alternatively, you can provide a function that returns whether to include each chunk (your function is given the chunk object as a parameter).
```
EugeneHlushko: It needs modification and rewording. Mention that it accepts a function and string with enum like we do with other params.

The function explanation needs to make more sense, needs example too.

src/content/plugins/split-chunks-plugin.md
chrisdothtml: I'm not sure what you want me to do about that

src/content/plugins/split-chunks-plugin.md
chrisdothtml: The docs currently don't mention that a function is a possible type for this parameter. If they had, it would have saved me quite a lot of time debugging an issue I was having. So any mention is better than what is currently out there.

> Mention that it accepts a function and string with enum like we do with other params

I agree that what I wrote isn't the best this could be, but without references to what you're referring to, I don't exactly know what you're looking for

EugeneHlushko: Sorry if that got you confused, here is a reference

For example, this https://webpack.js.org/configuration/optimization/#optimization-runtimechunk

can be set to string object and boolean, all of them have been mentioned and exampled, this needs to be at least of the same format and quality or better.

Function option requires an example, this is a must.

EugeneHlushko: otherwise we end up with another issue saying there is no example for the function option :/

chrisdothtml:
@EugeneHlushko this turned into a much larger PR than I planned. It was difficult to approach documenting one property the right way when the rest of them were a bit scattered and inconsistent, so I restructured the page and added a more complete list of the config properties available.

From my commit description:

> - organized config definitions in a more logical/readable format
> - simplified various descriptions
> - grouped examples together to avoid mixing the purposes of the different parts of the page
> - removed runtimeChunk section already mentioned on the optimization page (seemed irrelevant)

Sorry if this was too much at once; cleaning up one thing led to another, and another...

EugeneHlushko:
Thanks for quick update, i wont be able to look today but maybe someone else can review too (mentioning page authors): @jeremenichelli @sokra

chrisdothtml:
@jeremenichelli no problem!

jeremenichelli:
Already merged it since it wasn't a blocker to me. Good job!

PR #2198, opened 2018-05-26
Add Twig loader

webberig:
Add Twig loader to the list of templating loaders


webberig:
Added my secondary e-mail address that I used for the commit... should the clabot update its check automatically, or do I need to do something else ?

montogeek:
@webberig You would need to change commit author. Or create a new PR with the correct commit author.

webberig:
ok, think I succeeded :-)

montogeek:
Thanks!

PR #2201, opened 2018-05-27
Rebuild build memory allocation fix and build time improvements

jeremenichelli:
Hey folks, good news: I fixed memory allocation issues and improved build times 🎉 

Here's a quick catch up:

👉  **memory allocation improvements on static generation**
 - On production, build `_content.json` file ahead-of-time.
 - Flatten content tree to array ahead of time to avoid `crawl` option on SSG plugin.
 - Moved content to SSG `locals` to avoid multiple imports.

👉  **production build time improvements**
 - Moved source maps only for development.
 - Moved `react-hot-loader/babel` to only development.
 - Removed Clean plugin in favor of `rimraf`.
 - Excluded _server bundles_ from Uglify plugin.
 - Enabled parallel uglifying.

👉  **new project feature**
 - Added `serve` command for static site testing.

💪  In my machine production build times passed from ~58s to ~37s (almost 20% less).

💰 Shaved ~40kb from `index.bundle.js` .


jeremenichelli:
@montogeek @EugeneHlushko can you take a look at this? I've already finished the most import changes in the build flow and everything's working correlty both at development as for production.

src/components/Site/Site.jsx
```diff
@@ -5,7 +5,7 @@ import { hot as Hot } from 'react-hot-loader';
 import DocumentTitle from 'react-document-title';
 
 // Import Utilities
-import { ExtractPages, ExtractSections, GetPageTitle } from '../../utilities/content-utils';
+import { extractPages, extractSections, getPageTitle } from '../../utilities/content-utils';
```
montogeek: Why all these name changes?

jeremenichelli: I think that by convention methods that are capitalized in JavaScript are constructors, which is not the case for all these helper functions.

montogeek:
@jeremenichelli Why "On production, build _content.json file ahead-of-time."?

jeremenichelli:
@montogeek during development we watch the content and when an `.md` is deleted or added the `_content.json` file is updated through a webpack plugin. You will notice that I moved that from common to only dev config.

On production I build that file before webpack build kicks off, that means that we can access it before hand, build a static paths array and prevent excesive memory allocations when the content tree is big plus avoiding the crawl option in `StaticSitePlugin`, so it's not part of the build script itself.

src/components/Site/Site.jsx
skipjack: Yeah, I think that's definitely at least the convention in node. I got in the habit of capitalizing all `import`s a while back to delineate them from local variables but it definitely doesn't seem to be a standard practice.

PR #2200, opened 2018-05-27
Update Writing plugin and Plugin pattern page

nveenjain:
Fixes #2172, Updated writing plugins and plugin patterns page to include ES6 classes (as ES6 is supported by webpack completely) and updates related to new tapable library.

PR #2202, opened 2018-05-27
Update the “Writing a plugin” docs for the new plugin API

iamakulov:
* Update the docs to use the new Tapable API
* Add the “Under the hood” section
* Remove the “Hook types” section and link to the Tapable docs instead (depends on https://github.com/webpack/tapable/pull/63)
* Put sections talking about compiler and compilation next to each other
* Do some minor tuning

Depends on https://github.com/webpack/tapable/pull/63.

### DIFF
--- a/src/content/contribute/writing-a-plugin.md
+++ b/src/content/contribute/writing-a-plugin.md
@@ -14,7 +14,7 @@ A plugin for `webpack` consists of a named JavaScript class that:
 
 - Defines the `apply` method.
 - Specifies an [event hook](/api/compiler-hooks/) on which to bind itself.
-- Tunes the build using the plugin API provided by webpack.
+- Manipulates the build using the plugin API provided by webpack.
 
 ```javascript
 class MyExampleWebpackPlugin {
@@ -27,7 +27,7 @@ class MyExampleWebpackPlugin {
         console.log('This is an example plugin!');
         console.log('Here’s the `compilation` object which represents a single build of assets:', compilation);
 
-        // Tune the build using the plugin API provided by webpack
+        // Manipulate the build using the plugin API provided by webpack
         compilation.addModule(/* ... */);
 
         callback();
@@ -58,7 +58,7 @@ class HelloWorldPlugin {
 module.exports = HelloWorldPlugin;
 ```
 
-Then to use the plugin, just include an instance in your webpack config `plugins` array:
+Then to use the plugin, include an instance in your webpack config `plugins` array:
 
 ```javascript
 // webpack.config.js
--- a/src/content/contribute/writing-a-plugin.md
+++ b/src/content/contribute/writing-a-plugin.md
@@ -3,31 +3,30 @@ title: Writing a Plugin
 sort: 4
 contributors:
   - tbroadley
+  - iamakulov
 ---
 
 Plugins expose the full potential of the webpack engine to third-party developers. Using staged build callbacks, developers can introduce their own behaviors into the webpack build process. Building plugins is a bit more advanced than building loaders, because you'll need to understand some of the webpack low-level internals to hook into them. Be prepared to read some source code!
 
 ## Creating a Plugin
 
-A plugin for `webpack` consists of
+A plugin for `webpack` consists of a named JavaScript function that:
 
-- A named JavaScript function.
 - Defines `apply` method in its prototype.
 - Specifies an [event hook](/api/compiler-hooks/) on which to bind itself.
 - Manipulates webpack internal instance specific data.
-- Invokes webpack provided callback after functionality is complete.
 
 ```javascript
 // A named JavaScript function.
 function MyExampleWebpackPlugin() {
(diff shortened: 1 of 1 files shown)


dhruvdutt:
Related to #2200.

src/content/contribute/writing-a-plugin.md
```diff
@@ -3,125 +3,141 @@ title: Writing a Plugin
...
-  compiler.plugin('webpacksEventHook', function(compilation /* Manipulates webpack internal instance specific data. */, callback) {
-    console.log("This is an example plugin!!!");
-
-    // Invokes webpack provided callback after functionality is complete.
-    callback();
-  });
-};
+// A named JavaScript class...
+class MyExampleWebpackPlugin {
+  // ...that defines `apply` method in its prototype...
```
jeremenichelli: Can we change some of this comments? I suggest removing the first one since its redundant. Be more imperative, like _Define apply method in the class._ (without trailing periods and more connected with the requirements)

src/content/contribute/writing-a-plugin.md
```diff
@@ -3,125 +3,141 @@ title: Writing a Plugin
...
+  apply(compiler) {
+    // ...specifies webpack’s event hook to attach itself..
+    compiler.hooks.compile.tapAsync(
+      'afterCompile',
+      (compilation, callback) => {
+        // ...and manipulates webpack internal instance specific data.
+        console.log('This is an example plugin!!!');
+        console.log('Here’s the compilation object:', compilation);
```
jeremenichelli: Same as the comment abot the comments here (so meta), I would be more formal here and put something like _plugin runtime_ or something, and maybe describe shrotly the compilation object.

src/content/contribute/writing-a-plugin.md
```diff
@@ -3,125 +3,141 @@ title: Writing a Plugin
...
 
-    // Do something async...
-    setTimeout(function() {
-      console.log("Done with async work...");
-      callback();
-    }, 1000);
+    // tapPromise() is promise-based
+    compiler.hooks.emit.tapPromise('HelloAsyncPlugin', (compilation) => {
+      // Do something async...
+      return doSomethingAsync()
```
jeremenichelli: The comment from above feels redundant, we can put something like _Task that will run asynchronously_, or I will remove it, I think the context gives enough explanation.

src/content/contribute/writing-a-plugin.md
```diff
@@ -3,125 +3,140 @@ title: Writing a Plugin
...
-A plugin for `webpack` consists of
+A plugin for `webpack` consists of a named JavaScript class that:
 
-- A named JavaScript function.
-- Defines `apply` method in its prototype.
+- Defines the `apply` method.
 - Specifies an [event hook](/api/compiler-hooks/) on which to bind itself.
-- Manipulates webpack internal instance specific data.
-- Invokes webpack provided callback after functionality is complete.
+- Tunes the build using the plugin API provided by webpack.
```
montogeek: Manipulates is better than "Tunes"

src/content/contribute/writing-a-plugin.md
```diff
@@ -3,125 +3,140 @@ title: Writing a Plugin
...
-  compiler.plugin('done', function() {
-    console.log('Hello World!');
-  });
-};
-
 module.exports = HelloWorldPlugin;
 ```
 
-Then to install the plugin, just include an instance in your webpack config `plugins` array:
+Then to use the plugin, just include an instance in your webpack config `plugins` array:
```
montogeek: Please avoid words like "just"

src/content/contribute/writing-a-plugin.md
```diff
@@ -133,76 +148,63 @@ Once we can latch onto the webpack compiler and each individual compilations, th
...
+    // (This is an async one, so webpack passes a callback into it)
+    this.hooks.run.callAsync(() => {
+      // The callback is called when all tapped functions finish executing
+    });
+  }
+}
+```
 
-`applyPluginsParallelBailResult(name: string, args: any..., callback: (err: Error, result: any) -> void)`
+There’re multiple types of hooks which run tapped functions a bit differently. They are described [in the Tapable docs](https://github.com/webpack/tapable#hook-types).
```
montogeek: That anchor does not longer exists. https://github.com/webpack/tapable should be fine

src/content/contribute/writing-a-plugin.md
iamakulov: You’re right, good point!

src/content/contribute/writing-a-plugin.md
iamakulov: This depends on https://github.com/webpack/tapable/pull/63 – the anchor will go live as soon as that PR is merged.

If it’s unlikely the PR would be merged soon, just tell – I’ll remove the anchor :–)

PR #2206, opened 2018-05-31
Update mode.md

m4ss1m0g:
Added a simple example to use **_mode_** parameter inside the config


src/content/concepts/mode.md
```diff
@@ -80,3 +80,26 @@ module.exports = {
 -  ]
 }
 ```
+
+if you want to change the behavior according the **mode** variable inside the *webpack.config.js* you have to export a function instead of an object:
```
montogeek: Please capitalize `if`

src/content/concepts/mode.md
```diff
@@ -80,3 +80,26 @@ module.exports = {
...
+If you want to change the behavior according the **mode** variable inside the *webpack.config.js* you have to export a function instead of an object:
+
+```javascript
+var config = {
+  entry: "./app.js"
+  ..
+}
+
+module.exports = (env, argv) => {
+  console.log(argv.mode)        // outputs the mode
```
dhruvdutt: I don't think this line is needed since we already show the `(argv.mode === "development")`  check below which indicates what `argv.mode` is. Anyway, if we're keeping this line, please add the missing semi-colon.

src/content/concepts/mode.md
```diff
@@ -80,3 +80,26 @@ module.exports = {
 -  ]
 }
 ```
+
+If you want to change the behavior according the **mode** variable inside the *webpack.config.js* you have to export a function instead of an object:
```
dhruvdutt: Missing a comma here?

src/content/concepts/mode.md
jeremenichelli: I agree, this line can be removed for sure.

src/content/concepts/mode.md
m4ss1m0g: Since is my first pull request, I must correct the code and commit again ? 
If yes, what change I have to make ?
Thanks and sorry..

montogeek: @m4ss1m0g Yes, changes should be committed.
Not sure neither, care to explain @dhruvdutt ?

dhruvdutt: Not sure if there should be comma between "*webpack.config.js*" and "you".
Also "according" => "according to"?

m4ss1m0g: I'm italian and probably my english is not very good, fell free to correct the punctuation.

src/content/concepts/mode.md
montogeek: @m4ss1m0g Please remove it

PR #2213, opened 2018-06-04
Update API documentation about hooks

ooflorent:
- `tapName` was missing in several code examples
- add an example using `async`/`await`


montogeek:
Thanks!

PR #2224, opened 2018-06-06
Enable ESLint in code examples

### LINKED ISSUE #2151: Fix linter: throw on broken formatting in code examples
Code examples formatting seem to be broken in some places (eg. #2150 #2111 etc). 

I suggest linter could attempt to fix code formatting automatically or at least throw during CI.

byzyk:
Fixes #2151 
Depends on #2226 

### Notes

* Added separate ESLint config rules so it won't affect actual website source files
* Updated ESLint to the latest version to enable `--fix` functionality in ESLint plugins (for details check eslint/eslint#9090)
* Enabled autofix feature for `eslint-plugin-markdown`. Need to fix `package.json` dependency when eslint/eslint-plugin-markdown#97 gets merged and published.


montogeek:
Why the quotes change?

byzyk:
@montogeek just to make it consistent, because it used to be a mixture of `'` and `"` across all code examples. My personal preference is singlequote but can be easily changed to double if needed, let's see what other folks say :)

byzyk:
@montogeek wow, I just realized that we're not running tests before build in out Travis Job. Perhaps this should be fixed in another PR as not directly related to code examples. Also, I noticed some other issues with Travis, like using `npm` instead of `yarn`. I'll take care of it

package.json
```diff
@@ -56,9 +56,9 @@
     "copy-webpack-plugin": "^4.4.2",
     "css-loader": "^0.28.10",
     "duplexer": "^0.1.1",
-    "eslint": "4.5.0",
+    "eslint": "4.19.1",
     "eslint-loader": "^2.0.0",
-    "eslint-plugin-markdown": "^1.0.0-beta.7",
+    "eslint-plugin-markdown": "git://github.com/byzyk/eslint-plugin-markdown.git#5927ccb78aff8a5464663730800503c6de3704e1",
```
dhruvdutt: Why aren't we using semver here?

byzyk: @dhruvdutt please see the third note in the description

jeremenichelli:
This is so cool, Thanks!

Yes, I think that we need to improve both travis processes and the lint task that gets executed. Would love if you could open a new PR with those improvements.

byzyk:
Rebased on the latest `master` and resolved conflicts. It would be great if we could consider merging this as soon as we ready since almost every PR related to code example change will create conflict for this one.

Guys, do you think we should add/change anything here? ping @montogeek @jeremenichelli

montogeek:
@jeremenichelli Please next time squash and merge.

PR #2238, opened 2018-06-07
docs(tree-shaking): bundle minified problem

lumo10:
in my version of webpack 4.10.2 bundle comes already minified by default. I need to set the development mode

### DIFF
--- a/src/content/guides/tree-shaking.md
+++ b/src/content/guides/tree-shaking.md
@@ -10,6 +10,7 @@ contributors:
   - dmitriid
   - probablyup
   - gish
+  - lumo10
 related:
   - title: "webpack 4 beta — try it today!"
     url: https://medium.com/webpack/webpack-4-beta-try-it-today-6b1d27d7d7e2#9a67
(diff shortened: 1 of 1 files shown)


Legends:
?

montogeek:
This change is out of the scope of Tree shaking guide, it is also already well documented in other pages.
What do you think @EugeneHlushko @jeremenichelli ?

Legends:
> If not set, webpack sets production as the default value for mode
[Section mode](https://webpack.js.org/concepts/mode/#usage)

montogeek:
I guess it is important to be able to read the source code in order to make sure your tree shakeable modules are not there.

montogeek:
Thanks

Legends:
@lumo10 your last comment made it clear ;-)

src/content/guides/tree-shaking.md
```diff
@@ -57,6 +58,24 @@ export function cube(x) {
...
+const path = require('path');
+
+module.exports = {
+  entry: './src/index.js',
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, 'dist')
+- }
++ },
++ mode: "development"
```
jeremenichelli: Sorry for being late to the party but, is it necessary to specify entry and output when they have default values already?

Also, I think that I would add the mode by CLI and put a small warning on the guide maybe.

lumo10: You are right if you take the single page guide alone, but in the contest of the get started guide where you  build the concept step by step you already have the webpack.config.js

PR #2243, opened 2018-06-09
Add concurrently package to run scripts more conveniently

tryzniak:
This PR adds a [concurrently](https://www.npmjs.com/package/concurrently) package. I noticed that when running `yarn start` and then pressing Ctrl + C to quit could leave some processes (antwar, webpack) running in the background. To quit those processes you'd have to find out their pids with a `ps` command and then run `kill #pid`. This PR simplifies this workflow with `concurrently`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "node": ">=6.9"
   },
   "scripts": {
-    "start": "npm run init:generated && npm run fetch && node ./antwar.bootstrap.js develop",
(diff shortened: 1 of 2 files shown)


tryzniak:
Oh I've just learned about `npm-run-all`. Hmm, I could try to rework this PR to use that package. But have you had those issues with `yarn start` where it leaves zombie processes hanging?

byzyk:
To be honest I haven't tested it. Please try to use the existing package and see how it goes.

tryzniak:
Yeah, I'll try to replace `concurrently` with the existing package 👍

PR #2245, opened 2018-06-09
Include a tip for require.context and watching

spicalous:
Came across issues with webpack rebuilding project multiple times. I'd like to improve documentation for watching as it is not mentioned anywhere that using require.context can watch an entire directory

Thank you!


src/content/configuration/watch.md
```diff
@@ -57,12 +58,14 @@ For some systems, watching many file systems can result in a lot of CPU or memor
...
 
-It is also possible to use [anymatch](https://github.com/micromatch/anymatch) patterns:
+It is also possible to have and use multiple [anymatch](https://github.com/micromatch/anymatch) patterns:
 
 ```js
-ignored: "files/**/*.js"
+ignored: ["files/**/*.js", "node_modules"]
 ```
 
+T> If you use `require.context`, webpack will watch your entire directory. You will need to ignore files and/or directories so that unwanted changes will not trigger a rebuild
```
montogeek: Please add a dot at the end.

spicalous: Amended

montogeek:
Thanks!

PR #2226, opened 2018-06-06
Add `test` task to TravisCI, use `yarn`

byzyk:
as per discussion in #2224

PR #2249, opened 2018-06-11
feature: pre-commit hooks

byzyk:
As per discussion in #2226

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn test"
+      "pre-commit": "npm test"
     }
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


byzyk:
This one is ready to go.

montogeek:
What is the difference between this husky release candidate and the last stable one?

byzyk:
In the RC version, you define hooks in the separate `husky` object. In previous stable version hooks were part of `scripts`. Not sure if there are any other improvements like performance etc but this is what they recommend to install now.

[Here](https://github.com/typicode/husky/blob/dev/CHANGELOG.md) is the full changelog.

montogeek:
Thanks!

montogeek:
@byzyk If you have time, please add lint-staged too. We are losing time checking for files that shouldn't be linted, for example, bash scripts.

byzyk:
@montogeek sure man! will pick it up at the earliest, shouldn't take much time

PR #2250, opened 2018-06-11
chore: migrate from npm to yarn

byzyk:
* Switch to `yarn` in scripts
* Update contributing guides

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -8,20 +8,24 @@ page please create an issue first.
 
 ## Setup
 
-Start by ensuring you have Node installed and forking the repository:
+Start by ensuring you have Node and Yarn installed:
 
-- Install [Node.js][1] if you have not already.
+- Install [Node.js][1] if you haven't already.
+- Install [Yarn][13] if you haven't already.
+
+Fork the repository:
+
 - Fork the **webpack.js.org** repo from [the main repository][2].
 - `git clone <your-clone-url> && cd webpack.js.org`
 
 Once you are in the project directory, run the following commands:
 
-- `npm install` to pull all dependencies.
-- `npm run build` to create a production version of the site.
-- `npm start` to develop on a local webpack-dev-server: [localhost:3000][3].
-- `npm run fetch` to retrieve external documentation/data.
+- `yarn` to pull all dependencies.
+- `yarn build` to create a production version of the site.
+- `yarn start` to develop on a local webpack-dev-server: [localhost:3000][3].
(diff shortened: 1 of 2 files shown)


Legends:
I know that we have agreed to use `yarn` in the `webpack` repo, in `webpack.js.org` repo too?

byzyk:
@Legends not sure if anyone agreed on that but the current situation is a mixture of both, which is not good. I guess we started from `npm` and then switched to `yarn` but not completely. In addition to that, we already use `yarn.lock` file, we use `yarn` for CI and just for the sake of keeping it consistent with core repo. Sounds good?

package.json
```diff
@@ -24,10 +24,10 @@
     "node": ">=6.9"
   },
   "scripts": {
-    "start": "npm run init:generated && npm run fetch && node ./antwar.bootstrap.js develop",
+    "start": "yarn init:generated && yarn fetch && node ./antwar.bootstrap.js develop",
```
montogeek: This is a problem for new contributors, Yarn is not necessary to run build and run this repository.
Make `yarn` mandatory creates an unnecessary new barrier for new contributors.

Yes, we use `yarn.lock` for the CI, you changed CI to use Yarn, but new contributors shouldn't have in it order to contribute to the documentation.

.github/CONTRIBUTING.md
```diff
@@ -8,20 +8,24 @@ page please create an issue first.
...
 - Fork the **webpack.js.org** repo from [the main repository][2].
 - `git clone <your-clone-url> && cd webpack.js.org`
 
 Once you are in the project directory, run the following commands:
 
-- `npm install` to pull all dependencies.
-- `npm run build` to create a production version of the site.
-- `npm start` to develop on a local webpack-dev-server: [localhost:3000][3].
-- `npm run fetch` to retrieve external documentation/data.
+- `yarn` to pull all dependencies.
```
montogeek: Add a note stating that alternatively you can use Yarn instead of npm.

package.json
byzyk: Thanks for the review.

I get your concern, but currently, if you follow instructions from `CONTRIBUTING.md` step-by-step, running `npm install` is creating `package-lock.json` file which can also create confusion, for example, contributors might accidentally commit it and having two lock files it's not something we want to encourage, right?

Perhaps we can at least add `package-lock.json` and state in the guides that usage of yarn is more appreciated but not mandatory? What do you think?

montogeek: Please add `package-lock.json` to `.gitignore`, Same as webpack repo.

byzyk:
Closing in favor of #2256

PR #2251, opened 2018-06-11
Rebuild fix lint issues

montogeek:
This branch solves lint issues when merging latest `master` into `rebuild` branch.
- ESLint issues on new JS files.
- Markdownlint configuration for dynamically imported MD files.
- Deleted `webpack.config.js` file which mistakenly I included again when merging master last week.
- Set correct folder when deploying.
- Include again `.eslintignore` configuration file to ignore dynamically imported MD files.


package.json
```diff
@@ -34,13 +34,13 @@
     "test": "npm run lint",
     "lint": "run-s lint:*",
     "lint:js": "eslint . --ext .js,.jsx,.md",
-    "lint:markdown": "markdownlint --config ./.markdownlint.json *.md ./src/content/**/*.md",
+    "lint:markdown": "markdownlint --config ./.markdownlint.json *.md ./src/content/**/*.md --ignore './src/content/**/_*.md'",
```
byzyk: Why do we need `ignore` here? Do I understand correctly that plugins and loaders docs prefixed with an underscore are the ones we're fetching?

montogeek: We need to ignore them.
Yes, with an `_` before.

PR #2252, opened 2018-06-11
Minor SplitChunks Documentation Clean Up

jacobangel:
Originally, i was just trying to change defer => differ, but I ended up doing a bit more editing.


src/content/plugins/split-chunks-plugin.md
```diff
@@ -33,9 +34,9 @@ When trying to fulfill the last two conditions, bigger chunks are preferred.
 
 ## Configuration
 
-For developers that want to have more control over this functionality, webpack provides a set of options to better fit your needs. If you're changing the configuration, it's a good idea to measure the impact of your changes to ensure there's a real benefit.
+Webpack provides a set of options for developers that want more control over this functionality.
```
jacobangel: webpack should be lower case

jacobangel:
Oof, my apologies. I think by updating with master I must have reset the review.

montogeek:
@jacobangel No, What do you mean with reset the review?

jacobangel:
I meant I merged master, and my "approval" disappeared.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -94,7 +95,7 @@ module.exports = {
 };
 ```
 
-Alternatively, you can provide a function for more control. The return value will indicate whether to include each chunk.
+Alternatively, you may provide a function for more control. The return value will indicate whether to include each chunk.
```
EugeneHlushko: is this really a thing?

src/content/plugins/split-chunks-plugin.md
```diff
@@ -18,9 +19,9 @@ Since webpack v4, the `CommonsChunkPlugin` was removed in favor of `optimization
 
 ## Defaults
 
-Out of the box `SplitChunksPlugin` should work great for most users.
+Out of the box `SplitChunksPlugin` should work well for most users.
```
EugeneHlushko: feels like it didnt work great for most users and we had to change to to `well` 😕

jacobangel: In this case, "well" reads as the more natural adverb in this sentence.

montogeek:
Thanks!

PR #2254, opened 2018-06-12
Rebuild sitemap and others

montogeek:
I made a lot of mistakes when merging master into `rebuild` branch last week.
This PR rectifies those mistakes and also fix some TODOs along:
- Delete unused files.
- Update Alex configuration, https://github.com/webpack-contrib/npm-install-webpack-plugin/pull/124 was created to fix a warning.
- Include sitemap generation in build pipeline.
- Include CNAME generation in build pipeline.
- Removed unused dependencies.
- Add missing dependencies.
- Update SCSS import.


.alexignore
```diff
@@ -1 +1 @@
-generated/plugins/*
+src/content/plugins/_*.md
```
jeremenichelli: Shouldn't we ignore more paths, like _loaders_? Or all the fetched files go to this one?

package.json
```diff
@@ -25,19 +25,18 @@
...
     "clean": "rimraf ./dist",
-    "content": "node src/scripts/build-content-tree.js ./src/content ./src/_content.json",
     "prestart": "npm run clean",
     "start": "NODE_ENV=development webpack-dev-server --config webpack.dev.js --env.dev",
-    "prebuild": "npm run clean",
-    "build": "npm run content && npm run fetch && NODE_ENV=production webpack --config webpack.prod.js",
+    "content": "node src/scripts/build-content-tree.js ./src/content ./src/_content.json",
     "fetch": "sh src/scripts/fetch.sh",
+    "prebuild": "npm run clean",
+    "build": "npm run content && npm run fetch && NODE_ENV=production webpack --config webpack.prod.js && npm run sitemap && echo webpack.js.org > dist/CNAME",
```
jeremenichelli: Shouldn't _fetch_ run before content so the external markdown files are treated as content routes?

.alexignore
montogeek: Yeah

package.json
montogeek: Not sure, let me test

montogeek: Not, it is not the case. Routes are not created.

jeremenichelli: But, we don't need those routes? Or they aren't being created and that's a bug?

montogeek: No, we don't need them. I have compared each page to webpack.js.org and everything is there.

EugeneHlushko: I think the problem is, that `clean` doesnt remove fetched content, probably that is why you still get contents

EugeneHlushko: which should be the case for a clean build

montogeek: @EugeneHlushko I am not following you :D

EugeneHlushko: If you run `fetch` after `content` after pulling the repo, you wont get pages from fetch into directory tree

EugeneHlushko: content should be run after fetch

montogeek: @jeremenichelli @EugeneHlushko Absolutely right. Fixed in https://github.com/webpack/webpack.js.org/pull/2254/commits/fef8397c65a22dde49bfffbc6a97ccb73538aa4f

montogeek:
Thank you @EugeneHlushko

PR #2257, opened 2018-06-13
chore(rebuild): remove CNAME, remove unused antwar configs

byzyk:


montogeek:
Already done in https://github.com/webpack/webpack.js.org/pull/2254

byzyk:
Oh sorry, I didn't notice that. I'll update it to only remove `CNAME` file since it remains unused.

PR #2259, opened 2018-06-13
Making the examples consistent.

Spiral90210:
There was some inconsistency between referring to main.js in the html file, and bundle.js on other places. I assume that the default in previous versions was bundle.js, however following the examples with console output, webpack uses main.js, so I have updated the examples to reflect this, as it reflected quite badly that the getting started instructions followed exactly as examples presented wouldn't actually work. Happy to have my name removed as a contributed if deemed necessary, it's only a small change after all.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -179,7 +179,7 @@ __dist/index.html__
 
 In this setup, `index.js` explicitly requires `lodash` to be present, and binds it as `_` (no global scope pollution). By stating what dependencies a module needs, webpack can use this information to build a dependency graph. It then uses the graph to generate an optimized bundle where scripts will be executed in the correct order.
 
-With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `main.js` as the (default) [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
+With that said, let's run `npx webpack` with no arguments, which will take our script at `src/index.js` as the [entry point](/concepts/entry-points), and will generate `dest/main.js` as the [output](/concepts/output).. The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
 
 ``` bash
 npx webpack
@@ -190,7 +190,8 @@ Time: 287ms
 Built at: 13/06/2018 11:52:07
   Asset      Size  Chunks             Chunk Names
 main.js  70.4 KiB       0  [emitted]  main
-[1] (webpack)/buildin/module.js 497 bytes {0} [built][2] (webpack)/buildin/global.js 489 bytes {0} [built]
+[1] (webpack)/buildin/module.js 497 bytes {0} [built]
+[2] (webpack)/buildin/global.js 489 bytes {0} [built]
 [3] ./src/index.js 216 bytes {0} [built]
     + 1 hidden module
 
@@ -255,13 +256,13 @@ Time: 283ms
 Built at: 13/06/2018 11:53:51
   Asset      Size  Chunks             Chunk Names
 main.js  70.4 KiB       0  [emitted]  main
-[1] (webpack)/buildin/module.js 497 bytes {0} [built][2] (webpack)/buildin/global.js 489 bytes {0} [built]
+[1] (webpack)/buildin/module.js 497 bytes {0} [built]
+[2] (webpack)/buildin/global.js 489 bytes {0} [built]
 [3] ./src/index.js 216 bytes {0} [built]
     + 1 hidden module
 
 WARNING in configuration
-The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable d
-efaults for each environment.
+The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
 You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/concepts/mode/
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.md
```diff
@@ -178,25 +179,25 @@ __dist/index.html__
 
 In this setup, `index.js` explicitly requires `lodash` to be present, and binds it as `_` (no global scope pollution). By stating what dependencies a module needs, webpack can use this information to build a dependency graph. It then uses the graph to generate an optimized bundle where scripts will be executed in the correct order.
 
-With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `bundle.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
+With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `main.js` as the (default) [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
```
byzyk: why do we need parentheses here `(default)`?

src/content/guides/getting-started.md
```diff
@@ -178,25 +179,25 @@ __dist/index.html__
...
-Entrypoint main = bundle.js
-   [1] (webpack)/buildin/module.js 519 bytes {0} [built]
-   [2] (webpack)/buildin/global.js 509 bytes {0} [built]
-   [3] ./src/index.js 256 bytes {0} [built]
+Version: webpack 4.12.0
+Time: 287ms
+Built at: 13/06/2018 11:52:07
+  Asset      Size  Chunks             Chunk Names
+main.js  70.4 KiB       0  [emitted]  main
+[1] (webpack)/buildin/module.js 497 bytes {0} [built][2] (webpack)/buildin/global.js 489 bytes {0} [built]
```
byzyk: missing linebreak before `[2]`

src/content/guides/getting-started.md
```diff
@@ -178,25 +179,25 @@ __dist/index.html__
...
+[3] ./src/index.js 216 bytes {0} [built]
     + 1 hidden module
 
 WARNING in configuration
-The 'mode' option has not been set. Set 'mode' option to 'development' or 'production' to enable defaults for this environment.
+The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable d
+efaults for each environment.
```
byzyk: no need for linebreak here

src/content/guides/getting-started.md
```diff
@@ -249,19 +250,19 @@ Now, let's run the build again but instead using our new configuration file:
...
-Entrypoint main = bundle.js
-   [1] (webpack)/buildin/module.js 519 bytes {0} [built]
-   [2] (webpack)/buildin/global.js 509 bytes {0} [built]
-   [3] ./src/index.js 256 bytes {0} [built]
+Version: webpack 4.12.0
+Time: 283ms
+Built at: 13/06/2018 11:53:51
+  Asset      Size  Chunks             Chunk Names
+main.js  70.4 KiB       0  [emitted]  main
+[1] (webpack)/buildin/module.js 497 bytes {0} [built][2] (webpack)/buildin/global.js 489 bytes {0} [built]
```
byzyk: missing linebreak before `[2]`

src/content/guides/getting-started.md
```diff
@@ -249,19 +250,19 @@ Now, let's run the build again but instead using our new configuration file:
...
+[3] ./src/index.js 216 bytes {0} [built]
     + 1 hidden module
 
 WARNING in configuration
-The 'mode' option has not been set. Set 'mode' option to 'development' or 'production' to enable defaults for this environment.
+The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable d
+efaults for each environment.
```
byzyk: same as above

src/content/guides/getting-started.md
Spiral90210: It was to clarify the statement a little - it says to run the command with main.js as the output (and I suppose src/index.js as the input!), but the actual command invoked is just `webpack`, and you're not actually doing anything to cause the output file to be main.js, you're just running the command and it defaults that. Perhaps re-writing that would read more easily, something like this?

"With that said, let's run `npx webpack` with no arguments, which will take our script at `src/index.js` as the [entry point](/concepts/entry-points), and will generate `dest/main.js` as the [output](/concepts/output)."

src/content/guides/getting-started.md
Spiral90210: Very strange, I copied from my terminal! That's on me for assuming it was fine - same as the others :(

Spiral90210:
Ugh, I made those last edits on my work machine which has git configured with an internal account. I've added the email address to my account, hopefully this will clear up the cla problem.

src/content/guides/getting-started.md
```diff
@@ -178,25 +179,25 @@ __dist/index.html__
 
 In this setup, `index.js` explicitly requires `lodash` to be present, and binds it as `_` (no global scope pollution). By stating what dependencies a module needs, webpack can use this information to build a dependency graph. It then uses the graph to generate an optimized bundle where scripts will be executed in the correct order.
 
-With that said, let's run `npx webpack` with our script as the [entry point](/concepts/entry-points) and `bundle.js` as the [output](/concepts/output). The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
+With that said, let's run `npx webpack` with no arguments, which will take our script at `src/index.js` as the [entry point](/concepts/entry-points), and will generate `dest/main.js` as the [output](/concepts/output).. The `npx` command, which ships with Node 8.2 or higher, runs the webpack binary (`./node_modules/.bin/webpack`) of the webpack package we installed in the beginning:
```
byzyk: Looks good! One small thing, can you please remove `with no arguments` as it's already clear what we're running. Also, I guess it's `dist/main.js` based on config in the example above. Please remove unneeded period after `output` link.

Spiral90210: Oops! Will do that now.

Spiral90210:
Travis build seems to be failing for an excessive number of api calls somewhere, not to do with these changes (it has passed previously).

montogeek:
Thanks!

PR #2260, opened 2018-06-13
Update getting-started.md

andyturner:
Until the later point when a config file is applied, Webpack is outputting to main.js, not bundle.js

Sorry, no CLA.. you can fix this yourselves easily enough!


andyturner:
Apologies, but with it being such a tiny correction, I didn't imagine that there would already be a pending fix, especially one a month old.

PR #2261, opened 2018-06-13
Rebuild sidebar order

montogeek:
Sidebar items order were wrong:
image

They look fine after this fix:
image


montogeek:
I created this branch from `rebuild`, not sure about those 3 first commits :/

montogeek:
This doesn't work.

PR #2264, opened 2018-06-13
Fix import in typescript files

Nek-:
Since there is no default export in path and webpack definitions, we need to import modules this way.

### DIFF
--- a/src/content/configuration/configuration-languages.md
+++ b/src/content/configuration/configuration-languages.md
@@ -27,8 +27,8 @@ and then proceed to write your configuration:
 __webpack.config.ts__
 
 ```typescript
-import path from 'path';
-import webpack from 'webpack';
+import * as path from 'path';
+import * as webpack from 'webpack';
 
 const config: webpack.Configuration = {
   mode: 'production',


asapach:
A couple of lines below it says:
> Above sample assumes version >= 2.7 or newer of TypeScript is used with the new `esModuleInterop` and `allowSyntheticDefaultImports` compiler options in your tsconfig.json file.

That way you can use default imports.

EugeneHlushko:
I dont think this is viable, as @asapach mentioned the code works for TS 2.7+
cc @montogeek

montogeek:
I don't have experience with TypeScript, but from a documentation point of view. Maybe we could add `tsconfig.json` as well?

dhruvdutt:
@EugeneHlushko is right. This is not needed.
Not sure about `tsconfig.json` because that has nothing to do with webpack. The config is related to TypeScript in general and requires no special tuning for webpack.

Nek-:
I was not able to make things work properly by adding the following config to my tsconfig.json:

```
{
"esModuleInterop": true,
"allowSyntheticDefaultImports": true
}
```

dhruvdutt:
@Nek- What version of TypeScript are you using?

Nek-:
@dhruvdutt  `^2.9.2`

dhruvdutt:
If you send a file/repo to replicate, that would be convenient. I think this is fixed in latest versions.

Nek-:
It's here: https://github.com/Nek-/box2d.ts/blob/feature%2Fnew-build-system/Box2D/Testbed/webpack.config.ts

montogeek:
@dhruvdutt Can you please check it?

EugeneHlushko:
related: #3561

EugeneHlushko:
Thanks!

PR #2266, opened 2018-06-14
Update mode.md

daaain:
Update `mode` docs to be in line with current `module.exports` function signature.

At least for me a 4.8.3 → 4.12.0 upgrade removed the second parameter, so I'm guessing this should be reflected in the docs?

### DIFF
--- a/src/content/concepts/mode.md
+++ b/src/content/concepts/mode.md
@@ -90,13 +90,13 @@ var config = {
   //...
 };
 
-module.exports = (env, argv) => {
+module.exports = env => {
 
-  if (argv.mode === 'development') {
+  if (env.mode === 'development') {
     config.devtool = 'source-map';
   }
 
-  if (argv.mode === 'production') {
+  if (env.mode === 'production') {
     //...
   }
 


montogeek:
In the example, default module export is a function, you can get rest of parameters in `args` object.
`env` is a string parameter, not an object.

montogeek:
How exactly are you calling your webpack configuration?

daaain:
OK, so apparently it's a webpack-command issue / inconsistency, see https://github.com/webpack/webpack/issues/6460#issuecomment-397322772

daaain:
As mentioned in the linked issue thread, having no `env` but only `argv` returned is apparently an explicit design choice in webpack-command / config-loader: https://github.com/webpack-contrib/config-loader#gotchas

So what should be in the official Webpack docs now that webpack-cli is to be deprecated **this month**?

cc @shellscape

montogeek:
We could change it once webpack-command is official, until then, it works the way it is now.

montogeek:
PR is not longer valid. Thanks for your contribution.

PR #2268, opened 2018-06-14
Fix Notification bar wont close

montogeek:
Notification bar won't close on production.
This PR fixes it in a React way.


montogeek:
Would like to update React to use Fragment API to remove that empty `div`.

PR #2270, opened 2018-06-15
Add GitHub token to GitHub API calls

montogeek:
Follow up of https://github.com/webpack/webpack.js.org/pull/1963 that actually works.
Travis scripts runs in a child process which doesn't inherit environment variables, so you have to re export them.

See https://travis-ci.org/webpack/webpack.js.org/builds/392673634?utm_source=github_status&utm_medium=notification for logs regarding rate limits, it says 5000 instead of 60. I hope this will solve our constant failed builds.

### DIFF
--- a/src/scripts/fetch_package_names.js
+++ b/src/scripts/fetch_package_names.js
@@ -36,8 +36,8 @@ function fetchPackageNames(options, cb) {
   console.log("fetchPackageNames");
   const github = new GitHubApi();
 
-  console.log(process.env);
-  console.log(process.env.GITHUB_TOKEN);
+  console.log("processEnv", process.env);
+  console.log("GH token", process.env.GITHUB_TOKEN);
 
   if(process.env.GITHUB_TOKEN) {
     github.authenticate({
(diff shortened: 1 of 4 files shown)


src/scripts/fetch_package_names.js
```diff
@@ -1,58 +1,67 @@
...
-  }, function(err, d) {
-    if (err) {
-      return console.error(err);
-    }
+  fetchPackageNames(
+    {
+      organization: organization,
+      suffix: suffix
+    },
+    function(err, d) {
```
EugeneHlushko: can you define the error handler outside of function call and just pass it here pls? This would make it even more readable, same goes for following error handler

EugeneHlushko:
One stylistic comment otherwise LGTM

src/scripts/fetch_package_names.js
montogeek: Out of the scope of this PR, I just reformatted that file, actually I should revert it.
Also, we should change how fetch script works, one is reading stdout from the previous one, making it harder to test in isolation.

EugeneHlushko: nah dont revert, formatting is good now

PR #2273, opened 2018-06-16
docs(config) document removeAvailableModules option

EugeneHlushko:
* document removeAvailableModules option
* add missing examples

Also takes one off from #1968

### DIFF
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -158,8 +158,47 @@ Possible values:
 - any string: the value to set `process.env.NODE_ENV` to.
 - false: do not modify/set the value of `process.env.NODE_ENV`.
 
+__webpack.config.js__
+
+```js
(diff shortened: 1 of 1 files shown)


src/content/configuration/optimization.md
```diff
@@ -158,8 +158,47 @@ Possible values:
...
+    mangleWasmImports: true
+  }
+};
+```
+
+## `optimization.removeAvailableModules`
+
+`bool: true`
+
+Tells webpack to detect and remove modules from chunks when these modules are already included in all parents. Setting `optimization.removeAvailableModules` to `false` will disable this optimization.
```
montogeek: Default value is `true` then?

EugeneHlushko: yup i did mention that:
`bool: true`

montogeek:
Thanks!

PR #2282, opened 2018-06-20
docs(config): updated module.noParse example

willchb:
Updated the example in configuration doc to put noParse as a child of module in config.


byzyk:
Good catch! Seems like I missed this one while was working on code examples 😬

montogeek:
Thanks!

PR #2283, opened 2018-06-20
Fix typo in print.js diff

diogosalazar:
Fixed typo for console command in the diff snippet for the documentation on the dev server


diogosalazar:
My mistake, I was skimming through the documentation and spotted what seemed to be a typo. Only noticed it was an intentional typo after I submitted the PR. Please ignore this.

PR #2285, opened 2018-06-21
Add chunkGroups option to stats config docs

lencioni:
This option is picked up here:

  https://github.com/webpack/webpack/blob/4de3ce05/lib/Stats.js#L145-L148

And used here:

  https://github.com/webpack/webpack/blob/4de3ce05/lib/Stats.js#L474-L476

This controls the addition of `namedChunkGroups` in the stats output.


montogeek:
Thanks!

PR #2288, opened 2018-06-22
Fix Sidebar order alphabetically

montogeek:
Pages sidebar order was incorrect in all pages, this PR fixes it.

We are using `directory-tree` to get a JS Object representation of our content, at this moment it returns `index.md` files as any other page, but we have to use it as our main anchor in the sidebar.

I tried to change how `sort` function works on  `content-tree-enhancers.js`, without success, as last resort I manually moved the item to the beginning of the array. It works but it is a hack.

We should think in a better structure for our content, or at least create a different representation between system files and site content structure and navigation (sidebar).

### DIFF
--- a/src/components/Sidebar/Sidebar.jsx
+++ b/src/components/Sidebar/Sidebar.jsx
@@ -29,7 +29,7 @@ export default ({
           group = page.group;
 
           return (
-            <div key={`sidebar-item-${index}`}>
+            <React.Fragment key={`sidebar-item-${index}`}>
               {displayGroup ? <h4 className="sidebar__group">{group}</h4> : null}
 
               <SidebarItem
@@ -39,7 +39,7 @@ export default ({
                 anchors={page.anchors}
                 currentPage={currentPage}
               />
-            </div>
+            </React.Fragment>
           );
         })}
       </div>
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -32,7 +32,7 @@ import Content from '../../_content.json';
 class Site extends React.Component {
   state = {
     mobileSidebarOpen: false
-  }
+  };
 
   render() {
     let { location } = this.props;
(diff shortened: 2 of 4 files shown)


src/components/Sidebar/Sidebar.jsx
```diff
@@ -29,7 +29,7 @@ export default ({
           group = page.group;
 
           return (
-            <div key={`sidebar-item-${index}`}>
+            <React.Fragment key={`sidebar-item-${index}`}>
```
montogeek: Avoid unnecessary divs.

src/components/Site/Site.jsx
```diff
@@ -32,7 +32,7 @@ import Content from '../../_content.json';
 class Site extends React.Component {
```
montogeek: This file got Prettierified.

src/components/Site/Site.jsx
```diff
@@ -137,6 +127,14 @@ class Site extends React.Component {
    * @return {array}       - ...
    */
   _strip = array => {
+    let anchorTitleIndex = array.findIndex(item => item.name === 'index.md');
```
montogeek: This is the main change, I am finding the object which name is `index.md` and copy it at the beginning of the array, then proceed to remove it from its original position.

This is the result:
image

src/utilities/content-tree-enhancers.js
```diff
@@ -56,20 +56,25 @@ const sort = (a, b) => {
   if (group1 > group2) return 1;
   if (a.sort && b.sort) return a.sort - b.sort;
 
-  else return 0;
+  let aTitle = (a.title || '').toLowerCase();
```
montogeek: This will order alphabetically items in the sidebar.

src/components/Site/Site.jsx
EugeneHlushko: i would suggest going for `item.name.toLowercase()` in case if someone accidentally makes `index.MD` in the future

montogeek: @EugeneHlushko Updated

PR #2291, opened 2018-06-23
Fix the order of loaders

just-boris:
The actual behavior is that "inline" loaders are invoked after "normal"

fixes https://github.com/webpack/webpack/issues/7576


montogeek:
Thanks!

PR #2292, opened 2018-06-23
docs(api): fix cli example

chenxsan:
`entry` should begin with `./`, or we'll get error:

> Entry module not found: Error: Can't resolve 'src/index.js' in ...

Because webpack will try to resolve it from `node_modules` directory.


montogeek:
Thanks!

Issue #2244, opened 2018-06-09
Let's improve the accessibility of the website


tryzniak:
Hello. I ran Google Chrome's audit on accessibility and it turns out there are a number of issues. A lot of them it looks to me can be easily fixed like missing `alt` attributes, buttons without a name, low contrast. 
I can work on a PR btw :)

montogeek:
Send them :)

jeffin143:
![IMG_20191206_010530](https://user-images.githubusercontent.com/21286134/70267462-845df380-17c4-11ea-94b1-904ab413bb03.jpg)


It has reached 99 :)

Loved it , let's close this 
May be we could hit 100 some day :-p

PR #2294, opened 2018-06-24
fix(accessibility) Improve landmarking of the site

tryzniak:
Here is a PR for #2244 .

- adds a navigation landmark
- adds a search landmark
- adds some labels to links

### DIFF
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -37,7 +37,7 @@ export default class Navigation extends React.Component {
             }) }
           </nav>
 
-          <div className="navigation__search">
+          <div role="search" className="navigation__search">
             <input
               aria-label="Search documentation"
               type="search"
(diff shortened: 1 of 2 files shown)


src/components/Navigation/Navigation.jsx
```diff
@@ -78,7 +80,7 @@ export default class Navigation extends React.Component {
 
         { Links.filter(link => this._isActive(link) && link.children).map(link => (
           <div className="navigation__bottom" key={ link.title }>
-            <Container className="navigation__inner">
+            <nav className="container navigation__inner">
```
montogeek: What is the reason of this change?

tryzniak: Ah this one. So when using a screenreader there is a way quickly jump to navigations. By using <nav></nav> we add a navigation landmark.

montogeek:
Thanks!

PR #2299, opened 2018-06-26
Plugins/internals

### LINKED ISSUE #2275: ConstPlugin - Internal plugins
I saw this [wiki-page ](https://github.com/webpack/docs/wiki/internal-webpack-plugins) which describes internal plugins.
Could this page be enriched with infos and added to the official docs?

What does [ConstPlugin ](https://github.com/webpack/docs/wiki/internal-webpack-plugins#constplugin)actually do?
The wiki says:

>  Try to evaluate expressions in if(...) and replace it with true/false.

But this [issue ](https://github.com/webpack/webpack/issues/7519 )says that it should also remove dead branches....?

I thought `treeshaking` is done by the `UglifyJsPlugin`?
Does it mean that `ConstPlugin` does it too?

EugeneHlushko:
### Summary

* port [Plugins internals wiki page](https://github.com/webpack/docs/wiki/internal-webpack-plugins)
* update content for `ConstPlugin` ideally Fixes #2275
* reformat and refactor some of the MD
* do small rewording where possible (there are still places where we could make this great, but it can take too long within this single PR so i leave it as is and concentrate on `ConstPlugin`)

### Post deploy steps

Remove content from  [Plugins internals wiki page](https://github.com/webpack/docs/wiki/internal-webpack-plugins) and add a link to `https://webpack.js.org/plugins/internal-plugins/`

### DIFF
--- a/src/content/plugins/internal-plugins.md
+++ b/src/content/plugins/internal-plugins.md
@@ -8,9 +8,9 @@ contributors:
 
 This is a list of plugins which are used by webpack internally.
 
-W>You should only care about them if you are building your own compiler based on webpack, or introspect the internals.
+W> You should only care about them if you are building your own compiler based on webpack, or introspect the internals.
 
-categories of internal plugins:
+Categories of internal plugins:
 
 * [environment](#environment)
 * [compiler](#compiler)
@@ -27,7 +27,7 @@ Plugins affecting the environment of the compiler.
 
 `node/NodeEnvironmentPlugin()`
 
-Applies node.js style filesystem to the compiler.
+Applies Node.js style filesystem to the compiler.
 
 ## compiler
 
@@ -45,7 +45,7 @@ You can pass a `cache` object, where the modules are cached. Otherwise one is cr
 
 `ProgressPlugin(handler)`
 
-Hook into the compiler to extract progress information. The `handler` must have the signature `function(percentage, message)`. It's called with `0 <= percentage <= 1`. `percentage == 0` indicates the start. `percentage == 1` indicates the end.
+Hook into the compiler to extract progress information. The `handler` must have the signature `function(percentage, message)`. Percentage is called with a value between 0 and 1, where 0 indicates the start and 1 the end.
 
 ### RecordIdsPlugin
 
@@ -159,9 +159,9 @@ Add support for hot module replacement. Decorates the templates to add runtime c
 
 Plugins affecting the source code of modules.
 
-## APIPlugin
+### APIPlugin
 
-Make webpack_public_path## , webpack_require## , webpack_modules## , webpack_chunk_load##  accessible. Ensures that `require.valueOf` and `require.onError` are not processed by other plugins.
+Make webpack_public_path, webpack_require, webpack_modules and webpack_chunk_load accessible. Ensures that `require.valueOf` and `require.onError` are not processed by other plugins.
 
 ### CompatibilityPlugin
 
--- a/src/content/plugins/internal-plugins.md
+++ b/src/content/plugins/internal-plugins.md
@@ -173,7 +173,8 @@ Offers a pseudo `console` if it is not available.
 
 ### ConstPlugin
 
-Tries to evaluate expressions in `if (...)` conditions and replace them with `true`/`false` for further possible dead branch elimination.
+Tries to evaluate expressions in `if (...)` statements and ternaries to replace them with `true`/`false` for further possible dead branch elimination using hooks fired by the parser.
+
 There are multiple optimizations in production mode regarding dead branches:
 
 * The ones performed by UglifyJS
--- a/src/content/plugins/internal-plugins.md
+++ b/src/content/plugins/internal-plugins.md
@@ -21,7 +21,9 @@ categories of internal plugins:
 
 Plugins affecting the environment of the compiler.
 
-__`node/NodeEnvironmentPlugin`__
+### NodeEnvironmentPlugin
+
+`node/NodeEnvironmentPlugin()`
 
 Applies node.js style filesystem to the compiler.
 
@@ -29,17 +31,23 @@ Applies node.js style filesystem to the compiler.
 
 Plugins affecting the compiler
 
-__`CachePlugin([cache])`__
+### CachePlugin
+
+`CachePlugin([cache])`
 
 Adds a cache to the compiler, where modules are cached.
 
 You can pass a `cache` object, where the modules are cached. Otherwise one is created per plugin instance.
 
-__`ProgressPlugin(handler)`__
+### ProgressPlugin
+
+`ProgressPlugin(handler)`
 
 Hook into the compiler to extract progress information. The `handler` must have the signature `function(percentage, message)`. It's called with `0 <= percentage <= 1`. `percentage == 0` indicates the start. `percentage == 1` indicates the end.
 
-__`RecordIdsPlugin()`__
+### RecordIdsPlugin
+
+`RecordIdsPlugin()`
 
 Saves and restores module and chunk ids from records.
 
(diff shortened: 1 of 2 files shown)


EugeneHlushko:
Hmm probably i need to rephrase this, i suspect that import and dead if () {} gets removed by uglify @Legends

src/content/plugins/internal-plugins.md
```diff
@@ -163,104 +187,142 @@ if (a) {
 }
 ```
 
-Becausse `a` will evaluate as `false` in the `if (a)` condition and `foo()` call will never happen, the whole import will get pruned thanks to `ConstPlugin` run.
+Because `a` will evaluate as `false` in the `if (a)` condition and `foo()` call will never happen, the whole import will get pruned thanks to `ConstPlugin` run.
+
```
Legends: This doesn't sound that well:

>  the whole import will get pruned thanks to `ConstPlugin` run

..will get pruned by the `ConstPlugin`

.

EugeneHlushko: i am not quite sure, because import pruning maybe happens via uglify, but if not const plugin that wouldnt be possible

EugeneHlushko: thats what i was trying to say by "thanks to ..."

Legends: Probably yes, I would assume so.
`UglifyJS` is injected, when `--mode production` (treeshaking).

I guess `webpack` recognizes then, after the `ConstPlugin` has removed dead branches that there is an unused `import` and removes it, right?

EugeneHlushko: yeah so there are two options and would be nice if we can confirm:

a) by the time code gets to uglify it contains if (false) {...} produced by const plugin, so at this point uglify eliminates both import and if (false) {} branch.
b) by the time code gets to uglify it only contains unused import that uglify removes, because if (false) {} was evaluated and removed by const plugin

My guess is that option a) is correct and const plugin only does evaluate conditionals and tries to put false/true where possible, after that uglify removes both branches and imports. But as i said would be nice to turn uglify off and try

Legends: There is an open [issue](https://github.com/webpack/webpack/issues/7519) regarding ConstPlugin

EugeneHlushko: Yes but it doesnt really say that anything except ternary clearing isnt working, i still think its done by uglify. @sokra, @ooflorent can you clarify please

ooflorent: There are multiple optimizations in production mode regarding dead branches:

- The ones performed by UglifyJS
- The ones performed by webpack

webpack will try to evaluate conditional statements. If it succeeds then the dead branch is removed. webpack can't do constant folding unless the compiler knows it.

```js
const FOO = 1
if (FOO === 0) { /* dead */ }
```

In the above example, webpack is unable to prune the branch, but Uglify does. However, if `FOO` is defined using `DefinePlugin`, webpack will succeed.

```js
if (0 === 1) { /* dead */ }
```

Here webpack is able to reduce the comparison and will remove the branch.

> except ternary clearing isnt working

It also works for ternaries but not with _sloppy_ comparisons (see the issue mentioned by Legends).

EugeneHlushko: Pretty much what we though except for the define plugin detail, thank you very much @ooflorent for stepping by. i will update this one

EugeneHlushko:
@Legends i did update texts according to @ooflorent 's comments, looks good to me but wanted a double check

Legends:
I have only one question left:

`ConstPlugin` does only evaluation of expressions in `if` statements and the pruning logic is done, depending on the case, by `UglifyJS` or some other internal webpack logic !?
I am bit irritated by the title of the existing [issue](https://github.com/webpack/webpack/issues/7519):

> ConstPlugin should remove dead branches in LogicalExpressions

EugeneHlushko:
Yes at least this is what i understand from @ooflorent 's comment, also looking at plugin source code i dont see dead branch elimination

Legends:
Ok, lets keep it that way. If @ooflorent has any further comments on this, I am sure he will provide additional input if necessary.

ooflorent:
> @Legends 
> `ConstPlugin` does only evaluation of expressions in `if` statements and the pruning logic is done, depending on the case, by UglifyJS or some other internal webpack logic !?

The plugin handles pruning of `if` statements and ternaries using hooks fired by the parser.

> `ConstPlugin` should remove dead branches in `LogicalExpressions`

This issue has been created because the `Parser` does not have a hook prune `LogicalExpression` nodes.

> @EugeneHlushko 
> also looking at plugin source code i dont see dead branch elimination

https://github.com/webpack/webpack/blob/b06cca2371c3da0ede7866d33bb304a132072919/lib/ConstPlugin.js#L173-L184

https://github.com/webpack/webpack/blob/b06cca2371c3da0ede7866d33bb304a132072919/lib/ConstPlugin.js#L211-L219

EugeneHlushko:
thought it only replaces conditions, thanks again @ooflorent @Legends 
Now ConstPlugin section mentions both if statements and ternaries. Also pointed that dead branch elimination happens further down the line using hooks fired by the parser.
This is now ready for review @montogeek

src/content/plugins/internal-plugins.md
```diff
@@ -0,0 +1,335 @@
...
+`options` are the output options.
+
+If `options.pathinfo` is set, each module function is annotated with a comment containing the module identifier shortened to `context` (absolute path).
+
+### JsonpTemplatePlugin
+
+`JsonpTemplatePlugin(options)`
+
+Chunks are wrapped into JSONP-calls. A loading algorithm is included in entry chunks. It loads chunks by adding a `<script>` tag.
```
montogeek: Don't we have some docs regarding preload/prefetch? We are talking about the same right?

src/content/plugins/internal-plugins.md
```diff
@@ -0,0 +1,335 @@
...
+
+`options.publicPath` is uses as path for loading the chunks.
+
+`options.chunkFilename` is the filename under that chunks are expected.
+
+### NodeTemplatePlugin
+
+`node/NodeTemplatePlugin(options)`
+
+Chunks are wrapped into node.js modules exporting the bundled modules. The entry chunks loads chunks by requiring them.
```
montogeek: Node.js

montogeek:
What is this syntax? webpack_foo###

src/content/plugins/internal-plugins.md
```diff
@@ -0,0 +1,335 @@
...
+title: Internal webpack plugins
+contributors:
+  - EugeneHlushko
+  - ooflorent
+  - Legends
+---
+
+This is a list of plugins which are used by webpack internally.
+
+W>You should only care about them if you are building your own compiler based on webpack, or introspect the internals.
```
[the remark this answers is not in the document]
EugeneHlushko: Yes, it looks good, but i will add space 👍

src/content/plugins/internal-plugins.md
```diff
@@ -0,0 +1,335 @@
...
+
+Adds a cache to the compiler, where modules are cached.
+
+You can pass a `cache` object, where the modules are cached. Otherwise one is created per plugin instance.
+
+### ProgressPlugin
+
+`ProgressPlugin(handler)`
+
+Hook into the compiler to extract progress information. The `handler` must have the signature `function(percentage, message)`. It's called with `0 <= percentage <= 1`. `percentage == 0` indicates the start. `percentage == 1` indicates the end.
```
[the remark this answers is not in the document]
EugeneHlushko: yup makes sense

src/content/plugins/internal-plugins.md
EugeneHlushko: yes but this page just enlists all internals

EugeneHlushko:
@montogeek applied fixes, seems `webpack_smth###` was from corrupt copy paste buffer

src/content/plugins/internal-plugins.md
```diff
@@ -27,7 +27,7 @@ Plugins affecting the environment of the compiler.
 
 `node/NodeEnvironmentPlugin()`
 
-Applies node.js style filesystem to the compiler.
+Applies Node.js style filesystem to the compiler.
```
montogeek: There is 5 occurrences of this typo, didn't wanted to add a comment for each one, as advice, search in the entire document for typos.

src/content/plugins/internal-plugins.md
montogeek: Exactly, a link would be useful.

src/content/plugins/internal-plugins.md
EugeneHlushko: i think i did fix them all

Legends:
@EugeneHlushko 

If you have time, or we do it in another PR:

The categories and the category entries look the same, at first glance you cannot differentiate between them. They all have the same format and indentation. The only difference is categories start with a lowercase letter.

image

Legends:
Expandable sections with category entries, to not iterate over the whole document, this would be nice too.

image

Legends:
@EugeneHlushko should we put the additions into another PR?

EugeneHlushko:
I dont think we have this in the sidebar currently? if not lets move out @Legends @montogeek

montogeek:
@EugeneHlushko @Legends means in the content itself, we could do it later on rebuild branch :)

EugeneHlushko:
got ya, still think we dont have it implemented, lets create an issue describing feature request, can you please @Legends ?

Legends:
What I meant with `sidebar` is the following:

image

"Plugins" category is expanded, below that we could put another "Internal Plugins" category, which is closed initially but expands on click showing it's subcategories like `environment`, `compiler` ....

EugeneHlushko:
Yeah its a sidebar indeed, but the thing is the first level categories are separate pages (sections) and inside of them are the headings from MD. So in order to get these internals in such view would mean splitting them into different md files by categories, naming file titles according to categories and also to make it aligned well create whole different section for them e.g. `/internals/${page}`. As i said its not in scope here sorry :/

Legends:
@EugeneHlushko I will be off for the next 10 days.
Thanks for all the work done!

Of course the above would be another PR.  We'll see ....

montogeek:
Thanks!

PR #2003, opened 2018-04-09
Update contributors

renjithvk:
Update the contributors' list. It seems some of the contributor's usernames are updated and so wrong here.

PR #2301, opened 2018-06-26
Add contributor

renjithvk:
Update contributors list with renjithvk


montogeek:
Already done in https://github.com/webpack/webpack.js.org/pull/2003/

PR #2306, opened 2018-06-27
docs(config) document flagIncludedChunks optimization option

EugeneHlushko:
* document flagIncludedChunks optimization option

Takes one off #1968

### DIFF
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -236,3 +236,20 @@ module.exports = {
   }
 };
 ```
+
+## `optimization.flagIncludedChunks`
+
+`bool`
+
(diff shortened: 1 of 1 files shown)


src/content/configuration/optimization.md
```diff
@@ -236,3 +236,20 @@ module.exports = {
   }
 };
 ```
+
+## `optimization.flagIncludedChunks`
```
montogeek: What is its default value?

EugeneHlushko: unfortunately default is dynamic due to mode, if mode is production it will default to true as i mentioned.

montogeek:
Thanks!

PR #2308, opened 2018-06-27
See API Rate limit

montogeek:
Third take.


montogeek:
@jeremenichelli @EugeneHlushko @byzyk  I did a rather quick refactoring to fetch scripts, so I can debug GH API Rate limit easily, as you can see in the build logs, API Token is working.

byzyk:
Nice, let's see what causing the limit to exceed. Very likely it could be Travis issue.

byzyk:
@montogeek just to clarify, GH token is set as Travis env variable, right?

montogeek:
@byzyk Yes

src/scripts/fetch_package_names.js
```diff
@@ -12,6 +19,8 @@ if (require.main === module) {
 function main() {
   const organization = process.argv[2];
   const suffix = process.argv[3];
+  const file = process.argv[4];
+  const output = process.argv[5];
 
   if (!organization) {
```
EugeneHlushko: this is not for now but i would like to see an iteration vs 4 ifs

src/scripts/fetch_package_names.js
```diff
@@ -65,3 +105,98 @@ function fetchPackageNames(options, cb) {
...
+    function(pkg, cb) {
+      const branch = 'master';
+      const url = ['https://raw.githubusercontent.com', pkg.full_name, branch, file].join('/');
+
+      request(url, function(err, response, body) {
+        if (err) {
+          return cb(err);
+        }
+
+        if (body && file === 'README.md') {
```
EugeneHlushko: I think we had this in other PR but is it safe to assume that all the newly added webpack-contrib package readme files will be named exactly `README.md` and not `readme.md` or `README.MD` etc?

EugeneHlushko:
Guess it should work, sorry i couldnt test now

src/scripts/fetch_package_names.js
```diff
@@ -43,6 +78,11 @@ function fetchPackageNames(options, cb) {
       type: 'token',
       token: process.env.GITHUB_TOKEN
     });
+
+    github.misc.getRateLimit({}, (err, res) => {
```
jeremenichelli: What's the goal of this besides probably debugging on CI?

src/scripts/fetch_package_names.js
```diff
@@ -65,3 +105,98 @@ function fetchPackageNames(options, cb) {
...
+          }
+        );
+      });
+    },
+    finalCb
+  );
+}
+
+// TODO: push this type of to a script of its own to keep this generic
+function yamlHeadmatter(fields) {
```
jeremenichelli: Don't we have a util that already generates this based on a set of information? I think I did separate this logic, would love if you could adapt it so we don't go back to tech debt around this;

src/scripts/fetch_package_names.js
montogeek: That is the goal

src/scripts/fetch_package_names.js
montogeek: will try

montogeek: @jeremenichelli `rebuild` branch probably, it is not in `master`

src/scripts/fetch_package_names.js
montogeek: No

montogeek:
@EugeneHlushko @jeremenichelli Any comments about this?

src/scripts/fetch.sh
```diff
@@ -8,14 +8,14 @@ mkdir -p ./generated/plugins
 cp -rf ./src/content/plugins/ ./generated/plugins
 
 # Fetch webpack-contrib (and various other) loader repositories
-node ./src/scripts/fetch_package_names.js "webpack-contrib" "-loader" | node ./src/scripts/fetch_package_files.js "README.md" "./generated/loaders"
-node ./src/scripts/fetch_package_names.js "babel" "babel-loader" | node ./src/scripts/fetch_package_files.js "README.md" "./generated/loaders"
-node ./src/scripts/fetch_package_names.js "postcss" "postcss-loader" | node ./src/scripts/fetch_package_files.js "README.md" "./generated/loaders"
-node ./src/scripts/fetch_package_names.js "peerigon" "extract-loader" | node ./src/scripts/fetch_package_files.js "README.md" "./generated/loaders"
+node ./src/scripts/fetch_package_names.js "webpack-contrib" "-loader" "README.md" "./generated/loaders"
```
EugeneHlushko: i guess the last thing i would like changed is the `fetch_package_names.js` which becamse `fetch_packages` after merging with `fetch_package_files.js` functionality, makes sense to me to rename it accordingly

src/scripts/fetch_package_names.js
jeremenichelli: Oh sorry I thought this was on the rebuild one. Fine then :)

src/scripts/fetch.sh
montogeek: Done

PR #2316, opened 2018-06-28
Added explanation for splitChunks.minSize

alexnm:
First of all, I wanted to congratulate everyone on the great job of maintaining docs! Really useful.

For the splitChunks page, it took me a while to figure out what does the size represent (unit of measurement). I didn't find any reference on it, but hopefully I got this right since the rules are based on 30kb and the default value is 30000.

Cheers!
Alex


montogeek:
Thanks!

PR #2317, opened 2018-06-28
Concepts - Add "Bundling Fundamentals" resources

yairhaimo:
Module bundlers, and especially webpack (can) do a ton of stuff which would seem like magic to newcomers. I added a reference in the README to several resources that would contribute to the understanding of what webpack does under the hood and hopefully remove unbased fear and feeling of "magic".

I placed the links at the top because it seemed like the best fit. Let me know if it should be placed elsewhere.


montogeek:
Thanks!

PR #2318, opened 2018-06-29
docs(config): add linux troubleshoot solution for watch option

arkadyt:
Had to run webpack with **`sudo`** on Ubuntu (latest 18.04 LTS) to get the watcher to work. 
Tried all other applicable troubleshooting solutions beforehand.

Thought this could help other people who may encounter the same problem in the future.

### DIFF
--- a/src/content/configuration/watch.md
+++ b/src/content/configuration/watch.md
@@ -8,6 +8,7 @@ contributors:
   - EugeneHlushko
   - byzyk
   - spicalous
+  - arkadyt
 ---
 
 webpack can watch files and recompile whenever they change. This page explains how to enable this and a couple of tweaks you can make if watching does not work properly for you.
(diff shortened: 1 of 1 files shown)


src/content/configuration/watch.md
```diff
@@ -130,6 +131,14 @@ Verify that webpack is not being notified of changes by running webpack with the
 webpack --watch --progress
 ```
 
+### Changes Not Seen
+
+Make sure you launch webpack with root priveleges:
```
montogeek: Please change to `Make sure to execute webpack with root privileges`

asapach:
Frankly, it's a very strange recommendation: webpack should not require root and if it does, I would consider it a bug. You can do real damage to the system if for example you set incorrect [`output.path`](https://webpack.js.org/configuration/output/#output-path).

arkadyt:
Indeed, I agree with what @asapach says.
This requires more investigation.
I will close this PR for now.

PR #2325, opened 2018-07-02
fix(a11y) Make Dropdown more accessible

tryzniak:
- make it focusable
- allow navigation using arrow keys
- dismiss on escape

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -6,32 +6,59 @@ export default class Dropdown extends React.Component {
     active: false
   };
 
+  componentDidMount() {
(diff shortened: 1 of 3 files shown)


montogeek:
@tryzniak Thank you so much for this work, but I am going to ask you a big favor, Could you please work on `rebuild` branch instead?

Our current setup is complicated, with a lot of patching and using outdated and unmaintained dependencies, we are been working on a rebuild which will solve those issues, being the main one reducing build time from 20 mins to 1 min.

We can't accept more structural changes to a codebase that is already changing.

Hope you understand, thanks again!

tryzniak:
No problem, Fernando! I’ll rework my pr like you asked :)

PR #2326, opened 2018-07-02
fix(docs): Capitalize Windows and macOS

Calinou:
This uses the official casing for the Windows and macOS names.


montogeek:
Thanks!

PR #2329, opened 2018-07-03
(NotificationBar) Removed link to unixstickers

fALKENdk:
Removed link to unixstrickers, since it's not possible to buy webpack stickers anymore after they changed to only selling sticker packs.


montogeek:
Thanks!

PR #2330, opened 2018-07-03
Optimize SVGs using svgo

lencioni:
I was able to reduce the size of most of the SVGs in this repo, which
should give visitors a little bit better performance.


montogeek:
Unfortunately we have no visual regression tests (we will, soon). Did you notice any weird change on sizes or something?

lencioni:
> Unfortunately we have no visual regression tests (we will, soon).

Cool! You should consider happo.io. I bet @trotzig would help get you set up.

> Did you notice any weird change on sizes or something?

I did all of this visually in SVGOMG and it all looked very very similar to me, almost identical. You should be able to see these visually in GitHub by clicking on "View" when looking at the code changes.

trotzig:
I ran a quick check with happo (only looking at icons, in Chrome) and it looks like most diffs are subtle:
https://happo.io/a/23/compare/dev-76d8f9e7e371e59ba789/dev-876b368611dcedef48a6

I worry a little about the sizing change for `src/assets/by.svg` and `src/assets/cc.svg`. Depending on how they're used, we might want to address that. 

If you'd be interested, I could look at a more thorough happo integration.

lencioni:
> I worry a little about the sizing change for src/assets/by.svg and src/assets/cc.svg. Depending on how they're used, we might want to address that.

It looks like they are used in the footer. I grabbed their URLs from this PR and pasted them into the footer in my browser's inspector tool, and they look like they are still the same size. I can add height and width attributes to those images to be safe, and to also avoid jitter as the images come in.

lencioni:
@montogeek do you want me to make any changes to this PR?

montogeek:
Thanks!

PR #2331, opened 2018-07-03
Fix a broken markdown link in node.md

piticent123:
Not the biggest PR, but I like to help out where I can :)

It looked like the link for `WebpackOptionsDefaulter` was chopped in half to keep to a line length of 80, but markdown doesn't preserve links across new lines. I couldn't find a line length policy on the website or this repo, but since the line below it goes to 102 characters, I'm assuming you're willing to make an exception for links.  This brings line 77 up to 106 characters, but it's still safely below 120.


montogeek:
Thanks!

PR #2333, opened 2018-07-04
fix(a11y) Make Dropdown more accessible

tryzniak:
- make it focusable
- allow navigation using arrow keys
- dismiss on escape

This is a rework of #2325

### DIFF
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -46,7 +46,7 @@ export default class Navigation extends React.Component {
                 className="navigation__languages"
                 items={[
                   { title: 'English', url: 'https://webpack.js.org/' },
-                  { title: '中文', url: 'https://doc.webpack-china.org/' }
+                  { lang: 'zh', title: '中文', url: 'https://doc.webpack-china.org/' }
                 ]} />
             )
           }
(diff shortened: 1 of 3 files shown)


src/components/Navigation/Navigation.jsx
```diff
@@ -45,8 +45,8 @@ export default class Navigation extends React.Component {
               <Dropdown
                 className="navigation__languages"
                 items={[
-                  { title: 'English', url: 'https://webpack.js.org/' },
-                  { title: '中文', url: 'https://doc.webpack-china.org/' }
+                  { lang: 'en', title: 'English', url: 'https://webpack.js.org/' },
```
montogeek: What is the reason of this change?

tryzniak: It's my mistake. This shouldn't be there. Thanks for taking a look :)

src/components/Navigation/Navigation.jsx
```diff
@@ -46,7 +46,7 @@ export default class Navigation extends React.Component {
                 className="navigation__languages"
                 items={[
                   { title: 'English', url: 'https://webpack.js.org/' },
-                  { title: '中文', url: 'https://doc.webpack-china.org/' }
+                  { lang: 'zh', title: '中文', url: 'https://doc.webpack-china.org/' }
```
montogeek: It is fine to left English language with an empty `lang` attribute?

montogeek:
Thank you!

PR #2334, opened 2018-07-04
Implement responsive tables

montogeek:
Implementation of responsive tables.
This is how it looks on Desktop
image

and mobile:
image

- remark plugin will be moved to a external package once this PR is merged.


markdown.js
```diff
@@ -0,0 +1,103 @@
...
+              className: 'description desktop'
+            }
+          }
+        };
+      }
+    }
+  };
+}
+
+function logType(tree) {
```
EugeneHlushko: is this needed?

src/components/Markdown/Markdown.scss
```diff
@@ -138,125 +138,100 @@
     }
   }
 
-  .table {
+  table {
```
EugeneHlushko: TBH i would avoid tables in favor of flex or css grid via divs, but guess you had reasons for this

markdown.js
```diff
@@ -0,0 +1,103 @@
...
+
+var file = fs.readFileSync('./src/content/plugins/index.md', { encoding: 'utf8' });
+
+remark()
+  .use(parse)
+  .use(md)
+  .use(html)
+  .process(file, function(err, file) {
+    if (file) {
+      console.log(file.contents);
```
EugeneHlushko: this is weird, are we logging this for progress? maybe add some text

markdown.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+  });
+
+module.exports = md;
+
+function addClassName(node, className) {
+  node.data = {};
```
EugeneHlushko: since you are creating the whole data object on the node why not do the scaffolding in one declaration:
```
node.data = {
  hProperties: { className }
};
```

montogeek:
@EugeneHlushko From my original comment:
> remark plugin will be moved to a external package once this PR is merged.

All that code will be removed, plugin itself will be only `md` function exported.

src/components/Markdown/Markdown.scss
montogeek: That is exactly the problem here. All that CSS gimmicking tables with divs when you can use `table` directly, which is exactly the use case here.

64 additions and 89 deletions in this case, advantages of using the right element.

markdown.js
montogeek: Will be removed

markdown.js
montogeek: Will be removed, since this file is used as webpack plugin, it is not efficient to restart your webpack compilation to test it, so, I used the plugin as a standalone node program, faster to kill and restart.

markdown.js
montogeek: Will be removed

montogeek:
@EugeneHlushko I am montobot, please check latest update :D

EugeneHlushko:
ahahahah ... montobot

PR #2336, opened 2018-07-05
Show filename change from previous page

codexico:
On [Getting Started](https://webpack.js.org/guides/getting-started/index.html) uses `main.js`, and on  [Asset Management](https://webpack.js.org/guides/asset-management/) we need to show that changed to `bundle.js`.


src/content/guides/asset-management.md
```diff
@@ -29,7 +29,8 @@ __dist/index.html__
 +    <title>Asset Management</title>
     </head>
     <body>
-      <script src="./bundle.js"></script>
+-     <script src="./main.js"></script>
```
montogeek: Page starts with: `If you've been following the guides from the start, `, please remove this diff and only change `./bundle.js` to `main.js`.

montogeek:
Thanks!

PR #2337, opened 2018-07-05
Remove 404 link for NamedModulesPlugin

dmitrika:
I don't see any valid link for `NamedModulesPlugin`, so I removed it at all.


montogeek:
Thanks!

PR #2338, opened 2018-07-05
Rebuild Custom code blocks languages

montogeek:
After 4 hours doing this I finally managed to solve it.

This PR introduces support for custom `js-with-links` code blocks.

File `prism-links.js` is a copy of https://github.com/montogeek/remark-refractor, once this PR is merged I will update the dependency.

File `js-with-links.js` will be moved to https://github.com/montogeek/remark-refractor as well.

One changed is necessary, update `js-with-links` to `jslinks`, I am not 100% sure, but it seems Prism doesn't like composed names.

Also updated our process readme script to replace weird code examples (`console`, `sh`) to `bash`.

Update our content removing wrong and unnecessary code languages.


js-with-links.js
```diff
@@ -0,0 +1,78 @@
...
+//   };
+// }
+
+// function logType(tree) {
+//   console.log(JSON.stringify(tree, null, 2));
+// }
+
+module.exports = jslinks;
+
+jslinks.displayName = 'jslinks';
```
montogeek: It seems this property should be the same as the code snippet language.

package.json
```diff
@@ -41,11 +41,6 @@
     "serve": "npm run build && sirv start ./dist --port 4000",
     "deploy": "gh-pages -d dist"
   },
-  "husky": {
```
montogeek: Will revert

prism-links.js
```diff
@@ -0,0 +1,110 @@
...
+      let { data } = node;
+
+      if (!data) {
+        node.data = data = {};
+      }
+
+      try {
+        data.hChildren = refractor.highlight(node.value, lang);
+
+        for (let index = 0; index < data.hChildren.length; index++) {
```
montogeek: I find all comments in the tree and update every one by inject a link element.

prism-links.js
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      let { data } = node;
+
+      if (!data) {
+        node.data = data = {};
+      }
+
+      try {
+        data.hChildren = refractor.highlight(node.value, lang);
```
montogeek: `refractor` returns a [hast](https://github.com/syntax-tree/hast) tree.

prism-links.js
```diff
@@ -0,0 +1,110 @@
...
+
+        for (let index = 0; index < data.hChildren.length; index++) {
+          const element = data.hChildren[index];
+
+          if (element.properties && element.properties.className.includes('comment')) {
+            const { value } = element.children[0];
+
+            let match = /\[([^[\]]+?)\]\((.+?)\)/.exec(value);
```
montogeek: This Regex find all comments, it is from `markdown.js` file.

prism-links.js
```diff
@@ -0,0 +1,110 @@
...
+
+            if (match !== null) {
+              const [, text, url] = match;
+
+              element.children = [
```
montogeek: Node children are:
1. `//`
2. An empty space
3. Link

prism-links.js
```diff
@@ -0,0 +1,110 @@
...
+                      value: text
+                    }
+                  ]
+                }
+              ];
+            }
+          }
+        }
+
+        node.children = [node.data.hChildren];
```
montogeek: Will removed this line, it is not necessary.

src/utilities/process-readme.js
```diff
@@ -15,5 +15,8 @@ module.exports = function processREADME(body) {
     .replace(/<h2[^>]*>/g, '## ')
     .replace(/<\/h2>/g, '')
     // Drop any comments
-    .replace(/<!--[\s\S]*?-->/g, '');
+    .replace(/<!--[\s\S]*?-->/g, '')
+    // Replace code
+    .replace(/```console/g, '```bash')
```
montogeek: This new 2 lines save a lot of Prism warnings.

EugeneHlushko: do you think its worth it replacing content on the fly? i would suggest find replace ```console 
 -> ```bash, or is this issue coming from external readme files we fetch?

jeremenichelli: I kinda have the feeling that we should open a PR in projects using `console` and change it instead of this because it is not a valid language option in markdown regarding highlighting.

montogeek: @EugeneHlushko Yes, it is from external readmes.
@jeremenichelli Yep :D, will revert

js-with-links.js
jeremenichelli: Should we keep all the commented code?

package.json
jeremenichelli: 👍

prism-links.js
jeremenichelli: Can we rename the `hChildren` to something more meaningful. It can be just `hastChildren` or `children`, or `hastNodes`.

prism-links.js
jeremenichelli: Can we place a comment for this then?

prism-links.js
jeremenichelli: 👍

prism-links.js
```diff
@@ -0,0 +1,110 @@
...
+      } catch (e) {
+        if (!languages[lang]) {
+          console.warn('Prism does not support this language: ', lang);
+        } else {
+          console.warn('Prism failed to highlight: ', e);
+        }
+      }
+
+      data.hProperties = data.hProperties || {};
+      data.hProperties.className = ['hljs', ...(data.hProperties.className || []), `language-${lang}`];
```
jeremenichelli: Same as above, can we rename the `hProperties` to something more readable?

src/utilities/process-readme.js
jeremenichelli: We can keep this, until we replace it in all READMEs.

js-with-links.js
montogeek: From the original PR text:
> File js-with-links.js will be moved to https://github.com/montogeek/remark-refractor as well.

It means commented code will be deleted, as it is right now: https://github.com/montogeek/remark-refractor/blob/master/src/index.js

prism-links.js
montogeek: It can't, otherwise other libraries won't read it.
That is how it works: https://github.com/syntax-tree/mdast-util-to-hast/

prism-links.js
montogeek: Yeah

src/utilities/process-readme.js
montogeek: I already talked to @shellscape, console is better because GitHub support is richer than `bash`.
This is easily fixed with a plugin, just registering it using Prism:
`Prism.languages[`console`] = Prism.languages.extend('sh', {})`

prism-links.js
montogeek: We can't, same reason as `hChildren`, that is how unified ecosystem works.

montogeek:
I will revamp this PR, after working a bit more with `details` code blocks, I found some improvements.

montogeek:
Closing in favour of https://github.com/webpack/webpack.js.org/pull/2564

PR #2340, opened 2018-07-05
Update mode.md

Legends:
Added possible values for mode at the beginning.
Replaced `provides` with `sets`, sounds better imo.

### DIFF
--- a/src/content/concepts/mode.md
+++ b/src/content/concepts/mode.md
@@ -10,7 +10,7 @@ Providing the `mode` configuration option tells webpack to use its built-in opti
 
 `string`
 
-T> Possible values for `mode` are: `development` or `production`(default).
+T> Possible values for `mode` are: `none`, `development` or `production`(default).
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


Legends:
@montogeek

src/content/concepts/mode.md
```diff
@@ -10,7 +10,7 @@ Providing the `mode` configuration option tells webpack to use its built-in opti
 
 `string`
 
-T> The default value for `mode` is `production`.
+T> Possible values for `mode` are: `development` or `production`(default).
```
montogeek: What you think about adding `none`? It is a possible value as well.

montogeek:
>Setting --mode to none is the same as omitting it?
If you omit it, it will use `production`

Sorry, I don't understand your second question.

Legends:
Never mind, I just found the answer to my question, reading the docs helps in my case ;-)


> none --> Opts out of any default optimization options

montogeek:
Thanks!

PR #2350, opened 2018-07-12
Explicitly say `mode` sets properties to DefinePlugin

geeofree:
The docs for `mode` descriptions right now seems misleading as it currently says

> Sets `process.env.NODE_ENV` to value *mode value*

Which might lead people to think that it's being set like `NODE_ENV=mode value` and is unclear that it only sets the `process.env.NODE_ENV` on the `DefinePlugin` until we see the example.

With this, I think it might be better to explicitly say on the mode descriptions that it only modifies/sets the `process.env.NODE_ENV` on the `DefinePlugin`.  


montogeek:
Thanks!

PR #2356, opened 2018-07-16
Update code-splitting.md

ForJing:
the duplicate dependency removed from our `index.bundle.js` and `another.bundle.js`


montogeek:
Thanks!

PR #2357, opened 2018-07-16
Fixed grammar problems

Nick-Rivera:
Fixed comma splices
Removed commas when not necessary
Added a few missing words


montogeek:
Thanks!

PR #2359, opened 2018-07-17
docs(plugins) Relate SourceMapPlugins to default devtool config.

kremerd:
In development mode `devtool: eval` [is added automatically](https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L39) to the Webpack config. In particular it must be manually deactivated if one wants to use `SourceMapDevToolPlugin` or `EvalSourceMapDevToolPlugin`.

This PR documents this behavior, both when describing development mode, and - more importantly to me - when describing how to use the two relevant plugins.

### DIFF
--- a/src/content/plugins/eval-source-map-dev-tool-plugin.md
+++ b/src/content/plugins/eval-source-map-dev-tool-plugin.md
@@ -35,6 +35,8 @@ The following options are supported:
 
 T> Setting `module` and/or `columns` to `false` will yield less accurate source maps but will also improve compilation performance significantly.
 
+T> When using this plugin in [development mode](/concepts/mode/#mode-development), you must manually set `devTools: false`.
+
 
 ## Examples
 
--- a/src/content/concepts/mode.md
+++ b/src/content/concepts/mode.md
@@ -49,6 +49,7 @@ T> Please remember that setting `NODE_ENV` doesn't automatically set `mode`.
 // webpack.development.config.js
 module.exports = {
 + mode: 'development'
+- devtool: 'eval',
 - plugins: [
 -   new webpack.NamedModulesPlugin(),
 -   new webpack.NamedChunksPlugin(),
--- a/src/content/plugins/source-map-dev-tool-plugin.md
+++ b/src/content/plugins/source-map-dev-tool-plugin.md
@@ -41,6 +41,8 @@ The `fileContext` option is useful when you want to store source maps in an uppe
 
 T> Setting `module` and/or `columns` to `false` will yield less accurate source maps but will also improve compilation performance significantly.
 
+T> When using this plugin in [development mode](/concepts/mode/#mode-development), you must manually set `devTools: false`.
+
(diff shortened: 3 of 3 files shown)


src/content/plugins/eval-source-map-dev-tool-plugin.md
```diff
@@ -35,6 +35,8 @@ The following options are supported:
 
 T> Setting `module` and/or `columns` to `false` will yield less accurate source maps but will also improve compilation performance significantly.
 
+T> When using this plugin in [development mode](/concepts/mode/#mode-development), you must manually set `devTools: false`.
```
montogeek: Could you please add a complete example, like:
```js
module.exports = {
  // ...
  devTools: false,
}
```

montogeek: Also, it is `devtool`, not `devTools`

montogeek:
@EugeneHlushko What is your comment about?

EugeneHlushko:
You shouldnt be passing devtool false as this pr suggest, it is enough to say `devtool: 'source-map'` to use the mentioned plugin and so on for each corresponding plugin

EugeneHlushko:
Or is this about passing custom configuration object?

kremerd:
@EugeneHlushko The PR is about how to configure `EvalSourceMapDevToolPlugin` and `SourceMapDevToolPlugin` explicitly.

If it's sufficient for you to stick with one of the default `devtool` configurations, then sure you should do that. In that case I wouldn't expect you to look up the plugin configuration, though.

On the other hand if you need to configure one of the plugins manually, then you shouldn't forget to disable the `devtool` option as well. Otherwise two plugins will be loaded and this might screw up your source maps. (This happened to me, and it took me days to figure out.)

src/content/plugins/eval-source-map-dev-tool-plugin.md
```diff
@@ -35,11 +35,26 @@ The following options are supported:
 
 T> Setting `module` and/or `columns` to `false` will yield less accurate source maps but will also improve compilation performance significantly.
 
+T> When using this plugin in [development mode](/concepts/mode/#mode-development), you must manually set `devtool: false`.
```
EugeneHlushko: This is misleading, it should say that when you want to use this plugin with custom configuration, it is not possible to do so without setting `devtool: false` in development mode. Using this plugin in development mode through shortcut should work just fine e.g. `devtool: 'eval-source-map'`

src/content/plugins/eval-source-map-dev-tool-plugin.md
```diff
@@ -35,11 +35,26 @@ The following options are supported:
...
 
 ## Examples
 
 The following examples demonstrate some common use cases for this plugin.
 
+### Basic Use Case
+
+To write out `devtool: eval-source-map` using the plugin you can use the following configuration:
```
EugeneHlushko: Can you rephrase this? I would expect this to note that in development it is necessary to set `devtool: false` in order for custom configured plugin instance to work, default use case (`devtool: 'eval-source-map'`) as mentioned above should work just fine for both development and production modes.

Issue #2311, opened 2018-06-27
Add spellcheck to tests


byzyk:
Consider adding [`github-spellcheck-cli`](https://github.com/tbroadley/github-spellcheck-cli) instead of or complimentary to `alex`.

See discussion in #2303

madhavarshney:
Based on their online demo, it seems that `alex` does not catch spelling mistakes.  `spellcheck-cli` would probably be a good tool to compliment `alex`.

byzyk:
Yeah, I see, those two serve a slightly different mission so I guess we should consider `spellcheck-cli` as an addition rather than a replacement.

Regardless, the CLI is really cool and I think we should move forward with it. Although, from what I understood while reading the docs, it works in interactive mode. Meaning the moment it finds an opportunity to fix the typo it prompts for user input which is not exactly what we need in CI. Is there a way to run CLI, not in interactive mode, so in case it catches possible typos it just exits with a non-zero code?

@tbroadley would you be able to elaborate on the above, please?

tbroadley:
Hey! https://github.com/tbroadley/spellchecker-cli would be the tool to use for spellchecking documentation in CI. It does exactly what @byzyk described: it prints out any typos and exits with a non-zero error code if any are found.

If that sounds interesting to you, I'm happy to do the work to add `spellchecker-cli` to this repo. 😄

jeremenichelli:
I'm not sure honestly. I can see the benefits of course, but also I saw this going south because this gathers a lot of technical words and names that aren't real dictionary words. It might easily become something more to maintain or be unnecessary in the way when trying to collaborate. I have mixed feelings as you might see, so I wanna hear the rest :)

byzyk:
I guess we can always wrap those too techy words in code blocks and spellchecker will skip them, right?

madhavarshney:
There's actually a custom dictionary that you can maintain or generate automatically to make spellchecker-cli skip those words.  Check [the documentation](https://github.com/tbroadley/spellchecker-cli#personal-dictionary) for more information.

EugeneHlushko:
I dont see why not, when introducing this, a dictionary should be created so that lint passes after merge to master

byzyk:
I agree. That dictionary shouldn't be hard or time-consuming to maintain.

tbroadley:
I spent a little time to add `spellchecker-cli` to this repo, here's the result: https://github.com/webpack/webpack.js.org/pull/2360 Let me know what you think!

[...]

PR #2360, opened 2018-07-17
Add spellchecker-cli

tbroadley:
See https://github.com/webpack/webpack.js.org/issues/2311.

This PR adds `spellchecker-cli` to the set of tools that are run as part of the CI process and as pre-commit hooks. It also introduces a dictionary of whitelisted words for `spellchecker-cli` to use and fixes the actual mistakes that were found.


tbroadley:
CI seems to have failed because of rate-limiting from GitHub during the build step.

dictionary.txt
```diff
@@ -0,0 +1,656 @@
+250kb
```
EugeneHlushko: man this is a huge dictionary :( any way to set ignores for the cli? for example we can just ignore sections with contributors to not add all of them here in the dict

tbroadley: Yeah this is possible! The tool allows for programmatic dictionaries: JavaScript files that export lists of words to be ignored. I could write a programmatic dictionary that parses the frontmatter in each Markdown file and generates a list of all contributors' names to be ignored. The end result is that new contributors wouldn't have to add their name to the dictionary - it would be automatically ignored once added to the frontmatter of any Markdown file.

tbroadley:
@montogeek Sorry, I'm not sure what API you're talking about.

I agree that it's not great to have to ask new contributors to add their name to the dictionary. See [my comment](https://github.com/webpack/webpack.js.org/pull/2360#discussion_r203037017) for a way this could be resolved.

dictionary.txt
montogeek: That would be unnecessary work, it is better to avoid adding them, avoid adding a new file and avoid just doing it.

I am talking about this API: https://github.com/tbroadley/spellchecker-cli/blob/master/index.js#L43, your package doesn't export it, but it is there.

As you said, parse file's frontmatter, and only run the spellchecker in the title and content, of course, a dictionary may be necessary for all the tech/webpack related terms, but it would be significantly smaller and we won't have to update on almost every PR.

tbroadley: That makes sense! I'd be happy to add functionality to the tool to allow spell-checking only specific tags in the frontmatter.

Would you mind explaining why you think it would be better to use the internal API directly, rather than the CLI wrapping it?

montogeek: As my previous comment:
>  only run the spellchecker in the title and content

CLI is checking the entire file.

PR #2363, opened 2018-07-18
guide example config - add missing require('path')

nickhalvorsen:
environment-variables.md - added missing `require('path')`

config was not usable as given without this line.


montogeek:
Thanks!

PR #2367, opened 2018-07-18
infra(site) Avoid fetching packages in PRs

montogeek:
Currently we are fetching all webpack-contrib packages information in every PR to build our docs, but we don't really need to do it since we don't check for its content, we also don't have E2E test. This is a waste of time and resources, it is also consuming our GitHub API Limit rather quickly.

This PR removed fetching packages in PRs, this will reduce broken PRs builds and also make our build run faster.


montogeek:
This change saves us ~14 minutes in our builds.

montogeek:
Also, this will likely change once `rebuild` is merged.

PR #2368, opened 2018-07-19
docs(plugins) AutomaticPrefetchPlugin

### LINKED ISSUE #2267: AutomaticPrefetchPlugin not documented
I cannot see the AutomaticPrefetchPlugin under [plugins](https://webpack.js.org/plugins/prefetch-plugin/).
If it is very similar to the `PrefetchPlugin` we can add it there, create a separate menu entry and deep link it...

EugeneHlushko:
Fixes #2267


src/content/plugins/automatic-prefetch-plugin.md
```diff
@@ -0,0 +1,21 @@
+---
+title: AutomaticPrefetchPlugin
+contributors:
+  - sokra
+  - EugeneHlushko
+---
+
+The `AutomaticPrefetchPlugin` discovers __all modules__ from the previous compilation earlier when watching for changes to try and improve the incremental build times. Compared to [`PrefetchPlugin`](/plugins/prefetch-plugin/) which discovers a __single module__ earlier.
```
Legends: So it discovers all modules from the previous compilation  "upfront" ("earlier") ?
Perhaps "upfront" is more correct here?

And rephrasing the following to:  

>  ...while watching for changes, trying to improve the ...

What do you think?

EugeneHlushko: yeap earlier was a bit confusing. and rephrasing suggestion also lgtm. will update

montogeek:
Thanks!

PR #2369, opened 2018-07-19
infra(site) Fix sh unexpected token

montogeek:


montogeek:
It uses the same conditional from https://github.com/webpack/webpack.js.org/blob/master/src/scripts/deploy.sh#L8

PR #2372, opened 2018-07-19
infra(add) lint-staged

byzyk:
/cc @montogeek


package.json
```diff
@@ -41,9 +41,12 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm test"
+      "pre-commit": "lint-staged"
     }
   },
+  "lint-staged": {
+    "*.{js,jsx,md}": ["npm test"]
```
byzyk: Although I'm not sure if we should run tests here as well. Perhaps just linting would be enough?

montogeek: What does `test` run? :D

byzyk: LOL 😄 I guess I was thinking about core repo while writing this

montogeek: We should separate this config, we already have a lint command for each file type.
It doesn't make sense to lint JS if you only updated docs.

byzyk: Sounds wise, will update today

byzyk:
Hey @montogeek! Please have a look if it's ok to go now? I've enabled linter only for JS and MD files.

montogeek:
Thanks!

PR #2374, opened 2018-07-20
docs(api) correct get parser example

DeTeam:
Fix example in parser api section (it's a hook map, so `for` is needed).


DeTeam:
Rebased. Would appreciate a review & merge.

montogeek:
Thanks!

PR #2377, opened 2018-07-22
update Starter Kits

king-prawns:
Update starter kits guide fetched from @ahfarmer's [javascriptstuff-db package](https://github.com/ahfarmer/javascriptstuff-db).
Now the list will be consistent with that present in this file: data/source/react-starter-projects.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,13 +105,13 @@
   },
   "dependencies": {
     "ajv": "^5.5.2",
+    "javascriptstuff-db": "^1.12.0",
     "lodash.throttle": "^4.1.1",
     "prop-types": "^15.5.10",
     "react": "^15.6.1",
     "react-dom": "^15.6.1",
     "react-router": "^4.2.0",
     "react-router-dom": "^4.2.2",
-    "tool-list": "^0.11.0",
     "webpack.vote": "^0.1.2",
     "whatwg-fetch": "^2.0.3"
   }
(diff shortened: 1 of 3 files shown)


montogeek:
What about current tool-list packages?

package.json
```diff
@@ -105,13 +105,13 @@
   },
   "dependencies": {
     "ajv": "^5.5.2",
+    "javascriptstuff-db": "^1.12.0",
```
montogeek: This package author deprecated `tool-list`?

king-prawns: Current tool-list package fetch data from [this branch](https://github.com/ahfarmer/javascriptstuff-db/tree/v0).
So, at the moment [this](https://webpack.js.org/starter-kits) instructions not work properly because you refer to a file present in master branch.

montogeek: I see the page and links work too?
What is not working?

king-prawns: According with [this documentation](https://webpack.js.org/starter-kits) I need to submit a PR against this file:data/source/react-starter-projects.js. (master branch)
But the `tool-list` package fetch data from [another branch](https://github.com/ahfarmer/javascriptstuff-db/tree/v0) (branch v0).
So, there are 2 options:
- use `tool-list` package and update the description in webpack starters kit guide.
- use `javascriptstuff-db` package that fetch data from master branch. 

Note that the repo's list present in master branch is more updated than the other one (branch V0)

montogeek: Ok, I guess it is ok then.

montogeek:
Thanks!

PR #2384, opened 2018-07-24
fix(guides) Use proper JSON in author-libraries

tonikarppi:
Proper JSON syntax should not use single quotes, see http://www.json.org/.

Changes in this commit:
* Use json instead of javascript for codeblock language.
* Change single quotes to double quotes.
* Remove semicolon.
* Apply default formatting with prettier.


montogeek:
Thanks!

PR #2385, opened 2018-07-24
Update development.md

DaleSeo:
The `devServer` part in `webpack.config.js` that was added in the previous section titled `Using webpack-dev-server` suddenly disappeared in this section titled titled `Using webpack-dev-middleware`.
I suggesting keeping the `devServer` part in `webpack.config.js` since the `package.json` still has `"start": "webpack-dev-server --open"` so that the decument looks more consistent.


montogeek:
Thanks!

PR #2386, opened 2018-07-24
Update output-management.md (Minor)

DaleSeo:
I see the pattern of `npm install <option> <package_name>` being used across the guides, but this npm command has the option part at the end.
It would look more consistent if this line kept the same pattern although the result is the same.


montogeek:
Thanks!

PR #2388, opened 2018-07-24
fix(guides): update production guide, fix #2387

### LINKED ISSUE #2387: Documentation confusion: manually configure DefinePlugin with `mode` present or not?
As the [docs on Mode concept](https://webpack.js.org/concepts/mode/#usage) goes, when `mode` is set, webpack automatically configures some plugins like `DefinePlugin`.
image

But as the [production guide](https://webpack.js.org/guides/production/#specify-the-environment) goes, even with `mode` set to `'production'`, I should manually set `process.env.NODE_ENV` in `DefinePlugin`.
image

This really confused me. Which one should I follow?

---

This issue was moved from webpack/webpack#7779 by @montogeek. Original issue was by @SevenOutman.

SevenOutman:
Add explanation to [Specify the Environment](https://webpack.js.org/guides/production/#specify-the-environment) section in production guide, which is no longer effective with webpack v4. See advice here https://github.com/webpack/webpack.js.org/issues/2387.

P.S. I think it's better adding an instruction than removing the whole section, for there're users of previous major versions. If there's a separate branch and url for v3 docs, I'll add changes and remove the whole section in master.

### DIFF
--- a/src/content/guides/production.md
+++ b/src/content/guides/production.md
@@ -184,16 +184,13 @@ __webpack.prod.js__
 T> Avoid `inline-***` and `eval-***` use in production as they can increase bundle size and reduce the overall performance.
 
 
-## Specify the Environment
+## Specify the Mode
 
-T> Since webpack v4, you no longer need to manually specify the environment if you already have [`mode`](https://webpack.js.org/concepts/mode/) set correctly.
-
-Many libraries will key off the `process.env.NODE_ENV` variable to determine what should be included in the library. For example, when not in _production_ some libraries may add additional logging and testing to make debugging easier. However, with `process.env.NODE_ENV === 'production'` they might drop or add significant portions of code to optimize how things run for your actual users. We can use webpack's built in [`DefinePlugin`](/plugins/define-plugin) to define this variable for all our dependencies:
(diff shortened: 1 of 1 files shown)


montogeek:
It seems that we could drop the entire section.

SevenOutman:
@montogeek What about v3 users? Do we have v3 docs available for them?

EugeneHlushko:
I dont think we should remove it. It does need a big change, the title can be `Specify the mode`, and info about NODE_ENV relation to libraries is useful, just make it v4 compliant

EugeneHlushko:
Right now the v3 docs are only available locally, there is an outsdanding issue which (not without luck) can get resolved hopefully this year... Its not progressing quickly

SevenOutman:
@EugeneHlushko Agreed. An ideal solution could be removing the entire section from v4 docs with there being standalone v3 docs. Adding this instruction is just a temporary approach while v3 docs is still wip.

SevenOutman:
@montogeek Changes have been made according to https://github.com/webpack/webpack.js.org/pull/2388#pullrequestreview-139868481 and https://github.com/webpack/webpack.js.org/pull/2388#issuecomment-407393800

montogeek:
Thanks!

PR #2390, opened 2018-07-24
fix(guides) Update generated filename in PWA

tonikarppi:
The plugin `workbox-webpack-plugin` generates a service worker with name `service-worker.js` (see https://github.com/GoogleChrome/workbox/blob/master/packages/workbox-webpack-plugin/src/generate-sw.js#L54).
Replaced all instances of `sw.js` with `service-worker.js`.


montogeek:
Thanks!

PR #2392, opened 2018-07-25
docs(config) document optimization.sideEffects

EugeneHlushko:
as per title

Takes one off #1968

### DIFF
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -325,3 +325,22 @@ module.exports = {
   }
 };
 ```
+
+## `optimization.sideEffects`
+
+`bool`
+
+Tells webpack to recognise the sideEffects flag in `package.json` or rules to skip over modules which are flagged to contain no side effects when exports are not used. `optimization.sideEffects` depends on [`optimization.providedExports`](#optimization-providedexports) and [`optimization.usedExports`](#optimization-usedexports). These dependencies have a build time cost, but eliminating modules has positive impact on performance because of less code generation. Effect of this optimization depends on your codebase, try it for possible performance wins.
+
+By default `optimization.sideEffects` is enabled in `production` [mode](/concepts/mode/) and disabled elsewise. 
+
+__webpack.config.js__
+
+```js
+module.exports = {
+  //...
+  optimization: {
+    sideEffects: true
+  }
+};
+```
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -342,7 +342,7 @@ __package.json__
 }
 ```
 
-T> Please note that `sideEffects` should be in the big module's `package.json` file and doesn't mean that you need to set `sideEffects` to `false` in your own project's `package.json` which requires that big module.
+T> Please note that `sideEffects` should be in the npm module's `package.json` file and doesn't mean that you need to set `sideEffects` to `false` in your own project's `package.json` which requires that big module.
 
 `optimization.sideEffects` depends on [`optimization.providedExports`](#optimization-providedexports) to be enabled. This dependency has a build time cost, but eliminating modules has positive impact on performance because of less code generation. Effect of this optimization depends on your codebase, try it for possible performance wins.
 
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -344,7 +344,7 @@ __package.json__
 
 T> Please note that `sideEffects` should be in the big module's `package.json` file and doesn't mean that you need to set `sideEffects` to `false` in your own project's `package.json` which requires that big module.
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/optimization.md
```diff
@@ -324,4 +324,23 @@ module.exports = {
...
 };
-```
\ No newline at end of file
+```
+
+## `optimization.sideEffects`
+
+`bool`
+
+Tells webpack to recognise the sideEffects flag in `package.json` or rules to skip over modules which are flagged to contain no side effects when exports are not used. `optimization.sideEffects` depends on [`optimization.providedExports`](#optimization-providedexports) and [`optimization.usedExports`](#optimization-usedexports). These dependencies have a build time cost, but eliminating modules has positive impact on performance because of less code generation. Effect of this optimization depends on your codebase, try it for possible performance wins.
```
montogeek: > flags or rules in `package.json`

src/content/configuration/optimization.md
```diff
@@ -324,4 +324,23 @@ module.exports = {
...
+
+By default `optimization.sideEffects` is enabled in `production` [mode](/concepts/mode/) and disabled elsewise. 
+
+__webpack.config.js__
```
montogeek: Please add an example with `package.json`

src/content/configuration/optimization.md
EugeneHlushko: The rules maybe relate to the `module` rules?

src/content/configuration/optimization.md
EugeneHlushko: yup makes sense

src/content/configuration/optimization.md
EugeneHlushko: CC @ooflorent

src/content/configuration/optimization.md
```diff
@@ -324,4 +324,37 @@ module.exports = {
...
 };
-```
\ No newline at end of file
+```
+
+## `optimization.sideEffects`
+
+`bool`
+
+Tells webpack to recognise the [sideEffects](https://github.com/webpack/webpack/blob/75c6416cfe95217b5ce60d02ba5d24d88c50f538/examples/side-effects/README.md) flag in `package.json` or rules to skip over modules which are flagged to contain no side effects when exports are not used.
```
EugeneHlushko: not sure whether i want to link master or target current version via commit reference :S

ooflorent:
```diff
- sideEffects
+ `sideEffects`

File: src/content/configuration/optimization.md [hunk 4]
```diff
@@ -324,4 +324,37 @@ module.exports = {
...
+{
+  "name": "awesome npm module",
+  "version": "1.0.0",
+  "sideEffects": false
+}
+```
+
+T> Please note that `sideEffects` should be in the big module's `package.json` file and doesn't mean that you need to set `sideEffects` to `false` in your own project's `package.json` which requires that big module.
+
+`optimization.sideEffects` depends on [`optimization.providedExports`](#optimization-providedexports) and [`optimization.usedExports`](#optimization-usedexports). These dependencies have a build time cost, but eliminating modules has positive impact on performance because of less code generation. Effect of this optimization depends on your codebase, try it for possible performance wins.
```
ooflorent: I don't think it depends on these optimizations

sokra: The issue it that static knowledge of `export *` depends on these optimizations.

https://github.com/webpack/webpack/blob/a28f44f613276446fb764dec7fab38b7cff8a07c/lib/optimize/SideEffectsFlagPlugin.js#L72 depends on `safe-reexport`

https://github.com/webpack/webpack/blob/a28f44f613276446fb764dec7fab38b7cff8a07c/lib/dependencies/HarmonyExportImportedSpecifierDependency.js#L157 `export *` is only a `safe-reexport` when at least `providedExports` is enabled.

But you are probably right `usedExports` isn't really needed.

EugeneHlushko:
@montogeek i think its good for final look now

src/content/configuration/optimization.md
```diff
@@ -324,4 +324,37 @@ module.exports = {
...
+
+``` json
+{
+  "name": "awesome npm module",
+  "version": "1.0.0",
+  "sideEffects": false
+}
+```
+
+T> Please note that `sideEffects` should be in the big module's `package.json` file and doesn't mean that you need to set `sideEffects` to `false` in your own project's `package.json` which requires that big module.
```
montogeek: What is big module? Do you mean: main app?

EugeneHlushko: it means an npm module required by your project, follow the `sideEffects` url for more context

EugeneHlushko: maybe change it to:

Please note that `sideEffects` should be in the npm module's `package.json` file

montogeek: Yeah, that sounds better

EugeneHlushko: @montogeek done

Legends: Perhaps a link to each [library devs](https://github.com/webpack/webpack/tree/master/examples/side-effects) and [app devs](https://webpack.js.org/guides/tree-shaking/) could be added?

montogeek:
Thanks!

PR #2395, opened 2018-07-25
Update tree-shaking.md

aytee17:
Sentence clarification.

### DIFF
--- a/src/content/guides/tree-shaking.md
+++ b/src/content/guides/tree-shaking.md
@@ -60,7 +60,7 @@ export function cube(x) {
 }
 ```
 
-You will need to set the development mode to make sure that the bundle is not minified:
+Set the ```mode``` configuration option to [development](https://webpack.js.org/concepts/mode/#mode-development) to make sure that the bundle is not minified:
 
 __webpack.config.js__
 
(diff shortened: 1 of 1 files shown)


src/content/guides/tree-shaking.md
```diff
@@ -60,7 +60,7 @@ export function cube(x) {
 }
 ```
 
-You could need to set the development mode to be sure that bundle is not minified:
+You will need to set the development mode to make sure that the bundle is not minified:
```
EugeneHlushko: make `development mode` a link to mode docs

src/content/guides/tree-shaking.md
```diff
@@ -60,7 +60,7 @@ export function cube(x) {
 }
 ```
 
-You could need to set the development mode to be sure that bundle is not minified:
+Set the ```mode``` configuration option to [development](https://webpack.js.org/concepts/mode/#mode-development) to make sure that the bundle is not minified:
```
EugeneHlushko: A really minor change request here as we don't inline 3x ``` but instead use it for code blocks, single is enough for inline: `

montogeek:
Thanks!

PR #2399, opened 2018-07-26
added sass loader

debs-obrien:
added example of sass loader and explanation of why it goes at the end in order to work

### DIFF
--- a/src/content/concepts/loaders.md
+++ b/src/content/concepts/loaders.md
@@ -52,7 +52,7 @@ There are three ways to use loaders in your application:
 ### Configuration
 
 [`module.rules`](/configuration/module/#module-rules) allows you to specify several loaders within your webpack configuration.
-This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader:
+This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader. Loaders work in a backwards order, from the last one to the first one. The sass loader will work first compiling your sass to css and then it will call the css loader and then the style loader. In order for the sass loader to work it needs to be at the end as sass will be the first thing that gets compiled. If you are not working with sass then you won't need the sass-loader.
 
 ```js-with-links-with-details
 module.exports = {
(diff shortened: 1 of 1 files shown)


src/content/concepts/loaders.md
```diff
@@ -52,7 +52,7 @@ There are three ways to use loaders in your application:
 ### Configuration
 
 [`module.rules`](/configuration/module/#module-rules) allows you to specify several loaders within your webpack configuration.
-This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader:
+This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader. Loaders work in a backwards order, from the last one to the first one. The sass loader will work first compiling your sass to css and then it will call the css loader and then the style loader. In order for the sass loader to work it needs to be at the end as sass will be the first thing that gets compiled. If you are not working with sass then you won't need the sass-loader.
```
jeremenichelli: I think the explanation is valuable though it might need some tweaking. 

First, when you say _helps maintain clean code_ we need to specify why it's that, if not I would remove this part.

Before _The sass loader_ I would emphasize that this is how this would work in the code snippet below, like: _In the example above the sass loader..._.

Also, please leave a line break after the new paragraph so it's visible for others contributors that a new paragraph is kicking off.

src/content/concepts/loaders.md
```diff
@@ -52,7 +52,9 @@ There are three ways to use loaders in your application:
 ### Configuration
 
 [`module.rules`](/configuration/module/#module-rules) allows you to specify several loaders within your webpack configuration.
-This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader:
+This is a concise way to display loaders, and helps to maintain clean code. It also offers you a full overview of each respective loader. 
+
+In the example below note the order of the sass-loader. See ["Loader Features"](/concepts/loaders/#loader-features) for more information on order of loaders.
```
montogeek: What do you think about: `... for more information about loaders order.`?

debs-obrien: ok

montogeek:
Thanks!

PR #2394, opened 2018-07-25
infra(site) Fix project run on Windows

montogeek:
Fixes #2380

PR #2401, opened 2018-07-27
infra(site) Add cross-env to build script

montogeek:
Continuation of #2394

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "build-test": "npm run build && http-server build/",
     "fetch": "sh src/scripts/fetch.sh",
     "prebuild": "npm run clean",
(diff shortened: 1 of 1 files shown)


Legends:
@montogeek I have one question regarding the production version.

When I try to run the production build locally, it cannot resolve `css + js + svg` files (404) because the url's point to root folder and not to dist folder.

When I add the following to webpack.prod.js, it can resolve svg's but not css and js files...?
Do I have to do something else?
Btw. when using `yarn build` during build CPU goes up to 98%, it's not for long but nevertheless, it shouldn't go that high....

```
 output: {
    ...
      publicPath: "/dist/"
    }
```

image

montogeek:
Run your server from `dist`, no root.

PR #2402, opened 2018-07-27
Fix default hashFunction for HashedModuleIdsPlugin

lencioni:
According to the code, the default hashFunction is md4, not md5.

https://github.com/webpack/webpack/blob/482ff200/lib/HashedModuleIdsPlugin.js#L18


montogeek:
Thanks!

PR #2404, opened 2018-07-28
Travis optimization

Munter:
I've leveraged as much caching on Travis as possible, which have sped up yarn installs and proselint runs by a lot. 

I've switched the travis setup to run on the cheap and fast container setup, which means the test start faster.

The lint+build stage runs all linting except proselint, which runs in parallel in a separate container.

The lint+build stage saves build artefacts in a cache.

After the build has succeeded the external link check runs on the build artefacts in a separate container. This is a non-blocking test run to avoid massive blockage of PR's based on external links changing.

After the build has succeeded and if the build is of the `master` branch, a deploy will be initialized with the cached artefacts from the build in a previous container.

The fetching of the repositories for loaders and plugins, which previously caused our github api usage to exeed the rate limit has now been removed from the build pipeline. Instead it is a script you will run manually as a maintainer or contributor (`yarn update-repos`) , which will save the updated repos in version control. The build uses that list of repos to fetch the readmes without ever hitting the github api, and thus avoiding the rate limitation.

All of the above means we can increase the number of parallel builds on Travis, which are currently limited to `1`. I think 4 is the maximum we can get, which will give us significant improvement on feedback time.

Latest build of this branch: https://travis-ci.org/webpack/webpack.js.org/builds/409489808

Build times are around 5 minutes for the blocking part and 4 minutes for the non-blocking part. I've set travis up to `fast_finish, which should give a green light in Github as soon as the blocking part of the tests are finished.


montogeek:
It worked?

.travis.yml
```diff
@@ -1,15 +1,53 @@
...
+        - yarn lint:social
+
+    - stage: build
+      name: Build
+      before_script: source ./src/scripts/env.sh
+      script: yarn build
+
+    - stage: post-build
+      name: Deploy
+      if: branch = master
```
montogeek: This condition saves `deploy.sh` internal condition, right?

.travis.yml
```diff
@@ -1,15 +1,53 @@
...
+      script: bash ./src/scripts/deploy.sh
+
+    - stage: post-build
+      name: Link check
+      script: yarn linkcheck
+
+    - stage: post-build
```
montogeek: Why it this one post build? We shouldn't build if this command fails.

.travis.yml
Munter: I'll clean that up in a bit

.travis.yml
Munter: We can re-order as we see fit later. Right now I just need to get the pieces working, then we can discuss what the pipeline should do.

.travis.yml
Munter: Done

Munter:
@montogeek Ok, Now I think I'm ready to move stuff around in the pipeline. Should proselint be a blocker for a deployment? Should it run before or after a build, which has an influence on if we check fetched content or not?

src/utilities/fetch-package-repos.js
```diff
@@ -0,0 +1,77 @@
...
+  while (api.hasNextPage(response)) {
+    response = await api.getNextPage(response);
+    data = data.concat(response.data);
+  }
+
+  return data;
+}
+
+async function main() {
+  if (process.env.GITHUB_TOKEN) {
```
montogeek: I am not sure about leaving this here, this solution doesn't work because Travis doesn't export ENV variables to PRs from other repos (security measure), and deploying master wouldn't consume API limit.

repositories/loaders.json
```diff
@@ -0,0 +1,45 @@
+[
```
montogeek: Separation is well received, but we can also separate them using an object. What do you think?

.travis.yml
```diff
@@ -51,6 +51,7 @@ jobs:
       cache:
         pip: true
         directories:
+          - dist
```
montogeek: This should be revisited. At this point it was already deployed, right?

src/utilities/fetch-package-repos.js
Munter: I'll remove that bit

Munter: This specific script is not supposed to be run on travis anyway, so it doesn't really matter

.travis.yml
Munter: Yeah, I also forgot that the dist folder is actually not even checked with proselint at all. Just `src/content`. So this is just useless

repositories/loaders.json
Munter: That would make sense as well. I'd rather spend my time doing whatever needs to be done with the order of the steps. Where should proselint be run?

montogeek: Before building and deploying

Munter:
This branch is ready to go!

I've leveraged as much caching on Travis as possible, which have sped up yarn installs and proselint runs by a lot. 

I've switched the travis setup to run on the cheap and fast container setup, which means the test start faster.

The lint+build stage runs all linting except proselint, which runs in parallel in a separate container.

The lint+build stage saves build artefacts in a cache.

After the build has succeeded the external link check runs on the build artefacts in a separate container. This is a non-blocking test run to avoid massive blockage of PR's based on external links changing.

After the build has succeeded and if the build is of the `master` branch, a deploy will be initialized with the cached artefacts from the build in a previous container.

The fetching of the repositories for loaders and plugins, which previously caused our github api usage to exeed the rate limit has now been removed from the build pipeline. Instead it is a script you will run manually as a maintainer or contributor (`yarn update-repos`) , which will save the updated repos in version control. The build uses that list of repos to fetch the readmes without ever hitting the github api, and thus avoiding the rate limitation.

All of the above means we can increase the number of parallel builds on Travis, which are currently limited to `1`. I think 4 is the maximum we can get, which will give us significant improvement on feedback time.

Latest build of this branch: https://travis-ci.org/webpack/webpack.js.org/builds/409489808

Build times are around 5 minutes for the blocking part and 4 minutes for the non-blocking part. I've set travis up to `fast_finish, which should give a green light in Github as soon as the blocking part of the tests are finished.

Munter:
Not that this PR merges to `rebuild-hyperlink`, since the changes there are prerequisites

.travis.yml
```diff
@@ -1,15 +1,52 @@
...
+
+jobs:
+  fast_finish: true
+  allow_failures:
+    - name: External Link Check
+
+  include:
+    - stage: Build
+      name: Lint and Build
+      before_install: npm install --global yarn
```
byzyk: Travis will automatically use `yarn` once it sees `yarn.lock` file, we don't need to install it explicitly.

package.json
```diff
@@ -26,18 +26,24 @@
...
+    "fetch:supporters": "node src/utilities/fetch-supporters.js",
+    "fetch:starter-kits": "node src/utilities/fetch-starter-kits.js",
     "prebuild": "npm run clean",
-    "build": "npm run fetch && npm run content && NODE_ENV=production webpack --config webpack.prod.js && npm run sitemap && echo webpack.js.org > dist/CNAME",
+    "build": "run-s fetch content && NODE_ENV=production webpack --config webpack.prod.js",
+    "postbuild": "npm run sitemap",
     "test": "npm run lint",
     "lint": "run-s lint:*",
-    "lint:js": "eslint src --ext .js,.jsx,.md",
+    "lint:js": "eslint src --ext .js,.jsx,.md --cache true --cache-location .cache/.eslintcache",
```
byzyk: Nice one! 👍🏽

src/utilities/fetch-package-readmes.js
```diff
@@ -0,0 +1,66 @@
...
+        repo: htmlUrl
+      });
+
+      request(url)
+        .then(async content => {
+          const body = processReadme(content, { source: url });
+
+          await writeFile(fileName, headmatter + body);
+
+          console.log('Generated:', path.relative(cwd, fileName));
```
byzyk: Don't mean to sound too picky but perhaps we could add space after `Generated: ` here?

src/utilities/fetch-package-repos.js
```diff
@@ -0,0 +1,69 @@
...
+        .filter(name => suffixes.some(suffix => name.endsWith(suffix)));
+    }));
+
+    const json = JSON.stringify(_.flatten(result), undefined, 2);
+    const jsonPath = path.resolve(__dirname, `../../repositories/${type}.json`);
+
+    fs.writeFile(jsonPath, json, (err) => {
+      if (err) {
+        throw err;
+      }
```
byzyk: Do you think it will make sense to add some basic logging here, the same as one in `fetch-package-readmes` script?

byzyk:
Awesome work @Munter, can't wait to try this out! I like the idea of not having to fetch GH repos every 5 minutes and it's gonna solve the API limit issue finally 🎉

src/utilities/fetch-package-readmes.js
Munter: The console logging adds a space between arguments :)

src/utilities/fetch-package-repos.js
Munter: That would indeed make sense. I'd prefer if someone else adds that. I'm about to be offline for a week

montogeek:
@jeremenichelli @EugeneHlushko Any feedback? I would like to merge this one.

montogeek:
@Munter There is a cron job that deploys every day, How would this affect it?

Munter:
@montogeek since I kept the script names intact I think there will be no impact on any cron job

montogeek:
@Munter Yep

montogeek:
I will merge this one later today

PR #2408, opened 2018-07-29
Fix issue: Sub-Navigation disappears when sub-menu entry "Migrate" is clicked

Legends:
"migrate" added to:
` isActive: url => /^\/(api|concepts|configuration|guides|loaders|migrate|plugins)/.test(url),`

### DIFF
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -52,7 +52,7 @@ class Site extends React.Component {
             {
               content: 'Documentation',
               url: '/concepts',
-              isActive: url => /^\/(api|concepts|configuration|guides|loaders|plugins)/.test(url),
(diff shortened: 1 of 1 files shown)


montogeek:
Uhmm, not sure why it was never added. Maybe an oversight or by decision?

montogeek:
It doesn't happen in the website, right?

Legends:
No, this issue is **only** related to branch `rebuild`. By decision wouldn't make much sense to me :-), I guess it's a mistake.

Legends:
The only issue with "Migrate" page is that it has not enough content initially in order to have a scrollbar.
So when you click through the sub-navigation, you will see that the whole page moves left<-->right, when entering/leaving Migrate entry.

montogeek:
Yes, it is normal because of the scrollbar.

montogeek:
Thanks!

PR #2412, opened 2018-07-31
Update output.md to match the default options

petetnt:
As seen in the `WebpackOptionsDefaulter`, the `output.path` value defaults to `path.join(process.cwd(), "dist")` (and `process.cwd()` and `""`, before that 🤔).

https://github.com/webpack/webpack/blob/3366421f1784c449f415cda5930a8e445086f688/lib/WebpackOptionsDefaulter.js#L152


montogeek:
Thanks!

PR #2415, opened 2018-08-01
Add caveat for using http-proxy-middleware options

naomiajacobs:
When trying to use the `router` feature in [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) in our `webpack-dev-server` config, I was having trouble and couldn't find any documentation as to why it hadn't worked. I ended up looking in the source code and realized that I hadn't been including a `target` key. It hadn't been clear that that was required, since the `router` function overrides the `target` key. I figured others might run into the same issue and wanted to clarify.


montogeek:
Thanks!

PR #2418, opened 2018-08-02

aviyacohen:
Only added a parenthesis tip that will hopefully help future readers getting stuck on the same part I did. The tutorial updates the button.onclick printMe() only a couple of steps after enabling HMR and a focused user might think the code is not working when the output from the button click to the console.log does not update with the HMR changes.

### DIFF
--- a/src/content/guides/hot-module-replacement.md
+++ b/src/content/guides/hot-module-replacement.md
@@ -16,6 +16,8 @@ contributors:
   - caryli
   - xgirma
   - EugeneHlushko
+  - aviyacohen
+  
 related:
   - title: Concepts - Hot Module Replacement
     url: /concepts/hot-module-replacement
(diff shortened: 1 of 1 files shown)


src/content/guides/hot-module-replacement.md
```diff
@@ -103,7 +105,7 @@ __index.js__
 + }
 ```
 
-Start changing the `console.log` statement in `print.js`, and you should see the following output in the browser console.
+Start changing the `console.log` statement in `print.js`, and you should see the following output in the browser console (don't worry about that `button.onclick = printMe()` output for now, we also update that part later).
```
montogeek: `we will also...` should be instead.

EugeneHlushko:
I am not convinced that this is required, the guide is well structured and covers that part later on. Maybe someone benefits from these kind of hints, but only because they stop reading half way through the guide? I don't say i am against this addition, just would love to hear few more thumbs up for the addition before approving it

montogeek:
Thanks!

PR #2423, opened 2018-08-06
aliasFields is a string, not array

liamzebedee:
Fixed error in docs.


montogeek:
It is an array: https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L329
Why do you think it is a string?

liamzebedee:
Hmm, I was erroring out on the latest version of webpack-cli (3.1.0) with this. Sorry if I'm incorrect. In either case, the docs still confuse the reader - it says type "string" above an example which is an array.

montogeek:
Current webpack docs are for v4.x, please update the description to `array`.

EugeneHlushko:
Thanks!

PR #2426, opened 2018-08-07
docs(guides) update tree shaking guide, make syntax consistent

EugeneHlushko:
Update tree shaking guide
Update code to have consistent single quote in js examples

### DIFF
--- a/src/content/guides/tree-shaking.md
+++ b/src/content/guides/tree-shaking.md
@@ -13,6 +13,7 @@ contributors:
   - lumo10
   - byzyk
   - pnevares
+  - EugeneHlushko
 related:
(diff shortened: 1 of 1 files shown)


src/content/guides/tree-shaking.md
```diff
@@ -74,7 +75,7 @@ module.exports = {
     path: path.resolve(__dirname, 'dist')
 - }
 + },
-+ mode: "development"
++ mode: 'development'
```
byzyk: how come the linter didn't throw here?

byzyk: ah I see its `diff`, never mind

montogeek:
Thanks!

PR #2432, opened 2018-08-08
not necessary to use full URL for HMR

indefinitelee:
at least I don't think it is necessary.


montogeek:
Why do you think it is not necessary?

indefinitelee:
what i mean is that a full url is not necessary for HMR. 
since the default value of publicPath is "/"  and the previous section of the docs give this example 
```
module.exports = {
  //...
  devServer: {
    publicPath: '/assets/'
  }
}
```
it seems clear that a full URL is not necessary for HMR.

indefinitelee:
checking it. I know it's a small change but I think the text was trying to say that it is required to use the publicPath key in order to use HMR, not that publicPath must be a full URL. 

Or let me know if I'm wrong, I'll have learned something.

montogeek:
Thanks!

PR #2434, opened 2018-08-09
minimizer only accepts an array

mattkrick:
now complies to API


montogeek:
It also accepts an plugin function or object: https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L1635

montogeek:
Thanks!

PR #2448, opened 2018-08-15
added name to the contribution

debs-obrien:
sorry just a small thing but only just learnt how to add my name to the contribution that I made 2 weeks ago. hope thats ok to add it in here, if not no worries


montogeek:
Thanks!

PR #2451, opened 2018-08-17
fix(infra) Add return to actually replace readme links

montogeek:
We forgot to return :D
Fixes https://github.com/webpack/webpack.js.org/pull/2382/files#r210910668

### DIFF
--- a/src/utilities/process-readme.js
+++ b/src/utilities/process-readme.js
@@ -18,7 +18,7 @@ module.exports = function processREADME(body, options = {}) {
         href = href.replace('//www.npmjs.com');
       }
 
-      `[${content}](${url.resolve(options.source, href)})`;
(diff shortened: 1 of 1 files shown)


Munter:
So there are 3 errors in the link checks:

- json-loader is linking to `/guides/migrating/#json-loader-is-not-required-anymore`, which should be `migrate/3#json-loader-is-not-required-anymore`. I can't make a PR to update that README becasue the repo is archived

The two remaining ones might be false negatives. I'll have to look into that

Munter:
Yup. Last two errors of content-type-mismatch are false negatives. Fixed in the latest version of assetgraph, which hyperlink is not yet upgraded to because of a different regression in upgrading

montogeek:
We could setup a Redirect, it would solve the issue

PR #2453, opened 2018-08-17
Fix regressions from #2451

Munter:
Marked known false negatives from hyperlink as TODO and added a url rewrite mechanism for old urls while fetching README's

This should put the `rebuild` branch in the green again

### DIFF
--- a/src/utilities/process-readme.js
+++ b/src/utilities/process-readme.js
@@ -18,7 +18,7 @@ module.exports = function processREADME(body, options = {}) {
         href = href.replace('//www.npmjs.com');
       }
 
-      `[${content}](${url.resolve(options.source, href)})`;
+      return `[${content}](${url.resolve(options.source, href)})`;
     })
     // Modify links to keep them within the site
     .replace(/https?:\/\/github.com\/(webpack|webpack-contrib)\/([-A-za-z0-9]+-loader\/?)([)"])/g, '/loaders/$2/$3')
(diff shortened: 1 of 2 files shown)


src/utilities/process-readme.js
```diff
@@ -1,5 +1,11 @@
 const url = require('url');
 
+const beginsWithDocsDomainRegex = /^https?:\/\/webpack\.js\.org/;
+
+const redirects = {
```
montogeek: Uhmm, I was thinking of using webpack redirects instead of this string replacement
https://github.com/webpack/webpack.js.org/blob/rebuild/webpack.prod.js#L60

Munter: Those would not be evaluated when doing file based link checking I think?

montogeek: My idea is to put setup the redirect, that plugin creates meta tags with the redirect ()https://github.com/bebraw/redirect-webpack-plugin/blob/master/index.js#L15), so when hyperlink hit that link it would be redirected.

Munter: That might work. I'm pretty sure hyperlink models meta redirects. But it might show the redirect as an error and indicate that you should update the source url to the redirect target. We can probably mark that check as `--todo` or `--skip` though

Munter: image

image

montogeek: No PRs, just ask the maintainer to push a commit updating that README.

Munter: Project README updated. I'll remove the rewriting code. Though some of the refactorings will stay, since I discovered that fragment links had the original source prepended, so some of it actually fixes a bug

Munter:
I'm merging. Can't have the rebuild branch keep failing

PR #2459, opened 2018-08-22
docs(development): Add warning about dev server

GAumala:
In the Development guide, add a warning about the dev server not writing any output files.

I was recently using webpack for a new project, and had a little trouble setting up the dev server because of some misunderstandings about how it works. If the dev server says that it compiles files, I'd expect an output to be written somewhere, just like when you run webpack from the terminal. 

However this is not the case and it left me puzzled for a while. It seems [a lot of people](https://github.com/webpack/webpack-dev-server/issues/24#issuecomment-44366325) also find this not so obvious so I decided to try help a little with this PR. Let me know what you think.

### DIFF
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -8,6 +8,7 @@ contributors:
   - TheDutchCoder
   - WojciechKo
   - Calinou
+  - GAumala
 ---
 
 T> This guide extends on code examples found in the [Output Management](/guides/output-management) guide.
@@ -204,6 +205,8 @@ __webpack.config.js__
 ```
 
 This tells `webpack-dev-server` to serve the files from the `dist` directory on `localhost:8080`.
+
+W> webpack-dev-server doesn't write any output files after compiling. Instead, it keeps bundle files in memory and serves them as if they were real files mounted at the server's root path. If your page expects to find the bundle files in different path, you can change this with the [`publicPath`](/configuration/dev-server/#devserver-publicpath-) option in the dev server's configuration.
 
 Let's add a script to easily run the dev server as well:
 


montogeek:
I am not entirely sure this should be added: 
https://github.com/webpack/webpack-dev-server says in the 3rd line:
> It uses webpack-dev-middleware under the hood, which provides fast in-memory access to the webpack assets.

https://github.com/webpack/webpack-dev-middleware first item is:
> No files are written to disk, rather it handles files in memory

Also, that section already have a documentation link:
> The webpack-dev-server comes with many configurable options. Head over to the documentation to learn more.

That link should point to https://github.com/webpack/webpack-dev-server though.

EugeneHlushko:
On one hand the readme has it, on the other hand adding a note block wont hurt at least and maybe even helps % of readers. lets see what @jeremenichelli can say this is a 50/50 for me

GAumala:
Hello guys, thanks for the quick responses. My main issue is that the development guide, gives the illusion that Watch Mode, the dev server, and the dev middleware are three options that achieve the same thing and you can seamlessly switch between them.

My experience was like this. I first setup Watch Mode and it was a breeze. Everytime I made a change to my source files, a new bundle was written to the output directory. Everything just worked. But a bit later, I noticed the inconvenience of having to manually reload all the time, plus I needed to proxy some requests to my API, so I definitely needed some kind of server. So I read the guide, right after explaining Watch Mode, it goes on about the dev server. I think "this is just what I need" and read that block and update my config. But now my output directory isn't being updated and I'm confused trying to figure out what I did wrong. I did read the dev server's README, but "fast in-memory acess" does not imply that there won't be any output files written. It just means "this is super fast".

Certainly, the dev middleware's README is the only one that explains this.  But If I'm having trouble setting up the dev server I wouldn't think about reading that README unless I decide to ditch the dev server and roll out my own with the dev middleware. Besides, wouldn't you agree that having to read about the dev middleware in order to understand how the dev server works can be considered a leaky abstraction?

jeremenichelli:
I agree with adding it, the sooner a person finds out about a strange behavior the beetter for dev adoption. I would short it down a little bit though, like:

> webpack-dev-server doesn't write any output files. Instead, it keeps bundle files in memory and serves them as if they were real files mounted at the server's root path. If your page expects to find the bundle files in different path, you can change this by using [`publicPath`](../../configuration/dev-server/#devserver-publicpath-) option in the dev server's configuration.

GAumala:
Sounds good, I updated my changes.

src/content/guides/development.md
```diff
@@ -205,6 +206,8 @@ __webpack.config.js__
 
 This tells `webpack-dev-server` to serve the files from the `dist` directory on `localhost:8080`.
 
+W> webpack-dev-server doesn't write any output files after compiling. Instead, it keeps bundle files in memory and serves them as if they were real files mounted at the server's root path. If your page expects to find the bundle files in different path, you can change this with the [`publicPath`](../../configuration/dev-server/#devserver-publicpath-) option in the dev server's configuration.
```
montogeek: Please use absolute links: `/configuration/dev-server/#devserver-publicpath-`

GAumala: done! :)

montogeek:
@jeremenichelli @EugeneHlushko Please review

montogeek:
Thanks!

PR #2460, opened 2018-08-22
Fixes `console.log` typo

nodiel:
Just fixes a simple typo in `console.log` statement


montogeek:
Thanks for your contribution, unfortunately this is intended:
Following that guide, one line before your change it says:
> Now let's make sure we have something to debug, so let's create an error in our `print.js` file:

This is intended to trigger an error and show how they look.

Issue #2312, opened 2018-06-28
Add comparison with Parcel


dandv:
Just learned about Parcel from [Tensorflow](https://github.com/tensorflow/tfjs#getting-started), which mentions it before webpack.

Would be interesting to see Parcel at https://webpack.js.org/comparison/, since it claims half the bundling time of webpack at https://parceljs.org.

montogeek:
Send a PR

FadySamirSadek:
@montogeek Parcel currently supports source maps only in dev mode and in JS only not CSS. How can I mention this in the comparison and is it considered a Yes or a No in the debugging support section ?

FadySamirSadek:
Another thing what do Mangle path names and Replacement for browser mean?

PR #2465, opened 2018-08-24
Add parcel to bundlers/loaders comparison table <work in progress>

FadySamirSadek:
#2312


EugeneHlushko:
I think it makes sense to add versions (major ones) of the software used at least for webpack because the resulsts there could be for webpack v3...?

montogeek:
@FadySamirSadek Friendly ping.

PR #2470, opened 2018-08-26
fix(site): Improve loaders parsing of external link in headers

budmc29:
Improve the way external links (like licences) in headers are parsed and generated on the website.

Before:
image

After (the link will act as expected and point the user to the document etc):
image

This issue is encountered in multiple loaders pages (and possibly in other pages as well), but here's one example:
https://webpack.js.org/loaders/worker-loader/

### DIFF
--- a/src/utilities/markdown.js
+++ b/src/utilities/markdown.js
@@ -10,14 +10,18 @@ module.exports = function() {
 
   // Patch IDs (this.options.headerPrefix can be undefined!)
   renderer.heading = function(text, level, raw) {
-    var parsed = parseAnchor(raw);
-    var id = parsed.id;
+    let parsed = parseAnchor(raw);
+    let id = parsed.id;
+    let title = parsed.title;
+
+    let href = getHref(text);
+    let spanContent = href ? `<a href="${href}">${title}</a>` : title;
 
     return (
       `<h${level} class="header">
         <a class="anchor" aria-hidden="true" href="#${id}" id="${id}"></a>
-        <span class="text">${text}</span>
-        <a aria-label="${text}" class="icon-link" href="#${id}"></a>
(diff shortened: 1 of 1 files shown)


montogeek:
It is broken as well, but in a different way:
image

montogeek:
@budmc29 We are in an effort to modernize this repo infrastructure, our work is in the `rebuild` branch, Could you please fix the issue on that branch? Thanks

budmc29:
Yes, I can have a look at that, should we close this?

montogeek:
It was fixed by: https://github.com/webpack/webpack.js.org/pull/2453, I didn't have latest `rebuild` changes when taking the screenshot.

Although I am not sure if it is good to link the RAW file, it should link to the rendered markdown:
`https://github.com/webpack-contrib/worker-loader/blob/master/.github/CONTRIBUTING.md`
instead of
`https://raw.githubusercontent.com/webpack-contrib/worker-loader/master/.github/CONTRIBUTING.md`

What do you think @Munter ?

Munter:
@montogeek It does sound like a nicer experience to go the the Github UI for the markdown files. The problem is that the url we get in `options.source` is on `raw.githubusercontent.com`, and all relative links are resolved from the source URL. I don't know if there is a safe way to make that URL transformation. There's quite a few edge cases I had to cover to fix all the errors I found. And those links are mostly part of the external hyperlink check, so you'd have to manually inspect the output of that run to ensure everything still works

montogeek:
@Munter I can take a look, or do you want to do it @budmc29 ?

budmc29:
@montogeek I'd like to give it a try and see how it goes, and open another PR.

We can close this for now since it's not relevant anymore.

Thanks

PR #2471, opened 2018-08-27
Minor wording change

GregTrevellick:
Readability improvement in a sentence:

Before
   "Now we tell webpack to extract these source maps and into our final bundle:"

After
   "Now we tell webpack to extract these source maps and include in our final bundle:"


montogeek:
Thanks!

PR #2474, opened 2018-08-29
docs(configuration): reorder sections, improve wording, add references in concepts

byzyk:
* Moved `Use custom config file` section to the bottom since it's less relevant than actual options instead. 
* Added config reference in concepts
* Fixed wording in some places
* Minor adjustments to bring everything to common formatting in MD (perhaps we could add this to MD linter?)


src/content/concepts/configuration.md
```diff
@@ -45,8 +46,12 @@ module.exports = {
 };
 ```
 
+_See_: [Configuration section](/configuration) for the all supported config options
+
 ## Multiple Targets
 
+Along with exporting a single configuration object/function, webpack supports exporting multiple configurations.
```
EugeneHlushko: __webpack supports exporting multiple configurations__ - this line makes me think that webpack will export the configurations, when that is user's responsibility to write a configuration module with export statement.

suggest rewording into: webpack can accept multiple configurations

src/content/configuration/index.md
```diff
@@ -396,9 +383,23 @@ module.exports = {
 }
 ```
 
-T> Want to rapidly generate webpack configuration file for your project requirements with few clicks away.
+## Use custom config file
+
+If for some reason you want to use custom config file depending on certain situations you can change this via command line by using the `--config` flag.
```
EugeneHlushko: i understand why title was shortened, but in the text body suggest to change `config file` -> `configuration file`

src/content/configuration/index.md
```diff
@@ -396,9 +383,23 @@ module.exports = {
...
+
+__package.json__
+
+```json
+"scripts": {
+  "build": "webpack --config prod.config.js"
+}
+```
+
+## Config generators
```
EugeneHlushko: maybe `config` -> `configuration` in this title sits well in left hand side menu?

src/content/concepts/configuration.md
byzyk: Fair enough 👍🏽

src/content/configuration/index.md
byzyk: That's the thing, we've got both `config` and `configuration` all over the docs and sometimes it's hard to justify why one was picked over another one. I don't wanna seem too picky here but perhaps this can be brought to some common option? Also not sure if that's worth it

montogeek: @byzyk Yep, we have both, but we can start with this opportunity to change it to Configuration

src/content/concepts/configuration.md
```diff
@@ -4,25 +4,26 @@ sort: 6
...
 You may have noticed that few webpack configurations look exactly alike. This is because **webpack's configuration file is a JavaScript file that exports an object.** This object is then processed by webpack based upon its defined properties.
 
 Because it's a standard Node.js CommonJS module, you **can do the following**:
 
-* import other files via `require(...)`
-* use utilities on npm via `require(...)`
-* use JavaScript control flow expressions i. e. the `?:` operator
-* use constants or variables for often used values
-* write and execute functions to generate a part of the configuration
+- import other files via `require(...)`
```
montogeek: Please use title case in lists, same as #L24

byzyk: oops, seems like prettier formatted it automatically. will update tom

byzyk:
hey @montogeek, this one is ready to go

byzyk:
Hey @montogeek, I've resolved conflicts yesterday and now it seems ok. Do you have anything else on your mind to add here before we proceed?

src/content/configuration/index.md
```diff
@@ -23,25 +24,11 @@ All the available configuration options are specified below.
...
-}
-```
-
-
 ## Options
 
 Click on the name of each option in the configuration code below to jump to the detailed documentation. Also note that the items with arrows can be expanded to show more examples and, in some cases, more advanced configuration.
 
-W> Notice that throughout the configuration we use Node's built-in [path module](https://nodejs.org/api/path.html) and prefix it with the [__dirname](https://nodejs.org/docs/latest/api/globals.html#globals_dirname) global. This prevents file path issues between operating systems and allows relative paths to work as expected. See [this section](https://nodejs.org/api/path.html#path_windows_vs_posix) for more info on POSIX vs. Windows paths.
+W> Notice that throughout the configuration we use Node's built-in [path module](https://nodejs.org/api/path.html) and prefix it with the [\_\_dirname](https://nodejs.org/docs/latest/api/globals.html#globals_dirname) global. This prevents file path issues between operating systems and allows relative paths to work as expected. See [this section](https://nodejs.org/api/path.html#path_windows_vs_posix) for more info on POSIX vs. Windows paths.
```
EugeneHlushko: why do we need that change ? `[__dirname] => [\_\_dirname]`

byzyk: I guess prettier changed it accidentally, will revert later today

EugeneHlushko: ok, i think its good to go, fernando doesnt have time to review this so ping me

byzyk:
yo @EugeneHlushko, please take a look 🙂

src/content/concepts/configuration.md
```diff
@@ -46,8 +47,12 @@ module.exports = {
 };
 ```
 
+_See_: [Configuration section](/configuration) for the all supported configuration options
```
EugeneHlushko: can you please change to `/configuration/` so it also works on dev ? dev server doesn't work for non trailing slash urls

src/content/configuration/index.md
```diff
@@ -62,7 +49,7 @@ module.exports = {
     b: ["./app/entry-b1", "./app/entry-b2"]
   },
   </details>
-  // defaults to ./src
+  // defaults to "./src"
```
EugeneHlushko: single quotes

src/content/concepts/configuration.md
```diff
@@ -5,11 +5,12 @@ contributors:
   - TheLarkInn
   - simon04
   - EugeneHlushko
+  - byzyk
 ---
 
-You may have noticed that few webpack configurations look exactly alike. This is because **webpack's configuration file is a JavaScript file that exports an object.** This object is then processed by webpack based upon its defined properties.
+You may have noticed that few webpack configurations look exactly alike. This is because __webpack's configuration file is a JavaScript file that exports an object.__ This object is then processed by webpack based upon its defined properties.
```
EugeneHlushko: later on we mention that configuration is an object or a function please also mention it here in the `__` section

byzyk: I don't think it's necessary to mention it as it's covered in more details [here](https://webpack.js.org/configuration/configuration-types/#exporting-multiple-configurations). You can export a function, Promise or array but in the end, it's just a plain object returned. 

Maybe we can add it as a note below, saying something like `webpack configuration is always presented as an object but can be exported as an array, promise, function etc + include the link I shared above`. What do you think?

EugeneHlushko: Maybe change it to `... that exports a webpack [configuration](/configuration/)` because `javascript file that exports ....` is not about how its presented but about what it exports explicitly

byzyk: I guess the idea here was to show that webpack config is just a plain JS object and you can mess up with it in any way you like. Changing it to `exports webpack config` might lead to confusion, especially among beginners.

byzyk: Then there is more advanced guide on how to export function or array etc. and we can add that one as a note here, this will allow us to keep the existing guide more beginner friendly and have a reference that exporting an object is not the only option. Does it make sense?

EugeneHlushko: dunno man its plain straight to me that we can export w/e from javascript webpack config file, lets hear from more folks then

CC @jeremenichelli @montogeek

EugeneHlushko: on the other hand, the text was like this before (its not like you've updated it, just the syntax around it). we can change it later if more folks agree on it

byzyk:
@EugeneHlushko please see my last commit, is that something you had on your mind?

EugeneHlushko:
yup, lets wait for tests and LGTM

EugeneHlushko:
Thanks!

PR #2475, opened 2018-08-29
Removed a duplicate sentence

MrDorak:
Removed a duplicate sentence : 

Before : 
"When script files are wrapped by an IIFE, you can safely concatenate. When script files are wrapped by an IIFE, you can safely concatenate or safely combine files without concern of scope collision."

After : 
"When script files are wrapped by an IIFE, you can safely concatenate or safely combine files without concern of scope collision."


montogeek:
Thanks!

PR #2476, opened 2018-08-30
Update production.md

simioluwatomi:
The `devtool: 'source-map'` line is later added in the Source Mapping section of the guide. The Source Mapping section also explains why that line is added to `webpack.prod.js` file and shows the diff too.


montogeek:
Thanks!

Issue #1956, opened 2018-03-26
Improvements on SplitChunks article


jeremenichelli:
Besides the fact that the document might need some gardening and organization, there are some stuff that I couldn't completely get from the codebase or the config so there are some requests and gaps that might need to be filled.

Specially these:
 - include the different things that you can pass to runtimeChunk. _It can be a boolean, 'single', 'multiple' or an object that has property name whose value can be either a string or a function of signature (entrypoint: Chunk) => string._
 - a paragraph explaining better how enforces works and combines with other options. _The enforce allows the special cache group to ignore the minSize, minChunks, maxAsyncRequests and maxInitialRequests set in the overall splitChunks config and purely use test to decide what goes into the chunk. What I'm unclear on is how that fits in when used with reuseExistingChunk._
 - cacheGroups can receive a method. What are the arguments and what value needs to be returned.
 - function signature description for `name` option

Article pull request is https://github.com/webpack/webpack.js.org/pull/1955

jeremenichelli:
As always, I can help, but need the knowledge 😄

glen-84:
> The magic value `true` automatically chooses a name based on chunks and cache group key, otherwise a string or function can be passed.

What is the signature for this function?

jeremenichelli:
@glen-84 that's a missing thing right now.

thw0rted:
Would this be a good place to request documenting how to pass a function to `splitChunks.cacheGroups[xxx].test` or should that be its own ticket?  The documentation only mentions it in passing, and I haven't been able to reverse-engineer what the callback parameters are on my own at this time.

thw0rted:
Actually I'm not 100% clear what the regex version of `test` is testing against, and I'm sure that's complicated by my build chain where I start with Typescript and wind up with `name-shorthash.ext`.  Maybe just a little more detail on `test` generally?  Also: trackback to #2103.

chrisdothtml:
Fixed by #2183? Or are there still some parts that are unclear?

@glen-84 

> What is the signature for this function?

[relevant section](https://github.com/webpack/webpack.js.org/blob/fbffefd0e50bc05c1553c897e166146199a7ad1d/src/content/plugins/split-chunks-plugin.md#splitchunksname)

@thw0rted 

> Would this be a good place to request documenting how to pass a function to `splitChunks.cacheGroups[xxx].test`

[...]

PR #2478, opened 2018-08-30
fixing invalid splitChunks example

chrisdothtml:
From [this comment](https://github.com/webpack/webpack.js.org/issues/1956#issuecomment-417346913):

> The documentation (https://webpack.js.org/plugins/split-chunks-plugin/#splitchunks-cachegroups-test) shows a code example defining splitChunks.test (which I dont believe exists?) instead of splitChunks.cacheGroups[yourGroup].test.


montogeek:
Thanks!

PR #2480, opened 2018-09-01
Update index.md

gregoryagu:
Changed the default bundle name to 'main.js' as that is what is produced if no configuration file is used.


montogeek:
Closed by inactivity.

PR #2482, opened 2018-09-02
Update minimal config example

mlsteele:
I'm not the first person to forget this import https://github.com/webpack/webpack-dev-server/issues/823 from this [page](https://webpack.js.org/configuration/dev-server/#devserver).


montogeek:
@byzyk Shouldn't ESLint caught there errors?

byzyk:
@montogeek it's disabled in the config [here](https://github.com/webpack/webpack.js.org/blob/726cf8284026b59a2ce7e81294f57f1762163f8c/.eslintrc#L36) since we're missing `require` statements in a lot of places, not only `path` but also some plugins etc.

Do you think like we should address that to make examples more complete?

PR #2488, opened 2018-09-05
preset-es2015 is deprecated, please use preset-env

antonk52:
`preset-es2015` is deprecated and should not be used. It is encouraged by babel to use `preset-env` instead.

### DIFF
--- a/src/content/configuration/configuration-languages.md
+++ b/src/content/configuration/configuration-languages.md
@@ -135,14 +135,14 @@ In the example below JSX (React JavaScript Markup) and Babel are used to create 
 First install the necessary dependencies:
 
 ``` bash
-npm install --save-dev babel-register jsxobj babel-preset-es2015
+npm install --save-dev babel-register jsxobj @babel/preset-env
 ```
 
 __.babelrc__
 
 ``` json
 {
-  "presets": [ "es2015" ]
+  "presets": [ "env" ]
 }
 ```
 
--- a/src/content/configuration/index.md
+++ b/src/content/configuration/index.md
@@ -159,7 +159,7 @@ module.exports = {
         // -loader suffix is no longer optional in webpack2 for clarity reasons
         // see [webpack 1 upgrade guide](/guides/migrating)
         [options](/configuration/module#rule-options-rule-query): {
-          presets: ["es2015"]
+          presets: ["env"]
         },
         // options for the loader
       },


src/content/configuration/configuration-languages.md
```diff
@@ -135,14 +135,14 @@ In the example below JSX (React JavaScript Markup) and Babel are used to create
 First install the necessary dependencies:
 
 ``` bash
-npm install --save-dev babel-register jsxobj babel-preset-es2015
+npm install --save-dev babel-register jsxobj @babel/preset-env
```
asapach: `babel-register` is v6 and `@babel/preset-env` is v7 - you should not mix the two

montogeek: @antonk52 Please update, thanks @asapach !

antonk52: if I understood correctly, this should be as below, is that correct?

```
npm install --save-dev @babel/core @babel/register jsxobj @babel/preset-env
```

asapach: If you're using v7, then you need to set `presets` below as `@babel/preset-env` as well: https://babeljs.io/docs/en/babel-preset-env

src/content/configuration/configuration-languages.md
```diff
@@ -135,14 +135,14 @@ In the example below JSX (React JavaScript Markup) and Babel are used to create
...
 ```
 
 __.babelrc__
 
 ``` json
 {
-  "presets": [ "es2015" ]
+  "presets": [ "env" ]
```
asapach: Has anyone actually tried this example? I don't think it will work as advertised: https://github.com/developit/jsxobj/blob/master/.babelrc

montogeek: What do you mean?

asapach: It's not enough to set the preset for `jsxobj` to work, you also need to setup `transform-react-jsx` plugin correctly. See how they set the `pragma` setting to a different function in the linked `.babelrc`?

asapach: I'm not talking about the lib, I'm talking about JSX in the config: https://github.com/webpack/webpack.js.org/blame/master/src/content/configuration/configuration-languages.md#L161-L176

montogeek: Never used it, but I think it is handle when you name your webpack config __webpack.config.babel.js__.

I will try it later to see if that guide actually works

antonk52:
The PR should be closed, since it is jsxobj which does not support newer nanel version. Sorry guys.

PR #2493, opened 2018-09-06
Add .mjs to the defaults

philipwalton:
This should be updated once https://github.com/webpack/webpack/pull/7947 is released.


montogeek:
Thanks!

PR #2498, opened 2018-09-07
Actualise NEXT branch

EugeneHlushko:
We need to update the next branch, it is still using old linting etc.


EugeneHlushko:
There is a conflict now on parser.md, i could locally rebase next into master and resolve the conflict. Thoughts @montogeek ?

PR #2500, opened 2018-09-07
Update next branch with master updates

EugeneHlushko:
A conflictless next update by doing:
1. checkout next, pull latest
2. checkout from next to next-mastersync
3. merge master into next-mastersync, resolve conflicts
4. open pull request targeting next-mastersync into next


src/content/api/parser.md
```diff
@@ -4,6 +4,7 @@ group: Plugins
 sort: 4
 contributors:
   - byzyk
+  - DeTeam
```
montogeek: This was the conflict? xD

EugeneHlushko: yeah :D

montogeek:
Uhmm, about your original comment.
I think `next` branch is working under a git-flow model, which means that all that work isn't necessary, we just need to stick to it.

PR #2504, opened 2018-09-10
docs(manifest) corrections and small update to concept manifest page

EugeneHlushko:
Small updates and corrections for manifest concepts page


montogeek:
Thanks!

PR #2505, opened 2018-09-11
Document that concatenation is enabled by default

rluba:
I recently tried to add `new webpack.optimize.ModuleConcatenationPlugin();` to my webpack build and was very surprised that it had no effect. [It took me a while](new webpack.optimize.ModuleConcatenationPlugin()
https://github.com/aurelia/webpack-plugin/issues/153#issuecomment-420376814) to realize that it is auto-enabled by default in production.

I’ve added a corresponding note to prevent others from making the same mistake.


src/content/plugins/module-concatenation-plugin.md
```diff
@@ -10,12 +10,13 @@ related:
 
 In the past, one of webpack’s trade-offs when bundling was that each module in your bundle would be wrapped in individual function closures. These wrapper functions made it slower for your JavaScript to execute in the browser. In comparison, tools like Closure Compiler and RollupJS ‘hoist’ or concatenate the scope of all your modules into one closure and allow for your code to have a faster execution time in the browser.
 
-This plugin will enable the same concatenation behavior in webpack.
+This plugin will enable the same concatenation behavior in webpack. It is [enabled by default in production](/configuration/optimization/#optimization-concatenatemodules) but you can also manually add it:
```
montogeek: Uhmm, it is better to avoid plugin duplication.

rluba: One more reason to document the default :-)
What kind of wording would you prefer?
"It is enabled by default in production but you can add it manually if you need it in other circumstances"?

montogeek: "...in other mode".

src/content/plugins/module-concatenation-plugin.md
```diff
@@ -10,12 +10,13 @@ related:
 
 In the past, one of webpack’s trade-offs when bundling was that each module in your bundle would be wrapped in individual function closures. These wrapper functions made it slower for your JavaScript to execute in the browser. In comparison, tools like Closure Compiler and RollupJS ‘hoist’ or concatenate the scope of all your modules into one closure and allow for your code to have a faster execution time in the browser.
 
-This plugin will enable the same concatenation behavior in webpack.
+This plugin will enable the same concatenation behavior in webpack. It is [enabled by default in production](/configuration/optimization/#optimization-concatenatemodules) but you can also add it manually if you need it in other modes:
```
EugeneHlushko: The thing is, linked page already says that this optimization is enabled by default in `production` mode. 
```
Depends on `optimization.providedExports` and `optimization.usedExports`. By default `optimization.concatenateModules` is enabled in production mode and disabled elsewise.
```

Also if you go to mode page: https://webpack.js.org/concepts/mode/#usage is shows you which plugins are enabled by default in production mode.

If you still feel like adding it here, please use the same phrasing as we have:

```
By default this plugin is already enabled in production [`mode`](/concepts/mode/#mode-production) and disabled elsewise.
```

montogeek:
@rluba Friendly ping

src/content/plugins/module-concatenation-plugin.md
rluba: Done

rluba:
@montogeek Thanks for the ping. I’ve updated the wording.

montogeek:
@EugeneHlushko Please take another look

EugeneHlushko:
Thanks!

PR #2506, opened 2018-09-12
It's main.js not bundle.js from previous guide

bangzek:


bangzek:
Or you could add the diff from main.js to bundle.js in the getting-started.md

PR #2509, opened 2018-09-12
Fix diff for webpack.config.js in Dynamic Imports

bangzek:


bangzek:
Sorry, miss that comma

PR #2512, opened 2018-09-13
Docs Fix: fix naming consistence of runtime.js in example outputs

ro-savage:
This just changes `manifest.js` to `runtime.js` as is used in all other examples on the caching docs page.


montogeek:
Thanks!

PR #2515, opened 2018-09-13
Update externals.md - Combining syntaxes example

zefman:
Update the external docs with an example on how to combine the different syntaxes. 

I felt it was not clear how to achieve this from the existing docs.


montogeek:
Thanks!

PR #2516, opened 2018-09-15
Update example webpack output in asset-management.md

venelinpetrov:
- the webpack version in the console output is too old (2.6.1)
- the console output is too verbose, which doesn't seem to be the default setting


venelinpetrov:
Sorry, I forgot to update the rest of the article. I followed along and updated accordingly.

src/content/guides/asset-management.md
```diff
@@ -121,20 +121,19 @@ Now run your build command:
 ``` bash
 npm run build
 
-Hash: 9a3abfc96300ef87880f
```
EugeneHlushko: Hi @venelinpetrov 
To reduce the future maintance hussle, it was decided to only touch output when its `assets` information gets outdated. Also we will not show the rest of the output, only the assets information.

Please update your PR to only leave the assets info in the output logs like this:

```
...
                               Asset      Size  Chunks                    Chunk Names
da4574bb234ddc4bb47cbe1ca4b20303.png  3.01 MiB          [emitted]  [big]
                           bundle.js  76.7 KiB       0  [emitted]         main
Entrypoint main = bundle.js
...
```

Sorry if my example is misaligned.

montogeek: Let's merge this one as it is. We need to change all examples in one PR and then force it later.

EugeneHlushko: ok

montogeek:
Thanks!

PR #2518, opened 2018-09-15
docs(config): Example 3 - custom vendor chunks using RegEx

sakhisheikh:
An example added to SplitChunksPlugin in `split-chunks-plugin.md`

Not necessary.

No breaking change.

No update required.

When we have to split chunks on vendor basis we write test function in `splitChunks.cacheGroups.test`. I have added an example which demonstrate how we can use RegEx pattern to extract custom **node_modules** in vendor chunk.

For this I have gone through `SplitChunksPlugin.js` from `webpack/lib/optimize/`. I have tried so many to times to extract only react and react-dom at the first place with exact string matching pattern. But I failed to do this as soon as I figured it out while printing `module.nameForCondition())` and got to know the how modules are added in vendor chunks with RegEx test. I added `'/' `in the end of library name just to make sure it extracts all code under hood of the library.

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -8,6 +8,7 @@ contributors:
   - EugeneHlushko
   - byzyk
   - madhavarshney
+  - sakhisheikh
 related:
   - title: webpack's automatic deduplication algorithm example
     url: https://github.com/webpack/webpack/blob/master/examples/many-pages/README.md
@@ -344,3 +345,27 @@ module.exports = {
 ```
 
 W> This might result in a large chunk containing all external packages. It is recommended to only include your core frameworks and utilities and dynamically load the rest of the dependencies.
+
+### Split Chunks: Example 3
+
+ Create a `custom vendors` chunk, which includes only those `node_modules` from the whole application which passes the RegEx Test
+ 
+ __webpack.config.js__
+
+```js
+module.exports = {
+  //...
+  optimization: {
+    splitChunks: {
+      cacheGroups: {
+        vendor: {
+          test: /[\\/]node_modules[\\/](react|react-dom)[\\/]/,
+          name: 'vendor',
+          chunks: 'all',
+        }
+      }
+    }
+  }
+};
+```
+ T> This will result into a vendor chunk containing desired libraries provided in RegEx patter. A better understanding of chunks can be visualized by  [webpack-visualizer](https://chrisbateman.github.io/webpack-visualizer/).


src/content/plugins/split-chunks-plugin.md
```diff
@@ -344,3 +345,28 @@ module.exports = {
 ```
 
 W> This might result in a large chunk containing all external packages. It is recommended to only include your core frameworks and utilities and dynamically load the rest of the dependencies.
+
+### Split Chunks: Example 3
+
+ Create a `custom vendors` chunk, which includes only those `node_modules` from the whole application which passes the RegEx Test
```
byzyk: I think this can be slightly reworded to something like: 
Create a `custom vendor` chunk, which contains certain `node_modules` packages matched by `RegExp`.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -344,3 +345,28 @@ module.exports = {
...
+          name: 'vendor',
+          chunks: 'all',
+        }
+      }
+    }
+  }
+};
+```
+
+ T> This will result into a vendor chunk containing desired libraries provided in RegEx pattern. A better understanding of chunks can be visualized by  [webpack-visualizer](https://chrisbateman.github.io/webpack-visualizer/).
```
byzyk: Not sure we need to promote `webpack-visualizer` here as there might be more appropriate places in docs. Instead, we might wanna say:
This will result in splitting `react` and `react-dom` into a separate chunk.

sakhisheikh: @byzyk  Thanks for the needful input. I'll make recommended changes. I have one concern here sometimes users aren't sure if they custom vendor actually contains only those libraries in RegEx so they can verify it with weback visualizer. Anyways, I'm taking your feedback and making changes in the PR.

sakhisheikh: I have pushed the latest changes.

byzyk: This plugin is already [mentioned in docs](https://webpack.js.org/guides/code-splitting/#bundle-analysis), so if you feel like we need to add a note about that here you can just reference the links I shared, saying something like: 

If you're not sure what packages have been included in a chunk you may refer to Bundle Analysis section(link) for details.

sakhisheikh: Thanks @byzyk  added requested changes.

src/content/plugins/split-chunks-plugin.md
```diff
@@ -369,4 +369,4 @@ module.exports = {
 };
 ```
 
- T> This will result in splitting `react` and `react-dom` into a separate chunk.
+ T> This will result in splitting `react` and `react-dom` into a separate chunk. If you're not sure what packages have been included in a chunk you may refer to [Bundle Analysis](https://webpack.js.org/guides/code-splitting/#bundle-analysis) section for details.
```
montogeek: Please use relative URLs `/guides/code-splitting/#bundle-analysis`

sakhisheikh: @montogeek pushed the requested changes. Thanks

PR #2521, opened 2018-09-16
docs(guides): env should default to empty object

venelinpetrov:
`env.NODE_ENV` and `env.production` assumes that these are explicitly passed as parameters, which is not often the case. A real world example can be:

```
"build:dev": "webpack --config=webpack.config.js --display-modules ",
"build:prod": "webpack --config=webpack.config.js --env.production --display-modules ",
```

For dev, we don't set env.production=false.


src/content/guides/environment-variables.md
```diff
@@ -29,7 +29,7 @@ __webpack.config.js__
 ``` js
 const path = require('path');
 
-module.exports = env => {
+module.exports = (env = {}) => {
```
byzyk: Assigning to an empty object here is going to result in printing `undefined` in a console in case env vars are not set.

Maybe instead of defaulting value to an empty object add a comment, saying something like: Make sure to provide env vars when running webpack etc..

venelinpetrov: The problem is that in most cases developers don't provide all the env vars in every script, only where an explicit value is needed. As you can see in the example above, the `build:env` script doesn't set `--env.production`. I had a case where a colleague refactored the webpack config based on this example and all of the sudden it started throwing errors where no values are supplied. To me `env={}` is better than
```
"build:dev:win": "webpack --config=webpack.config.js --env.prod=false --mac=false --asar=false --"
```

byzyk: Thanks for the clarification. I get what you're saying, but my point is that I don't feel like defaulting an `env` to an empty object is a good thing either.

Going back to your example with a colleague. In this case, it won't throw an error, it'll just output `undefined` in the console which is no harm. But in the real world, you normally do more stuff with env vars than just log them out, and that is going to produce bugs which might be hard to track.

I wasn't denying your case and I completely agree with it, I just feel like it should be handled in a slightly different way.

venelinpetrov: I'm very open to suggestion, because I stem only from our perspective. Just to add something, somewhere in the config we might have

```js
if (env.production) {
    return merge(commponConfig, prodConfig);
}

return merge(commponConfig, devConfig);
```

If we have a task `"build": "webpack"` `env.production` will be undefined and thus the second path is taken, which is expected. If we have `"build": "webpack --env.production"`, then the first (production) path is taken.

I'm ok to not merge and close the PR if you are still not convinced :) Our case is just one of many.

byzyk: Oh no, you don't have to close the PR, I'm with you on this one totally.

My suggestion remains the same, to add a note or comment in code which would warn explicitly to not forget to run webpack with env vars set otherwise might result in an error, which is expected but might be confusing in some cases like yours.

Also let's see what other folks have to say @EugeneHlushko @montogeek

venelinpetrov: @EugeneHlushko Thanks for joining the conversation. I have the following question though. How do you suggest to refactor our `build:dev:win` task so that it doesn't throw (keep in mind that we have a lot of tasks that use webpack)? I also don't want to explicitly set all the parameters, because it becomes difficult to read and also when someone adds new parameter we need to go and fix all the npm tasks that use webpack to include that parameter. Maybe I'm not getting it right. You will be of great help if you can resolve this.

EugeneHlushko: For your case from what i understand, defaulting to an empty object is a solution or doing explicit checks before using `env` (if its in one place) `env && env.prod && merge...` or just an `if (env) { .... }` block and further logic

montogeek:
@EugeneHlushko Any feedback?

venelinpetrov:
I don't see why the build fails. The error is irrelevant to my change.

src/content/guides/environment-variables.md
```diff
@@ -29,6 +29,8 @@ __webpack.config.js__
 ``` js
 const path = require('path');
 
+// Note that if you don't pass all the env varialbles you'll get ReferenceError.
```
EugeneHlushko: This looks like describing JavaScript to readers. If you reference undefined variable, you'll get a `ReferenceError `. 

¯\_(ツ)_/¯

montogeek:
@venelinpetrov Friendly ping

EugeneHlushko:
I emphasize, this is confusing! "Note that if you don't pass all the env variables you'll get ReferenceError. "

Which all variables?

montogeek:
All the variables that you try to access, basic JS error.
Yeah, I also think that it is not necessary to add that note.

EugeneHlushko:
Yes you see we would write that into comment if want to merge otherwise it's confusing. It would be like:
```
Note that if you don't pass all the env the variables that you try to access, you'll get ReferenceError. 
```

EugeneHlushko:
Suggest we close this in favor of leaving it up to node to report a `ReferenceError`

PR #2522, opened 2018-09-16
docs(config): Replace UglifyJSPlugin references with TerserPlugin

### LINKED ISSUE #2520: Document webpack change: Switch from uglifyjs-webpack-plugin to terser-webpack-plugin
*A pull request by @edmorley was merged and maintainers requested a documentation change.*

See pull request: https://github.com/webpack/webpack/pull/8036

---

Some history:
* [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) < v1.0 used the minifier [uglify-js](https://github.com/mishoo/UglifyJS2)
* however `uglify-js` does not support ES6, which resulted in a fork called [uglify-es](https://www.npmjs.com/package/uglify-es) that was developed in the `uglify-js` repository, but under the [harmony](https://github.com/mishoo/UglifyJS2/tree/harmony) branch
* `uglifyjs-webpack-plugin` v1.x switched to `uglify-es` for ES6 support
* however `uglify-es` [stopped being maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058)
* which led to a fork called [terser](https://github.com/fabiosantoscode/terser) that has incorporated all of the unmerged PRs and will be where all new development occurs
* [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin) was created, which is the `terser` equivalent of `uglifyjs-webpack-plugin`
* `uglifyjs-webpack-plugin` v2.x [has switched back to uglify-js](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/blob/master/CHANGELOG.md#200-2018-09-14), so any project that needs to support ES6 now needs to switch to `terser-webpack-plugin` instead.

Fixes #7923.

Bugfix I guess? (since terser includes many correctness fixes)

There are existing tests which have been updated for the plugin name change.

Yes. The new plugin:
* requires the minifier options to be passed under [terserOptions](https://github.com/webpack-contrib/terser-webpack-plugin#terseroptions), rather than the previous [uglifyOptions](https://github.com/webpack-contrib/uglifyjs-webpack-plugin#uglifyoptions).
* has changed the method signature that should be used when specifying a `warningsFilter` function.

* The defaults listed for [optimization.minimizer](https://webpack.js.org/configuration/optimization/#optimization-minimizer) will need updating (and also made to link to the [terser plugin docs](https://webpack.js.org/plugins/terser-webpack-plugin/) instead)

edmorley:
Since as of webpack 5, the default `options.minimizer` is now `terser-webpack-plugin`:
https://github.com/webpack/webpack/pull/8036

I've also:
* updated the custom `minimizer` example to include all of the default options, to reduce the chance people forget to enable them.
* removed the `optimize/UglifyJsPlugin` reference, since it's no longer included in the webpack repository.
* removed the "Here are a few more popular ones" mention, since it's not accurate (they aren't more popular).
* removed manual plugin configuration from `configuration-languages.md` since the defaults for `minimizer` make it unnecessary.

Fixes #2520.

### DIFF
--- a/src/content/configuration/configuration-languages.md
+++ b/src/content/configuration/configuration-languages.md
@@ -118,7 +118,6 @@ config =
     use: 'babel-loader'
   } ]
   plugins: [
-    new (webpack.optimize.UglifyJsPlugin)
     new HtmlWebpackPlugin(template: './src/index.html')
   ]
 
@@ -167,10 +166,6 @@ export default (
       }} />
     </resolve>
     <plugins>
-      <uglify-js opts={{
-        compression: true,
-        mangle: false
-      }} />
       <CustomPlugin foo="bar" />
     </plugins>
   </webpack>
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -19,7 +19,7 @@ Since version 4 webpack runs optimizations for you depending on the chosen `mode
 
 `boolean`
 
-Tell webpack to minimize the bundle using the plugin(s) specified in `optimization.minimizer`.
+Tell webpack to minimize the bundle using the plugin(s) specified in [`optimization.minimizer`](#optimization-minimizer).
 
 This is `true` by default in `production` mode.
 
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -19,7 +19,7 @@ Since version 4 webpack runs optimizations for you depending on the chosen `mode
 
 `boolean`
 
-Tell webpack to minimize the bundle using the [UglifyjsWebpackPlugin](/plugins/uglifyjs-webpack-plugin/).
+Tell webpack to minimize the bundle using the plugin(s) specified in `optimization.minimizer`.
 
 This is `true` by default in `production` mode.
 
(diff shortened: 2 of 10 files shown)


src/content/configuration/optimization.md
```diff
@@ -19,7 +19,7 @@ Since version 4 webpack runs optimizations for you depending on the chosen `mode
 
 `boolean`
 
-Tell webpack to minimize the bundle using the [UglifyjsWebpackPlugin](/plugins/uglifyjs-webpack-plugin/).
+Tell webpack to minimize the bundle using the plugin(s) specified in `optimization.minimizer`.
```
EugeneHlushko: `optimization.minimizer` could be a link to next section

src/content/plugins/internal-plugins.md
```diff
@@ -312,14 +312,6 @@ Merges chunks until each chunk has the minimum size of `minChunkSize`.
 
 Adds chunk ids of chunks which are included in the chunk. This eliminates unnecessary chunk loads.
 
-### UglifyJsPlugin
```
EugeneHlushko: why remove and not replace with TerserPlugin mention ?

edmorley: This page is about internal webpack plugins (ie plugins in the webpack repository) - which `UglifyJsPlugin` used to be, but was since moved out. As such, the `UglifyJsPlugin` this section is referring to actually doesn't exist any more:
https://github.com/webpack/webpack/blob/2f78aae8d8f78ce9c9221ff25a0483d81809bd81/lib/webpack.js#L179-L184

EugeneHlushko: ah true, thx for

src/content/configuration/configuration-languages.md
```diff
@@ -118,7 +118,6 @@ config =
     use: 'babel-loader'
   } ]
   plugins: [
-    new (webpack.optimize.UglifyJsPlugin)
```
byzyk: shouldn't we add `Terser` here instead?

src/content/configuration/configuration-languages.md
```diff
@@ -167,10 +166,6 @@ export default (
       }} />
     </resolve>
     <plugins>
-      <uglify-js opts={{
-        compression: true,
-        mangle: false
-      }} />
```
byzyk: shouldn't we add Terser here instead?

src/content/configuration/configuration-languages.md
edmorley: This is explained in the commit message :-)

byzyk: Aw sorry didn't see it. Yeah makes sense, thanks!

montogeek:
Thanks!

edmorley:
Thank you for the reviews/merge :-)

PR #2524, opened 2018-09-18
Context Replacement Plugin

masives:
Add example on how to tree shake locales from date-fns.


montogeek:
Please sign CLA.

src/content/plugins/context-replacement-plugin.md
```diff
@@ -31,10 +31,17 @@ If the resource (directory) matches `resourceRegExp`, the plugin replaces the de
...
 new webpack.ContextReplacementPlugin(
   /moment[/\\]locale$/,
   /de|fr|hu/
 );
+
+// date-fns
+new ContextReplacementPlugin(
+  /date\-fns[\/\\]/, 
+  /[/\\](de|fr|hu)[/\\]/
+  );
```
montogeek: Please fix indentation.

masives: I believe it's now correct

masives:
I wonder though if it would be usefull to show also how the dynamic import is also happening in the place of use to avoid some possible confusion

src/content/plugins/context-replacement-plugin.md
masives: I haven't seen you reply but now I do believe it's correct. Do you have any other things you'd like to point out?

EugeneHlushko:
Wasn't moment example enough for a developer to figure out how to tree shake in similar cases? I mean after seeing two similar examples future contributors might want to put their use cases into example and we end up maintaining them (after example's lib releases breaking change to their structure and this regex wont fit anymore)

masives:
I had really hard time figuring this out as we're not checking against filename but filepath. 
moment.js

In moment we have a lot of files named after their locales
```js
const getLocale = locale => require(`moment/locale/${locale}.js`)
```


While in date-fns it is folder name that we need to check:
```js
const getLocale = locale => require(`date-fns/locale/${locale}/index.js`)
```

It took me some find to find a solution that would properly solve the issue of packing all locales so I decided to suggest this change to avoid this pain for someone that might be looking for this solution.

If you're afraid of maintaining each time library example maybe giving a link to solution would be a good option? [date-fns docs/webpack](https://github.com/date-fns/date-fns/blob/master/docs/webpack.md)?

EugeneHlushko:
Hi @masives 
I think afraid is a little too strong here but yes, there is a concern. The link (https://github.com/date-fns/date-fns/blob/master/docs/webpack.md) looks better to me. If you agree please put it into related section of .md file header (like here https://raw.githubusercontent.com/webpack/webpack.js.org/master/src/content/guides/code-splitting.md) with appropriate title

EugeneHlushko:
@montogeek pls review

montogeek:
Thanks!

PR #2525, opened 2018-09-18
Using vendors as entry is not recommended

sokra:
fixes webpack/webpack#6647

see https://github.com/webpack/webpack/issues/6647#issuecomment-422288194

> > You should not have a entrypoint for vendor. It's not an entry.

> This advice is in direct contradiction to official documentation: https://webpack.js.org/concepts/entry-points/


dhruvdutt:
Is there a need for an explicit note to not have an entrypoint for vendor?

montogeek:
Thanks!

PR #2531, opened 2018-09-20
fix(guides): Fixed example code typo

huang-x-h:
Update writing-a-plugin.md

PR #2535, opened 2018-09-20
Update filelist.md plugin example in write-a-plugin.md

danielna:
In the filelist.md the callback parameter should be `compilation`, not `compiler`.


montogeek:
Duplicate of https://github.com/webpack/webpack.js.org/pull/2531, please check PRs before submitting yours.

PR #2538, opened 2018-09-22
docs(NormalModuleReplacement) Update for WebPack 4 and TypeScript

robertmclaws:
- Updated the Advanced Example with the WebPack 4 exports signature
- Eliminated an unnecessary and confusing variable by basing replacement on the `env` argument.
- Added code to the replacement function to make it easier to troubleshoot.
- Made the config files examples less obtuse by creating a real-world situation.
- Updated the webpack shell commands with the v4 arguments.
- Added details for how to mitigate issues you'll encounter if you try to implement these instructions with TypeScript.


robertmclaws:
@montogeek Just wanted to put this on your radar... would love to get these changes into the docs, I think they would help a lot of people using WebPack and TypeScript. :)

src/content/plugins/normal-module-replacement-plugin.md
```diff
@@ -40,50 +41,78 @@ Conditional build depending on an [specified environment](/configuration/configu
...
 ``` javascript
-module.exports = function(env) {
-  var appTarget = env.APP_TARGET || 'VERSION_A';
+module.exports = function(env, args) {
   return {
     plugins: [
       new webpack.NormalModuleReplacementPlugin(/(.*)-APP_TARGET(\.*)/, function(resource) {
-        resource.request = resource.request.replace(/-APP_TARGET/, `-${appTarget}`);
+        var original = resource.request;
+        var replace = resource.request.replace(/-APP_TARGET/, `-${env}`);
```
montogeek: Shouldn't this be `-${env.APP_TARGET}`?

src/content/plugins/normal-module-replacement-plugin.md
```diff
@@ -40,50 +41,78 @@ Conditional build depending on an [specified environment](/configuration/configu
...
 console.log(config.title);
 ```
 
-And now you just get the right config imported depending on which target you're building for:
+And now the right config will be imported depending on which target you're building for:
 
 ``` shell
-webpack --env.APP_TARGET VERSION_A
-=> 'I am version A'
+webpack --mode production --env BETA
```
montogeek: `--env` is used when your webpack config is a function. Please use `--env.APP_TARGET`.
It could be also used to append more environment variables: https://webpack.js.org/guides/environment-variables/

src/content/plugins/normal-module-replacement-plugin.md
```diff
@@ -40,50 +41,78 @@ Conditional build depending on an [specified environment](/configuration/configu
...
-webpack --env.APP_TARGET VERSION_B
-=> 'I am version B'
+webpack --mode production --env PROD
+=> 'I am the Production configuration.'
 ```
+
+
+## TypeScript Considerations
+
+The TypeScript compiler requires that all files be physically-present in order to generate an output. If you attempt to
```
EugeneHlushko: why the line break? is this forced by linter or smth?

src/content/plugins/normal-module-replacement-plugin.md
```diff
@@ -40,50 +41,78 @@ Conditional build depending on an [specified environment](/configuration/configu
...
+complete the above tasks, you will likely receive the following error:
```
EugeneHlushko: Suggest changing to: After running the above example in TypeScript, you would likely receive the following error:

src/content/plugins/normal-module-replacement-plugin.md
```diff
@@ -40,50 +41,78 @@ Conditional build depending on an [specified environment](/configuration/configu
...
+
+``` shell
+TS2307: Cannot find module './config/config-APP_TARGET'.
+```
+
+To correct this issue, you'll need to create an additional file to satisfy the compiler without breaking the
```
EugeneHlushko: Doesnt sound right to me, suggest to change `To correct this issue` => `In order to make it work, ...`

src/content/plugins/normal-module-replacement-plugin.md
```diff
@@ -40,50 +41,78 @@ Conditional build depending on an [specified environment](/configuration/configu
 Say you want a configuration with specific values for different build targets.
 
 ``` javascript
-module.exports = function(env) {
-  var appTarget = env.APP_TARGET || 'VERSION_A';
+module.exports = function(env, args) {
```
EugeneHlushko: `args`?

robertmclaws: See the last code section on https://webpack.js.org/concepts/mode/#mode-production. It appears that the signature and use case has changed from earlier versions of Webpack, and some of the documentation throughout may not have been updated to match the change. 

Based on the new signature, from a DevOps perspective, `env` would specify the environment that you are **in**, and `args` (I believe `argv` is a typo) would be any additional configuration you pass to the build. Any other values that you needed that are *specific* to the environment would either be built *into the script itself based on the environment name*, or passed in through the `args`. **The way that `mode` shows up as a property of `args` and not `env` in this example (and I have confirmed this behavior in my own apps) appears to back up my observation on this change.**

Based on the fact that passing `args` is even *possible*, I would recommend updating **all** of the documentation to make sure the signature is consistent for all Webpack users moving forward.

src/content/plugins/normal-module-replacement-plugin.md
robertmclaws: Please see my notes here: https://github.com/webpack/webpack.js.org/pull/2538#discussion_r220957141

src/content/plugins/normal-module-replacement-plugin.md
robertmclaws: Well, you're not "running it in TypeScript". TypeScript is a compiler not a runtime engine. The problem happens when you run Webpack and ts-loader tries to finish its work compiling the code during that process. I can change to "After making the above changes and running Webpack, `ts-loader` is likely to throw the following error:"

src/content/plugins/normal-module-replacement-plugin.md
robertmclaws: I called it an "issue" because the pattern works in JavaScript and not in the TypeScript compiler, and the compiler throws an error. I'm opening an issue in the TypeScript project to see if there is a way to resolve it without my hack... at which point we could take out this whole section.

Happy to make the change if you still feel it's necessary.

src/content/plugins/normal-module-replacement-plugin.md
robertmclaws: I think that was just an artifact of editing it right in GitHub. I can remove it if you'd like.

src/content/plugins/normal-module-replacement-plugin.md
EugeneHlushko: yup much better, pls do

src/content/plugins/normal-module-replacement-plugin.md
EugeneHlushko: When i am reading "to correct an issue" i get confused, i had no issues and was following a guide, maybe we need more eyes on this cc @montogeek

src/content/plugins/normal-module-replacement-plugin.md
EugeneHlushko: CC @evenstensberg

robertmclaws: I should also point out that, if you do it this way, you get super clean and easy to understand build commands per environment. Consider the following elements in my Visual Studio 2017 project file:
```
    <Exec Command="npx webpack --config webpack.config.vendor.js --mode production" />
    <Exec Command="npx webpack --mode production --env dev" Condition="'$(Configuration)' == 'DEV'" />
    <Exec Command="npx webpack --mode production --env beta" Condition="'$(Configuration)' == 'BETA'" />
    <Exec Command="npx webpack --mode production --env prod" Condition="'$(Configuration)' == 'PROD'" />
```
Whether the concept was intentional or not, it's now really easy to understand what is going on, and wire the environment through your build pipeline.

HTH! Thanks for working to move this PR forward.

src/content/plugins/normal-module-replacement-plugin.md
robertmclaws: I've made the change as requested.

src/content/plugins/normal-module-replacement-plugin.md
robertmclaws: I've made changes to this as requested.

src/content/plugins/normal-module-replacement-plugin.md
evenstensberg: Yes I think args is possible to pass in. Has seen some people doing that. Is that not working out for you @robertmclaws ?

robertmclaws: @evenstensberg Not, it's definitely working. The concern on this part of the thread is that the documentation is inconsistent between sections with how it deals with `env` vs `args`. My change here is to make `env` a single string that specifies the environment name, which brings this page in line with documentation here: https://webpack.js.org/concepts/mode/#mode-production and has the side effect of being much simpler than the approach used throughout the rest of the docs.

I'm suggesting that my change be approved here, and then a separate issue opened to update documentation throughout the rest of the site to use the new `module.exports` signature.

evenstensberg: That might be a good thing, but people use the env variable for other things too, so that might be a diversion from the rest of the JS ecosystem, don't you think? I'm in for keeping env, but it has to be done similar to the rest of the ecosystem

EugeneHlushko: To summarize from what we have so far:
- addition of `args` to the signature is pure gold
- changing `env` to `string` is not that common, please roll back

src/content/plugins/normal-module-replacement-plugin.md
EugeneHlushko: please use any other param name to not mix with name and access it in example if still neded via `args.{param}`

EugeneHlushko: or also roll back this and keep `--env. APP_TARGET `

montogeek:
@EugeneHlushko Any feedback?

src/content/plugins/normal-module-replacement-plugin.md
EugeneHlushko: we would probably need to return the target declaration with fall back e.g. `const target = env || 'BETA';`
```suggestion
         var target = env || 'BETA';
         var replace = resource.request.replace(/-APP_TARGET/, `-${target}`);
```

montogeek:
What is the status of this PR?

EugeneHlushko:
Needs small tweaks, @robertmclaws can you share access to your fork i can throw final commit in?

montogeek:
@robertmclaws Friendly ping

snitin315:
Closing as stale.

PR #2544, opened 2018-09-25
docs(plugins) More info on HashedModuleIdsPlugin#hashDigestLength

jakub-g:
I was wondering what is the behavior in case of potential collisions, I think it's worth documenting.

See https://github.com/webpack/webpack/blob/91e9dd77062797e3e716485f295d32a5cab57042/lib/HashedModuleIdsPlugin.js#L50

(This doesn't handle the potential collisions of the hashing algorithm itself, only the collisions due to truncating the hash -- though the first type should virtually never happen in case of the input type being a module name / path string, so I think that's fine).


EugeneHlushko:
Thanks!

PR #2548, opened 2018-09-26
Update docs: adds JSON file to out of the box supported files

farskid:
webpack supports different file types and extensions by default out of the box. `.js`, `.mjs` and `.json` files. but in the old docs, it's been written as it only supports Javascript files.


montogeek:
Thanks!

PR #2555, opened 2018-09-27
Fix(utilities): Convert github raw links to rendered page links

budmc29:
As discussed in #2470


montogeek:
Thanks!

budmc29:
Thanks as well. Where can I see the status of rebuild and if are there any small tasks I can help with that?

PR #2556, opened 2018-09-28
docs(config): Add section for output.libraryTarget: "amd-require"

### LINKED ISSUE #2562: Document webpack change: Add "amd-require" libraryTarget, resulting in a require() wrapper
*A pull request by @eemeli was merged and maintainers requested a documentation change.*

See pull request: https://github.com/webpack/webpack/pull/8093

---

This is a partial fix for #8079.

This PR adds `"amd-require"` as a new library target. The target is otherwise the same as `"amd"`, except that it uses `require()` as a wrapper instead of `define()`. The specification for this API is [here](https://github.com/amdjs/amdjs-api/blob/master/require.md).

Specifically, we're using the global `require(dependencies, factory)` form, which will resolve the dependencies, and then immediately call the factory function with them as parameters. The library name is ignored in this case.

The rationale for including this target is that it enables compiling an immediately executed module that has AMD dependencies, which has not been previously possible with Webpack. As is, AMD dependencies are only resolved with the `"amd"` library target, which results in a `define()` wrapper. To execute the JavaScript of such output, a subsequent `require()` call is needed. If the intent is not to produce a library, this becomes clumsy and potentially difficult (see jantimon/html-webpack-plugin#1052).

Yes.

No.

The new target should be documented under the [Module Definition Systems](https://webpack.js.org/configuration/output/#module-definition-systems) section of the Output configuration page.

eemeli:
This adds a section to the Output configuration page for the "amd-require" library target that was added in webpack/webpack#8093.

Fixes #2562.

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -621,7 +621,7 @@ This bundle will not work as expected, or not work at all (in the case of the al
 
 `libraryTarget: "amd-require"` - This packages your output with an immediately-executed AMD `require(dependencies, factory)` wrapper.
 
-The "amd-require" target allows for the use of AMD dependencies without needing a separate later invocation. As with the `"amd"` target, this depends on the appropriate [`require` function](https://github.com/amdjs/amdjs-api/blob/master/require.md) being available in the environment in which the webpack output is loaded.
+The "amd-require" target allows for the use of AMD dependencies without needing a separate later invocation. As with the "amd" target, this depends on the appropriate [`require` function](https://github.com/amdjs/amdjs-api/blob/master/require.md) being available in the environment in which the webpack output is loaded.
 
 With this target, the library name is ignored.
 
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -621,7 +621,7 @@ This bundle will not work as expected, or not work at all (in the case of the al
 
 `libraryTarget: "amd-require"` - This packages your output with an immediately-executed AMD `require(dependencies, factory)` wrapper.
 
-This target allows for the use of AMD dependencies without needing a separate later invokation. As with the `"amd"` target, this depends on the appropriate [`require` function](https://github.com/amdjs/amdjs-api/blob/master/require.md) being available in the environment in which the webpack output is loaded.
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.md
```diff
@@ -618,6 +619,13 @@ define([], function() {
 This bundle will not work as expected, or not work at all (in the case of the almond loader) if loaded directly with a `<script>` tag. It will only work through a RequireJS compatible asynchronous module loader through the actual path to that file, so in this case, the `output.path` and `output.filename` may become important for this particular setup if these are exposed directly on the server.
 
 
+`libraryTarget: "amd-require"` - This packages your output with an immediately-executed AMD `require(dependencies, factory)` wrapper.
+
+This target allows for the use of AMD dependencies without needing a separate later invokation. As with the `"amd"` target, this depends on the appropriate [`require` function](https://github.com/amdjs/amdjs-api/blob/master/require.md) being available in the environment in which the webpack output is loaded.
```
EugeneHlushko: typo in word `invocation`

EugeneHlushko: Suggest to change `This target` to `"amd-require" target allows...` to not have two sentences starting from `This`

src/content/configuration/output.md
```diff
@@ -618,6 +619,13 @@ define([], function() {
 This bundle will not work as expected, or not work at all (in the case of the almond loader) if loaded directly with a `<script>` tag. It will only work through a RequireJS compatible asynchronous module loader through the actual path to that file, so in this case, the `output.path` and `output.filename` may become important for this particular setup if these are exposed directly on the server.
 
 
+`libraryTarget: "amd-require"` - This packages your output with an immediately-executed AMD `require(dependencies, factory)` wrapper.
+
+The "amd-require" target allows for the use of AMD dependencies without needing a separate later invocation. As with the `"amd"` target, this depends on the appropriate [`require` function](https://github.com/amdjs/amdjs-api/blob/master/require.md) being available in the environment in which the webpack output is loaded.
```
EugeneHlushko: i think it should be
```
`"amd-require"`
```

to be consistent with similar syntax in same paragraph where we mention
```
`"amd"`
```

montogeek:
Thanks!

PR #2558, opened 2018-09-28
Updated Creating a Plugin Section Example

NaviMarella:
Modified Creating a Plugin example for below changes.
1. Modified compile hook to emit, as compile is a sync hook and it doesn't have tapAsync method.


EugeneHlushko:
Thanks!

PR #2559, opened 2018-09-28
Add boolean as possible type for lineToLine

ndelangen:
I noticed a discrepancy between the typescript definitions and this documentation, and assumed the definition to be correct.

This changes doc to show that a boolean is possible for the lineToLine option


montogeek:
Thanks!

PR #2570, opened 2018-10-03
docs(mode) add missing pieces of mode magic

### LINKED ISSUE #2550: More going on with `mode: "development"` then in the docs!
# Bug report

These two configuration produce very different results, including wildly different bundle sizes, that are not documented properly or misleading in the [documentation](https://webpack.js.org/concepts/mode/#mode-development). I believe `mode: "development"` is also setting some optimization config parameters that prevent minification, but it's not documented. At minimum can the docs be updated to reflect exactly what mode is doing?

```
module.exports = {
  mode: "development
}
```

```
module.exports = {
  devtool: 'eval',
  plugins: [
    new webpack.NamedModulesPlugin(),
    new webpack.NamedChunksPlugin(),
    new webpack.DefinePlugin({ "process.env.NODE_ENV": JSON.stringify("development") }),
  ]
}
```

Try and replicate `mode: "development"` based on the [documentation](https://webpack.js.org/concepts/mode/#mode-development). Note the difference in package sizes and minification.

The documentation should be accurate. I should be able to replicate whatever `mode: "development"` is doing. Documentation:
https://webpack.js.org/concepts/mode/#mode-development

webpack version:  4.16.2
Node.js version: 9.8.0
Operating System: MacOS High Sierra 10.13.6
Additional tools:

---

This issue was moved from webpack/webpack#8091 by @montogeek. Original issue was by @mrichmond.

EugeneHlushko:
Fixes #2550

### DIFF
--- a/src/content/concepts/mode.md
+++ b/src/content/concepts/mode.md
@@ -4,6 +4,10 @@ sort: 4
 contributors:
   - EugeneHlushko
   - byzyk
+  - mrichmond
+related:
+  - title: 'webpack options defaulter source code'
+    url: https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js
 ---
 
 Providing the `mode` configuration option tells webpack to use its built-in optimizations accordingly.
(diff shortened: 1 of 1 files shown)


src/content/concepts/mode.md
```diff
@@ -4,6 +4,10 @@ sort: 4
 contributors:
   - EugeneHlushko
   - byzyk
+  - mrichmond
+related:
+  - title: 'webpack options defaulter source code'
```
byzyk: `options defaulter` is the name of the class which is not covered here. Perhaps a bit more generic title would fit better here, something like: "webpack default options (source code)" or something

EugeneHlushko: makes sense 👍

PR #2571, opened 2018-10-05
Mention @babel/preset-env transforming ES modules to CommonJS by default

wojtekmaj:
@babel/preset-env is extremely popular, but unless an additional configuration was added to it, it will break Webpack's tree shaking. It's worth mentioning it in the docs, I think.


montogeek:
Thanks!

ron23:
@montogeek is this still true?
I see it's defaulting to "auto" (no idea how how "auto" determines the type).
I ran my pretty big and complex app without `modules` at all and with `modules: false` and didn't see a different in bundle size.

Issue #1521, opened 2017-08-14
webpack-dev-server changeOrigin option?


blueshirts:
It seems like changeOrigin is a valid option for the webpack-dev-server proxy settings though I don't see it documented anywhere.

https://webpack.js.org/configuration/dev-server/

Should this be documented?  I am unsure what it's exact purpose is.

skipjack:
So where/how did you find that `changeOrigin` is a valid option?

skipjack:
`webpack-dev-server` passes the `proxy` options to the [`http-proxy-middleware`](https://github.com/chimurai/http-proxy-middleware) which is linked to and documented in the proxy section of the dev-server page:

https://webpack.js.org/configuration/dev-server#proxy

The `changeOrigin` option is documented in the other package in their README. If it's not documented well enough, I would open an issue over there. Also, if you want to highlight the relationship between the two packages a bit more then feel free to submit a PR here as well but we shouldn't re-document all options (i.e. linking out is the right way to go so we don't duplicate existing documentation).

blueshirts:
I saw the changeOrigin option in one of the examples in the old webpack documentation.

https://webpack.github.io/docs/webpack-dev-server.html#proxying-local-virtual-hosts

blueshirts:
I found that it was a valid option b/c my proxy would not work without it.  I initially did not see the link you mentioned, thanks for the insight.

skipjack:
Ah ok... yeah like I said if you wanted to submit a PR here I'd be happy to review but imo there's plenty of docs on the `proxy` option already and the link to `http-proxy-middleware` is good. Like I said if anything, I think the link should be highlighted more so that it's as clear as possible that these options are actually handled by another package.

Thanks for highlighting this though and definitely feel free to submit a PR. I'll keep my eyes out for anyone mentioning something similar (which would definitely indicate that link should be highlighted).

ebuildy:
``changeOrigin`` is a very important option, as you can see in http-proxy code source, it's change the host header, could be necessary according the targeted server configuration.

```
if (options.changeOrigin) {
    outgoing.headers.host =
      required(outgoing.port, options[forward || 'target'].protocol) && !hasPort(outgoing.host)
        ? outgoing.host + ':' + outgoing.port
        : outgoing.host;
  }
```

littlee:
[...]

PR #2572, opened 2018-10-06
Added some clarification about changeOrigin

Yiidiir:
Added some documentation regarding `changeOrigin` in `devServer.proxy`
Already discussed in #1521

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -836,6 +836,19 @@ module.exports = {
   }
 };
 ```
+The origin of the host header is kept when proxying by default, you can set `changeOrigin` to `true` to override it, this is useful in some situations like using [name-based virtual hosted sites](https://en.wikipedia.org/wiki/Virtual_hosting#Name-based).
+
+```js
+module.exports = {
+  //...
+  devServer: {
+    proxy: {
+      '/api': 'http://localhost:3000',
+      changeOrigin: true
+    }
+  }
+};
+```
 
 ## `devServer.progress` - CLI only
 


src/content/configuration/dev-server.md
```diff
@@ -836,6 +836,19 @@ module.exports = {
   }
 };
 ```
+The origin of the host header is kept when proxying by default, you can set `changeOrigin` to `true` to override it, this is useful in some situations like using [name-based virtual hosted sites](https://en.wikipedia.org/wiki/Virtual_hosting#Name-based).
+
```
EugeneHlushko: please add `__webpack.config.js__` before your code example

src/content/configuration/dev-server.md
```diff
@@ -836,6 +836,19 @@ module.exports = {
   }
 };
 ```
+The origin of the host header is kept when proxying by default, you can set `changeOrigin` to `true` to override it, this is useful in some situations like using [name-based virtual hosted sites](https://en.wikipedia.org/wiki/Virtual_hosting#Name-based).
```
EugeneHlushko: i suggest to rephrase a bit:
you can set `changeOrigin` to `true` to override it, ... **=>** you can set `changeOrigin` to `true` to override this behaviour. It is useful in some cases like ...

It is not exactly a situation but a use case, so a full stop before that sentence sounds better to me.

src/content/configuration/dev-server.md
```diff
@@ -836,6 +836,19 @@ module.exports = {
...
+
+```js
```
EugeneHlushko: js => javascript

Yiidiir:
@EugeneHlushko Updated, even though not sure about the `javascript` thing as i can see `js` in the rest of the parts.
Also should i update `contributors:` ?

EugeneHlushko:
Hi @Yiidiir ,

Yes please add yourself as contributor in the page

src/content/configuration/dev-server.md
```diff
@@ -917,6 +917,21 @@ module.exports = {
   }
 };
 ```
+The origin of the host header is kept when proxying by default, you can set `changeOrigin` to `true` to override this behaviour.It is useful in some cases like using [name-based virtual hosted sites](https://en.wikipedia.org/wiki/Virtual_hosting#Name-based).
```
EugeneHlushko: I think we are missing a space now in `behaviour.It is `

Yiidiir:
@EugeneHlushko Done!

PR #2575, opened 2018-10-10
Add 'name' field docs & minor fixes

vansosnin:
Added `name` field docs. I found it accidentally in the error message. And this is exactly what I needed in my work 🙂
Also added missing field types for sibling fields.

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -24,7 +24,9 @@ W> Help Wanted: This page is still a work in progress. If you are familiar with 
 
 Set the value of `require.amd` or `define.amd`:
 
-```js
+__webpack.config.js__
+
+```javascript
 module.exports = {
   //...
   amd: {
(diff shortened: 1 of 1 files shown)


src/content/configuration/other-options.md
```diff
@@ -146,3 +153,17 @@ module.exports = {
...
 ```
+
+
+## `name`
+
+`string`
+
+Name of the configuration. Used when loading multiple configurations.
+
+```js
```
EugeneHlushko: use `javascript` vs `js`

src/content/configuration/other-options.md
```diff
@@ -146,3 +153,17 @@ module.exports = {
   recordsOutputPath: path.join(__dirname, 'newRecords.json')
 };
 ```
+
+
+## `name`
+
+`string`
+
```
EugeneHlushko: please also add `__webpack.config.js__` above the code blocks

vansosnin: Both review fixes done. Also fixed inconsistency in other examples on this page.

montogeek:
Thanks!

PR #2577, opened 2018-10-11
docs(plugins) update scp test info

### LINKED ISSUE #1956: Improvements on SplitChunks article

EugeneHlushko:
Fixes #1956 

See https://github.com/webpack/webpack.js.org/issues/1956#issuecomment-417346913

In comments people also asked for examples, i think that is a separate to the original issue request, it is time to close this one as if syntax on test function is now correct.

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -30,10 +30,10 @@ By default it only affects on-demand chunks, because changing initial chunks wou
 
 webpack will automatically split chunks based on these conditions:
 
-* New chunk can be shared OR modules are from the `node_modules` folder
-* New chunk would be bigger than 30kb (before min+gz)
-* Maximum number of parallel requests when loading chunks on demand would be lower or equal to 5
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks!

PR #2578, opened 2018-10-11
docs(cs) enforce dashed ul style for markdown

EugeneHlushko:
enforce `MD004` dash style rule, and apply proper ul formatting to md files for future maintaining and consistency


montogeek:
@EugeneHlushko Dont merge master please, waste of resources.

montogeek:
Also, why add that rule now?

EugeneHlushko:
It was conflict resolve via github interface, it says master merge through ;]
Because we have it inconsistent, about 50/50%. Default value was `"consistent"` but in a scope of file, not in a scope of project...

EugeneHlushko:
There is another issue with formatting we have but i dont have a solution to that on automatic level, it is the bold/italic syntax. right now we use in older edits `**smth**` and in newer `__smth__` which should be set by a md lint rule, but there is no such rule currently for bold/italic consistency

montogeek:
I see

montogeek:
Thanks!

montogeek:
About the bold thing, just keep an eye open :D

EugeneHlushko:
yeah 👍

byzyk:
> There is another issue with formatting we have but i dont have a solution to that on automatic level, it is the bold/italic syntax. right now we use in older edits **smth** and in newer __smth__ which should be set by a md lint rule, but there is no such rule currently for bold/italic consistency

I created an issue DavidAnson/markdownlint#150 and was planning to implement it internaly as a custom rule. If it goes well it can be merged later in core repo.

PR #2597, opened 2018-10-14
docs(config) target updates to current webpack state

EugeneHlushko:
- atom and electron no longer allowed via schema
- outdated link and example

### DIFF
--- a/src/content/configuration/target.md
+++ b/src/content/configuration/target.md
@@ -9,13 +9,14 @@ contributors:
   - pastelsky
   - tbroadley
   - byzyk
+  - EugeneHlushko
 ---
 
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L2113
I dont have chance to check when @montogeek maybe tomorrow i can do git annotate and tell more if thats important

EugeneHlushko:
@montogeek confirmed with Tobias, this is good to go.

"sokra [11:19 AM]
Only stuff in the schema is usable and part of the public API. Can be removed from the docs. You can also remove them from the code if you want to. Or create an issue."

montogeek:
Thanks!

EugeneHlushko:
dat was quick, ty!

PR #2599, opened 2018-10-16
docs(config) output libraryExport updates, md formatting

### LINKED ISSUE #2598: Documented "_entry_return_" default-value for libraryExport doesn't work with "umd" libraryTarget
# Bug report

When building a package with,

```json5
"libraryTarget": "umd",
// ("libraryExport" is undefined)
```

I get a correct package; that is, something like this:

```js
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
	  // …
	else
		root["GameEngine"] = factory();
})(window, function() {
return /******/ (function(modules) { // webpackBootstrap

// … BUNDLE BODY …

/******/ });
});
//# sourceMappingURL=game-engine.js.map
```

I can, of course, change `libraryExport` to `"default"`, and things work as expected:

```json5
"libraryTarget": "umd",
"libraryExport": "default",
```

```diff
diff --git i/dist/game-engine.js w/dist/game-engine.js
@@ -50967,6 +50967,6 @@ function () {
 
 /***/ })
 
-/******/ });
+/******/ })["default"];
 });
 //# sourceMappingURL=game-engine.js.map
```

Now, if an upstream configuration changes the default value of this, I may wish to reset it to default. According to [the docs](https://webpack.js.org/configuration/output/#output-libraryexport), that's the value `"_entry_return_"` (o_O).

> 

Okay, let's try that:

```json5
"libraryTarget": "umd",
"libraryExport": "_entry_return_",
```

```diff
diff --git i/dist/game-engine.js w/dist/game-engine.js
@@ -50967,6 +50967,6 @@ function () {
 
 /***/ })
 
-/******/ });
+/******/ })["_entry_return_"];
 });
 //# sourceMappingURL=game-engine.js.map
```

Well, that's certainly not correct.

I spent a few hours debugging this under our build-system, trying to figure out why our UMD builds were exporting an `undefined` value. Turns out that's 'cuz our projects don't export an `_entry_return_` property. 🤣

After sufficient testing, I found out that using a `null` JSON value will *not* reset the algorithm to its default behaviour, but using the empty string *will*:

```json5
"libraryTarget": "umd",
"libraryExport": "",
```

```json5
// webpack.config.js
module.exports = {
  "output": {
    "library": "Blah",
    "libraryTarget": "umd",
    "libraryExport": "_entry_return_",
  }
}
```

Using the documented "default" value, verbatim, for `libraryExport`, should result in the exact same beahviour as omitting the value. (In particular, the string `_entry_return_` is pretty non-obvious; I'd suggest implementing, and then documenting, `null` as the default value.)

webpack version:  `v4.20.2`
Node.js version: `v8.12.0`
Operating System: macOS 10.14 “Mojave”

---

This issue was moved from webpack/webpack#8180 by @sokra. Original issue was by @ELLIOTTCABLE.

> `_entry_return_` doesn't exist in the implementation. In fact using `undefined` or an empty string for `libraryExport` will export the whole (namespace) object. It defaults to `undefined`

> This seems to be confusing in the documentation. There is a note below but that's difficult to find.

EugeneHlushko:
Fixes #2598

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -15,6 +15,7 @@ contributors:
   - madhavarshney
   - harshwardhansingh
   - eemeli
+  - EugeneHlushko
 ---
 
 The top-level `output` key contains set of options instructing webpack on how and where it should output your bundles, assets and anything else you bundle or load with webpack.
@@ -26,7 +27,9 @@ The top-level `output` key contains set of options instructing webpack on how an
 
 When used in tandem with [`output.library`](#output-library) and [`output.libraryTarget`](#output-librarytarget), this option allows users to insert comments within the export wrapper. To insert the same comment for each `libraryTarget` type, set `auxiliaryComment` to a string:
 
-```js
+__webpack.config.js__
+
+```javascript
 module.exports = {
   //...
   output: {
@@ -40,7 +43,9 @@ module.exports = {
 
 which will yield the following:
 
-```js
+__webpack.config.js__
+
+```javascript
 (function webpackUniversalModuleDefinition(root, factory) {
   // Test Comment
   if(typeof exports === 'object' && typeof module === 'object')
@@ -61,7 +66,9 @@ which will yield the following:
 
 For fine-grained control over each `libraryTarget` comment, pass an object:
 
-```js
+__webpack.config.js__
+
+```javascript
 module.exports = {
   //...
   output: {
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.md
```diff
@@ -437,36 +466,34 @@ T> Read the [authoring libraries guide](/guides/author-libraries) guide for more
 
 ## `output.libraryExport`
 
-`string` or `string[]` (since webpack 3.0.0)
-
-> Default: `_entry_return_`
+`string | string[]`
 
-Configure which module or modules will be exposed via the `libraryTarget`. The default `_entry_return_` value is the namespace or default module returned by your entry file. The examples below demonstrate the effect of this config when using `libraryTarget: "var"`, but any target may be used.
+Configure which module or modules will be exposed via the `libraryTarget`. It is `undefined` by default, same behaviour will be applied if you set `libraryTarget` to an empty string e.g. `''` it will export the whole (namespace) object. The examples below demonstrate the effect of this config when using `libraryTarget: 'var'`.
```
EugeneHlushko: this is the main change

montogeek:
Thanks!

PR #2604, opened 2018-10-19
docs(grammar) replace all its

EugeneHlushko:
FInally nail of those its -> it's


EugeneHlushko:
hmm you are right guys, i was a bit frustrated with a lot of minor changes coming into grammar recent weeks so wanted to make sure we are in a good shape, after reading through the sentences (without find+replace) i can see that this is not valid :) thanks for checking both

PR #2607, opened 2018-10-19
Fix broken link

numb86:
`/guides/migrating#automatic-loader-module-name-extension-removed` is broken link.
I think that the current link is `/migrate/3/#automatic-loader-module-name-extension-removed`.


EugeneHlushko:
Thanks!

PR #2614, opened 2018-10-24
docs(guides): Update var to let

anayadesign:
Updates the snippets for `src/index.js` to be consistent with the snippet from `/guides/getting-started.md`.

This consistency would benefit developers who are following along the walkthrough guides.

### DIFF
--- a/src/content/guides/hot-module-replacement.md
+++ b/src/content/guides/hot-module-replacement.md
@@ -81,8 +81,8 @@ __index.js__
   import printMe from './print.js';
 
   function component() {
-    let element = document.createElement('div');
-    let btn = document.createElement('button');
+    const element = document.createElement('div');
+    const btn = document.createElement('button');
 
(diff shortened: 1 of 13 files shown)


src/content/guides/hot-module-replacement.md
```diff
@@ -191,7 +192,7 @@ __index.js__
   }
 
 - document.body.appendChild(component());
-+ let element = component(); // Store the element to re-render on print.js changes
++ const element = component(); // Store the element to re-render on print.js changes
```
EugeneHlushko: You cannot reassign a const, this will throw when we do `element = component()` later.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

montogeek:
@anayadesign Friendly ping

montogeek:
Close because of inactivity

PR #2618, opened 2018-10-26
docs(config) stats pre defined template extension

### LINKED ISSUE #2616: src/content/configuration/stats.md
**Feature to document**

stats configuration - how to extend string-named with specific overrides

**Author(s)**

@SpaceK33z | @sallar | @jungomi | @ldrick | @jasonblanchard | @byzyk | @renjithvk | @Raiondesu | @EugeneHlushko

**Additional information**

My current stats configuration is `stats: 'minimal'` but I want to suppress a certain error without making any other changes.  How can we inherit 'minimal' or whatever and override just one setting?  Alternatively, what stats configurations to those string shortcuts stand for?

EugeneHlushko:
- explain on how to "extend" one of the stats pre-defined behaviours.

Fixes #2616

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -180,6 +180,27 @@ module.exports = {
 }
 ```
 
+If you want to use one of the pre-defined behaviours e.g. `'minimal'` but still override one or more of the rules, see [the source code](https://github.com/webpack/webpack/blob/master/lib/Stats.js#L1394-L1401). You would want to copy the configuration options from `case 'minimal': ...` and add your additional rules while providing an object to `stats`.
+
+__webpack.config.js__
+
+```javascript
+module.exports = {
+  //..
+  stats: {
+    // copied from `'minimal'`
+    all: false,
+    modules: true,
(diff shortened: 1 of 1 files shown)


src/content/configuration/stats.md
```diff
@@ -180,6 +180,27 @@ module.exports = {
 }
 ```
 
+If you want to use one of the pre-defined behaviours e.g. `'minimal'` but still override one or more of the rules, see [the source code](https://github.com/webpack/webpack/blob/master/lib/Stats.js#L1394-L1401). You would want to copy the configuration options from `case 'minimal': ...` and add your additional rules while providing an object to `stats`.
```
byzyk: I guess we could wrap **e.g. minimal** piece in parentheses `()`. What do you think?

EugeneHlushko: I know what you mean, but doesn't make much sense as the example is completely about `minimal`. The url, the case and webpack config

montogeek:
Thanks!

PR #2619, opened 2018-10-26
Small clarification in Tables section of writers-guide.md

carolstran:
When I first read the writers guide, it took me a while to realize that the "Same goes for tables" was referring to the fact that they must be in alphabetical order (or at least... I'm pretty sure that's what it is referring to 😅)

screen shot 2018-10-26 at 13 58 06

In my opinion, it would be better to give it its own context, especially because people can link directly to the Tables section.


EugeneHlushko:
Thanks!

PR #2626, opened 2018-10-30
docs(cli): document --no-color option

jayjun:
Document `--no-color` option, and clarify colours are enabled (for TTY output only) if no options are given. See [supports-color](https://github.com/chalk/supports-color) for details.


montogeek:
Thanks!

PR #2630, opened 2018-10-31
Update why-webpack.md

Stef733:
fix typo

### DIFF
--- a/src/content/concepts/why-webpack.md
+++ b/src/content/concepts/why-webpack.md
@@ -8,7 +8,7 @@ contributors:
 - EugeneHlushko
 ---
 
-To understand why you should use webpack lets do a recap of how we use JavaScript on the web before bundlers were a thing.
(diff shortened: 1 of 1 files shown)


montogeek:
Not sure that is a typo.
cc @EugeneHlushko

Stef733:
It's easy to confuse lets and let's :). 

Let's is the shortened version of the words let us which is what is happening in the context of the sentence:
`To understand why you should use webpack lets do a recap of how we...` 
You can check this works by changing the word to let us: 
`To understand why you should use webpack let us do a recap of how we...`

Lets without the apostrophe can be thought of as a synonym to allow. E.g. `Webpack lets us bundle modules` is the same as `Webpack allows us to bundle modules`.

If this is confusing, this link explains it better than I can :) - https://www.dictionary.com/e/lets/

EugeneHlushko:
Thanks!

PR #2633, opened 2018-11-05
Update html-webpack-plugin.md

userkang:
Use MiniCssExtractPlugin instead of ExtractTextWebpackPlugin

### DIFF
--- a/src/content/plugins/html-webpack-plugin.md
+++ b/src/content/plugins/html-webpack-plugin.md
@@ -55,7 +55,7 @@ If you have multiple webpack entry points, they will all be included with `scrip
 tags in the generated HTML.
 
 If you have any CSS assets in webpack's output (for example, CSS extracted
-with the [ExtractTextPlugin](/plugins/extract-text-webpack-plugin))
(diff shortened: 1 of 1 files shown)


src/content/plugins/html-webpack-plugin.md
```diff
@@ -55,7 +55,7 @@ If you have multiple webpack entry points, they will all be included with `scrip
 tags in the generated HTML.
 
 If you have any CSS assets in webpack's output (for example, CSS extracted
-with the [ExtractTextPlugin](/plugins/extract-text-webpack-plugin))
+with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin))
```
EugeneHlushko: Please add trailing slash to the URL, our dev server setup fails to open without it.

userkang: you mean it should be corrected like this ?
 ```suggestion
with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin/))
```

EugeneHlushko: yes please

userkang: I have added

PR #2638, opened 2018-11-12
docs(plugins) move code example related warning to the code blocks

EugeneHlushko:
Move related code example warning to the code block itself in order to avoid confusion, many come to see how to use `NODE_ENV` so the previous block location was too deep without giving appropriate warning immediately.


montogeek:
Thanks!

PR #2641, opened 2018-11-13
Additional info about Dynamic entry

smelukov:
I've used dynamic entries to collect entry points from external files and when I was implementing this, docs were not clear about some details (how dynamic entry works) and I was forced to dig webpack source code.

### DIFF
--- a/src/content/configuration/entry-context.md
+++ b/src/content/configuration/entry-context.md
@@ -81,9 +81,9 @@ module.exports = {
 
 For example: you can use dynamic entries to get the actual entries from an external source (remote server, file system content or database):
 
-**webpack.config.js**
+__webpack.config.js__
 
-```js
+``` js
 module.exports = {
   entry() {
     return fetchPathsFromSomeExternalSource(); // returns a promise that will be resolved with something like ['src/main-layout.js', 'src/admin-layout.js']
--- a/src/content/configuration/entry-context.md
+++ b/src/content/configuration/entry-context.md
@@ -61,7 +61,7 @@ If a string or array of strings is passed, the chunk is named `main`. If an obje
 
 If a function is passed then it will be invoked on every [make](/api/compiler-hooks/#make) event.
 
-> Note that the `make` event triggers when webpack starts and for every invalidation in watch-mode.
+> Note that the make event triggers when webpack starts and for every invalidation when [watching for file changes](/configuration/watch/).
 
 ```js
 module.exports = {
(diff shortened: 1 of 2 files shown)


src/content/configuration/entry-context.md
```diff
@@ -59,6 +59,10 @@ If a string or array of strings is passed, the chunk is named `main`. If an obje
 
 ### Dynamic entry
 
+If a function is passed then it will be invoked on every [make](https://webpack.js.org/api/compiler-hooks/#make) event.
```
montogeek: Please use relative URLs: 
```suggestion
If a function is passed then it will be invoked on every [make](/api/compiler-hooks/#make) event.
```

smelukov: Thanks

src/content/configuration/entry-context.md
```diff
@@ -59,6 +59,10 @@ If a string or array of strings is passed, the chunk is named `main`. If an obje
 
 ### Dynamic entry
 
+If a function is passed then it will be invoked on every [make](/api/compiler-hooks/#make) event.
+
+> Note that the `make` event triggers when webpack starts and for every invalidation in watch-mode.
```
EugeneHlushko: This is contradicting to other mode option in webpack configuration. Let's reword. Also make event shouldn't be quoted for consistency with first occurrence 
```suggestion
> Note that the make event triggers when webpack starts and for every invalidation when [watching for file changes](/configuration/watch/).
```

src/content/configuration/entry-context.md
```diff
@@ -75,4 +79,6 @@ module.exports = {
 };
 ```
 
+> For example: you can use dynamic entries to get the actual entries from an external source (remote server, file system content or database)
```
EugeneHlushko: If we are going to provide an example, don't put it into a note block:
```suggestion
For example: you can use dynamic entries to get the actual entries from an external source (remote server, file system content or database)
```

Also when we provide any example context, code block should follow. Please add abstract example in format of:

__webpack.config.js__

```javascript
// webpack config with abstract example corresponding to the given context
```

smelukov: fixed, thanks

EugeneHlushko:
Looks good, need to fix CLA.

smelukov:
@EugeneHlushko we can’t moving forward till you unblock merging :(

smelukov:
Can we just merge this PR?

montogeek:
Thanks!

smelukov:
@montogeek thank you 😊

PR #2649, opened 2018-11-18
docs(concepts) Remove redundant word in configuration.md

HusamIbrahim:
Removed 'following' from 'The following examples below' as it kind of seems redundant.


montogeek:
Thanks!

PR #2659, opened 2018-11-20
ReSubmit: Add ref-loader to loader's introduction page

leecjson:
Resumit since #2657


EugeneHlushko:
Hi @leecjson,

Thanks so much for adding test cases and code comments so that we could also analyze if the package is suspicious or not!

I think its good to go now 👍

leecjson:
Thanks for help

Issue #2646, opened 2018-11-14
Feature for pressing / to start the search akin to how it works in vim/GitHub.


maraisr:
What more to add?

<img width="378" alt="screen shot 2018-11-15 at 8 03 38 am" src="https://user-images.githubusercontent.com/599459/48515648-01ae8180-e8ad-11e8-85e8-672712b5132b.png">

montogeek:
PR is welcome!

PR #2661, opened 2018-11-20
feat(nav): allowing the / key to open the search 

maraisr:
Adds the functionality to open the search window, similarly to how GitHub/vim does it with the `/` key.

Points to consider:
- the `/` key pops the search
- the `esc` key closes the search window
- added a new method `_closeSearch`
- added a defaulted argument to the `_openSearch` method
- _might be worth consolidating the `_toggleSearch` method to use the "open and close" methods?_

resolves: #2646

### DIFF
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -124,18 +124,21 @@ export default class Navigation extends React.Component {
       });
 
       // Keydown so we can "stop" the event if need be. 
-      window.addEventListener("keydown", e => {
+      window.addEventListener('keydown', e => {
         // Short circuit, as we only care about the "naked" key
         if (e.shiftKey || e.ctrlKey || e.metaKey) return;
       
         switch (e.which) {
           case 9: // `tab` key
-            if (e.target.classList.contains("navigation__search-input"))
+            if (e.target.classList.contains('navigation__search-input'))
               this._openSearch();
             break;
           case 191: // `/` key
-            !this._searchOpen && this._openSearch();
-            e.preventDefault(); // to block entering a `/` if the input is already in focus.
+            // We wrap this all in an if, so that the `/` key can still be used normally when the search is open.
+            if (!this._searchOpen) {
+              this._openSearch();
+              e.preventDefault(); // to block entering a `/` if the input is already in focus.
+            }
             break;
           case 27: // `esc` key
             this._searchOpen && this._closeSearch();
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -8,7 +8,7 @@ import Links from './Links.json';
 // TODO: Migrate to React Banner
 export default class Navigation extends React.Component {
 
-   // As this varaible is purly to track the state of the search,
+  // As this varaible is purly to track the state of the search,
   // and it's state doesn't affect the renders' output. And to
   // save react a re-render, we will not be storing this in the
   // components state object. 
@@ -127,7 +127,7 @@ export default class Navigation extends React.Component {
       window.addEventListener('keydown', e => {
         // Short circuit, as we only care about the "naked" key
         if (e.shiftKey || e.ctrlKey || e.metaKey) return;
-      
+
         switch (e.which) {
           case 9: // `tab` key
             if (e.target.classList.contains('navigation__search-input'))
@@ -145,7 +145,6 @@ export default class Navigation extends React.Component {
             break;
         }
       });
-      
 
     }
   }
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -119,9 +119,19 @@ export default class Navigation extends React.Component {
       });
 
       window.addEventListener('keyup', e => {
-        if (e.which === 9 && e.target.classList.contains('navigation__search-input')) {
-          this._openSearch();
+
+        switch(e.which) {
+          case 9: // `tab` key
+            if (e.target.classList.contains('navigation__search-input')) this._openSearch();
+            break;
+          case 191: // `/` key
+            this._openSearch(true);
+            break;
+          case 27: // `esc` key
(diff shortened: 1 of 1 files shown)


src/components/Navigation/Navigation.jsx
```diff
@@ -119,9 +119,19 @@ export default class Navigation extends React.Component {
...
       window.addEventListener('keyup', e => {
-        if (e.which === 9 && e.target.classList.contains('navigation__search-input')) {
-          this._openSearch();
+
+        switch(e.which) {
+          case 9: // `tab` key
+            if (e.target.classList.contains('navigation__search-input')) this._openSearch();
+            break;
+          case 191: // `/` key
+            this._openSearch(true);
```
EugeneHlushko: Suggest we introduce a var to track search bar's state, and when its open dont call `this._openSearch(true);` when user presses `/` for example when typing `/` into search input 
```
!this.searchIsOpen  && this._openSearch(true);
```

src/components/Navigation/Navigation.jsx
```diff
@@ -119,9 +119,19 @@ export default class Navigation extends React.Component {
       });
 
       window.addEventListener('keyup', e => {
-        if (e.which === 9 && e.target.classList.contains('navigation__search-input')) {
-          this._openSearch();
+
+        switch(e.which) {
```
EugeneHlushko: I think it makes sense to check for `ctrlKey`, `shiftKey` etc status and if they are pressed don't even go into the switch eg. for example for `CMD + ESC` or any other reasons like `shift + /`

EugeneHlushko:
```suggestion
        if (e.shiftKey || e.ctrlKey || e.metaKey) return;
        switch(e.which) {
```

src/components/Navigation/Navigation.jsx
maraisr: Thanks for that, I have made some to reflect this.

src/components/Navigation/Navigation.jsx
maraisr: Thanks mate, all done :)

src/components/Navigation/Navigation.jsx
```diff
@@ -118,11 +123,27 @@ export default class Navigation extends React.Component {
...
+        if (e.shiftKey || e.ctrlKey || e.metaKey) return;
+      
+        switch (e.which) {
+          case 9: // `tab` key
+            if (e.target.classList.contains("navigation__search-input"))
+              this._openSearch();
+            break;
+          case 191: // `/` key
+            !this._searchOpen && this._openSearch();
+            e.preventDefault(); // to block entering a `/` if the input is already in focus.
```
EugeneHlushko: just wondering, if there is another input on the page. what will happen if they focus on it and type `api/cli`. It would mean the input receives `api` and then rest of it will go to search input: `cli` :D 
@montogeek do you think we have such input anywhere on the docs?

src/components/Navigation/Navigation.jsx
```diff
@@ -118,11 +123,27 @@ export default class Navigation extends React.Component {
         inputSelector: '.navigation__search-input'
       });
 
-      window.addEventListener('keyup', e => {
-        if (e.which === 9 && e.target.classList.contains('navigation__search-input')) {
-          this._openSearch();
+      // Keydown so we can "stop" the event if need be. 
+      window.addEventListener("keydown", e => {
```
EugeneHlushko: Please apply for single quotes
```suggestion
      window.addEventListener('keydown', e => {
```

src/components/Navigation/Navigation.jsx
```diff
@@ -7,9 +7,14 @@ import Links from './Links.json';
 
 // TODO: Migrate to React Banner
 export default class Navigation extends React.Component {
-  render() {
-    let { pageUrl = '' } = this.props;
 
+   // As this varaible is purly to track the state of the search,
```
dhruvdutt: Can you fix the indentation issue here?

src/components/Navigation/Navigation.jsx
```diff
@@ -118,11 +123,30 @@ export default class Navigation extends React.Component {
         inputSelector: '.navigation__search-input'
       });
 
-      window.addEventListener('keyup', e => {
-        if (e.which === 9 && e.target.classList.contains('navigation__search-input')) {
-          this._openSearch();
+      // Keydown so we can "stop" the event if need be. 
+      window.addEventListener('keydown', e => {
+        // Short circuit, as we only care about the "naked" key
+        if (e.shiftKey || e.ctrlKey || e.metaKey) return;
```
dhruvdutt: Not sure if this is really needed. Wouldn't the switch case event codes ignore it?

EugeneHlushko: yes you need those checks, please test in browser for experiment.

src/components/Navigation/Navigation.jsx
montogeek: Not that I know

montogeek:
@EugeneHlushko Good to merge?

EugeneHlushko:
@montogeek PR looks good to me now, but i havent tested it locally, needs a check before we merge

maraisr:
@EugeneHlushko who can do this testing?

EugeneHlushko:
Either of maintainers, i will try to test it tonight, thanks for reminding @maraisr

EugeneHlushko:
I have updated your branch for test

EugeneHlushko:
Unfortunately i found a bug straight off trying your branch.
1. Open homepage (`localhost:3000/`)
2. press on kb `/`
3. press `esc` (to close search bar)
4. try clicking `documentation` link, the search bar becomes visible and disappears again, blocking the click.

I did not proceed with checking other things.

maraisr:
Sorry for wasting your time.

EugeneHlushko:
@maraisr please reopen the PR, as it only needs bug fixes

PR #2670, opened 2018-11-28
docs(config): fix typo by removing extraneous 'which'

doersino:
There was an extra word. Now there isn't.


EugeneHlushko:
Thanks!

PR #2686, opened 2018-12-03
correct the regex

njleonzhang:
correct regex for loader usage.


njleonzhang:
@EugeneHlushko done.

montogeek:
Thanks!

PR #2689, opened 2018-12-04
Update public-path.md with an example how to override it in plugin

Knagis:
The plugin allows more flexible use cases than requiring `__webpack_public_path__` being set in just the right place. This was very useful in our project that has many entry chunks that are dynamically loaded by our own loader - so we need a mechanism to set the public path reliably in every entry chunk.

### DIFF
--- a/src/content/guides/public-path.md
+++ b/src/content/guides/public-path.md
@@ -67,17 +67,17 @@ Similar result can be achieved by using a plugin to set the public path directly
 
 ```js
 export class DynamicPublicPathPlugin {
-    apply(compiler) {
-        compiler.hooks.thisCompilation.tap("DynamicPublicPathPlugin", compilation => {
-            const mainTemplate = compilation.mainTemplate;
-            const hook = mainTemplate.hooks.requireExtensions;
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Thank you for confirming, closing this for now. We are still able to continue discussion if necessary. CC @Knagis @montogeek

Knagis:
The reason why I think `__webpack_public_path__` is not elegant, is the fact that what seems to be a global variable is actually replaced with `__webpack_require__.p` and thus only work inside of particular chunk. Basically one has to be very careful to use it exactly right (it has to be first import, it has to be in every entry chunk etc.).

Before I wrote the plugin, I actually had the `__webpack_public_path__` assignment in code, but since we have one "loader" chunk that manually (not via require/import) loads other chunks, my assumptions about how it works and is applied were wrong and thus for a while most webpack requests ended up missing CDN. (due to the one issue that use of runtime chunk in webpack 4 changes chunk naming we are not using that)

That being said I am not pushing for it to be included in docs if you think it should not be recommended for general use.

PR #2693, opened 2018-12-05
fix(fetch-packages): remove only titles from loaders and plugins docs

### LINKED ISSUE #2692: style-loader page bug
When reading the docs about `style-loader` there's a heading missing text.

screen shot 2018-12-05 at 7 25 15 am

ivandevp:
When fetching packages title is removed by a regular expressions, but in `README.md` files like the one from [`style-loader`](https://github.com/webpack-contrib/style-loader/blob/master/README.md), there's no a title heading using the pattern `# ...`. Instead, its first line that matches this pattern is a sub-header. Since the purpose is only to replace a title, this PR changes the RegExp to match this kind of issues and maintaining other docs working as usual.

Fixes #2692

### DIFF
--- a/src/scripts/fetch_packages.js
+++ b/src/scripts/fetch_packages.js
@@ -130,7 +130,7 @@ function fetchPackageFiles(options, finalCb) {
             })
             // Replace lone h1 formats
             .replace(/<h1.*?>.+?<\/h1>/, '')
-            .replace(/# .+/, '')
+            .replace(/^# .+/m, '')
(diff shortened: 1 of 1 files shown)


montogeek:
Done in https://github.com/webpack/webpack.js.org/pull/2693.
@ivandevp Sorry, we are working a new version of the website, that work is under `rebuild` branch, I moved your commit into a new branch: https://github.com/webpack/webpack.js.org/tree/rebuild-fix-process-readme.

PR #2694, opened 2018-12-05
Add documentation for `Rule.use` as a function

bew:
Adds some documentation for using a function for `Rule.use` configuration.

For the object passed to the function, I'm not surewhat `realResource`, `compiler` & `resourceQuery` are for, and I'd appreciate some help to explain them succinctly , thank you

Fixes #2691

Issue #2691, opened 2018-12-05
Rule.use as a function is not documented


bew:
`Rule.use` configuration is documented when used with a list of `UseEntry`, but not when used with a function.

I'll probably open a PR with something :smiley:

EugeneHlushko:
Please go ahead with pr

PR #2697, opened 2018-12-06
Note that `Rule.use` and `UseEntry` can be functions

glenjamin:
I discovered this earlier this week buried on a github comment somewhere, so wanted to add this into the docs.

I'm not massively happy with the wording here, there weren't any notes about how best to document function arguments or return values in the writing guide, so I figure I'd post this up as-is for feedback.

### DIFF
--- a/src/content/configuration/module.md
+++ b/src/content/configuration/module.md
@@ -19,7 +19,7 @@ These options determine how the [different types of modules](/concepts/modules) 
 
 ## `module.noParse`
 
-`RegExp | [RegExp] | function | string | [string]`
+`RegExp` `[RegExp]` `function(resource)` `string` `[string]`
 
 Prevent webpack from parsing any files matching the given regular expression(s). Ignored files __should not__ have calls to `import`, `require`, `define` or any other importing mechanism. This can boost build performance when ignoring large libraries.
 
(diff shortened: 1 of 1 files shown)


glenjamin:
Oh whoops, looks like this overlaps with #2691 and #2694 - although those don't mention that individual `UseEntry` items can also be functions.

EugeneHlushko:
Hello @glenjamin 

I would suggest to continue with your PR after the first one gets merged since it was submitted earlier.

EugeneHlushko:
@glenjamin we had merged #2694 please proceed here after updating the branch

src/content/configuration/module.md
```diff
@@ -404,12 +404,21 @@ module.exports = {
...
 ## `UseEntry`
 
-`object`
+`object | function`
 
 It must have a `loader` property being a string. It is resolved relative to the configuration [`context`](/configuration/entry-context#context) with the loader resolving options ([resolveLoader](/configuration/resolve#resolveloader)).
 
 It can have an `options` property being a string or object. This value is passed to the loader, which should interpret it as loader options.
 
+A `UseEntry` can also be a function which receives the object argument describing the module being loaded, and must return an options object. This can be used to vary the loader options on a per-module basis.
```
montogeek: Could you please add an example?

glenjamin:
I've added some subheadings to separate the types a little bit, and put in an example for each.

src/content/configuration/module.md
```diff
@@ -19,7 +19,7 @@ These options determine how the [different types of modules](/concepts/modules)
 
 ## `module.noParse`
 
-`RegExp | [RegExp] | function | string | [string]`
+`RegExp` `[RegExp]`  `function(resource)`  `string` `[string]`
```
EugeneHlushko:
```suggestion
`RegExp` `[RegExp]` `function(resource)` `string` `[string]`
```

glenjamin:
Hrm, I can't seem to see a good way to call out the separate docs for each type without having the markdown linter yell at me.

EugeneHlushko:
It actually says that header is duplicated, i suggest you dont use headers there but just go with bold text:
```
```

EugeneHlushko:
@montogeek please have a second look at it mate i thinks its good to go from looking at it.

EugeneHlushko:
@glenjamin please try not to rebase and force-push as commit history is very important for doing quick reviews

glenjamin:
Oh sorry, I thought github was able to keep track!

montogeek:
Thanks!

PR #2699, opened 2018-12-06
add period to end of first paragraph in SplitChunksPlugin article

mskalandunas:
Added a period to the end of the first paragraph.  Didn't add myself to the contributors list because this particular change is formatting and insignificant, but can do that if it's required even for changes of this size.


montogeek:
Thanks!

PR #2704, opened 2018-12-07
docs: add group prop in writer's guide

hemal7735:
### What is a change?
Currently we are using sub-sections using group property for `MODULES` and `PLUGINS` on `/api/` page.
This PR adds information about how to add sub-section.

### DIFF
--- a/src/content/contribute/writers-guide.md
+++ b/src/content/contribute/writers-guide.md
@@ -20,6 +20,7 @@ Each article contains a small section at the top written in [YAML Frontmatter](h
 ``` yaml
 ---
 title: My Article
+group: My Sub-Section
 sort: 3
 contributors:
   - [github username]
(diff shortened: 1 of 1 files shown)


src/content/contribute/writers-guide.md
```diff
@@ -1,6 +1,12 @@
 ---
 title: Writer's Guide
 sort: 2
+contributors:
```
montogeek: Only add yourself to contributors list, some people choose not to appear on it.

hemal7735: I can understand for @lhoff and @carolstran .
@skipjack and @byzyk , can you help me understand is there anything which I'm missing about this page that contribs don't want to add their name, as your name is there on other pages?

skipjack: @hemal7735 I touched almost every page in this site when I was maintaining it a while back, so I only added my name to ones that I wrote or made significant changes to. I appreciate you adding me but I agree with @montogeek that some folks may not want to be mentioned so it’s better to just add yourself. (and no aside from a desire to remain anonymous there’s no reason not to add yourself)

montogeek:
Thanks!

PR #2722, opened 2018-12-16
fix(fetch-packages): remove only titles from loaders and plugins docs

montogeek:
Implements https://github.com/webpack/webpack.js.org/pull/2693 to `rebuild` branch.


montogeek:
Thanks @ivandevp !

PR #2727, opened 2018-12-18
docs(config) document optimization.chunkIds

### LINKED ISSUE #2724: Document optimization.chunkIds
**Feature to document**

[`optimization.chunkIds`](https://github.com/webpack/webpack/blob/d5e26f728adb63a1fca080ef728fd627952a921d/schemas/WebpackOptions.json#L406-L409) was introduced in webpack/webpack#7686 and released in [webpack@4.16.0](https://github.com/webpack/webpack/releases/tag/v4.16.0).

**Author(s)**

@sokra 

**Additional information**

`optimization.moduleIds` is already documented, so the documentation for `optimization.chunkIds` will probably be fairly similar.

EugeneHlushko:
Fixes #2724

### DIFF
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -200,6 +200,32 @@ module.exports = {
 };
 ```
 
+## `optimization.chunkIds`
+
+`bool: false` `string: natural, named, size, total-size`
+
+Tells webpack which algorithm to use when choosing chunk ids. Setting `optimization.chunkIds` to `false` tells webpack that none of built-in algorithms should be used, as custom one can be provided via plugin. By default `optimization.chunkIds` is set to `false`.
+
+The following string values are supported:
+
+Option                | Description
+--------------------- | -----------------------
+`natural`             | Numeric ids in order of usage.
+`named`               | Readable ids for better debugging.
+`size`                | Numeric ids focused on minimal initial download size.
+`total-size`          | numeric ids focused on minimal total download size.
+
+__webpack.config.js__
+
+```js
+module.exports = {
+  //...
+  optimization: {
+    chunkIds: 'named'
+  }
+};
+```
+
 ## `optimization.nodeEnv`
 
 `string` `bool: false`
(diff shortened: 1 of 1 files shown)


src/content/configuration/optimization.md
```diff
@@ -200,6 +200,32 @@ module.exports = {
 };
 ```
 
+## `optimization.chunkIds`
+
+`bool: false` `string: natural, named, size, total-size`
```
montogeek: Don't get this notation, it allows `boolean` and `string`?
What do you think about putting it in a new line?

src/content/configuration/optimization.md
```diff
@@ -200,6 +200,32 @@ module.exports = {
...
+
+Tells webpack which algorithm to use when choosing chunk ids. Setting `optimization.chunkIds` to `false` tells webpack that none of built-in algorithms should be used, as custom one can be provided via plugin. By default `optimization.chunkIds` is set to `false`.
+
+The following string values are supported:
```
montogeek: What do you think about `are also supported`?

src/content/configuration/optimization.md
```diff
@@ -200,6 +200,32 @@ module.exports = {
 };
 ```
 
+## `optimization.chunkIds`
+
+`bool: false` `string: natural, named, size, total-size`
+
+Tells webpack which algorithm to use when choosing chunk ids. Setting `optimization.chunkIds` to `false` tells webpack that none of built-in algorithms should be used, as custom one can be provided via plugin. By default `optimization.chunkIds` is set to `false`.
```
nwoltman: From the current [code on master](https://github.com/webpack/webpack/blob/f2f62c76080dd34eb319a6031fe688b7a4716cf9/lib/WebpackOptionsApply.js#L413-L427), it looks like `optimization.chunkIds` defaults to `natural`.

It would also default to `named` in `development` and `total-size` in `production`, though that's only if the defaults for `optimization.namedChunks` and `optimization.occurenceOrder` aren't modified (which hopefully they wouldn't be since they were deprecated in [v4.16.0](https://github.com/webpack/webpack/releases/tag/v4.16.0)).

EugeneHlushko:
@nwoltman @montogeek its updated

nwoltman:
Awesome. Looks great @EugeneHlushko!

montogeek:
Thanks!

PR #2731, opened 2018-12-21
Added information on settings stats colors

grgur:
Added information on setting stats colors using ANSI escape sequences.


montogeek:
Thanks!

PR #2735, opened 2018-12-24
docs(guides): Update author-libraries.md

chenxsan:
Now that ExtractTextWebpackPlugin is no longer recommended since webpack 4, I think it's better to update the link.


montogeek:
Thanks!

PR #2740, opened 2018-12-27
docs(config): add stats option `outputPath`

terrierscript:
Append stats option `outputPath`
https://github.com/webpack/webpack/blob/master/lib/Stats.js#L214-L217


src/content/configuration/stats.md
```diff
@@ -146,6 +146,9 @@ module.exports = {
     // Show dependencies and origin of warnings/errors (since webpack 2.5.0)
     moduleTrace: true,
 
+    // Show outputPath
+    outputPath: true,
```
EugeneHlushko: is this true by default?

terrierscript: @EugeneHlushko 
This default value decided by `forToString`

And I found this document not collected like chunks and source.

* `chunks: true`
  * https://github.com/webpack/webpack/blob/master/lib/Stats.js#L156
* `source: false`
  * https://github.com/webpack/webpack/blob/master/lib/Stats.js#L188

Which case should I suit?

EugeneHlushko: By reading the source code it is not obvious, i would suggest running stats and checking in debugger the default value, can you try that @terrierscript ?

terrierscript:
This PR is stale. I close it

EugeneHlushko:
Hi @terrierscript if we cannot find out default value now, let's change it to `outputPath: bool` and merge. Someone can improve it later

terrierscript:
@EugeneHlushko Sorry, I haven't time to fix this.

src/content/configuration/stats.md
```diff
@@ -146,6 +146,9 @@ module.exports = {
     // Show dependencies and origin of warnings/errors (since webpack 2.5.0)
     moduleTrace: true,
 
+    // Show outputPath
+    outputPath: true | false,
```
terrierscript: @EugeneHlushko Does this mean what you say?

EugeneHlushko: Yes

EugeneHlushko:
Thanks!

PR #2743, opened 2018-12-28
document when optimization.minimizer is fn

hulkish:
Added docs for minimizer used as function. Related: https://github.com/webpack/webpack/pull/8573


src/content/configuration/optimization.md
```diff
@@ -40,27 +40,42 @@ T> Learn how [mode](/concepts/mode/) works.
 
 ## `optimization.minimizer`
 
-`[TerserPlugin]`
+`array` of `<plugin>` and/or `function`
```
EugeneHlushko: this is not the right way (to be consistent with the rest of the docs):
```suggestion
`[<plugin>]` and or `[function (compiler)]`
```

EugeneHlushko: that is if other plugins are supported if not, better to go with `[TerserPlugin]`

hulkish:
@EugeneHlushko As you can see in https://github.com/webpack/webpack/pull/8573 - This was actually merged to master. So, maybe I should also point this pr to master?

src/content/configuration/optimization.md
EugeneHlushko: Yes pointing to master makes total sense, please re-aplly PR fix that was lost

hulkish: @EugeneHlushko ok, all set

PR #2747, opened 2019-01-03
docs(devServer): add `writeToDisk` option

AlexDubok:
docs(devServer): add `writeToDisk` option

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -1226,3 +1226,36 @@ module.exports = {
 If this is too heavy on the file system, you can change this to an integer to set the interval in milliseconds.
 
 See [WatchOptions](/configuration/watch/) for more options.
+
+
+## `devServer.writeToDisk` 🔑
+
+`boolean` `function`
+
+Default: `false`. Tell dev-server to write generated assets to the disk.
+
+__webpack.config.js__
+
+```javascript
+module.exports = {
+  //...
+  devServer: {
+    writeToDisk: true
+  }
+};
+```
+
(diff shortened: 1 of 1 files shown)


src/content/configuration/dev-server.md
```diff
@@ -1226,3 +1226,36 @@ module.exports = {
 If this is too heavy on the file system, you can change this to an integer to set the interval in milliseconds.
 
 See [WatchOptions](/configuration/watch/) for more options.
+
+
+## `devServer.writeToDisk` 🔑
+
+`boolean` `function`
```
EugeneHlushko: please add default value here and remove below. please add function signature here e.g. `function (param1, param2)`

src/content/configuration/dev-server.md
```diff
@@ -1226,3 +1226,36 @@ module.exports = {
...
+
+Default: `false`. Tell dev-server to write generated assets to the disk.
```
```suggestion
Tells `devServer` to write generated assets to the disk.
```

src/content/configuration/dev-server.md
```diff
@@ -1226,3 +1226,36 @@ module.exports = {
...
+```javascript
+module.exports = {
+  //...
+  devServer: {
+    writeToDisk: true
+  }
+};
+```
+
+This option also accepts a Function value, which can be used to filter which files are written to disk. The function follows the same premise as [`Array#filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) in which a return value of false will not write the file, and a return value of true will write the file to disk.
```
```suggestion
Providing a `Function` to `devServer.writeToDisk` can be used for filtering. The function follows the same premise as [`Array#filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) in which a boolean return value tells if the file should be written to disk.
```

src/content/configuration/dev-server.md
AlexDubok: Done. thank you

EugeneHlushko:
Thanks!

PR #2749, opened 2019-01-04
Update installation.md to reference `npm build` vs. `npm start`

blimmer:
In the follow-up steps in the guide, `npm run build` is always referenced, not `npm run start`. It seems like in the initial installation section `build` should be the name of the npm task vs. `start`.

screen shot 2019-01-04 at 09 29 33


montogeek:
Thanks!

blimmer:
No problem - thanks for the quick review.

PR #2758, opened 2019-01-12
docs(Stats): Adds reference to Stats.presetToOptions static method.

jbottigliero:
- This actually seems like the best way for someone to consume one of these configuration levels (and would potentially lead to a bit more "evergreen" documentation)...

If desired, I can rework this section to promote this method with less emphasis on copying the `case` statements.

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -202,7 +202,7 @@ module.exports = {
 };
 ```
 
-Or, simply use the `Stats.presetToOptions` static method when using webpack programmatically.
+Or use the `Stats.presetToOptions` static method when using webpack programmatically.
 
 __webpack.config.js__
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/stats.md
```diff
@@ -202,6 +202,23 @@ module.exports = {
 };
 ```
 
+Or, simply use the `Stats.presetToOptions` static method when using Webpack programmatically.
```
montogeek: Typo `webpack`

jbottigliero: Fixed – thanks!

src/content/configuration/stats.md
```diff
@@ -202,6 +202,23 @@ module.exports = {
 };
 ```
 
+Or, simply use the `Stats.presetToOptions` static method when using webpack programmatically.
```
EugeneHlushko:
```suggestion
Or use the `Stats.presetToOptions` static method when using webpack programmatically.
```

jbottigliero:
@EugeneHlushko  – that works, thanks for the follow-up!

PR #2759, opened 2019-01-13
lint(links) disable link linter on travis

EugeneHlushko:
disable links linter on travis

```
ok 99 load dist/guides/public-path
ok 100 load dist/guides/integrations
ok 101 load dist/api/cli
ok 102 load dist/api/node
ok 103 load dist/api/module-methods
ok 104 load dist/loaders
ok 105 load dist/concepts/loaders
ok 106 load dist/api
ok 107 load dist/writers-guide/index.html
_http_client.js:114
      throw new ERR_UNESCAPED_CHARACTERS('Request path');
      ^

TypeError [ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped characters
    at new ClientRequest (_http_client.js:114:13)
    at Object.request (https.js:280:10)
    at Immediate.dispatchRequest (webpack.js.org/node_modules/teepee/lib/Teepee.js:605:79)
    at runCallback (timers.js:694:18)
    at tryOnImmediate (timers.js:665:5)
    at processImmediate (timers.js:647:5)
error Command failed with exit code 1.
```

PR #2762, opened 2019-01-14
Fix unencoded ✭ character in img src

papandreou:
Fix unencoded star character in img src url (`https://img.shields.io/github/stars/webpack/memory-fs.svg?label=✭&style=flat-square&maxAge=3600`)

Caused `npm run linkcheck` to break.

This is a hopefully nicer alternative to disabling the link check: https://github.com/webpack/webpack.js.org/pull/2759

We'll look into making `hyperlink` either tolerate this kind of invalid url (if browsers tolerate it), or throw a specific error message instead of an internal error.

### DIFF
--- a/src/components/Organization/Organization.jsx
+++ b/src/components/Organization/Organization.jsx
@@ -26,7 +26,7 @@ const Organization = props => {
               <h6>Downloads and Stars</h6>
               <Shield content={ `npm/dm/${org.npm}`} label="npm" />
               &nbsp;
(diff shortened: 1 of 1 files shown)


Munter:
When the tests could run without throwing, another problem showed itself: https://github.com/webpack-contrib/cache-loader/pull/60

I'll merge this one though, because it clearly fixes a problem, and the broken link that was introduced is a part of an external repository

PR #2766, opened 2019-01-16
docs(guides): Add "npm build" instruction on "Loading Data" section

Vanary:
Added instruction for `npm build` on the `guides/asset-management`.

This change would prevent reader from confusion if they forgot to build module before check upon `dist/index.html`.

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -483,19 +483,7 @@ __src/index.js__
   document.body.appendChild(component());
 ```
 
-Now once again, run your build command:
-
-``` bash
-npm run build
-
-...
-    Asset      Size  Chunks             Chunk Names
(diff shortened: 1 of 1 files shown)


src/content/guides/asset-management.md
```diff
@@ -483,6 +483,18 @@ __src/index.js__
   document.body.appendChild(component());
 ```
 
+Now once again, run your build command:
```
EugeneHlushko: We had already given example output of the build command and instruction to run the build itself, i would suggest adding this info to the next paragraph e.g.
```
Re-run the build command and open `index.html`....
```

Vanary:
I like the simplicity of suggestion. Followed review by:
- Removed build & CLI result from first commit
- Added information to re-run build command as suggested

EugeneHlushko:
Thanks!

PR #2768, opened 2019-01-17
docs(getting-started): use var instead to match the rest of the guides

kelvinau:
I understand that there's a PR for this similar issue already https://github.com/webpack/webpack.js.org/pull/2614, but it's been 3 months and we haven't had any updates from the last PR regarding to the `const` reassignment fix.

A simpler fix will be only changing **docs(getting-started)** to replace it `let` with `var`, as that's the only part inconsistent with the others.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -53,7 +53,7 @@ __src/index.js__
 
 ``` javascript
 function component() {
-  let element = document.createElement('div');
+  var element = document.createElement('div');
 
   // Lodash, currently included via a script, is required for this line to work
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Hello @kelvinau 
I understand how critical it is to update this example with the modern `let` syntax, but i would suggest we don't rush and wait two more weeks here and give original author a warning or something.

What do you think @montogeek ?

kelvinau:
Understood. I can make the changes for `let` and `const` instead too, if @anayadesign hasn't responded after.

montogeek:
@kelvinau Friendly ping

kelvinau:
This was fixed by #2962 which extended #2614 already.

PR #2769, opened 2019-01-17
Minor formatting tweaks

Zearin:
(Signed agreement, yadda yadda…)

I’m very used to editing markdown, but unfamiliar with the `T> ` syntax for “Tip” (and the others).

Dunno how to add a second paragraph in a single `T> ` block, so I just used 2 HTML line breaks.


src/content/concepts/entry-points.md
```diff
@@ -67,7 +68,7 @@ Below is a list of entry configurations and their real-world use cases:
 
 ### Separate App and Vendor Entries
 
-T> In webpack version < 4 it was common to add vendors as separate entrypoint to compile it as separate file (in combination with the `CommonsChunkPlugin`). This is discouraged in webpack 4. Instead the `optimization.splitChunks` option takes care of separating vendors and app modules and creating a separate file. __Do not__ create a entry for vendors or other stuff which is not the starting point of execution.
+T> In webpack versions < 4 it was common to add vendors as separate entrypoint to compile it as separate file (in combination with the `CommonsChunkPlugin`). <br><br>This is discouraged in webpack 4. Instead the `optimization.splitChunks` option takes care of separating vendors and app modules and creating a separate file. __Do not__ create a entry for vendors or other stuff which is not the starting point of execution.
```
EugeneHlushko: Can you please attach a screenshot of how it looks when you run it locally?

hemal7735: Hi @EugeneHlushko GitHub provides nice support for viewing `.md` file changes. For this change you can view at: https://github.com/webpack/webpack.js.org/pull/2769/files?short_path=1a429d5#diff-1a429d5163eda631c32ed927c408db47 

For webpack website we will still need a screenshot.

EugeneHlushko: Hi @hemal7735 
Thanks for highlighting github.com's feature to me, but i need a screenshot on the website when author does that change locally. We did not use html in md on other changes IIRC so would be nice to see the effect of it. Also it might not be a recommended way to maintain .md files so i suggest waiting for more reviews

montogeek:
This is how it looks
image

montogeek:
Thanks!

PR #2770, opened 2019-01-18
Update split-chunks-plugin.md

cymen:
splitChunks.chunks is a key within object so example should have it point to a function not be a function

The documentation has a function within the `splitChunks` object definition but as it is within an object, it needs to be `chunks` key that references the function -- it cannot be an inline function.


EugeneHlushko:
Hi @cymen 
Unfortunately you are not right, this is another ecmascript feature.

screen shot 2019-01-19 at 11 35 24 am

PR #2771, opened 2019-01-18
docs(stats): Set stats.assetsByChunkName[x] to arrays

superburrito:
Updating the sample stats file in response to #2745


EugeneHlushko:
Hi @superburrito 
We don't merge future release's content into `master` yet, we are currently on v4 of webpack. You can reopen such pr and target `next` branch, but your branch should also be created from `next`.

superburrito:
@EugeneHlushko Ah I see, thanks for the pointer, will keep that in mind for future PRs.

PR #2775, opened 2019-01-19
docs(stats): Set stats.assetsByChunkName[x] to arrays

superburrito:
Updating the sample stats file in response to #2745


EugeneHlushko:
Thanks!

PR #2778, opened 2019-01-24
docs(config) fix url and update format in plugins page

EugeneHlushko:
- fix an old url
- update type and add example file name
- reword `list` to `array`


montogeek:
Thanks!

PR #2779, opened 2019-01-24
docs(config) stats preset extension and overrides

EugeneHlushko:
In response to #2728 
It only takes one bit. So doesn't close the huge doc request issue itself.


montogeek:
Thanks!

PR #2797, opened 2019-02-07
added missing path.

nomadme:
Without this path, the script will fail to load.


nomadme:
didn't realize that the index.html was moved to `dist` folder earlier.

PR #2798, opened 2019-02-07
Adding minor getting started clarifications

jceipek:
I'm just trying out webpack and the installation and environment variables sections threw me off a bit. Hopefully these minor changes will help others avoid my confusion.

1. The installation section explains that you need to install the cli for webpack 4 but could be clearer that this is in addition to the `webpack` package. I added "also".

2. The environment variables section does not make clear that webpack's environment variables are separate from the environment variables I might set in `bash`, `fish`, `zsh`, etc... I added a tip to that effect.

### DIFF
--- a/src/content/guides/environment-variables.md
+++ b/src/content/guides/environment-variables.md
@@ -13,7 +13,9 @@ related:
     url: https://blog.flennik.com/the-fine-art-of-the-webpack-2-config-dc4d19d7f172#d60a
 ---
 
-To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use webpack "environment variables" (not to be confused with the [environment variables](https://en.wikipedia.org/wiki/Environment_variable) of the shell where you're running webpack).
+To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use webpack "environment variables".
+
+T> webpack's environment variables are different from the [environment variables](https://en.wikipedia.org/wiki/Environment_variable) of the shell where you're running webpack
 
 The webpack command line [environment option](/api/cli/#environment-options) `--env` allows you to pass in as many environment variables as you like. Environment variables will be made accessible in your `webpack.config.js`. For example, `--env.production` or `--env.NODE_ENV=local` (`NODE_ENV` is conventionally used to define the environment type, see [here](https://dzone.com/articles/what-you-should-know-about-node-env).)
 
--- a/src/content/guides/environment-variables.md
+++ b/src/content/guides/environment-variables.md
@@ -7,12 +7,13 @@ contributors:
   - tbroadley
   - legalcodes
   - byzyk
+  - jceipek
 related:
   - title: The Fine Art of the webpack 3 Config
     url: https://blog.flennik.com/the-fine-art-of-the-webpack-2-config-dc4d19d7f172#d60a
 ---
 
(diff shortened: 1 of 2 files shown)


src/content/guides/environment-variables.md
```diff
@@ -7,12 +7,15 @@ contributors:
...
+  - jceipek
 related:
   - title: The Fine Art of the webpack 3 Config
     url: https://blog.flennik.com/the-fine-art-of-the-webpack-2-config-dc4d19d7f172#d60a
 ---
 
-To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use environment variables.
+To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use "environment variables".
+
+T> webpack's environment variables are different from the [environment variables](https://en.wikipedia.org/wiki/Environment_variable) you might be used to from the shell where you're running webpack
```
EugeneHlushko: i don't think that quoting them before and putting war here is correct, can you remove the quotes above.

src/content/guides/environment-variables.md
```diff
@@ -7,12 +7,15 @@ contributors:
...
   - legalcodes
   - byzyk
+  - jceipek
 related:
   - title: The Fine Art of the webpack 3 Config
     url: https://blog.flennik.com/the-fine-art-of-the-webpack-2-config-dc4d19d7f172#d60a
 ---
 
-To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use environment variables.
+To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use "environment variables".
```
EugeneHlushko:
```suggestion
To disambiguate in your `webpack.config.js` between [development](/guides/development) and [production builds](/guides/production) you may use environment variables.
```

src/content/guides/environment-variables.md
jceipek: I was aiming to use quotations to indicate uncommon usage (see https://style.mla.org/quotes-when-nothing-is-being-quoted/), but I'm fine with getting rid of them.

EugeneHlushko: Yes i understand, but it looked weird when i read through that version of the page, same term with different styling in sibling paragraphs... Another thing here is that we are doing an assumption in the docs. Recently we'd received many PRs removing such statements. Can we get rid of `you might be ....` and just state the facts

jceipek: Changed. I understand the rationale (for example, this commit is fantastic: https://github.com/webpack/webpack.js.org/pull/1791/commits/be909876e9e84dfd41bc0cac888d773df71c9fc0), but in this case, I think it might be slightly worse, since the sentence is referring to a concept unrelated to webpack. My use of the phrase "you might be used to from" was intended to cause readers unfamiliar with the idea of environment variables and operating system shells to disregard the tip. Now such a reader needs to confront their own level of understanding with the concept of shells and environment variables.

jceipek: On a related note, the writing contribution guide makes no mention of writing style. It would be great if a maintainer could edit it to provide examples of phrasing that would or would not be appropriate, based on the design goals of the documentation.

EugeneHlushko: Very good point! that is tbd soon

EugeneHlushko:
Thanks!

PR #2799, opened 2019-02-08
docs(concepts) Remove redundant words in mode.md

mianlang:
Redundant words removed.
Seems like someone forgot to delete them in some previous edition.

Similar phrases can be found several lines above:
>Possible values for mode are: none, development or production(default).


EugeneHlushko:
Ouch! Good finding. Thanks

PR #2802, opened 2019-02-11
docs(config) mention cli config generator

EugeneHlushko:
Mention internal config generation capability


EugeneHlushko:
Will merge as its approved, ty for looking at 👍

PR #2805, opened 2019-02-12
Fix wrong alignment of green outline on Open Collective backers

Munter:
Changes method of creating green outline around Open Collective backers from an extra dom element with quirky positioning to a CSS box shadow.

**Before**
webpack js org_

**After**
localhost_3000_


montogeek:
Thanks!

PR #2806, opened 2019-02-13
infra(lint) make links check manual

EugeneHlushko:
as per title, to cut build times

PR #2807, opened 2019-02-13
Bring back linting of internal links

Munter:
https://github.com/webpack/webpack.js.org/pull/2806 disabled **all** link checking.

The slowest task we wanted to avoid was the external link check, which was allowed to fail anyway. But not the internal link check, which ensures we don't have any broken links internal to the documentation page.

This change brings back the internal checks. This is also the faster of the two link checks


montogeek:
And in just under 5 mins.
Thanks!

PR #2809, opened 2019-02-14
site(vote) fix vote landing url

### LINKED ISSUE #2808: https://webpack.js.org/vote JS failure
https://webpack.js.org/vote has a javascript error that blocks the page from being rendered, causing the default `Your browser is not supported` message from the raw HTML to appear.

image

If I load any other page and navigate to `/vote` through the JS-router the page shows up just fine

EugeneHlushko:
Fixes #2808


montogeek:
Why is this the fix? Something we should update in the webpack config?

EugeneHlushko:
hi @montogeek 

I didnt think of alias first, good catch! Check it out

montogeek:
That works too, but I mean the code splitting when using `import()`, We removed its support in the migration, What happened? :D

EugeneHlushko:
I couldnt see difference locally, let me try to play a full build when im free

montogeek:
Thanks @EugeneHlushko

PR #2810, opened 2019-02-15
docs(configuration): add a missing option

g-plane:
The `output.globalObject` is useful for UMD build, which should be documented.

Ref: https://github.com/webpack/webpack/issues/6522

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -16,6 +16,7 @@ contributors:
   - harshwardhansingh
   - eemeli
   - EugeneHlushko
+  - g-plane
 ---
 
 The top-level `output` key contains set of options instructing webpack on how and where it should output your bundles, assets and anything else you bundle or load with webpack.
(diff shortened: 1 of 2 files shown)


src/content/configuration/output.md
```diff
@@ -340,6 +341,11 @@ If using a function for this option, the function will be passed an object conta
 T> When using the [`ExtractTextWebpackPlugin`](/plugins/extract-text-webpack-plugin), use `[contenthash]` to obtain a hash of the extracted file (neither `[hash]` nor `[chunkhash]` work).
 
 
+## `output.globalObject`
+
```
EugeneHlushko:
```suggestion
`string: 'window'`

```

src/content/configuration/output.md
```diff
@@ -340,6 +341,11 @@ If using a function for this option, the function will be passed an object conta
...
+When targeting a library, especially the `libraryTarget` is `'umd'`, this option indicates what global object will be used to mount the library. Default value is `'window'`. If you want to make your UMD build available on both browsers and Node.js, you should set this option to `'this'`.
```
EugeneHlushko:
```suggestion
When targeting a library, especially the `libraryTarget` is `'umd'`, this option indicates what global object will be used to mount the library. To make UMD build available on both browsers and Node.js, set `output.globalObject` option to `'this'`.
```

g-plane:
@EugeneHlushko What's wrong with the Travis job?

EugeneHlushko:
@Munter can you check why svg throw content type mismatch?

Munter:
@EugeneHlushko 	It's an undesirable side effect of a strict matching of content types based on what type of relations are pointing at an asset. I think a newer version of assetgraph (hyperlink dependency) has an improvement here.

I'm marking this specific error as TODO in the tests for now so they should pass until I can update hyperlink

EugeneHlushko:
Got it, thank you @Munter 
I've also resolved the conflict and updated the branch

EugeneHlushko:
Thanks!

Issue #2804, opened 2019-02-12
Improve documentation for loader order


evilebottnawi:
**Feature to document**

Old docs https://github.com/webpack/docs/wiki/loaders#loader-order

In new docs we don't have information about `!!` and `-!`

**Author(s)**

No matter

**Additional information**

No

superburrito:
Hmm I can look into this

cc @EugeneHlushko

EugeneHlushko:
Go ahead @superburrito

EugeneHlushko:
Open a WIP PR with a mention of this issue so its trackable

superburrito:
@evilebottnawi This info can be found [here](https://webpack.js.org/configuration/module#ruleenforce), but I've tried to make that info more discoverable in [the above PR](https://github.com/webpack/webpack.js.org/pull/2811)

PR #2811, opened 2019-02-15
Add info about loader overrides

superburrito:
Fixes https://github.com/webpack/webpack.js.org/issues/2804

Update: Turns out that the docs do already talk about the overrides [here](https://webpack.js.org/configuration/module#ruleenforce), but this PR attempts to make this information more obvious.

@EugeneHlushko If you find that it's not necessary, we can close this PR

### DIFF
--- a/src/content/api/loaders.md
+++ b/src/content/api/loaders.md
@@ -17,7 +17,6 @@ The first loader is passed one argument: the content of the resource file. The c
 
 A single result can be returned in __sync mode__. For multiple results the `this.callback()` must be called. In __async mode__ `this.async()` must be called to indicate that the [loader runner](https://github.com/webpack/loader-runner) should wait for an asynchronous result. It returns `this.callback()`. Then the loader must return `undefined` and call that callback.
 
-TODO: Add info about loader overrides
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


src/content/api/loaders.md
```diff
@@ -95,7 +96,11 @@ module.exports.raw = true;
 
 ### Pitching Loader
 
-Loaders are __always__ called from right to left. There are some instances where the loader only cares about the __metadata__ behind a request and can ignore the results of the previous loader. The `pitch` method on loaders is called from __left to right__ before the loaders are actually executed (from right to left). For the following [`use`](/configuration/module#rule-use) configuration:
+Loaders are __always__ called from right to left. There are some instances where the loader only cares about the __metadata__ behind a request and can ignore the results of the previous loader. The `pitch` method on loaders is called from __left to right__ before the loaders are actually executed (from right to left). 
+
+Note that loaders may be added inline in requests and disabled via inline prefixes, which will impact the order in which they are "pitched" and executed. See the section on [`enforce`](/configuration/module#ruleenforce) for more details.
```
EugeneHlushko: How about we use an info block here?
```suggestion
T> Loaders may be added inline in requests and disabled via inline prefixes, which will impact the order in which they are "pitched" and executed. See [`Rule.enforce`](/configuration/module#ruleenforce) for more details.
```

superburrito: Sounds good to me

EugeneHlushko:
Thanks!

PR #2819, opened 2019-02-21
Update docs for module resolution

asayamakk:
This PR suggests to changing to document from `resolve.mainFields` to  `resolve.mainFiles` when resolving a module directory

Hello, I am new to webpack and I found that module resolution document might pointing wrong link.
In the resolve.mainFiles section, resolve.mainFiles is set for
> The filename to be used while resolving directories.

So I thought when resolving a module directory, webpack will refer `resolve.mainFiles` instead of `resolve.mainFields`

### DIFF
--- a/src/content/concepts/module-resolution.md
+++ b/src/content/concepts/module-resolution.md
@@ -65,7 +65,7 @@ Once the path is resolved based on the above rule, the resolver checks to see if
 
 If the path points to a folder, then the following steps are taken to find the right file with the right extension:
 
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Can you back this up with a link to related webpack source code please? Because i have an impression this is a guess, it looks right though i would like us to have a ground for this change outside of our guessings

asayamakk:
Thanks for comment. I will find the source code that reinforce my guessing. Or at least I try out these options and see how actual behavior changes.

EugeneHlushko:
Hi @asayamakk ,

Any updates on this PR?

montogeek:
I am not sure which one is correct: https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L1148

EugeneHlushko:
This needs testing, can anyone put up a reproduction repo?

PR #2822, opened 2019-02-22
Update LimitChunkCountPlugin docs

erykpiast:
Remove info about `minChunkSize` argument and refer to `MinChunkSizePlugin` plugin.


src/content/plugins/limit-chunk-count-plugin.md
```diff
@@ -43,23 +44,7 @@ module.exports = {
...
-module.exports = {
-  // ...
-  plugins: [
-    new webpack.optimize.LimitChunkCountPlugin({
-      minChunkSize: 1000
-    })
-  ]
-};
-```
+To keep chunk size above the specified limit by merging chunks that are smaller than the limit, use [MinChunkSizePlugin)[/plugins/min-chunk-size-plugin]
```
EugeneHlushko: Why is this needed here if its not done via this plugin anymore?

erykpiast: Aren’t plugins a valid way of configuring Webpack anymore?

EugeneHlushko: i mean if there is a plugin that does exactly this, move this info to its page?

erykpiast: There is already such info on the page I linked to. I've added the link to make migration from old API to new one a bit easier. It's more like "hey, this feature was moved somewhere else, if you search for it, look at here".

EugeneHlushko: Okay, that is still what i think is not really obvious from the text. The text says: "To do X, use plugin X", when you are saying it is intended to be "Hey, feature X has moved to plugin X". Or am i reading this wrong?

erykpiast: You're reading it right. I'll try to make the message clearer :)

src/content/plugins/limit-chunk-count-plugin.md
```diff
@@ -44,7 +44,7 @@ module.exports = {
 
 ### `minChunkSize`
 
-To keep chunk size above the specified limit by merging chunks that are smaller than the limit, use [MinChunkSizePlugin)[/plugins/min-chunk-size-plugin]
+Keeping chunk size above the specified limit is no longer a feature of this plugin. Use [MinChunkSizePlugin)[/plugins/min-chunk-size-plugin] instead.
```
EugeneHlushko: Worth mentioning a version since when that happened as a reference or is that sufficient enough? CC @montogeek

montogeek: It is enough

EugeneHlushko:
Thanks!

PR #2823, opened 2019-02-22
Update targets.md

dany74q:
Without the additional white space, the first letter (i.e. 'F') does not render.


EugeneHlushko:
Hi @dany74q,

Unfortunately, it doesn't really matter. Because this is a TODO item you are correcting. There is another TODO item beneath it as well. Maybe you want to actually expand the docs with more examples for popular targets instead or expand the docs on popular `target` values?

montogeek:
Thanks!

PR #2824, opened 2019-02-22
Updated index.mdx

dany74q:
Grammar: 'an extensions' -> 'an extension'

### DIFF
--- a/src/content/configuration/index.mdx
+++ b/src/content/configuration/index.mdx
@@ -286,7 +286,7 @@ module.exports = {
       // properites that are read from description file
       // to alias requests in this package
       enforceExtension: false,
-      // if true request must not include an extensions
+      // if true request must not include an extension
       // if false request may already include an extension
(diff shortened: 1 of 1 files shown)


src/content/configuration/index.mdx
```diff
@@ -286,7 +286,7 @@ module.exports = {
       // properites that are read from description file
       // to alias requests in this package
       enforceExtension: false,
-      // if true request must not include an extensions
+      // if true request must not include an extension
```
EugeneHlushko: I think this misses a comma too

hemal7735:
```suggestion
      // if true, request must not include an extension
```
like this @EugeneHlushko ?

EugeneHlushko: yes, but on the other hand i am not sure it is important, this is a comment in JavaScript example snippet

montogeek:
Thanks!

Issue #2815, opened 2019-02-19
https://webpack.js.org/api/  style confusing 


Aladingding:
The style is confusing when I expanded the two menus in https://webpack.js.org/api/

this bug screenshot
https://files.gitter.im/webpack/webpack/dev/LJX7/thumb/menu.saveimg.savepath20190219111622.jpg

montogeek:
@Aladingding Could you please tell us which menu looks weird? I can't see it in the screenshot.

Aladingding:
@montogeek  if i expend Compiler Hooks menu && Command Line Interface && Introduction the footer's style will be confusing of this page,with 1920*1080 windows 10

EugeneHlushko:
Oh I see so the footer is not pushed down when two are open

EugeneHlushko:
When there is not enough content*

PR #2825, opened 2019-02-22
fix(styles): don't let sidebar crossover footer

pierreneter:
Close: #2815

### DIFF
--- a/src/components/Sidebar/Sidebar.scss
+++ b/src/components/Sidebar/Sidebar.scss
@@ -6,7 +6,6 @@
   display: none;
   width: 100%;
   max-width: 280px;
-  max-height: 100%;
   will-change: transform;
 
   @include break {


EugeneHlushko:
Did not work for me locally, you sure it solved it on your end?

src/components/Site/Site.scss
```diff
@@ -11,5 +11,9 @@
 
   &__sidebar {
     flex: 0 0 280px;
+    &::after {
```
EugeneHlushko:
```suggestion

    &::after {
```

pierreneter:
Sure, didn't work, that's not the way to fix it. I'm confused, next commit is right thing....

EugeneHlushko:
@montogeek wanna give it a shot?

montogeek:
Thanks!

PR #2829, opened 2019-02-24
fix a little typo

aularon:
fixing a typo removing function's parenthesis, as we are assigning the function reference rather than its call return value


EugeneHlushko:
Thanks!

PR #2835, opened 2019-02-28
Enable horizontal scrolling for wide tables

dogafincan:
Currently, tables that are too wide aren't displayed completely. See https://webpack.js.org/comparison, for example.

The CSS already had overflow-x and overflow-y values in place, but they weren't doing much as the tables weren't rendered as a block.


EugeneHlushko:
Thanks! Great fix

PR #2836, opened 2019-02-28
Use let instead of var

mrseanbaines:
The previous ('getting started') section contains some of the same example code using `let` instead of `var`. So, to keep consistency, and also to use the latest JavaScript, this PR changes the use of `var` to `let`.

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -8,6 +8,7 @@ contributors:
   - sudarsangp
   - chenxsan
   - EugeneHlushko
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Hi @mrseanbaines 
There are already similar PRs:
https://github.com/webpack/webpack.js.org/pull/2768/files
https://github.com/webpack/webpack.js.org/pull/2614/files

My call is - support open PRs vs open new PRs on the same change. Unless you don't agree, i suggest closing this one

PR #2845, opened 2019-03-05
docs(guides): change CleanWebpackPlugin arguments usage

Ivocin:
CleanWebpackPlugin released  v2.0.0 and the API has a breaking change: No configuration is needed for standard usage (cleans webpack's output.path).

`new CleanWebpackPlugin(['dist/*'])` with the new version clean-webpack-plugin installed above will cause an error:
"Error: clean-webpack-plugin only accepts an options object. See: https://github.com/johnagan/clean-webpack-plugin#options-and-defaults-optional "

### DIFF
--- a/src/content/guides/output-management.md
+++ b/src/content/guides/output-management.md
@@ -194,7 +194,7 @@ __webpack.config.js__
       print: './src/print.js'
     },
     plugins: [
-+     new CleanWebpackPlugin(['dist/*']),
++     new CleanWebpackPlugin(),
       new HtmlWebpackPlugin({
         title: 'Output Management'
       })
--- a/src/content/guides/output-management.md
+++ b/src/content/guides/output-management.md
@@ -194,7 +194,7 @@ __webpack.config.js__
       print: './src/print.js'
     },
     plugins: [
-      // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
++     // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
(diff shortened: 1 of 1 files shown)


src/content/guides/output-management.md
```diff
@@ -194,7 +194,7 @@ __webpack.config.js__
       print: './src/print.js'
     },
     plugins: [
-+     new CleanWebpackPlugin(['dist/*']),
++     new CleanWebpackPlugin(),
```
EugeneHlushko:
```suggestion
       // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
+     new CleanWebpackPlugin(),
```

src/content/guides/output-management.md
```diff
@@ -194,7 +194,8 @@ __webpack.config.js__
       print: './src/print.js'
     },
     plugins: [
-+     new CleanWebpackPlugin(['dist/*']),
+       // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
```
EugeneHlushko: I'm sorry but editing in github got an extra space there
```suggestion
      // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
```

Ivocin:
> It looks good to me, i am hesitant if we should also add a `+` to the comment line so its highlighted....
> 
> screen shot 2019-03-06 at 11 43 38 am

add a `+` is a good idea, because the previous demo code above  doesn't have this line.

montogeek:
Thanks!

PR #2847, opened 2019-03-05
docs(concepts): add missing space in targets.md

dogafincan:
The missing space causes the first letter of the line to disappear.


montogeek:
Thanks!

PR #2849, opened 2019-03-06
Update for change in CleanWebpackPlugin

nikolajbaer:
CleanWebpackPlugin now requires an "options" object, or no args. For the tutorial, it is sufficient to pass in no arguments to achieve the same 'dist/*' functionality.

This change prevents an error from occurring while following the tutorial.


montogeek:
Thanks for your contribution, unfortunately this is duplicate of https://github.com/webpack/webpack.js.org/pull/2845, please check existing PRs before submitting yours.

PR #2850, opened 2019-03-06
docs: improve api docs

wizardofhogwarts:


src/content/api/compilation-hooks.md
```diff
@@ -15,7 +16,7 @@ During the compilation phase, modules are loaded, sealed, optimized, chunked,
 hashed and restored.
 
 The `Compilation` class also extends `Tapable` and provides the following
-lifecycle hooks. They can be tapped the same way as compiler hooks:
+life-cycle hooks. They can be tapped the same way as compiler hooks:
```
EugeneHlushko: I think this can also be `life cycle` and `lifecycle`. Why updating it now?

src/content/api/loaders.md
```diff
@@ -185,12 +186,12 @@ require('./loader1?xyz!loader2!./resource?rrr');
 
 ### `this.version`
 
-__Loader API version.__ Currently `2`. This is useful for providing backwards compatibility. Using the version you can specify custom logic or fallbacks for breaking changes.
+__Loader API version.__ Currently `2`. This is useful for providing backward compatibility. Using the version you can specify custom logic or fallbacks for breaking changes.
```
EugeneHlushko: Why is this necessary?

src/content/api/loaders.md
```diff
@@ -337,7 +338,7 @@ T> Loaders were originally designed to also work as Babel transforms. Therefore
 
 ### `this.sourceMap`
 
-Should a source map be generated. Since generating source maps can be an expensive task, you should check if source maps are actually requested.
+Should a source map be generated? Since generating source maps can be an expensive task, you should check if source maps are actually requested.
```
EugeneHlushko: This might be better imo
```suggestion
Tells if source map should be generated. Since generating source maps can be an expensive task, you should check if source maps are actually requested.
```

src/content/api/loaders.md
```diff
@@ -480,7 +481,7 @@ A boolean flag. It is set when in debug mode.
 
 ### `this.minimize`
 
-Should the result be minimized.
+Should the result be minimized?
```
EugeneHlushko:
```suggestion
Tells if result should be minimized.
```

src/content/api/module-methods.md
```diff
@@ -167,7 +168,7 @@ W> Module ID is a number in webpack (in contrast to NodeJS where it is a string
 
 ### `require.cache`
 
-Multiple requires to the same module result in only one module execution and only one export. Therefore a cache in the runtime exists. Removing values from this cache cause new module execution and a new export.
+Multiple requires in the same module result in only one module execution and only one export. Therefore a cache in the runtime exists. Removing values from this cache cause new module execution and a new export.
```
EugeneHlushko: I think this actually ment: "multiple requires of a specific module", so please revert

EugeneHlushko:
```suggestion
Multiple requires of the same module result in only one module execution and only one export. Therefore a cache in the runtime exists. Removing values from this cache cause new module execution and a new export.
```

EugeneHlushko: Or change to of as suggested

src/content/api/stats.mdx
```diff
@@ -176,7 +177,7 @@ The `errors` and `warnings` properties each contain a list of strings. Each stri
 ``` bash
 ../cases/parsing/browserify/index.js
 Critical dependencies:
-2:114-121 This seem to be a pre-built javascript file. Even while this is possible, it's not recommended. Try to require to original source to get better results.
+2:114-121 This seems to be a pre-built javascript file. Even while this is possible, it's not recommended. Try to require the original source to get better results.
```
EugeneHlushko: Please revert, this is example output of an error which is not controlled by the docs website.

src/content/api/compilation-hooks.md
wizardofhogwarts: I just felt that life-cycle is better.

src/content/api/loaders.md
wizardofhogwarts: In this context, backward makes more sense.

src/content/api/stats.mdx
wizardofhogwarts: oh! I hadn't noticed that.

src/content/api/module-methods.md
wizardofhogwarts: @EugeneHlushko yeah, it looks better with `of`.

src/content/api/stats.mdx
wizardofhogwarts: @EugeneHlushko I just committed to change it back to seem

src/content/api/loaders.md
EugeneHlushko: I have read an article on it, it seems both versions are correct. Unless there is more reading to it?

src/content/api/compilation-hooks.md
EugeneHlushko: Same goes for life-cycle, i had found few resources on the internet which ensure me that all three versions can be used, and be correct. Unless this is wrong/misused word i would suggest reverting the changes to not change for the sake of change

wizardofhogwarts: Ok i'll revert it.

src/content/api/loaders.md
wizardofhogwarts: Well, what I just figured out is that `backward` is American variant. So, changing it won't create extraordinary improvement. So, reverting it.

src/content/api/compilation-hooks.md
```diff
@@ -541,6 +541,8 @@ compilation.hooks.afterOptimizeChunkAssets.tap('MyPlugin', chunks => {
 ```
 
 
+
```
EugeneHlushko: remove this extra 2 blank lines please

src/content/api/stats.mdx
```diff
@@ -176,7 +177,7 @@ The `errors` and `warnings` properties each contain a list of strings. Each stri
 ``` bash
 ../cases/parsing/browserify/index.js
 Critical dependencies:
-2:114-121 This seem to be a pre-built javascript file. Even while this is possible, it's not recommended. Try to require to original source to get better results.
+2:114-121 This seem to be a pre-built javascript file. Even while this is possible, it's not recommended. Try to require the original source to get better results.
```
EugeneHlushko: this looks like output from other software, please revert if so

wizardofhogwarts: @EugeneHlushko to two times was lookin awkward that's why i did this. What u say?

src/content/api/compilation-hooks.md
wizardofhogwarts: @EugeneHlushko Oh! my bad. Reverted. Thanks.

src/content/api/stats.mdx
wizardofhogwarts: @EugeneHlushko just reverted the output to it's original state

wizardofhogwarts:
@EugeneHlushko Now check the suggested changes. They are implemented as per your suggestions.

src/content/api/stats.mdx
```diff
@@ -5,6 +5,7 @@ contributors:
   - skipjack
   - franjohn21
   - byzyk
+  - wizardofhogwarts
```
EugeneHlushko: This isn't necessary yet :)

wizardofhogwarts: ohh yeah!! @EugeneHlushko thanks for figuring out that.

EugeneHlushko:
Lets see the tests pass after updating to latest master

EugeneHlushko:
Thanks!

PR #2863, opened 2019-03-12
docs(contribute): Adds contributors for fix #2862

### LINKED ISSUE #2862: Unexpected behaviour of markdowns when contributors is null
### Introduction
Visit this [page](https://webpack.js.org/contribute/writers-guide) and expand the `Writer's Guide` section in the sidebar. You might see something like this :point_down: 
Screenshot from 2019-03-12 19-19-27

There's an unexpected `sort :2`.

### Cause
This is because for this doc the contributors filed is not defined.

### Solution
Adding `contributors` to the doc fixed the issue.Might be some issue with the parsing.

pranshuchittora:
Temporary fix to #2862


EugeneHlushko:
Thanks!

PR #2870, opened 2019-03-14
infa(ignore) ignore links checker artifact

EugeneHlushko:
Add links checker artifact to .gitignore


EugeneHlushko:
Merging without extra review as it is minor.

PR #2857, opened 2019-03-11
Remove links inside the code block in the loaders.md.

1010543618:
Remove links inside the code block.

I‘m confused about that does JS have a new syntax, when I frist see that. Finally I realized that's Markdown syntax.

PR #2872, opened 2019-03-14
Fix syntax mistake in loaders.md

scriptype:
Loader links removed from sample configuration. They were appearing as literal markdown statements. And since the pages they were linking to don't seem to exist, I just removed them.

### DIFF
--- a/src/content/concepts/loaders.md
+++ b/src/content/concepts/loaders.md
@@ -12,6 +12,7 @@ contributors:
   - byzyk
   - debs-obrien
   - EugeneHlushko
+  - scriptype
 ---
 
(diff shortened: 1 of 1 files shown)


scriptype:
Oh, looks like those pages exist at the root level of documentation. But anyways, I don't think anyone will want to follow links in some source code sample, or whether or not doing so is even possible with markdown.

1010543618:
Duplicate of #2857

1010543618:
My be we should expend [remark-refractor](https://github.com/montogeek/remark-refractor) make these links available.

EugeneHlushko:
We do support links in code blocks. Sorry for late response.

PR #2873, opened 2019-03-14
Use webpack bundle default path in example code

yaph:
Changed the script source in example code from ``main.js`` to ``./dist/main.js``, which is the default path without a configuration file.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -171,7 +171,7 @@ __dist/index.html__
    </head>
    <body>
 -    <script src="./src/index.js"></script>
-+    <script src="main.js"></script>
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.md
```diff
@@ -171,7 +171,7 @@ __dist/index.html__
    </head>
    <body>
 -    <script src="./src/index.js"></script>
-+    <script src="main.js"></script>
++    <script src="./dist/main.js"></script>
```
rishabh3112: `main.js` is in `dist`, So as `index.html`. Therefore the previous path is the one we should have here IMO

EugeneHlushko:
Closing this for now, please shout if we should reopen

PR #2875, opened 2019-03-15
Add correct reference for closure-webpack-plugin

yankouskia:
add correct reference for closure-webpack-plugin


EugeneHlushko:
Thanks!

PR #2883, opened 2019-03-17
Update asset-management.md

jprado:
Adds missing comma.


jprado:
So the user enters the trailing comma. On second thought this is just too trivial to recommend. I'm just closing this out instead.

PR #2887, opened 2019-03-19
infra(eslint) force single quote for js?x

EugeneHlushko:
- force single quotes for strings in js?x for consistency reasons
- fix current occurence


montogeek:
Thanks!

PR #2901, opened 2019-03-26
Doc: remove not existing cli options

kohver:
As I can see, webpack-dev-server doesn't have these options https://github.com/webpack/webpack-dev-server/blob/master/lib/options.json.

But it would nice to have them by the way.


EugeneHlushko:
Doesn't it forward that to webpack?

kohver:
> Doesn't it forward that to webpack?

Sorry, not sure I understand you.

EugeneHlushko:
You are stating CLI does not forward to webpack these options: --entry, --output-path
Right?

kohver:
Does it have to?

EugeneHlushko:
I was under impression it does forward them to webpack and they would override the defaults, my question is: is that what you find out? cli ignores these options now?

kohver:
> cli ignores these options now

Sorry, I checked it again and looks like doesn't ignore. Not sure how it actually works, but if I add an invalid argument it'll be an error:

```
npm run webpack-dev-server -- --entry=./src/index2.js

> package-name@1.0.0 webpack-dev-server /Users/user/package-name
> webpack-dev-server "--entry=./src/index2.js"

ℹ ｢wds｣: Project is running at http://localhost:9000/webpack-dev-server/
ℹ ｢wds｣: webpack output is served from /
✖ ｢wdm｣: Hash: 1c850c2ddb86940ab16e
Version: webpack 4.29.6
Time: 54ms
Built at: 04/08/2019 12:00:07 PM

ERROR in Entry module not found: Error: Can't resolve './src/index2.js' in '/Users/user/package-name'
```

So I'm closing this request.

PR #2904, opened 2019-03-28
Fix clipped header on homepage

superburrito:
Fixes https://github.com/webpack/webpack.js.org/issues/2854

Before:

Screen Shot 2019-03-28 at 9 53 44 PM

After:

Screen Shot 2019-03-28 at 9 53 18 PM

cc @Ajacmac


EugeneHlushko:
Thanks!

PR #2909, opened 2019-03-31
Update loaders.mdx

subugupta:
minor update to document implying the order is either (right to left) or (bottom to top)


montogeek:
Thanks!

PR #2911, opened 2019-03-31
Handle cases when the scaffold doesn't have a git repo

anshumanv:
Presently if a listed scaffold isn't a git repo, the link redirects to `/` so fixed that and just rendered a non-clickable card.

@evenstensberg


src/components/Scaffolding/Scaffolding.jsx
```diff
@@ -22,31 +22,54 @@ import {
...
-                <div className="product-title">
-                <h3>{hit.name}</h3> 
-                </div>
-                <div className="product-avatar">
-                  <img src={hit.owner.avatar} height="40" width="40"/>
-                </div>
+       {
+          hit.repository && hit.repository.url ? (
+            <a target="_blank" rel="noopener noreferrer"
+              href={hit.repository && hit.repository.url ? hit.repository.url: '/'}>
```
evenstensberg: No need to check twice? :)

evenstensberg: Try to find the npm link if gh is not present. Agnolia grabs data from npm registry

anshumanv: Fixed the approach, first check is needed if hit is available so we can render because if hit's not available list won't be rendered, 2nd is to check is gh link is present, if not use npm link.

src/components/Scaffolding/Scaffolding.jsx
```diff
@@ -22,30 +22,33 @@ import {
...
-              </div>
-              <div className="product-name">
-              <h3>{hit.owner.name}</h3>
-              </div>
-              
-              <div className="product-desc">
-                <p>{htmlDecode(hit.description)}</p>
-              </div>
+       {
+          hit && (
```
evenstensberg: Hit already has the metadata you need, no need for hard coded value

anshumanv: It doesn't have the package link, so I'm using the name from the metadata to get to the package.

PR #2913, opened 2019-04-01
Clarify behavior of resolve.extensions.

### LINKED ISSUE #2899: The order of `resolve.extensions` matters, but should not
# Bug report

The order of `resolve.extensions` matters. Specifically, this line of configuration:

    resolve : {extensions: ['.js', '.jsx', '.ts', '.tsx'] }

leads to `.ts` files not being compiled, even if there is a `rule` for Typescript. On the other hand, this line:

    resolve : {extensions: ['.ts', '.tsx', '.js', '.jsx'] }

works, and leads to `.ts` files being compiled.

Take a project that is currently setup to use webpack, and change the order of `resolve.extensions` to put the `.ts` extension(s) after the `.js` extension(s). Then notice that the `.ts` files are no longer compiled.

EDIT: Sorry, apparently the order only _sometimes_ matters. I haven't been able to reproduce this with a small example.

Either these two lines should behave identically to each other:

    resolve : {extensions: ['.ts', '.tsx', '.js', '.jsx'] }
    resolve : {extensions: ['.js', '.jsx', '.ts', '.tsx'] }

Or the documentation should be _very clear_ about how and why the order of  `resolve.extensions` matters.

webpack version:  4.29.6
Node.js version: NA
Operating System: Linux
Additional tools:  typescript

justinpombrio:
I [was confused](https://github.com/webpack/webpack.js.org/issues/2899) about the behavior of resolve.extensions. This is my attempt to clarify the docs.

Please only accept this PR if you believe what I wrote is accurate.

EDIT: 
Fixes #2899

### DIFF
--- a/src/content/configuration/resolve.md
+++ b/src/content/configuration/resolve.md
@@ -198,7 +198,7 @@ module.exports = {
 
 Attempt to resolve these extensions in order.
 
-If multiple files share the same name but have different extensions, resolve the one with the extension listed first and skip the rest.
+W> If multiple files share the same name but have different extensions, webpack will resolve the one with the extension listed first in the array and skip the rest.
 
 __webpack.config.js__
 
--- a/src/content/configuration/resolve.md
+++ b/src/content/configuration/resolve.md
@@ -196,7 +196,9 @@ module.exports = {
 
 `[string]: ['.wasm', '.mjs', '.js', '.json']`
 
-Automatically resolve certain extensions.
+Attempt to resolve these extensions in order.
+
+If multiple files share the same name but have different extensions, resolve the one with the extension listed first and skip the rest.
 
 __webpack.config.js__
 


src/content/configuration/resolve.md
```diff
@@ -196,7 +196,9 @@ module.exports = {
 
 `[string]: ['.wasm', '.mjs', '.js', '.json']`
 
-Automatically resolve certain extensions.
+Attempt to resolve these extensions in order.
+
+If multiple files share the same name but have different extensions, resolve the one with the extension listed first and skip the rest.
```
EugeneHlushko:
```suggestion
W> If multiple files share the same name but have different extensions, webpack will resolve the one with the extension listed first in the array and skip the rest.
```

EugeneHlushko:
Thanks!

PR #2914, opened 2019-04-01
docs(infra) format scss files for maintainability

EugeneHlushko:
- format scss files for maintainability


montogeek:
We should add prettier

montogeek:
Thanks!

PR #2917, opened 2019-04-02
Escape backslashes in mdx files

lukasgeiter:
This PR fixes several not escaped backslashes in `.mdx` files. Without a second escaping backslash they are either not visible or interpreted as special character (e.g. `\n`)


EugeneHlushko:
Tested, approved

EugeneHlushko:
Thanks!

Issue #2922, opened 2019-04-07
new Dev Server option: http2


Loonride:
**Feature to document**

New Dev Server `http2` option: https://github.com/webpack/webpack-dev-server/pull/1721

**Author(s)**

Myself. Mainly reviewed by @evilebottnawi 

**Additional information**

N/A

EugeneHlushko:
Closing via #2923

PR #2923, opened 2019-04-07
docs(dev-server): New http2 option

Loonride:
_describe your changes..._

Issue: https://github.com/webpack/webpack.js.org/issues/2922

I documented the new `devServer.http2` option.
- Included API/CLI examples for self-signed and custom certificates
- Detailed relations of this with `https` option

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -479,7 +479,9 @@ webpack-dev-server --hot-only
 
 Serve over HTTP/2 using [spdy](https://www.npmjs.com/package/spdy). This option is ignored for Node 10.0.0 and above, as spdy is broken for those versions. The dev server will migrate over to Node's built-in HTTP/2 once Express supports it.
 
-When this option is enabled but the server is unable to serve over HTTP/2, the server defaults to HTTPS. If this option is not explicitly set to `false`, it will default to `true` when using HTTPS.
+If this option is not explicitly set to `false`, it will default to `true` when using HTTPS. When this option is enabled but the server is unable to serve over HTTP/2, the server defaults to HTTPS.
+
+HTTP/2 with a self-signed certificate:
 
 __webpack.config.js__
 
@@ -492,7 +494,7 @@ module.exports = {
 };
 ```
 
-Provide your own certificate using the [https](#devserver-https) option:
+Provide your own certificate using the [https](#devserverhttps) option:
 
 __webpack.config.js__
 
@@ -514,6 +516,12 @@ Usage via the CLI
 
 ```bash
 webpack-dev-server --http2
+```
+
+To pass your own certificate via the CLI use the following options
+
+```bash
+webpack-dev-server --http2 --key /path/to/server.key --cert /path/to/server.crt --cacert /path/to/ca.pem
 ```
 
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/dev-server.md
```diff
@@ -472,6 +473,58 @@ webpack-dev-server --hot-only
 ```
 
 
+## `devServer.http2`
+
+`boolean`
```
EugeneHlushko: This means default: `true`

```suggestion
`boolean: true`
```

src/content/configuration/dev-server.md
```diff
@@ -472,6 +473,58 @@ webpack-dev-server --hot-only
...
+
+Serve over HTTP/2 using [spdy](https://www.npmjs.com/package/spdy). This option is ignored for Node 10.0.0 and above, as spdy is broken for those versions. The dev server will migrate over to Node's built-in HTTP/2 once Express supports it.
```
```suggestion
Serve over HTTP/2 using [spdy](https://www.npmjs.com/package/spdy). This option is ignored for Node 10.0.0 and above, as spdy is broken for those versions. The dev server will migrate over to Node's built-in HTTP/2 once [Express](https://expressjs.com/) supports it.
```

src/content/configuration/dev-server.md
```diff
@@ -472,6 +473,58 @@ webpack-dev-server --hot-only
...
+
+If this option is not explicitly set to `false`, it will default to `true` when using HTTPS. When this option is enabled but the server is unable to serve over HTTP/2, the server defaults to HTTPS.
```
```suggestion
If `devServer.http2` is not explicitly set to `false`, it will default to `true` when using HTTPS. When this option is enabled but the server is unable to serve over HTTP/2, the server defaults to HTTPS.
```

EugeneHlushko: We already said this option once in previous paragraph, lets keep the context with explicitly naming it here.

src/content/configuration/dev-server.md
```diff
@@ -472,6 +473,58 @@ webpack-dev-server --hot-only
...
+    https: {
+      key: fs.readFileSync('/path/to/server.key'),
+      cert: fs.readFileSync('/path/to/server.crt'),
+      ca: fs.readFileSync('/path/to/ca.pem'),
+    }
+  }
+};
+```
+
+Usage via the CLI
```
wizardofhogwarts:
```suggestion
Usage via CLI
```
sounds better to me

src/content/configuration/dev-server.md
```diff
@@ -472,6 +473,58 @@ webpack-dev-server --hot-only
...
+
+```bash
+webpack-dev-server --http2
+```
+
+To pass your own certificate via the CLI use the following options
```
wizardofhogwarts:
```suggestion
To pass your own certificate via CLI, use the following options
```

src/content/configuration/dev-server.md
Loonride: @EugeneHlushko it is not entirely `true` by default, it only defaults to `true` when `https` is enabled, which is not a default. If the `https` option is not included, it will default to `false`

EugeneHlushko: In that case we mark it as `false` by default and mention the combination case (which you already did later in the text)
```suggestion
`boolean: false`
```

EugeneHlushko:
Thanks!

PR #2929, opened 2019-04-10
docs(api) fix links on api loaders page

EugeneHlushko:
Fixing link anchors on api loaders page


montogeek:
Why this wasn't catch by the link checker?

EugeneHlushko:
Hard to say, but i've bumped onto similar PR last week on different page, looks like plugin update changed anchors or smth

EugeneHlushko:
Thanks for checking it 👍

PR #2931, opened 2019-04-11
Change Heading Level of "Magic Comments" Section

dhurlburtusa:
The "Magic Comments" section applies to the `import()` (level-3) sub-section of the `ES6` (level: 2) section.
Changing to level-4 will make it clear that the magic comments apply to `ES6` > `import()`.


montogeek:
Thanks!

PR #2937, opened 2019-04-12
Update define-plugin.md

liuuu:
This PR fixed documentation for the DefinePlugin plugin.

### DIFF
--- a/src/content/plugins/define-plugin.md
+++ b/src/content/plugins/define-plugin.md
@@ -68,8 +68,7 @@ After passing through webpack with no minification results in:
 if (!true) {
   console.log('Debug info');
 }
-if (true) {
-  console.log('Production log');
+if (fasle) {
 }
 ```
 


src/content/plugins/define-plugin.md
```diff
@@ -68,8 +68,7 @@ After passing through webpack with no minification results in:
 if (!true) {
   console.log('Debug info');
 }
-if (true) {
-  console.log('Production log');
+if (fasle) {
```
montogeek: Shouldn't be `false`?

liuuu: the compiled file' content is `if(false){}`.  there's no content inside the curly brace

EugeneHlushko:
How does it fix it @liuuu 
We are setting `PRODUCTION` to `JSON.stringify(true)`. Why would that become false?

Suggest we close it.

PR #2941, opened 2019-04-13
Grammar correction

craigdanj:
Small grammar correction. Should be either "content has changed" or "contents have changed". 'Content has changed' sounds more appropriate.


EugeneHlushko:
Thanks!

PR #2943, opened 2019-04-15
docs(plugins) Correct tag

EugeneHlushko:
Use real tag as opposed to the next similar reference to `<link>` tag...

### DIFF
--- a/src/content/plugins/html-webpack-plugin.md
+++ b/src/content/plugins/html-webpack-plugin.md
@@ -54,7 +54,7 @@ This will generate a file `dist/index.html` containing the following:
 
 If you have multiple webpack entry points, they will all be included with `<script>` tags in the generated HTML.
 
-If you have any CSS assets in webpack's output (for example, CSS extracted with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin/)) then these will be included with `<link>` tags in the HTML head.
+If you have any CSS assets in webpack's output (for example, CSS extracted with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin/)) then these will be included with `<link>` tags in the `<head>` element of generated HTML.
 
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


src/content/plugins/html-webpack-plugin.md
```diff
@@ -51,12 +52,9 @@ This will generate a file `dist/index.html` containing the following:
 </html>
 ```
 
-If you have multiple webpack entry points, they will all be included with `script`
-tags in the generated HTML.
+If you have multiple webpack entry points, they will all be included with `<script>` tags in the generated HTML.
```
pranshuchittora: **in** the generated HTML -> **inside** the generated HTML

src/content/plugins/html-webpack-plugin.md
```diff
@@ -51,12 +52,9 @@ This will generate a file `dist/index.html` containing the following:
...
 
-If you have any CSS assets in webpack's output (for example, CSS extracted
-with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin/))
-then these will be included with `<link>` tags in the HTML head.
+If you have any CSS assets in webpack's output (for example, CSS extracted with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin/)) then these will be included with `<link>` tags in the HTML head.
```
pranshuchittora: If you have any CSS assets in webpack's output (for example, CSS extracted with the [MiniCssExtractPlugin](/plugins/mini-css-extract-plugin/)) then those will be included with `<link>` tags inside the `<head>` of HTML.

src/content/plugins/html-webpack-plugin.md
EugeneHlushko: I would keep **in**, why change?

src/content/plugins/html-webpack-plugin.md
EugeneHlushko: i dont see a benefit of changing **in** to **inside**, but i like the idea of the rephrasing

src/content/plugins/html-webpack-plugin.md
pranshuchittora: Hmm... Ok

montogeek:
Thanks!

pranshuchittora:
Awesome 🚀🤘

PR #2945, opened 2019-04-16
docs(config) correct caching default, fix minors

### LINKED ISSUE #2934: Cache documention is wrong
**Feature to document**

`cache` option in webpack config has incorrect documentation. It states that cache is automatically enabled if the compiler is in watch mode. 

It turns out turning on watch mode doesn't automatically turn on the cache option. It looks like this is actually only the case if `mode` is set to `development`: https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsDefaulter.js#L39

webpack-dev-middleware doesn't automatically enable it either; it puts the compiler in watch mode here: https://github.com/webpack/webpack-dev-middleware/blob/master/index.js#L51

This doesn't affect the cache option in any way. So unless `mode` is development, it looks like you have to explicitly set `cache` to true. 

**Author(s)**

@sokra ?

**Additional information**

N/A

EugeneHlushko:
- fixes #2934 
- minor fixes (types, rephrase `simply`)

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -47,7 +47,7 @@ As it happens, the AMD support in webpack ignores the defined name anyways.
 
 ## `bail`
 
-`boolean`
+`bool`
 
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks!

Issue #1548, opened 2017-08-25
Dev - OpenCollective Image Lag


pierreneter:
![loading time](https://user-images.githubusercontent.com/12815768/29734226-30ee8536-8a1b-11e7-986e-593e91062876.PNG)

Location: Vietnam,
My speed:
![speedtest](http://beta.speedtest.net/result/6570012365.png)

Most of the time it takes to load the home page from the avatars of the backer.
Avatars are provided from `opencollective` and we can not manage the cache settings on the header.
Of course, these avatars will also be able to change frequently, so we should not cache it.

![backer avatars](https://user-images.githubusercontent.com/12815768/29734345-e3d86fd6-8a1b-11e7-84a2-57534c76e2b3.PNG)

I think browser cache management is just a temporary solution.
For the first time, when the cache is not saved it still takes a lot of time to load the page.
So, skip the cache control problem.

We really need to solve this problem in a different way.

So...

What is the best way for this?

skipjack:
@pierreneter I'll add a note about this in #1525, but I'm really not sure what we can do to resolve this seeing as were not serving the content. Being an open source project, there's only so much we can do in terms of hosting infrastructure and in this case we aren't even serving those images ourselves (as you pointed out). Also, the page is interactive/scrollable before those images load (in my experience at least) so I'm not sure how much we gain by refactoring.

For the reasons listed above, my initial inclination was to close this and mark as `wontfix` so we can keep the discussions focused on content improvements but, with #1400 in progress (and hopefully #1525 soon as well) I think another load time optimization wouldn't hurt. 

What do you think of [this approach](https://varvy.com/pagespeed/defer-images.html)? We could use a placeholder image and defer loading the rest until after the initial page load. It would be nice if `<img>` tags supported a `defer` attribute out of the box but I'm not sure anything like that exists as of yet.

I am going to move this to a note in #1525, but feel free to keep discussing here and potentially open a PR. Once we've plowed through a bit more of the content-related backlog, I wouldn't be against re-opening as long as it stays active. 

Also note that you should have gotten/be getting invites to be a collaborator and join our slack channel. Thanks again for your work so far!

pierreneter:
About this issue, I think I should focus on finishing #1400 and #1525 before coming back to it.
Thank you for making me part of the team.

`Deferring images` is a good way. We will touch upon it after completing the unfinished work that we are doing and discuss it further.

TheDutchCoder:
[...]

PR #2956, opened 2019-04-19
speed(homepage) offload images until in viewport

EugeneHlushko:
Because of the terrible load on the homepage visits, i am offloading img loading until supporters are in the viewport.

Related: #1548
Takes one off #1525


EugeneHlushko:
What can i say... If you are not sure, check the source code.

pranshuchittora:
@EugeneHlushko  I meant, I am not sure about the viability of the package. Hence I prefer to implement the feature using Intersection Observer API.

EugeneHlushko:
Really concerned with browser support around Intersection Observer.
So i've looked at our visitors browser/OS versions for the past 90 days in GA, and it looks like we don't really have IE 11 or Safari pre Intersection Observer visitors (not even 0.05%).

I am planning to make a HOC and wrap them into it.

EugeneHlushko:
Decided to not go with another HOC, rather extend my reusable component ;] @byzyk

src/components/Support/Support.jsx
```diff
@@ -61,9 +62,10 @@ function formatMoney(number) {
   return str;
 }
 
-export default class Support extends React.Component {
+export default class Support extends VisibilitySensor {
```
montogeek: 2 things:
1. And most important one, this break the goal of this project, @skipjack vision when rebuilding this project is to focus solely on  **content**, meaning that custom plugins/components/logic/code etc should be outside of it, as dependencies, that's why my custom MDX plugins are dependencies.2

2. This way of writing React components is not recommended by React itself (https://reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance), meaning that it is not idiomatic, which is important if you want to attract new contributors, especially in an Open Source project like this one.

src/components/VisibilitySensor/VisibilitySensor.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+
+export default class VisibilitySensor extends React.Component {
```
montogeek: It is nice component, good work

src/components/Support/Support.jsx
EugeneHlushko: this looks like an overkill to me, @skipjack thoughts?

skipjack: I agree with @montogeek in this case.

> ...when rebuilding this project is to focus solely on content, meaning that custom plugins/components/logic/code etc should be outside of it, as dependencies, that's why my custom MDX plugins are dependencies.

Yeah it seems there are already some pretty popular open source packages that do similar things so we should consider those first instead of bloating this codebase with more components:

- https://github.com/joshwnj/react-visibility-sensor
- https://github.com/fkhadra/react-on-screen
- https://github.com/dazld/react-on-visible

Even if none of those work, you could always write this new component as a separate package the same way we've done for other things.

> This way of writing React components is not recommended by React itself (https://reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance), meaning that it is not idiomatic, which is important if you want to attract new contributors, especially in an Open Source project like this one.

Yeah I agree with this as well. However, if you use one of the open source packages I listed above then the new component (and inheritance) wouldn't even be necessary.

EugeneHlushko: I've started this PR with a `react-visibility-sensor` which doesn't benefit from Intersection Observer. After looking at our user base i was tempted to go ahead and use that. Which seems to be widely supported nowadays.
So i had did a re-work here within PR.

having that said i will move the visibility sensor into a separate package and rework it from inheritance type module into a HOC

EugeneHlushko: Thanks for dropping by btw @skipjack 👍

EugeneHlushko:
Apparently there is a: https://www.npmjs.com/package/react-intersection-observer

EugeneHlushko:
Its updated CC @montogeek

EugeneHlushko:
I love this bot lol

montogeek:
Yeah, this PR works great!

montogeek:
Thank you!!

EugeneHlushko:
Cheers

PR #2957, opened 2019-04-19
markup(html) add missing required attribute lang

EugeneHlushko:
- Looking through reports it seems we can fix this trivial html markup error
- add DOCTYPE to our html


montogeek:
Thanks!

PR #2964, opened 2019-04-22
infa(fetch) add hide option to the fetch script, hide deprecated package

### LINKED ISSUE #2573: Config loader is shown in webpack docs
Config loader was deprecated some time ago: https://webpack.js.org/loaders/config-loader/#extending-configuration-files

Lets stop showing it to users? CC @montogeek

EugeneHlushko:
Fixes #2573

### DIFF
--- a/src/utilities/fetch-package-repos.js
+++ b/src/utilities/fetch-package-repos.js
@@ -9,7 +9,7 @@ const fetch = {
     {
       organization: 'webpack-contrib',
       suffixes: ['-loader'],
-      hides: ['config-loader']
+      hides: ['webpack-contrib/config-loader']
     },
     'babel/babel-loader',
     'postcss/postcss-loader',
@@ -53,10 +53,8 @@ async function main() {
       const repos = await paginate(organization);
       return repos
         .map(repo => repo.full_name)
-        .filter(name =>
-          suffixes.some(suffix => name.endsWith(suffix)) &&
-          !hides.some(hide => name.endsWith(hide))
-        );
+        .filter(name => suffixes.some(suffix => name.endsWith(suffix)))
+        .filter(name => !hides.includes(name));
     }));
 
     const json = JSON.stringify(_.flatten(result), undefined, 2);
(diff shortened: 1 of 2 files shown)


src/utilities/fetch-package-repos.js
```diff
@@ -18,15 +19,16 @@ const fetch = {
...
+      hides: []
     }
   ]
 };
 
 const api = new GithubAPI();
 
 async function paginate (org) {
-  let response = await api.repos.getForOrg({ org, type: 'public', per_page: 100});
+  let response = await api.repos.listForOrg({ org, type: 'public', per_page: 100});
```
EugeneHlushko: The old one is deprecated.

src/utilities/fetch-package-repos.js
```diff
@@ -46,13 +48,15 @@ async function main() {
...
+      const { organization, suffixes, hides } = item;
 
       const repos = await paginate(organization);
-
       return repos
         .map(repo => repo.full_name)
-        .filter(name => suffixes.some(suffix => name.endsWith(suffix)));
+        .filter(name =>
+          suffixes.some(suffix => name.endsWith(suffix)) &&
+          !hides.some(hide => name.endsWith(hide))
```
montogeek: I would rather use 2 different filter, but this is fine.
It would also be great to have a test to make sure we are displaying the same number of repos

EugeneHlushko: This hide filter isn't scary, we already put them into `loaders.hides[]` and `plugins.hides[]`, so should be pretty safe except for cases when a plugin is called strangely: `super-config-loader` 

Let me change filter to full name instead so there is no risk

EugeneHlushko: @montogeek now thats against full match

montogeek:
We don't have access to add Netlify bot to publish a comment saying that the preview is ready, right?

anshumanv:
> We don't have access to add a Netlify bot that publishes a comment saying that the preview is ready, right?

We don't, it just needs to be enabled, I think @EugeneHlushko can take care of that. :)

EugeneHlushko:
Just added, lets see how it plays out @montogeek @anshumanv

PR #2965, opened 2019-04-23
fix(contribute): Uncaught error in a compiler example.

keidrun:
This PR fixes an uncaught error in a compiler example code in a document about writing a loader.


montogeek:
Thanks!

PR #2877, opened 2019-03-15
docs(guides): Make all demo context consistently

Ivocin:
I read the guides and find that some demos aren't consistent.
This pr contains three parts:
-  #2409 change var to let in `index.js` file
-  #2845 change CleanWebpackPlugin arguments usage in `webpack.config.js` file
-  delete extra part in `package.json` to avoid changing the dependency vesion （now the webpack in `package.json` version is `^3.0.0`)

PR #2967, opened 2019-04-24
docs(plugins): add < v2 CleanWebpackPlugin caveat to guides

EugeneHlushko:
Follow up to #2877


montogeek:
Do we really need to support old versions of plugins? They changed a lot, why support this one and not others?

This is something that could be solved using releases in a more strict way, one day :D

EugeneHlushko:
Yeah there are people hitting the bug, i suggest us keeping this for a quarter at least

montogeek:
Thanks!

PR #2968, opened 2019-04-24
site(misc) simplify rendering logic, clean dist on dev, clean config

EugeneHlushko:
- Fixed a case when`npm run start` command to possibly serve old `./dist/index.html` after working with `npm run build-test` command for example
- simplified rendering logic (tested, vote works)
- cleanup unused vars in config

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,8 +24,9 @@
     "node": ">=8.9.4"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


montogeek:
It seems fine, hope nothing breaks

EugeneHlushko:
Yeah i've tested the pages as well, all looks good (except some things but they are known issues on master as well)

montogeek:
Thanks!

PR #2970, opened 2019-04-25
feat(app): Add website manifest, minify CSS for production

EugeneHlushko:
- Add website manifest and respective icons
- Copy assets into ./dist when building the website


src/server.jsx
```diff
@@ -48,6 +48,7 @@ export default locals => {
           { Object.keys(assets).filter(asset => /\.css$/.test(asset)).map(path => (
             <link key={ path } rel="stylesheet" href={ `/${path}` } />
           ))}
+          <link rel="manifest" href="/manifest.json" />
```
anshumanv: Will this be able to read manifest in dev as well? 🤔

EugeneHlushko: No this file is only used for prod builds. On dev we are using HTML plugin for the HTML boilerplate.

montogeek:
This is nice to have, but do we need it? How many users are installing PWA?

src/server.jsx
anshumanv: Sounds good! 💯

EugeneHlushko:
Its very nice to have for cases when you are commuting and losing connectivity. I am looking at service worker implementation now, so will continue in further PRs when have time

EugeneHlushko:
oops forgot to mention @montogeek ;]

EugeneHlushko:
Generally my idea is to get all pages work offline, there a way to get to that point, but its possible.

montogeek:
I guess it it nice to have :)
We need to make sure though to publicize it, tell users that it is now possible to read the docs offline, maybe a bagde or one time notification.

EugeneHlushko:
Yeah we can, but when its finished ;]

montogeek:
Thanks!

pranshuchittora:
Awesome :rocket: :clap:

PR #2971, opened 2019-04-25
build(minimize): minimize css on prod

EugeneHlushko:
- minimize css for production
- had to include terser as css minimizer would override existing terser


webpack.prod.js
```diff
@@ -1,12 +1,12 @@
 // Import External Dependencies
-const path = require('path');
```
EugeneHlushko: these were unused

montogeek:
Thanks

PR #2972, opened 2019-04-26
docs(configuration) Fix an expression

zjffun:
I think there means using symlink packages, but not useing tools.


zjffun:
Thanks for your patient explanation. Current expression is better. I misunderstood 'tools that symlink packages' before.

EugeneHlushko:
Glad i could help @zjffun 
Cheers!

PR #2973, opened 2019-04-26
Fixing dist path

acaparrelli:
As I see there is an error in the path.


wizardofhogwarts:
> Yes, the path was not correct. Great job. Thanks @acaparrelli .

Sorry to say but I think the path is correct

acaparrelli:
> > Yes, the path was not correct. Great job. Thanks @acaparrelli .
> 
> Sorry to say but I think the path is correct

@wizardofhogwarts You are right, I got it. In the tutorial the index.html is manually created directly into the dist folder.

PR #2975, opened 2019-04-26
misc(updates) Upgrade react to 16.6, stop loading gitter for all user…

EugeneHlushko:
- update react, dom, prop types
- only load gitter after clicking icon. Only a small % of users does click that gitter icon. Now its loaded on-demand

According to speed audit, this is 1100ms faster to interactive than before.


EugeneHlushko:
According to speed audit, this is **1100MS** faster to interactive than before.

anshumanv:
I think you can set netlify to update preview link in a single comment rather than a new comment for all commits pushed. 🤔

EugeneHlushko:
It is because i've pushed another commit later :)

montogeek:
Very good! (Hattori Hanzō voice)

anshumanv:
> It is because i've pushed another commit later :)

Yes @EugeneHlushko it basically updates the original comment with the new link and commit, I think it's default behavior though, not entirely sure why it didn't happen here 🤔

PR #2981, opened 2019-04-30
Removed the relative path of bundle and main

akhil-gautam:
In the Getting Started section **<script src="main.js"></script>** is written but in asset-management src is replaced with relative path which can be misleading/confusing for beginners. As both the relative path and filename will work fine, it is better to have it same across the documentation.


EugeneHlushko:
We already have `+    <script src="main.js"></script>` in our Getting starting guide. So this was a leftover after updating the starting point guide last time, great catch @akhil-gautam

EugeneHlushko:
Thanks!

PR #2983, opened 2019-05-01
misc(phrasing) drop "its that simple" from homepage

### LINKED ISSUE #2378: Drop "it's that simple"
Let's switch up the header to something that might leave a first timer feeling frustrated if they didn't end up getting webpack working on the first try. 

This is something I had brought to me as a small but really impactful change! Some suggestions:

* Let's dive in/deeper!!!
* <Open for suggestions>

EugeneHlushko:
Fixes #2378


EugeneHlushko:
Thanks for double checking everyone

PR #2984, opened 2019-05-01
Update development.md

nakhodkiin:
Since the article gives an example of usage `mode` property within `webpack.config.js`, but the `package.json` refers to the `"webpack": "^3.0.0"`, but to make the `mode` property works, it is needed to use webpack v4 major version at least.

### DIFF
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -162,7 +162,7 @@ __package.json__
       "file-loader": "^0.11.2",
       "html-webpack-plugin": "^2.29.0",
       "style-loader": "^0.18.2",
-      "webpack": "^3.0.0",
+      "webpack": "^4.30.0",
       "xml-loader": "^1.2.1"
     }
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Please sign CLA, @nakhodkiin

nakhodkiin:
@EugeneHlushko 
I've already signed CLA, however, it seems that @jsf-clabot perform only initial CLA verification once PR is created, although it might be the case when CLA is signed in later, the bot won't pick that update, unfortunately. 

Screen Shot 2019-05-02 at 7 14 57 PM

EugeneHlushko:
You sure that you’re signing CLA with the right email, compared to email under which you’ve done the commit?

montogeek:
Thanks!

nakhodkiin:
@EugeneHlushko 
Oops, you're right! Email addresses are different!

PR #2985, opened 2019-05-02
docs(guides) update wp version as follow up to #2984

EugeneHlushko:
Following #2984 found another occurence


montogeek:
Thanks!

PR #2986, opened 2019-05-02
fix(homepage): Fix a homepage animation bug.

zjffun:
Add a condition to the homepage animation.

Reproduction:
When you come back after leave the homepage for a long time , the cube will spin quickly and then become normal.

### DIFF
--- a/src/components/Cube/Cube.jsx
+++ b/src/components/Cube/Cube.jsx
@@ -93,7 +93,7 @@ export default class Cube extends React.Component {
         let nowTime = performance.now();
         let deltaTime = nowTime - lastTime;
 
-        if (repeatDelay <= deltaTime) {
+        if (repeatDelay <= deltaTime && !document.hidden) {
           let obj = {};
           obj[axis] = degrees += 90;
 


EugeneHlushko:
requestAnimationFrame is paused when tab is inactive, are you on older FireFox? @zjffun

zjffun:
@EugeneHlushko 

This is indeed the case in the documentation. [window.requestAnimationFrame() - Web APIs | MDN](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)

But it doesn't work fine in FireFox 66.0.3 (64 bit). Puzzled @_@

EugeneHlushko:
Maybe worth checking firefox bug backlog, if not found open an issue. But i'd expect this to be known

zjffun:
I print some log just now and find it really a bug of Firefox...

zjffun:
Bug reported. [1548561 - requestAnimationFrame() can't paused when running in background tabs](https://bugzilla.mozilla.org/show_bug.cgi?id=1548561)

Close this

zjffun:
@EugeneHlushko 

Err.. Should we using `document.hidden` condition temporary? Even though this animation doesn't make a big difference.

EugeneHlushko:
its fine, edge case not directly related to reading docs. lets just let ff fix their bug instead. Thanks for looking into this anyway

zjffun:
All right

PR #2987, opened 2019-05-02
docs(loaders) document prefixing inline imports

### LINKED ISSUE #2966: Documentation on disabling (pre)loaders inline
**Feature to document**
Documentation on disabling (pre)loaders inline
* adding ! to a request will disable configured preLoaders
```require("!raw!./script.coffee")```

* adding !! to a request will disable all loaders specified in the configuration
```require("!!raw!./script.coffee")```

* adding -! to a request will disable configured preLoaders and loaders but not the postLoaders
```require("-!raw!./script.coffee")```

***Current documentation status***
https://webpack.js.org/concepts/loaders#inline

**Author(s)**
Don't know.

**Additional information**
After reading https://stackoverflow.com/questions/47860007/what-does-mean-in-webpack-loader-string I thought the inline prefixes (!! and -!) were deprecated. After asking on https://gitter.im/webpack/webpack it seems this is not the case (which would be very good news for me). Could the documentation be clarified/extended so everyone knows this awesome feature is still very much alive and kicking? I can't be the only one with an edge-case where declaring a loader inline with the !!-prefix is the most straight-forward solution :).

### PR DESCRIPTION
**Pull request by EugeneHlushko**
Document inline import prefixing

Fixes #2966

### DIFF
--- a/src/content/concepts/loaders.mdx
+++ b/src/content/concepts/loaders.mdx
@@ -96,7 +96,7 @@ import Styles from 'style-loader!css-loader?modules!./styles.css';
 
 It's possible to override any loaders, preLoaders and postLoaders from the [configuration](/configuration) by prefixing the inline `import` statement:
 
-- Prefixing with `!` will disable configured preLoaders
+- Prefixing with `!` will disable all configured preLoaders
 
 ```js
 import Styles from '!style-loader!css-loader?modules!./styles.css';
--- a/src/content/concepts/loaders.mdx
+++ b/src/content/concepts/loaders.mdx
@@ -94,7 +94,25 @@ It's possible to specify loaders in an `import` statement, or any [equivalent "i
 import Styles from 'style-loader!css-loader?modules!./styles.css';
 ```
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/content/concepts/loaders.mdx [hunk 1]
```diff
@@ -94,7 +94,25 @@ It's possible to specify loaders in an `import` statement, or any [equivalent "i
 import Styles from 'style-loader!css-loader?modules!./styles.css';
 ```
 
-It's possible to override any loaders in the configuration by prefixing the entire rule with `!`.
+It's possible to override any loaders, preLoaders and postLoaders from the [configuration](/configuration) by prefixing the inline `import` statement:
+
+- Prefixing with `!` will disable all configured preLoaders
```
EugeneHlushko: This is correct, thanks Tobias for checking
```suggestion
- Prefixing with `!` will disable all configured loaders
```

EugeneHlushko:
Trying to find the diff between `!` and `!!`, on hold.

EugeneHlushko:
@montogeek this is now clarified, ready to go

montogeek:
Thanks!

Issue #2974, opened 2019-04-26
Add  documentation of sockPort, injectClient, and injectHot


hiroppy:
**Feature to document**

Add new options's descriptions of webpack-dev-server.

`sockPort`, `injectClient`, and `injectHot`

**Author(s)**

/cc @evilebottnawi 

**Additional information**

related: https://github.com/webpack/webpack-dev-server/pull/1821

EugeneHlushko:
@hiroppy wanna file a PR or should we pick it up?

EugeneHlushko:
Picking this up:
Lets track progress on checkboxes:
- [] `sockPort`:
- [] `injectClient`:
- [] `injectHot`:

hiroppy:
I missed the message. I'm sorry. Thank you for adding these to the document.

EugeneHlushko:
done, closing

IronSean:
I still have no idea what these do though, even with the documentation

lili21:
I still have no idea what these do though, even with the documentation

jpierson:
Agreed @IronSean. I just ran into an issue getting a library working with webpack-dev-server and ran into the `injectClient` following the thread for https://github.com/webpack/webpack-dev-server/issues/2484. As it turns it, it seems to have fixed the problem I was running into by setting this to `false` but I have no idea why really. Better documentation here would most likely help.

PR #2995, opened 2019-05-05
docs(config) document devServer sockPort option

EugeneHlushko:
Takes one off #2974

Documents devServer `sockPort`

ON HOLD due to waiting dev server release.


src/content/configuration/dev-server.md
```diff
@@ -1192,6 +1192,23 @@ module.exports = {
 };
 ```
 
+## `devServer.sockPort`
+
+`number` `string`
```
hiroppy: Maybe we should add null?

https://github.com/webpack/webpack-dev-server/blob/master/lib/options.json#L239

EugeneHlushko: `null` doesn't affect anything, it should be removed from the schema imho:
https://github.com/webpack/webpack-dev-server/blob/cf4d0d07b8ffba9d28cd627b48f74e5f4f307fc6/lib/utils/addEntries.js#L20

hiroppy:
We have released webpack-dev-server@3.4.1, so I remove `on hold` label. /cc @EugeneHlushko

EugeneHlushko:
Thanks everyone involved!

PR #2998, opened 2019-05-06
docs(resolve): Remove outdated '*' string comment

### LINKED ISSUE #2997: `resolve.extensions` docs say that `*` is needed to resolve imports with extensions
https://webpack.js.org/configuration/resolve/#resolveextensions says:

> For modules that are imported with their extension, e.g. import SomeFile from './somefile.ext', to be properly resolved, a string containing "*" must be included in the array.

This change landed in https://github.com/webpack/webpack.js.org/commit/c5a12dfeb5da53f839c30649473c7d9aa301e70b around webpack 2

However AFAICT, it is no longer needed - I don't have `'*'` nor `''` in my webpack config and the imports with explicit extensions like `.js` work fine. So I'm wondering if this piece of documentation is still valid? Can someone confirm?

jakub-g:
Close #2997

### DIFF
--- a/src/content/configuration/resolve.md
+++ b/src/content/configuration/resolve.md
@@ -217,7 +217,7 @@ which is what enables users to leave off the extension when importing:
 import File from '../path/to/file';
 ```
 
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Thanks!

PR #3000, opened 2019-05-07
feat(pwa) use offline-plugin for service worker setup and offline sup…

EugeneHlushko:
- add offline plugin for utilizing SW and offline support
- takes one off #1400
- the whole website works offline thanks to our existing flatten content tree

Note:
- moved ssg config out as webpack was compiling both in a race (at least on my machine it seemed so.) to now run ssg and then run prod and include ssg run's css that was printed into html files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,7 @@
     "modularscale-sass": "^3.0.3",
     "node-sass": "^4.5.3",
     "npm-run-all": "^4.1.1",
+    "offline-plugin": "^5.0.7",
     "optimize-css-assets-webpack-plugin": "^5.0.1",
     "postcss-loader": "^2.1.3",
     "redirect-webpack-plugin": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "fetch:supporters": "node src/utilities/fetch-supporters.js",
     "fetch:starter-kits": "node src/utilities/fetch-starter-kits.js",
     "prebuild": "npm run clean",
-    "build": "run-s fetch content && cross-env NODE_ENV=production webpack --config webpack.prod.js",
+    "build": "run-s fetch content && cross-env NODE_ENV=production webpack --config webpack.ssg.js && cross-env NODE_ENV=production webpack --config webpack.prod.js",
     "postbuild": "npm run sitemap",
     "test": "npm run lint",
     "lint": "run-s lint:*",
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -31,6 +31,11 @@ import './Site.scss';
 // Load Content Tree
 import Content from '../../_content.json';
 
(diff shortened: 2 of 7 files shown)


webpack.prod.js
```diff
@@ -43,7 +44,8 @@ const prod = {
         to: './assets/'
       },
       'CNAME'
-    ])
+    ]),
+    new OfflinePlugin()
```
EugeneHlushko: Might be worth including a few externals e.g. google fonts?

montogeek:
How does this work?

EugeneHlushko:
It registers a service worker who then catches network requests to assets and caches them. There are defaults enlisted in offline-plugin's readme on options section.

EugeneHlushko:
A good ref: https://github.com/NekR/offline-plugin#docs

EugeneHlushko:
So looks like SSG prod version doesnt need it at all, instead of doing it twice we are doing it once now

EugeneHlushko:
On the preview its still not perfect, for some reason not all css are there when offline

montogeek:
> On the preview its still not perfect, for some reason not all css are there when offline

Good luck in your adventures :D

Services Workers are hard.

EugeneHlushko:
@montogeek ready for review, seem to work as expected

package.json
```diff
@@ -35,7 +35,7 @@
     "fetch:supporters": "node src/utilities/fetch-supporters.js",
     "fetch:starter-kits": "node src/utilities/fetch-starter-kits.js",
     "prebuild": "npm run clean",
-    "build": "run-s fetch content && cross-env NODE_ENV=production webpack --config webpack.prod.js",
+    "build": "run-s fetch content && cross-env NODE_ENV=production webpack --config webpack.ssg.js && cross-env NODE_ENV=production webpack --config webpack.prod.js",
```
montogeek: This `run-s` doesn't work if you append the ENV variable at the beginning?

src/components/Site/Site.jsx
```diff
@@ -77,6 +82,7 @@ class Site extends React.Component {
                   <Route path="/vote" component={Vote} />
                   <Route path="/organization" component={Organization} />
                   <Route path="/starter-kits" component={StarterKits} />
+                  <Route path="/app-shell" component={() => <React.Fragment />} />
```
montogeek: Why is this needed?

montogeek:
Thanks!

src/components/Site/Site.jsx
EugeneHlushko: This is "empty" state of the page for service worker to serve when offline, our bundle can then render content onto it on clientside. We have two different page templates: home and content pages. Homepage is always saved in SW and this one is used as fall back for the rest to not download all of them to disk when installing service worker, as client doesnt really need SSR version of the page.

PR #3004, opened 2019-05-09
feat(infra) Fix image sizes of the PWA

EugeneHlushko:
Our logo wasn't actually squared due to its shape. Icon generation service failed to insert transparent edges and center it rather cut the image:
Screen Shot 2019-05-09 at 2 20 45 PM

so that an error was thrown:
Screen Shot 2019-05-09 at 2 18 55 PM

I've hand crafted images to be exact square sized. 
Screen Shot 2019-05-09 at 2 19 42 PM


montogeek:
Thanks!

PR #3008, opened 2019-05-10
fix(pwa) fix versioned assets from SSG run 404-ing via SW

EugeneHlushko:
OG image wasn't known to SW as its emitted during SSG phase, this fixes it.
Currently og:image 404s


montogeek:
Thanks!

PR #3009, opened 2019-05-10
misc(style) fix gitter right position for mobile breakpoint

EugeneHlushko:
Gitter spacing on right was off, fixed it to stay aligned to content on mobile as well

scr


montogeek:
Thanks

PR #3011, opened 2019-05-10
Update link to new Write a Loader doc

Timmehs:
Reading up on loaders, I noticed a broken link toward the bottom of https://webpack.js.org/concepts#loaders to an old "How to write a loader?" article. I think https://webpack.js.org/contribute/writing-a-loader is the most up to date doc on this. Updated href and link name.


montogeek:
Thanks!

PR #3016, opened 2019-05-12
Fixed page not autoscrolling to top bug

### LINKED ISSUE #3015: Page scroll remains fixed when navigating
**Feature to document**
When navigating, if I've scrolled to the bottom of one page and navigate to another page, then the new page loads with scroll fixed at the bottom, One has to scroll up to start reading from starting of the page.
To reproduce, scroll down to the bottom in any page, and then click on any link to load another page.

**Author(s)**
N/A

**Additional information**
window.scrollTo(0,0) might save the problem. I'll send a PR.

nveenjain:
Fixes #3015  
Added window.scrollTo(0,0) on page mounting.

### DIFF
--- a/src/components/Page/Page.jsx
+++ b/src/components/Page/Page.jsx
@@ -1,5 +1,6 @@
 // Import External Dependencies
 import React from 'react';
+import { Prompt } from 'react-router';
 
 // Import Components
 import PageLinks from '../PageLinks/PageLinks';
@@ -11,6 +12,7 @@ import Configuration from '../Configuration/Configuration';
 // Load Styling
 import './Page.scss';
 
+let positionMap = new Map();
 class Page extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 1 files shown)


src/components/Page/Page.jsx
```diff
@@ -25,6 +25,7 @@ class Page extends React.Component {
   }
 
   componentDidMount() {
+    window.scrollTo(0,0);
```
EugeneHlushko: What about anchor links?

e.g. `page#section` ?

montogeek: What about if you reopen the last closed tab?

EugeneHlushko: Yeah this approach isn't good enough. What if we bind onto menu link?

nveenjain: Anchor links work fine, for last closed tab, I'm thinking of storing scroll positions when navigating the code and restoring it back to the stored position. This way if a user has read upto a certain position, his/her position will be preserved when he/she comes back to the same page.

montogeek:
This definitely breaks the page when:
1. Open https://deploy-preview-3016--webpackjsorg-netlify.netlify.com/contribute/writing-a-loader/#testing
2. Close the tab.
3. Re open it using reopen last tab option of your browser, page stays at the top.
4. It should open in the same anchor when it was closed.

This behaviour works fine in webpack.js.org.

montogeek:
@nveenjain Friendly ping

montogeek:
Closing because it got stalled.

PR #3018, opened 2019-05-12
chore(import): minor fixing in import statement

pranshuchittora:
Fixed import of scss in `Sidebar` component


montogeek:
How does this affect the site? I can't see a difference in the preview when comparing it with the production site.

pranshuchittora:
There won't be any difference. Because, previously the `import` statement goes one folder back & comes to the same folder. 

The required file exists in the same directory, so the changes proposed does the required fixes.

montogeek:
Thanks!

PR #3026, opened 2019-05-14
Update main.js file path.

sadick254:
Update the generated `main.js` path from `main.js` to `dist/main.js`
Webpack by default generates the bundle to a `dist` folder

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -173,7 +173,7 @@ __dist/index.html__
    </head>
    <body>
 -    <script src="./src/index.js"></script>
-+    <script src="main.js"></script>
++    <script src="dist/main.js"></script>
    </body>
   </html>
 ```


src/content/guides/getting-started.md
```diff
@@ -173,7 +173,7 @@ __dist/index.html__
    </head>
    <body>
 -    <script src="./src/index.js"></script>
-+    <script src="main.js"></script>
++    <script src="dist/main.js"></script>
```
EugeneHlushko: HTML file is moved into `./dist` in previous step, this is wrong

sadick254: You are right. closing this

montogeek:
It is the third time we get this PR, Maybe the docs could use some improvement?

sadick254:
I think the docs should be updated to make it clear that it is using the generated `./dist/index.html` from the `./index.html` template.

montogeek:
That would be good indeed. Mind sending a PR? Thank you!

sadick254:
Sure no worries.

PR #3027, opened 2019-05-14
add default value [/node_modules/] for ignored in watchOptions

EslamHiko:
set the default value of ignored in watchOptions as in this PR
PR : https://github.com/webpack/webpack-dev-server/pull/1794


hiroppy:
This feature was reverted, so I'll close.

PR #3030, opened 2019-05-14
fix(seo) canonical mustn't be relative

EugeneHlushko:
- Fixes canonical from being relative

Screen Shot 2019-05-14 at 7 50 57 PM


montogeek:
Thanks!

PR #3036, opened 2019-05-14
feat(foit) use font display for google fonts

EugeneHlushko:
fix foit using font-display now available also for google fonts


montogeek:
Boom!

PR #3041, opened 2019-05-15
Integrate prettier, setup precommit hooks, format all files

anshumanv:
* Integrated prettier with ESLint
* Formatted all files
* Added precommit hooks to format files before commit.

@montogeek @EugeneHlushko Looking forward to your feedback! :)


package.json
```diff
@@ -55,8 +57,10 @@
     }
   },
   "lint-staged": {
-    "*.{js,jsx,md}": [
-      "npm run lint:js"
+    "*.{js,jsx,md,scss}": [
```
montogeek: Given that you created scripts for each type of file, please separate this as well, editing a SCSS file then run prettier for JS files is a waste of time

package.json
```diff
@@ -83,6 +87,7 @@
     "directory-tree-webpack-plugin": "0.3",
     "duplexer": "^0.1.1",
     "eslint": "4.19.1",
+    "eslint-config-prettier": "^4.2.0",
```
montogeek: Please pin exact version as you did with `prettier`

EugeneHlushko:
From the top of my head:
- change jsx element closing tag to not be on the newline for multiline cases
- allow trailing comma (we can enfroce as well)
- change the order of calling the format tasks, format -> lint -> throw if something wasnt fixed automatically

montogeek:
> change jsx element closing tag to not be on the newline for multiline cases

I prefer it in a new line, thanks James for prettier :D

package.json
anshumanv: Pinned

package.json
anshumanv: Ah missed this, thanks!

anshumanv:
Enforced trailing comma, updated order and keeping multiline cased in new line

anshumanv:
@montogeek @EugeneHlushko formatting md with prettier seems to use emphasis style as `*` but markdownlink used `-` I can't seem to find a way to enforce md style via prettier, how about we switch to `*` as emphasis style or are there some tradeoffs which I'm missing?

anshumanv:
You can try running `npm run format:markdown` for a quick check

chenxsan:
Closed in favor of https://github.com/webpack/webpack.js.org/pull/4518. Thanks anyway.

anshumanv:
At least this got closed now after 2 years, thanks for picking it up!

PR #1317, opened 2017-06-20
docs(config/node): add `false` for `node` option

simon04:
Docs for https://github.com/webpack/webpack/pull/4995

PR #3050, opened 2019-05-17
docs(config/node): add `boolean` to the type

foolip:
Follow up for https://github.com/webpack/webpack.js.org/pull/1317.

_describe your changes..._

### DIFF
--- a/src/content/configuration/node.md
+++ b/src/content/configuration/node.md
@@ -16,7 +16,7 @@ This feature is provided by webpack's internal [`NodeStuffPlugin`](https://githu
 
 ## `node`
 
-`object`
+`boolean | object`
 
 This is an object where each property is the name of a Node global or module and each value may be one of the following...
 


src/content/configuration/node.md
```diff
@@ -16,7 +16,7 @@ This feature is provided by webpack's internal [`NodeStuffPlugin`](https://githu
 
 ## `node`
 
-`object`
+`boolean | object`
 
 This is an object where each property is the name of a Node global or module and each value may be one of the following...
```
anshumanv: Please update this too since it can accept bool :)

foolip: The documentation of the `false` value is at the end:
> Since webpack 3.0.0, the `node` option may be set to `false` to completely turn off the `NodeStuffPlugin` and `NodeSourcePlugin` plugins.

Would you like me to put that up front instead?

anshumanv: Ah, missed that one, this looks good! ✨

src/content/configuration/node.md
```diff
@@ -16,7 +16,7 @@ This feature is provided by webpack's internal [`NodeStuffPlugin`](https://githu
 
 ## `node`
 
-`object`
+`boolean | object`
```
EugeneHlushko: For enums we specify values, you cannot set it to `true` as schema wont let it go. So only mark false.

```suggestion
`bool: false | object`
```

EugeneHlushko:
Thanks

PR #3054, opened 2019-05-21
Fetch supporters with Open Collective GraphQL API

znarf:
- fetch supporters with Open Collective GraphQL API
  - likely more reliable and future proof, ESLint and Babel recently migrated to it for similar case: 
    - https://github.com/eslint/eslint.github.io/pull/564
    - https://github.com/babel/website/pull/2031
- refactor path computation with `absoluteFilename`
- stop relying on `id` as it's not exposed anymore
- order supporters by totalDonations, pretty JSON format for easy troubleshooting

Expected behavior changes:
- the data is not computed the same, so it can be there is small differences with the current data. before merging we need to assess if the result is what's expected
- backers without an avatar get a default one with their initials

Todo:
- [x] support `firstDonation`
- [x] de-duplication (due to supporters with multiple orders)
- [x] pagination (limited to 1000 now)

### DIFF
--- a/src/components/Support/Support.jsx
+++ b/src/components/Support/Support.jsx
@@ -144,7 +144,7 @@ export default class Support extends React.Component {
 
           {
             supporters.map((supporter, index) => (
-              <a key={ supporter.id || supporter.slug || index }
+              <a key={ supporter.slug || index }
                 className="support__item"
                 title={ `$${formatMoney(supporter.totalDonations / 100)} by ${supporter.name || supporter.slug}` }
                 target="_blank"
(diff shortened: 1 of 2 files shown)


EugeneHlushko:
Well, it works as expected i would say, code looks good too.
But `+80 −12`, is that what future is 😅 

I'll leave it to the rest of maintainers

montogeek:
It is fine

montogeek:
I don't see the request in the preview version, but the supporters list is there, maybe cache?

EugeneHlushko:
We fetch them during build time, this is only run on node

EugeneHlushko:
Thanks, well structured PR too!

znarf:
Thank you for merging!

@EugeneHlushko Sorry, GraphQL is a bit more verbose but I think this is worth it! 

Don't hesitate if you have any feedback and feature request for the Open Collective API.

PR #3059, opened 2019-05-23
fix errors in caching.md

chenyinkai:
1. use `new CleanWebpackPlugin()` instead of `new CleanWebpackPlugin(['dist'])`
2. use `hash` instead of `contenthash `

### DIFF
--- a/src/content/guides/caching.md
+++ b/src/content/guides/caching.md
@@ -53,7 +53,7 @@ __webpack.config.js__
     entry: './src/index.js',
     plugins: [
       // new CleanWebpackPlugin(['dist/*']) for < v2 versions of CleanWebpackPlugin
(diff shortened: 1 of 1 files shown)


chenyinkai:
Sorry, this seems to be my fault. [stackoverflow](https://stackoverflow.com/questions/50217480/cannot-use-chunkhash-or-contenthash-for-chunk-in-name-chunkhash-js-us).

1. use new CleanWebpackPlugin() instead of new CleanWebpackPlugin(['dist'])
2. ~~use hash instead of contenthash~~

@EugeneHlushko

montogeek:
Thanks!

EugeneHlushko:
Yup all good thx for ping @chenyinkai

PR #3064, opened 2019-05-24
misc(url) enforce trailing slash, make sure we are building trailing …

### LINKED ISSUE #3063: Enforce trailing slash in internal linking
After looking at GA i've found an issue with our page views, server side links were different to many front-end links via react router navigations.

Screen Shot 2019-05-24 at 3 37 26 PM

Sometimes on dev we didnt have proper item open in sidebar or anchor links were'nt working.

EugeneHlushko:
Fixes #3063

### DIFF
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -1,6 +1,6 @@
 // Import External Dependencies
 import React from 'react';
-import { Switch, Route } from 'react-router-dom';
+import { Switch, Route, Redirect } from 'react-router-dom';
 import { hot as Hot } from 'react-hot-loader';
 import DocumentTitle from 'react-document-title';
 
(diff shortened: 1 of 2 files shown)


src/components/Site/Site.jsx
```diff
@@ -74,6 +74,7 @@ class Site extends React.Component {
           toggle={this._toggleSidebar} /> : null}
 
         <Switch>
+          <Route exact strict path="/:url*" render={props => <Redirect to={`${props.location.pathname}/`}/>} />
```
montogeek: Why is this needed?

EugeneHlushko: The context is in the issue, we have two types of Urls, and ssr urls are different from react router urls. When users copy links and share them they share react router links without trailing slashes, which gets into redirect and sometimes anchors didnt work after that e.g. `/concepts#something` ended up on `/concepts/#something`. Also reporting in GA is broken, all page views are duplicated with `/concepts` (landed via react-router) and `/concepts/` (landed from search)

montogeek:
Thanks!

PR #3068, opened 2019-05-24
ui(mobile) fix headings being cut off screen

EugeneHlushko:
It has been reported on twitter that headings are invisible for mobile resolutions when heading is long enough.

We can break words for such headings. It looks like this now:
Screen Shot 2019-05-24 at 11 47 29 PM

Or we can allow horizontal scroll for such headings so that user can scroll them (as we do for code blocks):
Screen Shot 2019-05-24 at 11 49 14 PM

### DIFF
--- a/src/components/Markdown/Markdown.scss
+++ b/src/components/Markdown/Markdown.scss
@@ -37,6 +37,7 @@
     line-height: 1.4;
     margin: 1.5em 0 0.25em;
     color: getColor(fiord);
+    word-break: break-all;
 
     tt,
     code {


EugeneHlushko:
performance is the key to success of documentation website. Accessibility, performance, quality of content. Having performance impact would be a mean of last resort. 

@montogeek pls review

rishabh3112:
It's not that much of decrease performance as you have mentioned. FYI only a bunch of CSS media queries are needed to implement this.

Dropping a link for your reference (that will most probably clear your misunderstanding about this) at [CSS tricks](https://css-tricks.com/snippets/css/fluid-typography/).

montogeek:
Thanks!

PR #3070, opened 2019-05-26
fix: update name in manifest

rishabh3112:
Update application name to be displayed on PWA to comply with branding guidelines of webpack


montogeek:
Thanks!

PR #3071, opened 2019-05-26
Making resolve extension for ts as well

hamidre13:
Generally, before migration, the package should not resolve .ts or .tsx files so I think it is a good idea to highlight the resolve extensions as well.

_describe your changes..._

### DIFF
--- a/src/content/guides/typescript.md
+++ b/src/content/guides/typescript.md
@@ -78,8 +78,8 @@ module.exports = {
       }
     ]
   },
-  resolve: {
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks for your contribution.
That `+` is used to inform that a new line was added from a previous example, in this case that snippet is the first example of that configuration, there is no previous examples do make a difference.

This is not needed.

montogeek:
@EugeneHlushko Please review

PR #3075, opened 2019-05-28
docs(config) node config boolean only false allowed

EugeneHlushko:
- only false is allowed via schema, clarifying it here.

ref: https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L2005


montogeek:
Thnaks!

PR #3078, opened 2019-05-29
docs: replace chunk.modules with chunk.getModules()

tricoder42:
Replace `chunk.modules` with `chunk.getModules()`. `_modules` are private property on `chunk` object and should be accessed through `getModules()` method (or `modulesIterable` property, but that returns `Set`, instead of `Array`).


EugeneHlushko:
Thanks!

PR #3081, opened 2019-05-30
Update dev-server.md

Sawtaytoes:
This gave me a cryptic error, and I spent a long time trying to figure out why it was happening. I'm sure this would affect others as well.

When passing in the second argument to `webpack(config, callback)`, it will return a different object then when called with just the config. This causes problems when using webpack-dev-server because it's expecting a compiler, not the webpack config with a `compiler` prop.

You would see the error `TypeError: Cannot read property 'plugins' of undefined` when the callback is added to `webpack()`.

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -53,7 +53,7 @@ Content not from webpack is served from /path/to/dist/
 
 that will give some background on where the server is located and what it's serving.
 
-If you're using dev-server through the Node.js API, the options in `devServer` will be ignored. Pass the options as a second parameter instead: `new WebpackDevServer(compiler, {...})`. [See here](https://github.com/webpack/webpack-dev-server/tree/master/examples/api/simple) for an example of how to use webpack-dev-server through the Node.js API.
+If you're using dev-server through the Node.js API, the options in `devServer` will be ignored. Pass the options as a second parameter instead: `new WebpackDevServer(compiler, {...})`. [See here](https://github.com/webpack/webpack-dev-server/tree/master/examples/api/simple) for an example of how to use webpack-dev-server through the Node.js API. **Note:** You cannot use the second compiler argument (a callback) when using `WebpackDevServer`.
 
 W> Be aware that when [exporting multiple configurations](/configuration/configuration-types/#exporting-multiple-configurations) only the `devServer` options for the first configuration will be taken into account and used for all the configurations in the array.
 


src/content/configuration/dev-server.md
```diff
@@ -53,7 +53,7 @@ Content not from webpack is served from /path/to/dist/
 
 that will give some background on where the server is located and what it's serving.
 
-If you're using dev-server through the Node.js API, the options in `devServer` will be ignored. Pass the options as a second parameter instead: `new WebpackDevServer(compiler, {...})`. [See here](https://github.com/webpack/webpack-dev-server/tree/master/examples/api/simple) for an example of how to use webpack-dev-server through the Node.js API.
+If you're using dev-server through the Node.js API, the options in `devServer` will be ignored. Pass the options as a second parameter instead: `new WebpackDevServer(compiler, {...})`. [See here](https://github.com/webpack/webpack-dev-server/tree/master/examples/api/simple) for an example of how to use webpack-dev-server through the Node.js API. **Note:** You cannot use the second compiler argument (a callback) when using `WebpackDevServer`.
```
EugeneHlushko: If i am reading correct, it says: `Pass me two arguments: first one is compiler, second is object with props`. What exactly should this mean in that case? "You cannot use the second compiler argument (a callback) when using `WebpackDevServer`."

Sawtaytoes: When using `WebpackDevServer` you pass it the compiler.

But if you pass 2 args into the webpack compiler (as in `new WebpackDevServer(webpack(first, second))`, it gives a cryptic error.

Sawtaytoes: Looks like the CI build fails now when running since this PR's been sitting here for a while.

Was my reasoning incorrect? I just want to help other developer who'll understandably end up in the same situation.

EugeneHlushko:
```
If you don’t pass the webpack runner function a callback, it will return a webpack Compiler instance. This instance can be used to manually trigger the webpack runner or have it build and watch for changes, much like the CLI.
```

EugeneHlushko: The linked example also doesnt pass the runner function, so thats ok.
Based on this, i'd suggest to update your changes to:
```
If you're using .......... Node.js API. Remember to [avoid providing a runner function to `webpack()`](/api/node/#compiler-instance) as a second argument in order to get a `compiler` instance.
```

EugeneHlushko:
CI will pass when you remove `**`, we emphasize with `__`

montogeek:
@Sawtaytoes Friendly ping

Sawtaytoes:
@montogeek Sadly, I had removed the patch branch in my GitHub account so I cannot make further changes. I'm closing this PR as I created a new one instead #3395.

PR #3083, opened 2019-05-30
docs(plugins) remove unmaintained outdated plugin

EugeneHlushko:
The plugin at question is not maintained since2014, time to remove it from the docs

ref: https://github.com/webpack-contrib/component-webpack-plugin/commits/master

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   "devDependencies": {
     "@mdx-js/loader": "0.15.7",
     "@mdx-js/mdx": "0.15.7",
-    "@octokit/rest": "^15.9.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
   "devDependencies": {
     "@mdx-js/loader": "0.15.7",
     "@mdx-js/mdx": "0.15.7",
-    "@octokit/rest": "^15.9.4",
+    "@octokit/rest": "^16.27.1",
```
EugeneHlushko: this needed update because of the api

montogeek:
Thanks!

PR #3087, opened 2019-06-02
Update reference to SplitChunksPlugin

ewolfe:
Since webpack v4, the CommonsChunkPlugin was removed in favor of optimization.splitChunks


montogeek:
Thanks!

PR #3089, opened 2019-06-02
update: service worker docs

benschac:
users need to write dist to disk in order to user `http-server`
https://stackoverflow.com/questions/41928358/webpack-how-to-compile-write-on-disk-and-serve-static-content-js-css-using-w/41928812#41928812
and
https://github.com/webpack/webpack-dev-server/releases/tag/v3.1.10


hiroppy:
CI fails. @benschac please check `yarn lint:markdown `.

PR #3096, opened 2019-06-04
fix(guides): Update caching.md

rsstdd:
There is an error in the punctuation of the following compound/complex sentence:
`This allows sites to load faster with less unnecessary network traffic, however it can also cause headaches when you need new code to be picked up.` I propose the grammar changes in `caching.md`. The following sentence would also be correct. 

`This allows sites to load faster with less unnecessary network traffic; however, it can also cause headaches when you need new code to be picked up.`

Edited for accuracy.


montogeek:
Thanks!

Issue #3095, opened 2019-06-04
Dev Server Documentation: sockHost option


Loonride:
**Feature to document**

`sockHost`: https://github.com/webpack/webpack-dev-server/pull/1858

**Author(s)**

Myself, reviewed by @evilebottnawi @hiroppy 

**Additional information**

`sockHost` is only used by the client of the Dev Server. It is the host that the client connects to. This makes it possible for the server to exist on a different host than the client, allowing the client to know where to remotely connect to.

EugeneHlushko:
Via #3101

PR #3101, opened 2019-06-06
docs(dev-server): document sockHost, change sockPort docs

Loonride:
_describe your changes..._

Issue: https://github.com/webpack/webpack.js.org/issues/3095

- Added docs for `sockHost` option
- Fixed `sockPort` docs to specify that it is specific to connected clients of the `devServer`


EugeneHlushko:
Thanks!

PR #3103, opened 2019-06-06
Added PWA compact

kuldeepkeshwar:
This brings consistent behavior(App Icon) when a user does `Add to Home Screen`. 
Sometimes it doesn't happen automatically on old/iOS devices.
e.g 
[Happened with me](https://twitter.com/kuldeepkeshwar/status/1126572346096082945/photo/1) 😉

### DIFF
--- a/src/components/Print/Print.js
+++ b/src/components/Print/Print.js
@@ -15,38 +15,31 @@ const PRINTABLE_SECTIONS = [
 ];
 
 export default function Print (props) {
-    const { url } = props;
-    const printUrl = _printPageUrlFromUrl(url);
+  const { url } = props;
+  const printUrl = _printPageUrlFromUrl(url);
 
-    // not in a printable section
-    if (!url) {
-      return null;
-    }
+  // not in a printable section
+  if (!printUrl) {
+    return null;
+  }
 
-    return (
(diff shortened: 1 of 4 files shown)


EugeneHlushko:
I am against adding a third party .js for this. Please consider working on manifest or <head>

kuldeepkeshwar:
Can you provide more context on third party js ? Is it being served from a different domain or is it just using a third party library ?

pranshuchittora:
@kuldeepkeshwar IMO adding an external library will increase the bundle size, hence loading time.
On the other hand, adding pwacompact is unnecessary as similar behaviour can be achieved by modifying manifest file.

EugeneHlushko:
@kuldeepkeshwar yes this doesnt require third party javascript, or any javascript at all, Please try and fix by different means (manifest.json, or inside `<head>`).

kuldeepkeshwar:
@EugeneHlushko @pranshuchittora [its](https://github.com/GoogleChromeLabs/pwacompat) 3kb(5.6kb) .(can be included into the bundle if we are not comfortable with the third-party hoisting)
Screen Shot 2019-06-11 at 9 18 04 PM

Plus it does take care of browsers nuisance, taking that nuisance & maintaining it yourself (which we can always do) will be like reinventing the wheel

Issue #3100, opened 2019-06-06
Dev Server Documentation: serverMode option


Loonride:
**Feature to document**

`serverMode`: https://github.com/webpack/webpack-dev-server/pull/1937

**Author(s)**

Myself, @evilebottnawi @hiroppy 

**Additional information**

N/A

Loonride:
Closing as this option was removed in favor of `transportMode`: https://github.com/webpack/webpack.js.org/issues/3239

PR #3104, opened 2019-06-06
docs(dev-server): document serverMode

Loonride:
_describe your changes..._

Closes https://github.com/webpack/webpack.js.org/issues/3100

This will need to be updated once `'ws'` implementation is added.

As far as the full class implementation example, I see a few options:
- keep it as is, full class example using `sockjs`
- make it generic, not using `sockjs`, just omitting some parts
- remove completely, make an example in dev server and link to it instead


Loonride:
Closing as this option was removed in favor of `transportMode`: #3239

PR #3106, opened 2019-06-07
Update entry-context.md

mnieber:
Use the term "application bundling" instead of "application execution" to describe entry points.


montogeek:
Thanks!

PR #2247, opened 2018-06-09
Add PrettierJS to format code

tryzniak:
The title says it all :) This PR adds prettierjs and a npm script to format the codebase.

PR #3109, opened 2019-06-10
Add Prettier formatter for JS files

montogeek:
Continues https://github.com/webpack/webpack.js.org/pull/2247/


anshumanv:
lmk if that needs to be closed 😄

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "printWidth": 100
+}
```
evilebottnawi: I recommend use this too:
```
  trailingComma: 'es5',
  arrowParens: 'always',
```
- Less changes in review
- Use parens for consistently output
- Also i recommend don't use printWidth: `100` with spaces, it is can generate unreadable code

montogeek:
@evilebottnawi @anshumanv @EugeneHlushko Ping

@anshumanv I would like to include all your changes, let's do it after merging this one.

anshumanv:
WIll update 👍

jeffin143:
What is the blockage here ?

evilebottnawi:
@montogeek Why prettier is blocked? 

@chenxsan I think we need focus on this after webpack@5 upgrading

chenxsan:
Sure, I can do it after the webpack v5 upgrade if @montogeek doesn't have time to continue it.

montogeek:
There is no blocker at this moment, I will restart it instead, can do it soon :)

chenxsan:
@montogeek Hi, do you still have time doing this? Otherwise I can take care of it.

chenxsan:
Closed via https://github.com/webpack/webpack.js.org/pull/4518

PR #3115, opened 2019-06-12
feat(manifest) reduce maintenance amount of PWA files by using a plug…

EugeneHlushko:
- reduce maintenance amount of PWA files by using a plugin to cut icons and generate manifest.json
- fix majority of potential issues with adding to home screens across devices

Also renders #3103 's approach redundant


montogeek:
Thanks!

PR #3126, opened 2019-06-16
docs(api) dynamic expressions in import()

### LINKED ISSUE #1699: API - Delineate and Elaborate on Dynamic Expressions
I searched and searched but couldn't find any mention of this feature in the new website.

https://webpack.github.io/docs/context.html

It speaks of *dynamic imports*:
```js
require("./template/" + name + ".jade");
```

So I was wondering first of all - is that on purpose? ie. is this feature deprecated? If not, and if it really isn't in the docs, then I can start a PR for updating the docs with this useful information.

EugeneHlushko:
Fixes #1699

### DIFF
--- a/src/content/api/module-methods.md
+++ b/src/content/api/module-methods.md
@@ -9,6 +9,7 @@ contributors:
   - byzyk
   - debs-obrien
   - wizardofhogwarts
+  - EugeneHlushko
 related:
   - title: CommonJS Wikipedia
     url: https://en.wikipedia.org/wiki/CommonJS
@@ -58,11 +59,12 @@ export default {
 
 ### `import()`
 
-`import('path/to/module') -> Promise`
+`function(String path):Promise`
 
 Dynamically load modules. Calls to `import()` are treated as split points, meaning the requested module and its children are split out into a separate chunk.
 
 T> The [ES2015 Loader spec](https://whatwg.github.io/loader/) defines `import()` as method to load ES2015 modules dynamically on runtime.
+
 
 ``` javascript
 if ( module.hot ) {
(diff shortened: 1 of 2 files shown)


src/content/api/module-methods.md
```diff
@@ -58,12 +59,13 @@ export default {
 
 ### `import()`
 
-`import('path/to/module') -> Promise`
+`function(String path):Promise`
```
montogeek: Let's stick with `string` https://github.com/webpack/webpack.js.org/search?q=string&unscoped_q=string

src/content/api/module-methods.md
```diff
@@ -74,6 +76,24 @@ if ( module.hot ) {
...
+
+```javascript
+// imagine we had a method to get language from cookies or other storage
+const language = detectVisitorLanguage();
+import(`./locale/${language}.json`).then(module => {
+  // do something with the translations
+});
+```
+
+T> Using the [`webpackInclude` and `webpackExclude`](/api/module-methods/#magic-comments) options allows us to add regex patterns that reduce the files that webpack will bundle for this import.
```
montogeek: > allow you

or

> allow

src/content/guides/code-splitting.md
```diff
@@ -280,6 +280,8 @@ __src/index.js__
   });
 ```
 
+T> It is possible to provide a [dynamic expression](/api/module-methods/#dynamic-expressions-in-import) to `import()` when you might need to import specific module based on a computed variable later.
```
montogeek: Last part sounds weird, What you think about:
> based on a variable computed later.

src/content/api/module-methods.md
EugeneHlushko: my god where did i grab that Camel case :)

src/content/guides/code-splitting.md
EugeneHlushko: we are trying to say: "use `import()` to import something later (maybe based on a variable that you will compute)"

src/content/api/module-methods.md
EugeneHlushko: updated

src/content/api/module-methods.md
EugeneHlushko: changed to `allows you` and also improved the next statement

montogeek:
Thanks!

PR #3130, opened 2019-06-18
fix: loader overridden explanation

liximomo:


EugeneHlushko:
@liximomo what is the reason of this change. e.g. pull request or issue link for context? previous version was reviewed by Tobias recently

liximomo:
@EugeneHlushko 
image These two things can't match.

EugeneHlushko:
In that case you are fixing it in the wrong place. Change the code comment instead?

EugeneHlushko:
Thanks!

PR #3131, opened 2019-06-18
site(ux) Fix the menu issue caused by printable sidebar item index sw…

### LINKED ISSUE #3127: Sidebar - Incorrect active navigation links in sidebar
It seems the sidebar in guide page might have incorrect navigation link generated and incorrect active link when page initially opened.

For example, I opened <https://webpack.js.org/guides/development/#using-watch-mode> directly, and get the following:

image

Seems that every sidebar menus are shifted by one position.

EugeneHlushko:
Fixes #3127

### DIFF
--- a/src/components/Print/Print.js
+++ b/src/components/Print/Print.js
@@ -15,38 +15,31 @@ const PRINTABLE_SECTIONS = [
 ];
 
 export default function Print (props) {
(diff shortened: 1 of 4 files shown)


montogeek:
Thanks!

PR #3139, opened 2019-06-23
docs(configuration languages): remove reference to deleted tweet

DanielRuf:
This PR removes the reference to the tweet as it seems it was deleted.


EugeneHlushko:
Thanks!

PR #3142, opened 2019-06-25
docs(guides): add bundle-stats link to code-splitting

vio:
Added link to bundle-stats.

Webpack plugin source - https://github.com/bundle-stats/bundle-stats/blob/master/packages/bundle-stats/src/webpack-plugin.js


montogeek:
Thanks!

montogeek:
Thanks!

PR #3144, opened 2019-06-26
Add webpack-dev-server dependency to exmple

masives:
Added missing dependency in example at https://webpack.js.org/guides/development/#using-webpack-dev-server


src/content/guides/development.md
```diff
@@ -260,6 +260,7 @@ __package.json__
       "html-webpack-plugin": "^2.29.0",
       "style-loader": "^0.18.2",
       "webpack": "^4.30.0",
++     "webpack-dev-server": "^3.7.2"
```
corg: Since this line is not added manually when following the guide (it is auto-generated), does it make sense to mark it as an addition in the guide? The paragraph before it says "Let's add a script to easily run the dev server", so, to the reader, it may seem that they should be the one adding this line.

```suggestion
      "webpack-dev-server": "^3.7.2"
```

masives: My experience was that after adding the script it didn't work because the dependency wasn't there and I figured it out from errors. Maybe changing paragraph to "Let's add a script to easily run the dev server. We have to also include dependency for webpack-dev-server".

corg: This `package.json` part in question is in the "Using webpack-dev-server" section, and that section tells to install the dependency in the [first paragraph](https://webpack.js.org/guides/development/#using-webpack-dev-server):

> ### Using webpack-dev-server
>
>The `webpack-dev-server` provides you with a simple web server and the ability to use live reloading. Let's set it up:
>
>``` bash
>npm install --save-dev webpack-dev-server
>```

I'm sorry, I'm not the contributor here, just a passer-by. When I followed the guide, it was just as it says: 1) install `webpack-dev-server` dependency, 2) update `webpack.config.js`, 3) update `package.json`. There weren't any errors since `webpack-dev-server` was installed in step 1, at the start of this guide section.

masives:
So if I understand you all at this point we can assume that people have this dependency installed already? I'll remove the diff plus

montogeek:
Friendly ping

masives:
Thank you for reminder @montogeek. Given that it seemed not necessary for others and nobody have seen this as an issue I'm convinced that it's ok to leave it as is.

PR #3152, opened 2019-06-28
docs(config) Fix types and info on output section of config docs

EugeneHlushko:
Wanted to just quick fix the type on `output.filename` function option, but found many inconsistencies

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -105,32 +105,26 @@ Number of milliseconds before chunk request expires, defaults to 120 000. This
 
 ## `output.crossOriginLoading`
 
-`boolean` `string`
+`boolean: false` `string: ['anonymous', 'use-credentials']`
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.md
```diff
@@ -431,31 +430,17 @@ For details see [`output.jsonpFunction`](#outputjsonpfunction).
...
-`string` `function`
-
-Customize the main hot update filename. See [`output.filename`](#outputfilename) option for details on the possible values.
+`string: '[hash].hot-update.json'` `function`
 
-`[hash]` is the only available placeholder, the default being:
-
-__webpack.config.js__
-
-```javascript
```
EugeneHlushko: Removed example, it was default. Doesn't look like the function is being used at all, maybe we need to flag this in issues: https://github.com/webpack/webpack/search?q=hotUpdateMainFilename&unscoped_q=hotUpdateMainFilename

montogeek:
Thanks!

PR #3164, opened 2019-07-03
misc(web) add robots txt; change a to button for print; apply styling…

### LINKED ISSUE #3161: Remove printable pages from search entries
It seems that just about every search entry is duplicated with a regular page and a printable page. Aside from the confusion that duplicates alone bring, it can also be very confusing to click on a printable page because the print dialog keeps relentlessly popping up and the reader might not have any idea what's happening.

Removing printable pages from search entries would significantly improve viewing the docs.

EugeneHlushko:
… accordingly

Closes #3161 

- add robots txt
- make SW ignore robots.txt
- change <a> to <button> for print this page
- style button as link to keep design unchanged

### DIFF
--- a/src/assets/robots.txt
+++ b/src/assets/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /**/printable/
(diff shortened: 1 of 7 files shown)


montogeek:
Thanks!

PR #3169, opened 2019-07-05
Fix inaccurate description of callback in MultiCompiler

loilo:
The docs currently state that the callback passed to the `MultiCompiler` is executed after each configuration in the options array has been compiled.

This is not the case. The callback is executed once, after all compilations have finished.


montogeek:
Thanks!

loilo:
My pleasure. 🙂

Issue #3168, opened 2019-07-05
Potential error in proxy doc


bard:
https://webpack.js.org/configuration/dev-server/ says:

```js
module.exports = {
  //...
  devServer: {
    proxy: {
      '/api': 'http://localhost:3000',
      changeOrigin: true
    }
  }
};
```

That doesn't work for me. The following does:

```js
module.exports = {
  //...
  devServer: {
    proxy: {
      '/api': {
        target: 'http://localhost:3000',
        changeOrigin: true
      }
    }
  }
};
```

montogeek:
Please send a PR.

bard:
I would have gladly sent a PR — in fact, I would have sent one right away if certain that this is a bug and not my misunderstanding — but getting assigned work on a project that I didn't opt into joining or othewise made myself available to is, well... unusual.

montogeek:
Sorry

EugeneHlushko:
related pr is merged

PR #3173, opened 2019-07-06
fix(site): correction on proxy config

bard:
Addresses https://github.com/webpack/webpack.js.org/issues/3168


montogeek:
Thanks!

PR #3177, opened 2019-07-08
Clarify the affected Node versions

AshKyd:
The original text implies this is a current issue, but the fix has long been made.

This whole heading could possibly be removed since latest Node 8.x.x is fixed, and v9 doesn't appear to be a thing. Thoughts?


src/content/guides/build-performance.md
```diff
@@ -199,9 +199,11 @@ module.exports = {
 };
 ```
 
-### Node.js Version
+### Node.js Versions 8.9.10-9.11.1
 
-There has been a [performance regression](https://github.com/nodejs/node/issues/19769) in the latest stable versions of Node.js and its ES2015 `Map` and `Set` implementations. A fix has been merged into master, but a release has yet to be made. In the meantime, to get the most out of incremental build speeds, try to stick with version 8.9.x (the problem exists between 8.9.10 - 9.11.1). webpack has moved to using those ES2015 data structures liberally, and it will improve the initial build times as well.
+There was a [performance regression](https://github.com/nodejs/node/issues/19769) in Node.js versions 8.9.10 - 9.11.1 in the ES2015 `Map` and `Set` implementations. Webpack uses those data structures liberally, so this regression affects compile times.
```
montogeek:
```suggestion
There was a [performance regression](https://github.com/nodejs/node/issues/19769) in Node.js versions 8.9.10 - 9.11.1 in the ES2015 `Map` and `Set` implementations. webpack uses those data structures liberally, so this regression affects compile times.
```

montogeek:
@EugeneHlushko I don't get why this is failing :/

EugeneHlushko:
Looks like bunch of svg's 500-ed, lets restart by updating the branch

EugeneHlushko:
Thanks!

PR #3179, opened 2019-07-09
docs: added a note about alias resolution

Aghassi:
We came across this when working with our tooling where `alias` takes precedence over all other module resolution patterns. It was confusing at first, so I wanted to add it to the docs as a callout for others.


src/content/configuration/resolve.md
```diff
@@ -112,6 +113,8 @@ The following table explains other cases:
 
 `/abc/node_modules` may resolve in `/node_modules` too.
 
+W> It should be noted that using `alias` will take precedence over any other module resolution as Webpack interprets it to be an exact location of a file.
```
montogeek:
```suggestion
W> It should be noted that using `alias` will take precedence over any other module resolution as webpack interprets it to be an exact location of a file.
```

EugeneHlushko: Taking it further
```suggestion
W> `resolve.alias` will takes precedence over other module resolutions.
```

Aghassi:
@EugeneHlushko Can you point me to where you want this instead? I thought I did edit the configurations page...

src/content/configuration/resolve.md
```diff
@@ -112,6 +113,8 @@ The following table explains other cases:
 
 `/abc/node_modules` may resolve in `/node_modules` too.
 
+W> `resolve.alias` will takes precedence over other module resolutions.
```
EugeneHlushko: It can either be `X will take precedence over Y` or `X takes precedence over Y` but not both (will, takes).
```suggestion
W> `resolve.alias` takes precedence over other module resolutions.
```

Aghassi:
@EugeneHlushko Thanks! I've updated the PR :). Thanks for the review!

EugeneHlushko:
Thanks!

PR #3180, opened 2019-07-09
ui(search) improve mobile search ux/ui; close button on notification

### LINKED ISSUE #3052: 🐞[search]: Search result overflows.
**Description**
Algolia search results overflow from the page in mobile devices.

Screenshot_2019-05-16-15-19-12

**How to reproduce**
1. Go to https://webpack.js.org/ in a mobile device or simulate one.
2. In the search field type `module.iden`.
3. You can see the search results overflowing from the page in X-direction.

**Expected Behaviour**
Shouldn't overflow.

**Reason for this issue**
IMO it's due to the character size of the title of this [doc](https://github.com/webpack/webpack.js.org/blame/d5bff764a00ff5c71076571cbe1b7c0b94914211/src/content/plugins/eval-source-map-dev-tool-plugin.md#L2)

EugeneHlushko:
Search feels terrible on mobile right now... So i took a shot on trying to not create our own docsearch.css but to add a few overrides, most of them are to inline style therefore hold the !important flag. If we'd get rid of this would mean to clone and compile our own docsearch.css.

- Make sure search box fits screen
- Use full width for screens under docsearch's max-width of search box (600px;)
- Different offset for case with notification and without.
- Fix notification's close button, it is covered with text now :(

Search before:
Screen Shot 2019-07-09 at 8 37 51 PM

Search after: (notice how it now fits the screen, try more on preview)
Screen Shot 2019-07-09 at 8 38 30 PM

And notification bar, close button.
Before:
Screen Shot 2019-07-09 at 8 39 46 PM

After:
Screen Shot 2019-07-09 at 8 39 57 PM

Might actually close #3052 but not sure

### DIFF
--- a/src/components/Navigation/Search.scss
+++ b/src/components/Navigation/Search.scss
@@ -1,13 +1,35 @@
 @import 'functions';
 
+// use full width for search results
+.navigation-search .algolia-autocomplete .ds-dropdown-menu {
+  @media (max-width: 600px) {
+    position: fixed !important;
+    left: 0 !important;
+    right: 0 !important;
+    top: 54px !important;
+    min-width: 0 !important;
+  }
+}
+
+// when notification is shown
+.notification-bar + header .navigation-search .algolia-autocomplete .ds-dropdown-menu {
+  @media (max-width: 600px) {
+    top: 110px !important;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


montogeek:
Thanks!

PR #3183, opened 2019-07-10
ui(search) fix mobile auto zoom on search input

EugeneHlushko:
Fixes the auto zoom when user focuses on search input. IOS zooms on any input which has font size < 16px;

To reproduce, open https://webpack.js.org/ and use search input, screen will zoom in
To test the fix, go to preview and repeat steps

### DIFF
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -4,7 +4,6 @@ import Banner from 'react-banner';
 
 // Import Components
 import Link from '../Link/Link';
-import Container from '../Container/Container';
 import Logo from '../Logo/Logo';
 import Dropdown from '../Dropdown/Dropdown';
 
--- a/src/components/Navigation/Navigation.scss
+++ b/src/components/Navigation/Navigation.scss
@@ -107,7 +107,7 @@
 }
 
 .navigation-search__input {
-  font-size: 14px;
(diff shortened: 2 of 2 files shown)


src/components/Navigation/Navigation.jsx
```diff
@@ -4,7 +4,6 @@ import Banner from 'react-banner';
 
 // Import Components
 import Link from '../Link/Link';
-import Container from '../Container/Container';
```
EugeneHlushko: Cleanup, apparently it was imported but not used

anshumanv:
Hi, is it possible to test this on the devices tab or can this be only tested on iOS?

EugeneHlushko:
Hi @anshumanv 
This is ios "feature", only on real devices

EugeneHlushko:
a small ref from quick google, i had a better source but i know about this for years now i dont remember from where: https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone

montogeek:
Fix is just to change the font size?

EugeneHlushko:
Yes, i am surprised that nobody did fix this before. This is pretty known behavior tbh. By the way, it affects different kinds of inputs. Beware, on your other projects

montogeek:
Thanks!

PR #3187, opened 2019-07-12
docs: improve visibility of scaffolds

rishabh3112:
## Motivation
Scaffolds are like project starters that made on top of the scaffolding system provided by at webpack-cli. These scaffolds will also be supported by [webpack-ui](https://github.com/webpack-ui/webpack-ui) Currently, there are not so many scaffolds so this might help to get more scaffolds to npm.
- To get an idea of how scaffolds would be working (kind of PoC) [rishabh3112/webpack-ui](https://github.com/rishabh3112/webpack-ui)

## Further Plans
- Add [webpack-scaffold-starter](https://github.com/rishabh3112/webpack-scaffold-starter) under `writing a scaffold` section (if it makes sense).
- Adding available webpack scaffolds under [Starter Kits](https://webpack.js.org/starter-kits/) section.

cc @evenstensberg

### DIFF
--- a/src/content/contribute/index.md
+++ b/src/content/contribute/index.md
@@ -46,6 +46,7 @@ The remainder of this section of the site is dedicated to developers such as you
 
 - [Writing a Loader](/contribute/writing-a-loader)
 - [Writing a Plugin](/contribute/writing-a-plugin)
+- [Writing a Scaffold](/contribute/writing-a-scaffold)
 - [Plugin Patterns](/contribute/plugin-patterns)
 - [Release Process](/contribute/release-process)
 


rishabh3112:
cc @EugeneHlushko

EugeneHlushko:
Makes sense to list it there, despite the fact that scaffolds arent widely used, there is no counterargument against that.

EugeneHlushko:
Thanks!

rishabh3112:
Well yes, Currently they aren't used at all I must say!
Although as webpack ecosystem lacks offering default configurations for projects officially, this can be a great addition to the webpack ecosystem in future when we build tools over scaffolds (like we have UI under development) so as to make sure it actually increases the efficiency of the users!

PR #3190, opened 2019-07-13
docs: add introduction and quickstart

rishabh3112:
Extends #3187 
It updates the intro section with an introduction to scaffolds. It also adds a quick start section.


src/content/contribute/writing-a-scaffold.md
```diff
@@ -8,9 +8,20 @@ contributors:
   - rishabh3112
 ---
 
-Welcome to the demonstration of the new `webpack init` command! To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
+A scaffold is a node module which exports an `yeoman generator` based upon [Scaffolding API](/guides/scaffolding/#api). Scaffolds are used to initialise a new webpack project, tailored to a specific use case. To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
```
jamesgeorge007:
```suggestion
A scaffold is a node module which exports a `yeoman generator` based upon [Scaffolding API](/guides/scaffolding/#api). Scaffolds are used to initialise a new webpack project, tailored to a specific use case. To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
```

rishabh3112: Thanks!

rishabh3112:
cc @EugeneHlushko

src/content/contribute/writing-a-scaffold.md
```diff
@@ -8,9 +8,20 @@ contributors:
   - rishabh3112
 ---
 
-Welcome to the demonstration of the new `webpack init` command! To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
+A scaffold is a node module which exports a `yeoman generator` based upon [Scaffolding API](/guides/scaffolding/#api). Scaffolds are used to initialise a new webpack project, tailored to a specific use case. To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
```
EugeneHlushko: node module is not correct, also scaffold could do more than that, so maybe rephrase a bit

```suggestion
A scaffold is an ECMAScript Module which exports a `yeoman generator` based on [Scaffolding API](/guides/scaffolding/#api). Scaffolds can be used but not limited to initialize a new webpack project, tailored to a specific use case. To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
```

src/content/contribute/writing-a-scaffold.md
```diff
@@ -8,9 +8,20 @@ contributors:
...
 
-## Create a basic structure
+## Quick Start
+
+[`webpack-scaffold-starter`](https://github.com/rishabh3112/webpack-scaffold-starter) can be used to setup a new scaffold project. To do so, follow the following commands.
```
EugeneHlushko: To recommend a personal project on official docs website, it'd need to get a full code review. I will not approve this nor do i have the time. maybe @evenstensberg has the time to do it and take the responsibility, especially considering this will get file system access.

src/content/contribute/writing-a-scaffold.md
```diff
@@ -8,9 +8,20 @@ contributors:
...
+
+```bash
+npm install -g webpack-scaffold-starter
```
EugeneHlushko: It doesnt have to be -g, you could use it inside of the future project, especially considering you are leveraging `mkdir`.

That could be 
```
mkdir myscaffold && cd myscaffold
npm install webpack-scaffold-starter
./node_modules/.bin/webpack-scaffold-starter
```

src/content/contribute/writing-a-scaffold.md
rishabh3112: We can have it under webpack or webpack-contrib if that helps in resolving this issue.

EugeneHlushko: That would solve the concern of linking a tool from docs

src/content/contribute/writing-a-scaffold.md
rishabh3112: I wouldn't recommend because by that way `webpack-scaffold-starter` would be added to the user's dependency, which I think shouldn't be recommended, considering the fact it will be of no use after the initial scaffolding of the project.

src/content/contribute/writing-a-scaffold.md
rishabh3112: So what you want me to do in this regard?
P.S.
No offence, but as every node package can have the file system access, No personal project should be on the official docs ( even the plugins/loaders under personal accounts ) if going by the same argument.

src/content/contribute/writing-a-scaffold.md
rishabh3112: > To add dependencies and devDependencies to a package.json file from the command line, **you can install them in the root directory of your package using the --save-prod flag for dependencies (the default behavior of npm install)** or the --save-dev flag for devDependencies.

from https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file. Tested it for your reference:
**Command**
```bash
 $ npm install webpack-scaffold-starter
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN a@1.0.0 No description
npm WARN a@1.0.0 No repository field.

+ webpack-scaffold-starter@1.1.1
added 37 packages from 33 contributors and audited 38 packages in 36.127s
found 0 vulnerabilities

```
**package.json**
```json
{
  "name": "a",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "webpack-scaffold-starter": "^1.1.1"
  }
}
```
Also Preference order for usage in context to DX IMO `webpack-scaffold` > `npx webpack-scaffold` > `./node_modules/.bin/webpack-scaffold-starter`

EugeneHlushko: It shouldnt save by default though, anyway npx > -g thats why it was invented

rishabh3112: 1. it's saving by default see above ☝
2. I don't want to fall into a large discussion on DX and npm, I would do the changes you suggested if you want it that way. 😅🙌

TIL: Thanks for suggesting `./node_modules/.bin/webpack-scaffold-starter`  I wasn't aware of this before that  bin for local packages are at `./node_modules/.bin/*` ❤

EugeneHlushko: Haha glad it helped :)

On the command itself lets ask them to do npx as we already say to use npx in guides section

src/content/contribute/writing-a-scaffold.md
EugeneHlushko: Yeah i dont mean any offence either, but since you are creating this for webpack-cli, why cannot it be under webpack-cli ownership, then there are no risks in such linking

EugeneHlushko: @sokra what do you think? If you say we should do it as is, i don't mind.

src/content/contribute/writing-a-scaffold.md
```diff
@@ -8,9 +8,20 @@ contributors:
...
-Welcome to the demonstration of the new `webpack init` command! To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
+A scaffold is an ECMAScript Module which exports a `yeoman generator` based on [Scaffolding API](/guides/scaffolding/#api). Scaffolds can be used but not limited to initialize a new webpack project, tailored to a specific use case. To view what we are building today, run `webpack init webpack-scaffold-demo`. This demo will show you how to build your own webpack scaffold. Let's start by creating a file named `generator.js`.
 
-## Create a basic structure
+## Quick Start
+
+[`webpack-scaffold-starter`](https://github.com/rishabh3112/webpack-scaffold-starter) can be used to setup a new scaffold project. To do so, follow the following commands.
+
+```bash
+mkdir myscaffold && cd myscaffold
```
evenstensberg: maybe follow the convention of hyphens, `my-scaffold`

src/content/contribute/writing-a-scaffold.md
rishabh3112: >  why cannot it be under webpack-cli ownership, then there are no risks in such linking

This is package is supposed to be standalone, thus the solutions includes
1. Have under either `webpack` or `webpack-contrib`
2. Let it be as it is now

P.S.
Let me know if this PR should be closed. Thanks.

EugeneHlushko:
Thanks!

PR #3192, opened 2019-07-14
dosc(guides) mention unstable node caveat in the guide

### LINKED ISSUE #3182: Bug in Getting Started Guide (easy workaround)
# Bug report

The Getting Started guide fails at the end of the Creating a Bundle section.

Follow the guide as is.

At the end of that section, the browser should display Hello webpack, but instead you get a syntax error in the middle of the minified javascript.

So it looks like there's some kind of compatibility issue between the lodash library and the default minimizer used in production mode (which the guide uses since it's the default mode and it's not specified.) I think the escaped values in lodash/_deburrLetter.js are getting translated into their literal values, and in the process of doing that one of the values ends up missing quotes for some reason and causes a syntax error.

An easy work-around in the guide is to set the mode to development, which is probably fine for the purposes of the guide.

Figured I'd report this for the benefit of people like me who are just getting started with webpack.

webpack version:  4.35.3
Node.js version: 12.5.0
Operating System: macos (darwin)
Additional tools:

EugeneHlushko:
Closes #3182

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -201,7 +201,7 @@ T> Your output may vary a bit, but if the build is successful then you are good 
 
 Open `index.html` in your browser and, if everything went right, you should see the following text: 'Hello webpack'.
 
-W> If you are getting a syntax error in the middle of minified JavaScript when opening `index.html` in the browser, set [`developemnt mode`](/configuration/mode/#mode-development) and run `npx webpack` again. This is related to running `npx webpack` on latest Node.js (v12.5+) instead of [LTS version](https://nodejs.org/en/).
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.md
```diff
@@ -201,6 +201,8 @@ T> Your output may vary a bit, but if the build is successful then you are good
 
 Open `index.html` in your browser and, if everything went right, you should see the following text: 'Hello webpack'.
 
+W> If you are getting a syntax error in the middle of minified JavaScript when opening `index.html` in the browser, set [`developemnt mode`](/configuration/mode/#mode-development) and run `npx webpack` again. This is related to running `npx webpack` on latest Node.js (v12.5+) instead of [LTS version](https://nodejs.org/en/).
```
montogeek:
```suggestion
W> If you are getting a syntax error in the middle of minified JavaScript when opening `index.html` in the browser, set [`development mode`](/configuration/mode/#mode-development) and run `npx webpack` again. This is related to running `npx webpack` on latest Node.js (v12.5+) instead of [LTS version](https://nodejs.org/en/).
```

EugeneHlushko:
Right, great catch mate!

montogeek:
Thanks!

PR #3196, opened 2019-07-16
Update docs to explain what I assume "Dll" means

adamloving:
I'm just assuming this definition. I just discovered Webpack DLL, and find this terminology a little awkward. DLLs are traditionally binary files linked at run time, not build time (or maybe both, I can't remember). In any case, when I discovered Webpack DLL I couldn't find a definition for "Dll."

### DIFF
--- a/src/content/plugins/dll-plugin.md
+++ b/src/content/plugins/dll-plugin.md
@@ -13,7 +13,7 @@ related:
     url: https://github.com/webpack/webpack/blob/master/examples/explicit-vendor-chunk/README.md
 ---
 
-The `DllPlugin` and `DllReferencePlugin` provide means to split bundles in a way that can drastically improve build time performance. The term "DLL" abbreviates Dynamic Link Library (taken from Microsoft's implementation linking binary code after it is compiled).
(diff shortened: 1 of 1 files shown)


src/content/plugins/dll-plugin.md
```diff
@@ -13,7 +13,7 @@ related:
     url: https://github.com/webpack/webpack/blob/master/examples/explicit-vendor-chunk/README.md
 ---
 
-The `DllPlugin` and `DllReferencePlugin` provide means to split bundles in a way that can drastically improve build time performance.
+The `DllPlugin` and `DllReferencePlugin` provide means to split bundles in a way that can drastically improve build time performance. The term "DLL" abbreviates Dynamic Link Library (taken from Microsoft's implementation linking binary code after it is compiled).
```
EugeneHlushko: From wiki: https://en.wikipedia.org/wiki/Dynamic-link_library#Background_for_DLL

```suggestion
The `DllPlugin` and `DllReferencePlugin` provide means to split bundles in a way that can drastically improve build time performance. The term "DLL" stands for Dynamic-link library which was originally introduced by Microsoft.
```

montogeek:
Thanks!

PR #3199, opened 2019-07-17
docs(config) document Rule.resolve

EugeneHlushko:
This is the missing configuration i mentioned on the #3197 
Although to completely close the issue we need to figure our where to add the documentation on the `getResolve` resolver method. Currently its not on webpack.js.org and not on https://github.com/webpack/enhanced-resolve

I guess i could be tracked in #3197 and future prs

### DIFF
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -403,6 +403,27 @@ module.exports = {
 
 See [UseEntry](#useentry) for details.
 
+## `Rule.resolve`
+
+Resolving can be configured on module level. For options see [resolve configuration page](/configuration/resolve/#resolve). All applied resolve options are merged.
+
+__webpack.js.org__
+
+```javascript
+module.exports = {
+  //...
+  module: {
+    rules: [
+      {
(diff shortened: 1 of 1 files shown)


src/content/configuration/module.mdx
```diff
@@ -403,6 +403,27 @@ module.exports = {
 
 See [UseEntry](#useentry) for details.
 
+## `Rule.resolve`
+
+Resolving can be configured on module level. For options see [resolve configuration page](/configuration/resolve/#resolve). All applied resolve options are merged.
```
montogeek: > All applied resolve options are merged.

What does this mean? Merged where?

EugeneHlushko: With configuration resolve (high level)

EugeneHlushko: previous link in text for reference

EugeneHlushko: hope the new version is more accessible

montogeek:
Nice example, thanks!

PR #3200, opened 2019-07-17
Remove statement about ES module browser support

taymonbeal:
[All modern browsers](https://caniuse.com/#feat=es6-module) support them now.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -206,7 +206,7 @@ W> If you are getting a syntax error in the middle of minified JavaScript when o
 
 ## Modules
 
-The [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) statements have been standardized in [ES2015](https://babeljs.io/docs/en/learn/). Although they are not supported in most browsers yet, webpack does support them out of the box.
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks for your contribution. I think this note is still necessary, we have to acknowledge the existence of other browsers and specially those that doesn't support features.
Something like this would be better:

> Although they are not supported by all browsers, some modern browsers do, webpack does support them out of the box.

src/content/guides/getting-started.md
```diff
@@ -206,7 +206,7 @@ W> If you are getting a syntax error in the middle of minified JavaScript when o
 
 ## Modules
 
-The [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) statements have been standardized in [ES2015](https://babeljs.io/docs/en/learn/). Although they are not supported in most browsers yet, webpack does support them out of the box.
+The [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) statements have been standardized in [ES2015](https://babeljs.io/docs/en/learn/). webpack supports them out of the box.
```
EugeneHlushko:
```suggestion
The [`import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) statements have been standardized in [ES2015](https://babeljs.io/docs/en/learn/). Although they are not supported in some browsers yet, webpack supports them out of the box.
```

montogeek:
@taymonbeal Friendly ping

chenxsan:
Closed as they're already improved as per https://github.com/webpack/webpack.js.org/blob/master/src/content/guides/getting-started.md#modules

PR #3202, opened 2019-07-19
docs(api) document compilation object

EugeneHlushko:
Takes a bit off #3148 

Part of the methods dont have types and descriptions yet.

Reference: https://github.com/webpack/webpack/blob/master/lib/Compilation.js

### DIFF
--- a/src/content/api/compilation-object.md
+++ b/src/content/api/compilation-object.md
@@ -22,6 +22,8 @@ Returns Stats object for the current compilation.
 
 Returns Stats object for the current compilation.
 
+Parameters:
+
 - `module` - module to be added
 - `cacheGroup` - `cacheGroup` of the module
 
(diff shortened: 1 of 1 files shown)


src/content/api/compilation-object.md
```diff
@@ -0,0 +1,322 @@
...
+`function`
+
+Returns Stats object for the current compilation.
+
+### addModule
+
+`function (module, cacheGroup)`
+
+Returns Stats object for the current compilation.
+
```
rishabh3112:
```suggestion

Parameters:

```

EugeneHlushko: ty! great catch

EugeneHlushko:
@montogeek a reminder to take a look :) would to get this in as i have a page to add to API, which will mean a conflict in `sort` field

montogeek:
Thanks!

PR #3206, opened 2019-07-23
docs(concepts): fix typos in the Entry Points section

kyarik:
Added a missing space in "entrypoint" and changed the verb "increase" to "increases" since the subject is "the number of entry points".


EugeneHlushko:
Thanks!

PR #3207, opened 2019-07-23
Updating with new babel-present-env modules link

fokusferit:
I wanted to read something about the ModuleConcatenationPlugin and realized that the URL to the Babel modules configuration is leading to the old 2015 preset.

Just Updating this with the new https://babeljs.io/docs/en/babel-preset-env#modules.

This PR is only updating the modules Url for the Babel reference.


montogeek:
Thanks!

PR #3208, opened 2019-07-23
docs(config) document logging feature

EugeneHlushko:
This is on master, review asap pls

Takes off configuration part from #3205


montogeek:
Very good, thanks!

PR #3210, opened 2019-07-24
 docs(api) document infrastructureLog and log compiler hooks 

EugeneHlushko:
Continuing to work towards documenting bunch of changes from #3205

Taken care of __Plugins__ task from the list which is around documenting `Compiler.hooks.infrastructureLog` and` Compilation.hooks.log` hooks


montogeek:
Thanks!

PR #3224, opened 2019-08-01
Use words instead of ASCII for speed ratings

tigt:
I found this table very hard to scan (dashes are often used in tables to indicate no data), and the emojis seem more intuitive since you don't see the legend first.


tigt:
Bikeshedding on what emojis seem faster than others is totally encouraged. Other candidates:
* 🏎
* 🚄
* 🛳
* 🚜
* 🚶‍♀️
* 🏃‍♀️
* 🐢
* 🐇
* ☄️
* 🌩
* 😃🙂😌😐😒😣 (alternate metaphor than travel velocity)

montogeek:
Personally I don't find this intuitive at all, neither the + symbols, it would be better to use the legend directly instead of any symbol.

tigt:
I agree with that. I assumed the symbols were chosen for compactness
reasons, but inspecting the table shows there are some malfunctioning
.markdown .description styles that fixing would restore more than enough
space to put in the full words.

EugeneHlushko:
I like this, @montogeek ?

montogeek:
Definitely better.

PR #3231, opened 2019-08-04
docs(cli): fix multiple-entry cli command

vsramanujan:
- The `-o` tag was missing in the multiple-entry CLI command


montogeek:
Thanks!

PR #3232, opened 2019-08-06
docs(loaders) document loaders logging usage

EugeneHlushko:
- As per title
- Takes one of last items in #3205

### DIFF
--- a/src/content/api/loaders.md
+++ b/src/content/api/loaders.md
@@ -648,6 +648,6 @@ module.exports = function(source) {
 
 Logging API is available since the release of webpack 4.37. When `logging` is enabled in [`stats configuration`](/configuration/stats/#stats) and/or when [`infrastructure logging`](/configuration/other-options/#infrastructurelogging) is enabled, loaders may log messages which will be printed out in the respective logger format (stats, infrastructure).
 
-- Loaders should prefer to use `this.getLogger()` for logging which is a shortcut to `compilation.getLogger()` with loader path and processed file. This kind of logging is stored to the Stats and formatted accordingly. It can be filtered and exported by the user.
+- Loaders should prefer to use `this.getLogger()` for logging which is a shortcut to `compilation.getLogger()` with loader path and processed file. This kind of logging is stored to the Stats and formatted accordingly. It can be filtered and exported by the webpack user.
 - Loaders may use `this.getLogger('name')` to get an independent logger with a child name. Loader path and processed file is still added.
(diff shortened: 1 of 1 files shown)


src/content/api/loaders.md
```diff
@@ -643,3 +643,11 @@ module.exports = function(source) {
   return match[0];
 };
 ```
+
+## Logging
+
+Logging API is available since the release of webpack 4.37. When `logging` is enabled in [`stats configuration`](/configuration/stats/#stats) and/or when [`infrastructure logging`](/configuration/other-options/#infrastructurelogging) is enabled, loaders may log messages which will be printed out in the respective logger format (stats, infrastructure).
+
+- Loaders should prefer to use `this.getLogger()` for logging which is a shortcut to `compilation.getLogger()` with loader path and processed file. This kind of logging is stored to the Stats and formatted accordingly. It can be filtered and exported by the user.
```
montogeek:
```suggestion
- Loaders should prefer to use `this.getLogger()` for logging which is a shortcut to `compilation.getLogger()` with loader path and processed file. This kind of logging is stored to the Stats and formatted accordingly. It can be filtered and exported by the loader author for the user.
```

User you mean the loader author or the final user of the loader?

src/content/api/loaders.md
```diff
@@ -643,3 +643,11 @@ module.exports = function(source) {
...
+- Loaders may use `this.getLogger('name')` to get an independent logger with a child name. Loader path and processed file is still added.
+- Loaders may use special fallback logic for detecting logging support `this.getLogger() ? this.getLogger() : console` to provide a fallback for cases when an older webpack version is used which does not support `getLogger` method.
```
montogeek:
```suggestion
- Loaders may use special fallback logic for detecting logging support `this.getLogger() ? this.getLogger() : console` to provide a fallback when an older webpack version is used which does not support `getLogger` method.
```

src/content/api/loaders.md
EugeneHlushko: the final user via webpack config

EugeneHlushko:
```suggestion
- Loaders should prefer to use `this.getLogger()` for logging which is a shortcut to `compilation.getLogger()` with loader path and processed file. This kind of logging is stored to the Stats and formatted accordingly. It can be filtered and exported by the webpack user.
```

EugeneHlushko: changed to `webpack user` @montogeek

Issue #3239, opened 2019-08-09
Dev Server documentation: transportMode option


Loonride:
__Feature to document__

transportMode option: https://github.com/webpack/webpack-dev-server/pull/2116

__Author(s)__

Myself

/cc @evilebottnawi @hiroppy 

__Additional information__

This option allows the user to specify which socket transport mode to use for communication between the dev server and client. It is still an experimental option. Usage:

string:
```
transportMode: 'sockjs'
```
```
transportMode: 'ws'
```

object, with `server` (can be string mode/path or class), `client` (string mode/path) properties:

equivalent to `ws` mode above:
```
transportMode: {
  server: 'ws',
  client: 'ws'
}
```

custom implementation:
```
transportMode: {
  server: require('path/to/custom/server'), // can be path or class (class seen here)
  client: require.resolve('path/to/custom/client') // must be path
}
```

[x] I am willing to work on this issue and submit a pull request.

jeffin143:
@montogeek , This is done and dusted, Let's close this :)

PR #3240, opened 2019-08-09
docs(dev-server): add transportMode docs and examples

Loonride:
_describe your changes..._

`devServer.transportMode`: https://github.com/webpack/webpack.js.org/issues/3239

I think that, rather than having large custom implementation examples in here, we should link to implementation examples in `webpack-dev-server/examples`.


src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
 ```
 
 
+## `devServer.transportMode`
+
+`string: 'sockjs' | 'ws'` `object`
```
EugeneHlushko: The default value is object, lets move it to the first place and mention the default
```suggestion
`object: { client: 'sockjs',  server: 'sockjs' }` `string: 'sockjs' | 'ws'`
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+
+W> `transportMode` is an experimental option, meaning its usage could potentially change without warning.
+
+This option allows you to choose the current `devServer` client/server transport mode, or provide your own custom client/server implementation. This allows you to specify how your browser, or other client, communicates with the `devServer`. The `devServer` then uses this to inform the browser client when the page should reload, for example.
```
```suggestion
This option allows to choose the current `devServer` transport mode for client/server individually or to provide custom client/server implementation. This allows to specify how browser or other client communicates with the `devServer`.

Providing a string to `devServer.transportMode` is a shortcut to setting both `devServer.transportMode.client` and `devServer.transportMode.server` to the given string value.
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+```javascript
+module.exports = {
+  //...
+  devServer: {
+    transportMode: 'ws'
+  }
+};
+```
+
+You can also create custom client/server `transportMode` implementations. The client and server implementations must be compatible with one another to communicate successfully.
```
EugeneHlushko: We already said this earlier.
```suggestion
When providing a custom client and server implementation make sure that they are compatible with one another to communicate successfully.
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+
+To create a custom server implementation, you must create a class that extends [`BaseServer`](https://github.com/webpack/webpack-dev-server/blob/master/lib/servers/BaseServer.js), found in the `lib/servers/` directory of the `webpack-dev-server` module.
```
```suggestion
To create a custom server implementation, create a class that extends [`BaseServer`](https://github.com/webpack/webpack-dev-server/blob/master/lib/servers/BaseServer.js).
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+
+<!-- TODO: add link to a custom server implementation example -->
```
EugeneHlushko: dont put TODO here, file an issue

```suggestion
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+
+To create a custom client implementation, you must create a class that extends [`BaseClient`](https://github.com/webpack/webpack-dev-server/blob/master/client-src/clients/BaseClient.js), found in the `client-src/clients/` directory of the `webpack-dev-server` module.
```
```suggestion
To create a custom client implementation, create a class that extends [`BaseClient`](https://github.com/webpack/webpack-dev-server/blob/master/client-src/clients/BaseClient.js).
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+
+<!-- TODO: add link to a custom client implementation example -->
```
```suggestion
```

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,108 @@ webpack-dev-server --stdin
...
+
+Using path to `CustomServer.js`, a custom WebSocket server implementation, along with the compatible `'ws'` client:
```
EugeneHlushko: From here, create two h3 tags by using `### devServer.transportMode.server` and giving related info per item, mention possible types per value. E.g. client `string` `path`, server `string` `function` `path`

src/content/configuration/dev-server.md
Loonride: I think that it makes more sense to say that the default is 'sockjs', even if it is internally that equivalent object. Most basic users I suspect would only make use of the string options.

src/content/configuration/dev-server.md
Loonride: Can it be `class` instead of `function`?

src/content/configuration/dev-server.md
Loonride: Maybe something like: `string: 'sockjs' | 'ws'` `object: { client: ... ,  server: ... }`

and then elaborate on the object properties later?

Loonride:
I did a bit a bit of minor reorganization for the new `transportMode.client` and `transportMode.server` sections.

src/content/configuration/dev-server.md
EugeneHlushko: depends on the validator, check json schema

src/content/configuration/dev-server.md
EugeneHlushko: okay as long as we say that setting `devServer.transportMode` to a string is a shortcut to setting both `devServer.transportMode.client | server` to that value

src/content/configuration/dev-server.md
Loonride: @EugeneHlushko Yes the schema will accept a class, and that is what users should be passing in. What I meant was, are you fine with me listing it as a `class` in the docs rather than `function`.  I will make the change, if it is a problem let me know.

EugeneHlushko: got it, please use `class` and if schema will also accept a function use both e.g. `class` `function`

Loonride: Done. It accepts both. The validator will always accept `class` when the schema is `"instanceof": "Function"`, for this reason: https://stackoverflow.com/questions/42100659/why-in-javascript-class-a-instanceof-function-but-typeof-class-a-is-not-an-obje

> classes are just syntactic sugar for constructor functions. I.e. the evaluation of class A {} produces a function.

EugeneHlushko: Deeply apologize, but we had finally have defined the types and defaults definitions, please use `function` here as per: https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types

Loonride:
@EugeneHlushko Currently the types I have are:

devServer.transportMode:
`string = 'sockjs': 'sockjs' | 'ws'` `object`

devServer.transportMode.client:
`string` `path`

devServer.transportMode.server:
`string` `path` `function`

Should `path` be there, or should we just let that fall under `string`? If you think these are not specific enough, could you suggest how to make it fit in with the Writer's Guide guidelines?

- Could `string` for the client and server be changed to `string = 'sockjs': 'sockjs' | 'ws' | path`?
- Could I do `function: class`? I just need something making it clear that providing a plain function like `server: () => {}` will not be sufficient

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,113 @@ webpack-dev-server --stdin
...
 
+## `devServer.transportMode`
+
+`string = 'sockjs': 'sockjs' | 'ws'` `object`
+
+W> `transportMode` is an experimental option, meaning its usage could potentially change without warning.
+
+T> Providing a string to `devServer.transportMode` is a shortcut to setting both `devServer.transportMode.client` and `devServer.transportMode.server` to the given string value.
+
+This option allows to choose the current `devServer` transport mode for client/server individually or to provide custom client/server implementation. This allows to specify how browser or other client communicates with the `devServer`.
```
wizardofhogwarts:
```suggestion
This option allows us either to choose the current `devServer` transport mode for client/server individually or to provide custom client/server implementation. This allows to specify how browser or other client communicates with the `devServer`.
```

EugeneHlushko: Looks better, CC @Loonride

src/content/configuration/dev-server.md
```diff
@@ -1353,6 +1353,113 @@ webpack-dev-server --stdin
...
+
+The current default mode is [`'sockjs'`](https://www.npmjs.com/package/sockjs). This mode uses [SockJS-node](https://github.com/sockjs/sockjs-node) as a server, and [SockJS-client](https://www.npmjs.com/package/sockjs-client) on the client.
+
+Use `'sockjs'` mode:
```
EugeneHlushko: I just realized, we dont give examples of default behavior, can you please get rid of that?

EugeneHlushko: The rest looks good

Loonride: You meant remove the example, right?

EugeneHlushko:
Yes, i think it looks very good now

EugeneHlushko:
Thanks!

PR #3242, opened 2019-08-11
docs(config) fix up inconsistency

EugeneHlushko:
This is very minor but caught my eye when checking config pages


montogeek:
Thanks!

PR #3245, opened 2019-08-12
docs(guides): fix typos and a link hash

kyarik:
In the [*Development*](https://webpack.js.org/guides/development/) guide,

- Replaced "lets" with "let's" when it was meant to be a contraction of "let us".
- Added a missing comma.
- Changed the capitalization of an anchor link's text so it's consistent with the title of the section that is being pointed to.
- Replaced "find the bundle files in different path" with a more appropriate "find the bundle files on a different path".
- Added backticks around `webpack-dev-middleware` where missing.
- ~~Fixed the link to the next guide, which is "Code Splitting" and not "Hot Module Replacement".~~ See https://github.com/webpack/webpack.js.org/pull/3245#pullrequestreview-273809991.
- Fixed the hash in the link to `devServer.publicPath`, which had an extra `-` (`#devserver-publicpath-` --> `#devserverpublicpath-`). The actual link is indeed https://webpack.js.org/configuration/dev-server/#devserverpublicpath-

### DIFF
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -231,7 +231,7 @@ __webpack.config.js__
 
 This tells `webpack-dev-server` to serve the files from the `dist` directory on `localhost:8080`.
 
-W> webpack-dev-server doesn't write any output files after compiling. Instead, it keeps bundle files in memory and serves them as if they were real files mounted at the server's root path. If your page expects to find the bundle files on a different path, you can change this with the [`publicPath`](/configuration/dev-server/#devserver-publicpath-) option in the dev server's configuration.
+W> webpack-dev-server doesn't write any output files after compiling. Instead, it keeps bundle files in memory and serves them as if they were real files mounted at the server's root path. If your page expects to find the bundle files on a different path, you can change this with the [`publicPath`](/configuration/dev-server/#devserverpublicpath-) option in the dev server's configuration.
 
 Let's add a script to easily run the dev server as well:
 
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -21,7 +21,7 @@ If you've been following the guides, you should have a solid understanding of so
 
 W> The tools in this guide are __only meant for development__, please __avoid__ using them in production!
 
-Before proceeding, let's first set [`mode` to `'development'`](/configuration/mode/#mode-development).
+Let's start by setting [`mode` to `'development'`](/configuration/mode/#mode-development).
 
 __webpack.config.js__
 
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -420,4 +420,4 @@ To disable this feature in some common editors, see the list below:
 
 ## Conclusion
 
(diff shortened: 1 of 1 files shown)


src/content/guides/development.md
```diff
@@ -420,4 +420,4 @@ To disable this feature in some common editors, see the list below:
 
 ## Conclusion
 
-Now that you've learned how to automatically compile your code and run a simple development server, you can check out the next guide, which will cover [Hot Module Replacement](/guides/hot-module-replacement).
+Now that you've learned how to automatically compile your code and run a simple development server, you can check out the next guide, which will cover [Code Splitting](/guides/code-splitting/).
```
EugeneHlushko: Logically, next is HMR, as this is development guide, despite the sidebar order.

src/content/guides/development.md
```diff
@@ -21,7 +21,7 @@ If you've been following the guides, you should have a solid understanding of so
 
 W> The tools in this guide are __only meant for development__, please __avoid__ using them in production!
 
-Before proceeding lets first set [`mode` to `'development'`](/configuration/mode/#mode-development).
+Before proceeding, let's first set [`mode` to `'development'`](/configuration/mode/#mode-development).
```
EugeneHlushko:
```suggestion
Let's start from setting [`mode` to `'development'`](/configuration/mode/#mode-development).
```

src/content/guides/development.md
kyarik: Yes, the HMR guide is mentioned and linked several times in this guide.

So, should the order in the sidebar change or should it be left as is and change the link to Code Splitting?

src/content/guides/development.md
kyarik: Maybe "Let's start **by** setting"? Sounds better...

EugeneHlushko: yup, sounds better, please go ahead and change

src/content/guides/development.md
EugeneHlushko: its logical next, not order next

We could change this to further reading similar as we have here: 
https://webpack.js.org/guides/hot-module-replacement/

Annotation 2019-08-12 210225

But i don't think this is really necessary unless there are people raising issues with it

kyarik: But shouldn't the guides be read in the order in which they are in the sidebar? I believe that's how most people read them. So, if the intent is to make people read about HMR next, shouldn't that be next in the sidebar?

kyarik: All right, I will set the next guide link back to HMR. However, I do suggest that the order in the sidebar is changed so that the HMR guide comes after the *Development* guide for the reason I mentioned in my previous comment.

kyarik:
I noticed that the hash in the link to `devServer.publicPath` had an extra `-`, so I removed it (`#devserver-publicpath-` --> `#devserverpublicpath-`). See https://webpack.js.org/configuration/dev-server/#devserverpublicpath-

EugeneHlushko:
Thanks!

PR #3257, opened 2019-08-16
docs(config) refactor stats options, update links, update info

EugeneHlushko:
Finally got my hands into refactoring the non accessible version of the stats config. Now all the options are properly documented (based on the available info and somewhere enhanced) and have new type system as a demonstration. If we decide to change the types notation i will update this PR.

- stats options have types and defaults in a common format
- stats options can now be linked to
- some of the stats options are now explained better 
- provided direct links to stats.logging\]

P.S. Sorry for the big diff but dont see how i'd split that work.

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -54,7 +54,7 @@ module.exports = {
   stats: {
     all: undefined
   }
-}
+};
 ```
 
 ### `stats.assets`
@@ -69,7 +69,7 @@ module.exports = {
   stats: {
     assets: false
   }
-}
+};
 ```
 
 ### `stats.assetsSort`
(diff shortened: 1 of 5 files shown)


src/content/configuration/stats.md
```diff
@@ -44,176 +44,686 @@ module.exports = {
...
-<!-- eslint-skip -->
+### `stats.all`
+
+A fallback value for stats options when an option is not defined. It has precedence over local webpack defaults.
+
+```javascript
+module.exports = {
+  //...
+  stats: {
+    all: undefined
```
EugeneHlushko: https://github.com/webpack/webpack/blob/master/lib/Stats.js#L112
this is so confusing, i've resued the previous information but would love to improve.

montogeek:
I am not sure about the `Tell` verb, but also not sure what to use instead

EugeneHlushko:
Lets revisit Tell later if there are ideas from us or other contributors

PR #3260, opened 2019-08-18
docs(config) apply consistent types format, small example added to co…

EugeneHlushko:
- consistent types across configuration section
- one more example to contrib for array mixed type

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -83,7 +83,7 @@ module.exports = {
 
 ## `devServer.allowedHosts`
 
-`array`
+`[string]`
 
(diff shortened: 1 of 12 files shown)


montogeek:
Really nice :D

montogeek:
Thanks!

EugeneHlushko:
@montogeek i am thinking to link our types to that contrib section explaining types in a new tab, what do you think?

EugeneHlushko:
not via <a href> but rather onlick

montogeek:
How would that work? A link every page that uses that notation? That is too much.

EugeneHlushko:
no, a global on click handler on a document was my initial thought @montogeek

PR #2606, opened 2018-10-19
docs(config) document configuration cache for webpack v5

EugeneHlushko:
Adds documentation to configuration of cache for webpack v5

Fixes #2605

PR #3261, opened 2019-08-19
docs(cache): remove cache.loglevel

skovy:
Remove `cache.loglevel` originally added in https://github.com/webpack/webpack.js.org/pull/2606.

This was removed in https://github.com/webpack/webpack/pull/9563/files.

Resolves https://github.com/webpack/webpack.js.org/issues/3246.

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -9,6 +9,7 @@ contributors:
   - liorgreenb
   - vansosnin
   - EugeneHlushko
+  - skovy
 related:
   - title: Using Records
     url: https://survivejs.com/webpack/optimizing/separating-manifest/#using-records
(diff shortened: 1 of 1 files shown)


src/content/configuration/other-options.md
```diff
@@ -143,33 +144,6 @@ module.exports = {
 };
 ```
 
-### `cache.loglevel`
```
EugeneHlushko: This isnt a way to go, it's a breaking change, therefore we add
```
W> Removed in webpack 5
```

EugeneHlushko: This feature is available in webpack 4 currently

skovy: Got it, currently undocumented though? 👍 

I will fix!

EugeneHlushko: Just confirmed that it is not available in latest webpack, for some reason i was sure it was added alongside logging at 4.37.0, can you please re-use the first commit? I see there was a force push

EugeneHlushko: sorry for the confusion there, logging caught me off

skovy: Updated 👌

EugeneHlushko:
Thanks for quick update @skovy and for PR!

PR #3264, opened 2019-08-20
chore(site): Upgraded remark-extract-anchors

montogeek:
Fixes https://github.com/montogeek/remark-extract-anchors/issues/2


Munter:
Awesome. Works like a charm! 💃

PR #3265, opened 2019-08-20
Fix broken fragment links

Munter:
Fixes a whole bunch of broken fragment links. These fixes are baseed on new reports from hyperlink, which correctly check fragment links in a new version.

I'll PR an update to the new version when the master branch has fewer errors, since it's a build blocker


Munter:
The problem is that the link checker in its current form is failing the build on errors, and some errors come from external repositories. So I need to get those externals in line before we can enable the blocking behavior. I think I'm pretty close though

Munter:
I think I got it now. I'll rebase and fix whatever has been introduced while this branch was alive

Munter:
Rebase successful. And all links and fragment check are passing :D

Munter:
I bypassed the problem with the external dependencies linking to the wrong place by adding a sort of redirect map for the fragment links I knew there were problems with. This means those specific wrong links will be fixes as part of the readme processing.

I'm outputting on the console when I fix an actual broken link with that map. In the ideal case we would pick these warnings up and start opening pull requests against the repositories in question so we could get the entire ecosystem in line with correct links

src/utilities/process-readme.js
```diff
@@ -1,6 +1,70 @@
...
+    const fragmentLinkMapMatch = Object.keys(fragmentLinkMap).find(mapFrom => href.includes(mapFrom));
+    if (fragmentLinkMapMatch) {
+      href = href.replace(fragmentLinkMapMatch, fragmentLinkMap[fragmentLinkMapMatch]);
+      console.error(`DEPRECATED EXTERNAL README LINK:
+  URL: ${sourceUrl}
+  ACTUAL: ${oldHref}
+  EXPECTED: ${oldHref.replace(fragmentLinkMapMatch, fragmentLinkMap[fragmentLinkMapMatch])}`);
+    }
+
+    // Lowercase all fragment links, since markdown generstors do the same
```
EugeneHlushko:
```suggestion
    // Lowercase all fragment links, since markdown generators do the same
```

src/utilities/process-readme.js
```diff
@@ -1,6 +1,70 @@
 const url = require('url');
 
-const beginsWithDocsDomainRegex = /^https?:\/\/webpack\.js\.org/;
+const beginsWithDocsDomainRegex = /^(?:https?:)\/\/webpack\.js\.org/;
+const inlineLinkRegex = /\[[^\]]*\]\(([^)]+)\)/g;
+const externalLinkRegex = /^[ \t]*\[[^\]]+\]:[ \t]+([^ \n]+)/gm;
+
+const fragmentLinkMap = {
```
EugeneHlushko: Can this include the faulty readme file source so we know where to put a PR to fix it?

src/utilities/process-readme.js
```diff
@@ -1,6 +1,70 @@
...
+
+    // Modify absolute documenation links to be root relative
+    if (beginsWithDocsDomainRegex.test(href)) {
+      href = href.replace(beginsWithDocsDomainRegex, '');
+    }
+
+    const fragmentLinkMapMatch = Object.keys(fragmentLinkMap).find(mapFrom => href.includes(mapFrom));
+    if (fragmentLinkMapMatch) {
+      href = href.replace(fragmentLinkMapMatch, fragmentLinkMap[fragmentLinkMapMatch]);
+      console.error(`DEPRECATED EXTERNAL README LINK:
```
EugeneHlushko: or here? (Can this include the faulty readme file source so we know where to put a PR to fix it?)

src/utilities/process-readme.js
Munter: The link map itself is source agnostic. It's just a mapping. The source will be highlighted in whatever code uses the map

src/utilities/process-readme.js
Munter: The url of the readme is shown here: https://github.com/webpack/webpack.js.org/pull/3265/files/3de50b8b70331e4943e8c9f6022069001378b7ce#diff-de191ad0d27673dbfaf4f71e79e397c0R49

See output here: https://travis-ci.org/webpack/webpack.js.org/jobs/576537198#L294-L297

Maybe I should make that a clickable edit link instead, to help us out a bit

Munter: Would actually be nice to pipe those errors into a Slack channel or something similar. No one is ever going to inspect this log

Munter: Github allows direct edit URL's with prefilled title and description: https://github.com/webpack-contrib/bundle-loader/edit/master/README.md?message=Fix%20broken%20link&description=This%20broken%20URL%20was%20found%20by%20the%20webpack.js.org%20link%20checker

This means the workflow could be reduced to clicking a link and manually making the replacement. Maybe we can prefill even more from the outside

montogeek:
Why this fragment generation broke?

Munter:
@montogeek there was a regression in hyperlink a while back when I added the switch to only check internal links. That also disabled the fragment checks, even though I thought I had test cases for it. The new version brings them back

src/utilities/process-readme.js
EugeneHlushko: thats enough, ty

EugeneHlushko: but the nice to haves are welcome at any time i suppose, they sound pretty handy

EugeneHlushko:
@montogeek want to take a look at the code or fine if i merge this? Looks good to me

montogeek:
@Munter Sorry, I don't understand how it is related to these changes in the hashes : from `#module-hot-webpack-specific-` to `#modulehot-webpack-specific`.

montogeek:
I think we broke it in the rebuild branch :/

EugeneHlushko:
Yeap or another thought i have is that we break it when generate on different dep versions on different systems or node versions

Munter:
@montogeek 	I'm pretty sure one of the changes we made on the rebuild branch change the way the fragments were generated. And because of the bug in hyperlink we never noticed the regression. Hopefully this should at least be an informed choice going forward, as it would explode with errors if we change the slugification of fragments now

PR #3268, opened 2019-08-24
Add units

aryzing:
I can hear my science teacher in the back of my head: "Always add units!"


aryzing:
Thanks for reviewing this PR. I'm going to go out on a limb here and say that I think that you interpreted that this PR was adding units to `minSize` (which as you correctly point out already states the units) rather than adding units to `maxSize`.

Please let me know if that's not the case, or how this PR can be otherwise improved.

EugeneHlushko:
Hi @aryzing 
Unfortunately i still think that we dont want to add units everywhere where the option is mentioned. When we would speak about timeout we wouldnt add ms e.g. "After the `timeout` milliseconds is triggered, `X` is called"

EugeneHlushko:
We can keep the PR in current shape and wait for more contributors reviews if you like

aryzing:
> we dont want to add units everywhere

I very much agree with you, which is why this PR adds units once. I think we can both agree that each variable's units should be stated at least once.

Bear in mind that the docs aren't necessarily for a seasoned devs like yourself that are already familiar with the config, but rather for folks like me just trying to figure out how to set up webpack properly.

The experience I had while browsing this section of the documents was not great: I was left puzzled not knowing exactly how to make this option work. I then invested some time using trial-and-error builds, and finally, by chance, stumbled upon the `minSize` section, all of which could have been avoided by stating the option's units.

It's a single line PR that does make a positive difference.

EugeneHlushko:
Thanks for elaborating, still i would suggest adding it here once: https://webpack.js.org/plugins/split-chunks-plugin/#splitchunksmaxsize

montogeek:
This is a good addition, it only says "number", but it could be anything.

PR #3274, opened 2019-09-05
Fix aspect ratio open collective sponsors

MartijnCuppens:
Some sponsors on the homepage have stretched logos. This PR fixes this by using `max-height` instead of `height`. Images are also aligned in the middle for a more natural look. I've also limited the width of platinum & gold members, this has no effect at the moment but is just a fix just in case someone has a superwide logo in the future.

## Before

webpack js org_(iPad Pro)

## After

webpack js org_(iPad Pro) (1)

The difference is most noticeable in the **intracto** and **TEKHATTAN** logos (Bronze sponsors).

### DIFF
--- a/src/components/Support/Support.scss
+++ b/src/components/Support/Support.scss
@@ -25,26 +25,37 @@
   }
 
   &__latest-avatar {
-    height: 64px;
+    max-height: 64px;
     max-width: 192px;
+    vertical-align: middle;
   }
 
   &__bronze-avatar {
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Hi @MartijnCuppens 
Can you please include screenshots of the before/after state in the description

MartijnCuppens:
Sure, done.

EugeneHlushko:
Thanks!

PR #3279, opened 2019-09-12
docs(stats): fix link to 'minimal' preset

Armanio:


src/content/configuration/stats.md
```diff
@@ -724,7 +724,7 @@ For `assetsSort`, `chunksSort` and `modulesSort` there are several possible fiel
 
 ### Extending stats behaviours
 
-If you want to use one of the pre-defined behaviours e.g. `'minimal'` but still override one or more of the rules, see [the source code](https://github.com/webpack/webpack/blob/master/lib/Stats.js#L1394-L1401). You would want to copy the configuration options from `case 'minimal': ...` and add your additional rules while providing an object to `stats`.
+If you want to use one of the pre-defined behaviours e.g. `'minimal'` but still override one or more of the rules, see [the source code](https://github.com/webpack/webpack/blob/master/lib/Stats.js#L1618-L1626). You would want to copy the configuration options from `case 'minimal': ...` and add your additional rules while providing an object to `stats`.
```
montogeek: It changed again: 
```suggestion
If you want to use one of the pre-defined behaviours e.g. `'minimal'` but still override one or more of the rules, see [the source code](https://github.com/webpack/webpack/blob/master/lib/stats/DefaultStatsPresetPlugin.js#L68-L74). You would want to copy the configuration options from `case 'minimal': ...` and add your additional rules while providing an object to `stats`.
```

EugeneHlushko: Yup this cannot go in, we need a reword like i suggested earlier or another similar solution without line binding: https://github.com/webpack/webpack.js.org/pull/3279#pullrequestreview-287697497

snitin315:
Fixed in master, this is no longer valid.

PR #3285, opened 2019-09-13
docs(config) document performance side effect of optimization.removeA…

EugeneHlushko:
Document performance side effect of `optimization.removeAvailableModules`


montogeek:
Thanks!

PR #3286, opened 2019-09-13
chore(deps) update and remove deps to kill of a few security warnings

EugeneHlushko:
upgrades and cleanup in deps motivated by security warnings in github admin interface


montogeek:
What about remark-mermaid?

EugeneHlushko:
Hi @montogeek 
I've removed it, couldn't find where we used it... Maybe you remember?

montogeek:
webpack configuration
https://github.com/webpack/webpack.js.org/blob/master/webpack.common.js

webpack.common.js
```diff
@@ -3,7 +3,6 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
 const mdPlugins = [
   require('remark-slug'),
-  require('remark-mermaid'),
```
EugeneHlushko: @montogeek it is removed here too, what is the point?

montogeek: That is the point, we need it.

EugeneHlushko: We need it for? the communication seems broken, i assume it is not used, therefore it is deleted. Can you mention where it is used and why we need it please :D

montogeek: It is used in the webpack config and we need it to render our logos and icons

EugeneHlushko: I dont see any effect by removing it not locally, not on ssg's result not on preview. Are you sure it has effect? Because it feels like its a b it different tool: https://github.com/temando/remark-mermaid#usage

montogeek:
You are right (https://github.com/webpack/webpack.js.org/commit/713f78be663b5222b3d805d9e4ec09fcbeb38f6c), it is doing nothing.

If it is, we will find out soon :)

EugeneHlushko:
i will monitor now

PR #3288, opened 2019-09-14
fix(links) fix hydration overriding wrong elements of the ssr html

### LINKED ISSUE #3287: [Bug] Incorrect url and title in side navigation of guide
I found the bug on this page: https://webpack.js.org/guides/asset-management/. 
It looks like the bug occurs for the "Output Management" section and below.

The incorrect "title" and "href" attribute values are set for the sections of the side navigation. The values for the links and their containing list elements are set as the ones for the proceeding section. 

For example, the list and anchor elements under "Output Management" have the values for their respective "title" and "href" attributes set to the values that should be set on the list and anchor elements under the "Development" section. The "Output Management" heading links to the "Development section".

EugeneHlushko:
Closes #3287

hydration with index based keys maps wrong data to ssred html elements, wrong indexes come from printable.md being in the content during ssg config run.


montogeek:
Please try to keep formatting changes out of normal PRs.
Also, I don't get issue error at all right now in production. Did you reproduce it?

ejez:
Issue still in production and also after this PR:

Screenshot from 2019-09-20 11-20-32

EugeneHlushko:
@ejez i cannot reproduce on netlife preview nor locally on this branch :O

ejez:
Delete browser cache/history related to the site, then open guides main url, reload it once or twice, then hover over a sub topic below "asset management".

Issue is not present when already on a sub topic url.

Try different devices/browsers

ejez:
(do not click on the subtopic links, just hover. clicking will take you to the correct suptopic, but hovering on or inspecting the sidebar item shows incorrect link)

EugeneHlushko:
@ejez i cannot reproduce with cache disabled, in incognito, etc. @montogeek its your call now

ejez:
webpack org js-incorrect-links-issue

firefox 69.0 / ubuntu 18.04

ejez:
Also when in a subtopic page and after refreshing the page, the subtree will show a different subtopic open:

Screenshot from 2019-09-25 17-52-06


Screenshot from 2019-09-25 17-58-55

ejez:
@EugeneHlushko you can reproduce here:
https://www.browserling.com/browse/win/7/firefox/68/https%3A%2F%2Fwebpack.js.org%2Fguides%2F

and here:
https://www.browserling.com/browse/win/7/firefox/68/https%3A%2F%2Fdeploy-preview-3288--webpackjsorg-netlify.netlify.com%2Fguides%2F

EugeneHlushko:
Yup its reproducible in that firefox, will take a look into it again, ty @ejez

EugeneHlushko:
Interestingly i cannot reproduce it in latest firefox either ( 69.0.2 (64-bit) )

montogeek:
I just installed a new Firefox channel, which is basically a fresh installation and it is reproducible.

Screenshot 2019-10-10 at 07 43 21


It is also reproducible in this PR Netflix preview deploy. As you can see in above image, "Output management" is missing from the server rendering, it is hydrated by React client side. That gives us 2 ideas:
1. Why is "Output management" not rendering in the server side like the rest of the links? This would be the correct fix.
2. What are we missing when building the app that is creating this _mismatch_ which is confusing React hydration?

ejez:
made a local test:

the issue disappears if we set `mode: 'none',` or `mode: 'development',` in `webpack.prod.js`

The issue might be narrowed down by checking individually the plugins enabled by `mode: 'production',` or setting `process.env.NODE_ENV` to `development`

will try to check later

ejez:
The issue disappears if we set `process.env.NODE_ENV` to `development` (while keeping `mode: 'production'`)

ejez:
While waiting to identify the issue, we can get the site working properly by disabling ssr for the sidebar,
I am not familiar with "react" but I tested the following and it is working:
https://deploy-preview-3314--webpackjsorg-netlify.netlify.com/guides/development/
PR #3314 

`src/components/Site/Site.jsx` :
```diff
 class Site extends React.Component {
   state = {
+    hasMounted: false,
     mobileSidebarOpen: false
   };
 
+  componentDidMount() {
+    this.setState({ hasMounted: true });
+  }
+
   render() {
     let { location } = this.props;
     let { mobileSidebarOpen } = this.state;

...

                         return (
                           <React.Fragment>
                             <Sponsors />
+                            { this.state.hasMounted ? (
                            <Sidebar
                              className="site__sidebar"
                              currentPage={location.pathname}
                              pages={this._strip(
                                section
                                  ? section.children
                                  : Content.children.filter(item => item.type !== 'directory' && item.url !== '/')
                              )}
                            />
+                            ) : '' }
                             <Page {...page} content={content} />
                             <Gitter />
                           </React.Fragment>
```

EugeneHlushko:
Disabling SSR for the sidebar will remove the links, this will mess up Google ranking of the pages and established internal websites linking.
I will look into what you guys found

EugeneHlushko:
@montogeek it looks like Output management is in SSR's HTML, trying to understand why its not visible, maybe HTML is invalid?

EugeneHlushko:
It is now fixed (writing this before preview lands so 🤞 )

I found that sort: 0 was causing a weird issue:
after sort with 0

When we had sort: 0 item sort was behaving strange...

After i changed it to -1:

after sort with -1

EugeneHlushko:
For details i suggest you see last commit

EugeneHlushko:
Yup its fixed: https://www.browserling.com/browse/win/7/firefox/68/https%3A%2F%2Fdeploy-preview-3288--webpackjsorg-netlify.netlify.com%2Fguides%2F

@montogeek or @ejez please confirm

ejez:
Working! thanks @EugeneHlushko

src/scripts/concatenate-docs.js
```diff
@@ -49,7 +49,16 @@ function getDirectoryRecursive(basePath) {
 			}
 
 		// write compound target file
-		const targetFile = fs.createWriteStream(targetFilePath);
+    const targetFile = fs.createWriteStream(targetFilePath);
+
+    targetFile.write(`---
```
montogeek: It this needed? If so, why?

src/utilities/content-tree-enhancers.js
```diff
@@ -76,6 +76,7 @@ function restructure(item, options) {
   enhance(item, options);
 
   if (item.children) {
+    // item.children.push(...item.children.splice(item.children.findIndex(child => child.name === 'printable.md'), 1));
```
montogeek: Please explain

src/scripts/concatenate-docs.js
EugeneHlushko: So this makes sure even the `printable.md` files have sort. Without title and contributors it fails to parse so i've added generic ones as they are harmless to printable files

src/utilities/content-tree-enhancers.js
EugeneHlushko: this is redundant, cleaning up 
```suggestion
```

montogeek:
> When we had sort: 0 item sort was behaving strange...

Why exactly? Something related with 0 array indexes?

Also from your screenshots, `21` is missing before but present in the second screenshot. Maybe another fix will be to guarantee the consecutive order of the sort property?

montogeek:
Also wonder if this is the same in other sections of the website.

EugeneHlushko:
Yeah so i've tested all sections as well which had 0 on sort, checking if its working is quite easy just hovering over links like we did to reproduce this one

EugeneHlushko:
Why sort failed, i am not sure, we have interesting sort function which is pretty straightforward, but for some reason it only worked (without my fix) if i did .sort(sort) twice in a row [here](https://github.com/webpack/webpack.js.org/blob/master/src/utilities/content-tree-enhancers.js#L82)... After fix 1 sort is enough.

montogeek:
Thanks!

EugeneHlushko:
🎉 finally

EugeneHlushko:
i think this was third tracked attempt to track down the issue, although the culprit is react's hydration that fails when sort order changes (there is an issue in react repo about this linked in previous pr)

PR #3290, opened 2019-09-16
guides: add trailing commas to last items in js objects/arrays

ejez:
guides: correct diffs to account for adding a comma before the last item when appending items to an object

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -48,8 +48,8 @@ __webpack.config.js__
     output: {
 -     filename: 'main.js',
 +     filename: 'bundle.js',
-      path: path.resolve(__dirname, 'dist')
-    }
+      path: path.resolve(__dirname, 'dist'),
+    },
   };
 ```
 
@@ -71,7 +71,7 @@ __webpack.config.js__
     entry: './src/index.js',
     output: {
       filename: 'bundle.js',
-      path: path.resolve(__dirname, 'dist')
+      path: path.resolve(__dirname, 'dist'),
     },
 +   module: {
 +     rules: [
@@ -80,10 +80,10 @@ __webpack.config.js__
 +         use: [
 +           'style-loader',
 +           'css-loader'
-+         ]
-+       }
-+     ]
-+   }
++         ],
++       },
++     ],
++   },
   };
 ```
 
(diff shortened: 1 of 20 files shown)


src/content/guides/asset-management.md
```diff
@@ -72,7 +72,8 @@ __webpack.config.js__
     output: {
       filename: 'bundle.js',
       path: path.resolve(__dirname, 'dist')
-    },
+-    }
++    },
```
EugeneHlushko:
```suggestion
+   },
```

ejez:
@EugeneHlushko , don't merge yet, the issue is repeated many times in the docs, let's create one PR for all

ejez:
Checked the guides directory, and using a regex I added a comma to the last item in an object when a new item is appended to the end of that object.
@EugeneHlushko

montogeek:
Me too, this problem is so common that there is a JS proposal for it :D

ejez:
Another related suggestion:

the diff signs (+/-) make it difficult for the reader to copy/paste, if the diff is going to be kept, it would be better to remove the diff signs, I think the colors (green/red) would be enough for the reader.

something like this should work:
```js
function removeDiffSigns() {
  const diffElements = document.querySelectorAll(
    "code.language-diff > span.inserted, code.language-diff > span.deleted"
  );

  diffElements.forEach(function(diffElement) {
    diffElement.innerHTML = diffElement.innerHTML.replace(/^[+-]/, '');
  });
}
```

ejez:
> This makes me want to have trailing comma rule in our configs

Agree, here is a quote from Mozilla: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

> Trailing commas (sometimes called "final commas") can be useful when adding new elements, parameters, or properties to JavaScript code. If you want to add a new property, you can simply add a new line without modifying the previously last line if that line already uses a trailing comma. This makes version-control diffs cleaner and editing code might be less troublesome.
> 
> JavaScript has allowed trailing commas in array literals since the beginning, and later added them to object literals (ECMAScript 5) and most recently (ECMAScript 2017) to function parameters.

EugeneHlushko:
@ejez do you want to re-work this PR into changing all "diff" code snippets in guides to have trailing comma and reduce the amount of diff size on them?

ejez:
Will do when I get home... 

You mean all code snippets, not only those having diffs.

ejez:
@EugeneHlushko 
done

PR #3296, opened 2019-09-22
[MDF] Creating a plugin steps.

sigfriedCub1990:
The code shows an example of a JS `class` while in the description only suggests a JS `named function`, I know this sounds pedantic but for the sake of completeness I think it's good to list both options.


sigfriedCub1990:
Thanks @montogeek

montogeek:
Thanks!

PR #3299, opened 2019-09-25
docs(concepts) fix e.g. spelling

EugeneHlushko:
We usually use `e.g.` across the docs and it seems pretty off here


montogeek:
Thanks!

PR #3301, opened 2019-09-27
docs(concepts) entry points concepts typing according to styleguide; …

EugeneHlushko:
…rework single entry list into continuation of paragraph

- types updated to be in line with styleguide
- there was single entry list, reworked to make sense
- link to mentioned config


src/content/concepts/entry-points.md
```diff
@@ -44,7 +44,7 @@ This is a great choice when you are looking to quickly setup a webpack configura
 
 ## Object Syntax
 
-Usage: `entry: {[entryChunkName: string]: string|Array<string>}`
+Usage: `entry: {[string entryChunkName]: string | [string]}`
```
EugeneHlushko: Actually this has to be corrected according to the writers guide
```suggestion
Usage: `entry: { entryChunkName string [string] }`
```

src/content/concepts/entry-points.md
```diff
@@ -44,7 +44,7 @@ This is a great choice when you are looking to quickly setup a webpack configura
 
 ## Object Syntax
 
-Usage: `entry: {[entryChunkName: string]: string|Array<string>}`
+Usage: `entry: { entryChunkName string [string] }`
```
EugeneHlushko:
```suggestion
Usage: `entry: { [entryChunkName] string [string] }`
```

PR #3303, opened 2019-09-30
docs(api) change optimise to optimize (prefer U.S. spelling)

kswedberg:
Only one change: optimise to optimize


EugeneHlushko:
Thanks!

PR #3305, opened 2019-09-30
docs(DevServer): document writeToDisk output dir

### LINKED ISSUE #3304: Document the output directory for devServer.writeToDisk
Regarding the `webpack-dev-server` documentation for `devServer.writeToDisk`, the docs would be improved (especially for newcomers) if we were told where to find the files that get written to disk.

I think I figured out that the files go to `./dist`. However I'm not sure if that is true for everyone, or if that's just for my own setup. The docs should tell the reader exactly what to expect. I'll make a PR if the answer is always `./dist`, or if someone tells me what the correct answer is.

FYI, the current documentation for `devServer.writeToDisk` looks like this:

https://webpack.js.org/configuration/dev-server/#devserverwritetodisk-

image

bseib:
Tell users of `DevServer.writeToDisk` where to find the output that gets written to disk.

fixes #3304

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -1533,7 +1533,7 @@ See [WatchOptions](/configuration/watch/) for more options.
 
 `boolean = false` `function (filePath)`
 
-Tells `devServer` to write generated assets to the disk.
+Tells `devServer` to write generated assets to the disk. The output is written to the [output.path](/configuration/output/#outputpath) directory, which is `./dist` by default.
 
 __webpack.config.js__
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/dev-server.md
```diff
@@ -1533,7 +1533,7 @@ See [WatchOptions](/configuration/watch/) for more options.
 
 `boolean = false` `function (filePath)`
 
-Tells `devServer` to write generated assets to the disk.
+Tells `devServer` to write generated assets to the disk. The output is written to the [output.path](/configuration/output/#outputpath) directory, which is `./dist` by default.
```
EugeneHlushko: Please avoid duplicating defaults, we already have a link.
```suggestion
Tells `devServer` to write generated assets to the disk. The output is written to the [output.path](/configuration/output/#outputpath) directory.
```

EugeneHlushko:
Thanks!

PR #3307, opened 2019-10-04
docs(concepts) clarify array syntax of the entry option

opl-:
This PR splits off the tip about passing an array to the entry point into its own section, in order to make it more visible and to provide more elaborate explanation with an example.

This PR repeats some of the changes from #3301

### DIFF
--- a/src/content/concepts/entry-points.md
+++ b/src/content/concepts/entry-points.md
@@ -8,6 +8,7 @@ contributors:
   - sokra
   - EugeneHlushko
   - Zearin
+  - opl-
 ---
 
 As mentioned in [Getting Started](/guides/getting-started/#using-a-configuration), there are multiple ways to define the `entry` property in your webpack configuration. We will show you the ways you __can__ configure the `entry` property, in addition to explaining why it may be useful to you.
@@ -37,8 +38,6 @@ module.exports = {
 };
 ```
 
-T> __What happens when you pass an array to `entry`?__ Passing an array of file paths to the `entry` property creates what is known as a __"multi-main entry"__. This is useful when you would like to inject multiple dependent files together and graph their dependencies into one "chunk".
-
 This is a great choice when you are looking to quickly setup a webpack configuration for an application or tool with one entry point (i.e. a library). However, there is not much flexibility in extending or scaling your configuration with this syntax.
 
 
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Hi @opl- 
can you please update your branch and apply the updated type syntax first?

opl-:
I just realized that I need to also update my type now, but I feel like the types in the other PR have a typo in them. What's what in this declaration: `entry: { [entryChunkName] string [string] }`? Shouldn't there at least be a `:` separator between the prop name and value? Shouldn't the two possible types be separated by a `|`?

I think the correct syntax for the object syntax section would be `entry: { [entryChunkName]: string | [string] }` and for the array syntax `entry: [string] | { [entryChunkName]: string | [string] }`. Is that correct?

EugeneHlushko:
@opl- the colon means possible values enumeration, the square bracket around the prop means its free-form key, we need to add this to https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types
Different types are separated by space e.g. `boolean` `number`, but when already in object we cannot really use ` sign anymore to capture the type so that if object's prop can be multi-type i separated it with space. Its not perfect, another way would be to create to separate records there e.g.

`{ [entryChunkName] string, [entryChunkName] [string]}` Which still allows to show the default value and enumeration e.g. if there was a default and a range of possible values it would still be doable e.g.

`{ [entryChunkName] string = 'index.js': 'another-allowed.js', 'none', [entryChunkName] [string]}`
Which would say for the type `string`, default value is `'index.js'` but it is also possible to set it to `'another-allowed.js'` and to `'none'`

I am tempted to file a PR now to update the contribution guide, give me an hour i will also ping @montogeek

opl-:
The `[key]` syntax matches the normal js syntax for computed prop keys, so it's fine, but the lack of a separator between the prop key and value/type is extremely confusing in my opinion.

The character used to separate allowed value types seems to be `|` according to the writer's guide right now, unless we are in an array, in which case its a comma. A comma (`,`) is obviously not a character that can be used to separate different possible types, so you're suggesting using a space instead...

We now end up in a situation where 3 different characters can be used for achieving the same thing. Wouldn't it make more sense to instead use a TypeScript-like syntax that many developers will already be used to instead of making up our own, confusing standard?

opl-:
There's also the problem of presenting arrays that contain different types at different indexes. With the current syntax we end up in a problematic situation. Assuming index 0 takes in one of constant strings and index 1 takes in an object, how do we write the type? `['const0', 'const1', {}]` is plain wrong, `[('const0', 'const1'), {...}]` is confusing.

TypeScript has already solved this problem and one would simply write it as `['const0' | 'const1', {...}]`. Obviously there are some cases where TS still doesn't have solutions, but in a vast majority of cases the syntax they use is already defined and easy enough to understand.

EugeneHlushko:
@opl- JavaScript is not typed language, also TypeScript is not the industry standard. While you may be familiar with TypeScript, regular documentation reader might just be starting to learn JavaScript. This is why we didn't go for it to begin with. As you can see adding types to the documentation is not an easy task when you try to cover all of the edge cases like enumerations, defaults, function returns especially when they are inside of an object already

We use `[TerserPlugin]` syntax to say that an array of Terser Plugin instances is allowed as a configuration property, so i think thats why `<key>` was used to clearly distinguish between them.

The `,` symbol was already held for object properties, thats why we had to use pipe, also colon is used in the object, so that to make it clear colon means same thing everywhere it was reserved for enumerations. e.g.

The current solution we have in here: https://webpack.js.org/configuration/entry-context/#entry
It probably doesn't work well for complex example cases when property has enum and is multi-type...

webpack doesn't have the index related constraints iirc or we'd bump into it earlier

EugeneHlushko:
Suggest we continue discussion in #3323 which will affect writer's guide

snitin315:
Closing as stale

PR #3309, opened 2019-10-07
docs(configuration): modify consumer definition in the context of ext…

jask84:
…ernals option

Modifies what appears to be a typo in the definition of consumer in the context of externals?

### DIFF
--- a/src/content/configuration/externals.md
+++ b/src/content/configuration/externals.md
@@ -12,7 +12,7 @@ contributors:
 
 The `externals` configuration option provides a way of excluding dependencies from the output bundles. Instead, the created bundle relies on that dependency to be present in the consumer's environment. This feature is typically most useful to __library developers__, however there are a variety of applications for it.
 
-T> __consumer__ here is any end-user application that includes the library that you have bundled using webpack.
+T> __consumer__ here is any end-user application that includes the library that you have excluded from the bundle using webpack.
 
 
 ## `externals`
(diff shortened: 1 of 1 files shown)


src/content/configuration/externals.md
```diff
@@ -12,7 +12,7 @@ contributors:
 
 The `externals` configuration option provides a way of excluding dependencies from the output bundles. Instead, the created bundle relies on that dependency to be present in the consumer's environment. This feature is typically most useful to __library developers__, however there are a variety of applications for it.
 
-T> __consumer__ here is any end-user application that includes the library that you have bundled using webpack.
+T> __consumer__ here is any end-user application that includes the library that you have excluded from the bundle using webpack.
```
EugeneHlushko: I think it was added to explain consumer keyword, the context about includes or not includes is extra, we can reduce this to have less confusion in the feature
```suggestion
T> __consumer__ here is any end-user application.
```

jask84: I'm happy with this too, as the current definition did confuse me at least. I would question the need for a `consumer` definition in this case at all then, but I guess it does fit in with the guidelines of not assuming knowledge etc.

EugeneHlushko: In that case commit suggestion or create your own change with the same content, after that tag me again please so i see notification

jask84:
@EugeneHlushko committed the suggestion

montogeek:
Thanks!

PR #3311, opened 2019-10-10
docs(concepts) loaders command fix and message re-worked

EugeneHlushko:
npm install command wasnt running without `&&` notation so it makes sense to merge it.
The loaders ecosystem empowerment phrase was not backed up nor it wasn't straightforward, reworded it

### DIFF
--- a/src/content/concepts/loaders.mdx
+++ b/src/content/concepts/loaders.mdx
@@ -25,8 +25,7 @@ Loaders are transformations that are applied on the source code of a module. The
 For example, you can use loaders to tell webpack to load a CSS file or to convert TypeScript to JavaScript. To do this, you would start by installing the loaders you need:
 
 ``` bash
-npm install --save-dev css-loader
-npm install --save-dev ts-loader
+npm install --save-dev css-loader ts-loader
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/concepts/loaders.mdx
```diff
@@ -25,8 +25,7 @@ Loaders are transformations that are applied on the source code of a module. The
 For example, you can use loaders to tell webpack to load a CSS file or to convert TypeScript to JavaScript. To do this, you would start by installing the loaders you need:
 
 ``` bash
-npm install --save-dev css-loader
-npm install --save-dev ts-loader
+npm install --save-dev css-loader ts-loader
```
montogeek: Do we have more cases like this one?

EugeneHlushko: Just quickly checked guides section, looks like we have a few there and maybe more. I will create a follow-up PR to change those occurrences

montogeek: Can put them here too

EugeneHlushko: In the recent years breaking work down and keeping it isolated and focused helped a lot to move quicker, did not you notice? :)

PR #3314, opened 2019-10-11
fix inconsistent sidebar links

ejez:
fix inconsistent sidebar links by disabling ssr for sidebar.
see https://github.com/webpack/webpack.js.org/pull/3288


ejez:
issue fixed by https://github.com/webpack/webpack.js.org/pull/3288
closing

PR #3327, opened 2019-10-22
docs(concepts) Modules page actualization

EugeneHlushko:
- Remove webpack 1 note
- html -> HTML
- remove unnecessary parenthesis


montogeek:
What about the project about having actual releases per documentation? We should do it this winter.

PR #3328, opened 2019-10-22
docs(multiple) remove webpack 1-2 notes as webpack 5 is closing in

EugeneHlushko:
Removing very old references.
Another question i have is should we remove LoaderOptionsPlugin and its references too or should we update its description.


montogeek:
I don't like to remove older versions information, I am pretty sure that out there is a lot of projects still working with older versions of webpack.

Correct solution is to implement versioning in the docs.

EugeneHlushko:
This information is too old though, meaning that using webpack 1-2 should be completely discouraged. I have caught up with @sokra and he agreed we'd clear up 1-2 notes and warnings when webpack 5 is live
We can hold this and removal of certain pages up until its live and migration guide for 4->5 is up

PR #3318, opened 2019-10-14
docs(caching) : Update broken links

melbarch:
Updated 3 broken links with hash routes that were not pointing on the right section of the page

- [X] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [X] Make sure your PR complies with the [writer's guide][2].
- [X] Review the diff carefully as sometimes this can reveal issues.

 
[1]: https://cla.js.foundation/webpack/webpack.js.org
[2]: https://webpack.js.org/contribute/writers-guide/

PR #3332, opened 2019-10-24
docs(links) update the links to have trailing slash in order for anch…

EugeneHlushko:
…ors to work consistently

follow up to #3318 

P.S. @Munter is it possible to add that check to linkcheck? e.g. all internal links that have anchor but dont start with `#` ( e.g. same page anchor links ) would have to have trailing slash i.e. a `/#` combination

### DIFF
--- a/src/content/api/cli.md
+++ b/src/content/api/cli.md
@@ -127,7 +127,7 @@ In every other case, webpack prints out a set of stats showing bundle, chunk and
 
 ### Environment Options
 
-When the webpack configuration [exports a function](/configuration/configuration-types#exporting-a-function), an "environment" may be passed to it.
+When the webpack configuration [exports a function](/configuration/configuration-types/#exporting-a-function), an "environment" may be passed to it.
 
 ```bash
 webpack --env.production    # sets env.production == true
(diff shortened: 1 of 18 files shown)


Munter:
That's an interesting case. @papandreou is this actually a bug in the way assetgraph resolves index html files from directories? Or is it one of those nasty things where it depends on the server configuration?

papandreou:
Hmm, so the issue is that `/configuration/configuration-types#exporting-a-function` gets redirected to `/configuration/configuration-types/`, losing the fragment identifier, and that hyperlink doesn't point that out?

I think assetgraph's modelling of this is correct. When a href of `/configuration/configuration-types` is encountered during population, and `stat` says that `configuration-types` is a directory, it will add a `FileRedirect` relation to the graph, which assumes that there's an `index.html` inside. That's a de facto standard for static web sites that also holds here, and the webpack.js.org server is also set up to correctly redirect, adding the slash to the url.

We should probably update hyperlink itself to complain when a link with a fragment but without a trailing slash points at an asset that gets redirected (has an outgoing `FileRedirect` or `HttpRedirect`). I can take a look at that.

Actually, I think I had convinced myself that browsers carried over the fragment when following a redirect, but maybe that changed, or I was imagining things :)

montogeek:
Why is this needed now? or was is always broken?

EugeneHlushko:
Not sure @montogeek 
I am sure it was working some time ago, but recently it stopped scrolling to an anchor at certain scenario (follow internal link vs full page load when paste url to the browser)

@papandreou thanks for filing a PR so quick :)

PR #3334, opened 2019-10-27
docs(configuration): fix links

sibiraj-s:
Fixes links to `output.outputlibrarytarget` and `cli.environment-options`


EugeneHlushko:
I think this was fixed globally already

PR #3337, opened 2019-10-29
Upgrade to hyperlink 4.4.0

Munter:
Hyperlink 4.4.0 has a new feature that will detect broken fragment links of the form `/page#fragment` where `/page` is actually a redirect to `/page/` because it's a directory with an index file. In this case the server redirect will strip the fragment, thus breaking the intended link.

See errors in the current `master` branch here: https://travis-ci.org/webpack/webpack.js.org/jobs/604644796#L2041-L2159

The following commit fixes all those errors.

Relates to https://github.com/webpack/webpack.js.org/pull/3332

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "html-webpack-template": "^6.1.0",
     "http-server": "^0.10.0",
     "husky": "^1.0.0-rc.8",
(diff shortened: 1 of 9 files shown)


montogeek:
Thank you very much!

PR #3345, opened 2019-11-07
Add `: false` to `bail` documentation

lencioni:
This will make it clearer that this is `false` by default.

https://github.com/webpack/webpack/blob/408a4ae77eb3f7441d6dd2dc011f23e491e8fbfb/lib/Compilation.js#L501

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -48,7 +48,7 @@ As it happens, the AMD support in webpack ignores the defined name anyways.
 
 ## `bail`
 
-`boolean: false`
+`boolean = false`
 
 Fail out on the first error instead of tolerating it. By default webpack will log these errors in red in the terminal, as well as the browser console when using HMR, but continue bundling. To enable it:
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/other-options.md
```diff
@@ -48,7 +48,7 @@ As it happens, the AMD support in webpack ignores the defined name anyways.
 
 ## `bail`
 
-`boolean`
+`boolean: false`
```
EugeneHlushko: Please change according to https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types
```suggestion
`boolean = false`
```

lencioni: Ok! I've updated with this change.

Separately, it might be a good idea to change the other default values in this document to match.

EugeneHlushko: @lencioni can you mention which ones? by looking at it on live i dont see any issues

EugeneHlushko:
Thanks!

src/content/configuration/other-options.md
lencioni: https://github.com/webpack/webpack.js.org/pull/3355

PR #3349, opened 2019-11-11
Correct `stats.entrypoints` default value

lewislbr:
While configuring the stats to display, without using the entrypoints option, I found out the entrypoints were appearing to the log, even when in the documentation was shown with a default value of `false`. I checked in the [file where defaults are defined](https://github.com/webpack/webpack/blob/beec753201763fa3724bd965a93106126d13b271/lib/stats/DefaultStatsPresetPlugin.js), and in line 112 it implicitly appears the default value is `true`, so I've edited the documentation to reflect it.


EugeneHlushko:
Thanks!

PR #3353, opened 2019-11-13
chore: updated the docs to mention correct loader

anikethsaha:
`html-loader` doesnt uses `addDependency` method, _Might be outdated with the changes_
I changed the example with `sass-loader` and `less-loader` which does use it

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -288,7 +288,7 @@ module.exports = {
 
 ### `stats.entrypoints`
 
-`boolean = false`
+`boolean = true`
 
 Tells `stats` whether to display the entry points with the corresponding bundles.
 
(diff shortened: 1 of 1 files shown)


anikethsaha:
cc @EugeneHlushko

montogeek:
Thanks for your contribution, What happened with the yarn.lock? Could you please revert those changes?

anikethsaha:
that is using yarn install at the beginning.
nothing explicit.

EugeneHlushko:
Yes please revert lockfile, can you also mention OS and node version to understand why so many changes in the lockfile after your install

anikethsaha:
node : v10.15.3
OS : windows 10

EugeneHlushko:
Yeap do looks like the changes are related to mac vs win. As soon as its reverted should be good to go

PR #3354, opened 2019-11-13
chore: updated the start script to fetch at first

anikethsaha:
currently as per the contribution docs, running `start` script will not work when running the website in the browser as it required some of the fetched files like for the `_supporter.json`.
So `fetch` command need to be called beforehand in order to make it consistent.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -19,6 +19,9 @@ Once you are in the project directory, run the following commands:
 - `npm install` to pull all dependencies. Make sure to check notes below if you prefer Yarn.
 - `npm run build` to create a production version of the site.
 - `npm start` to develop on a local webpack-dev-server: [localhost:3000][3].
+
+> NOTE: run `npm fetch` before running `npm start` command for first time
+
 - `npm run fetch` to retrieve external documentation/data.
 
 The final command, `npm run fetch`, is optional as both `npm run build` and `npm start`


EugeneHlushko:
Fetched files are not required every time you run start, better add a note to the contribution document

anikethsaha:
> , better add a note to the contribution document

Yea I thought of the same, will do it
although, `start` is a one-time command only right!!, we can run this script and start editing.
I don't think it will run every time whenever there is a change in the code.

EugeneHlushko:
You are right, although i meant when ever you do a commit and switch to different work you dont need to re-fetch but sometimes want to stop sev server and in a but run it again. Fir single time effort contributors this would be handy, but nit for the longer term workflow

anikethsaha:
yeah 👍

Should I change the docs then ?

EugeneHlushko:
@anikethsaha yup, please go ahead and mention that prior to the first start command, developer should run fetch

.github/CONTRIBUTING.md
```diff
@@ -1,11 +1,9 @@
-Contributing
-------------
+## Contributing
```
EugeneHlushko: Screen Shot 2019-11-19 at 8 36 57 AM
It looks all right, please revert

.github/CONTRIBUTING.md
```diff
@@ -19,22 +17,24 @@ Once you are in the project directory, run the following commands:
 - `npm install` to pull all dependencies. Make sure to check notes below if you prefer Yarn.
 - `npm run build` to create a production version of the site.
 - `npm start` to develop on a local webpack-dev-server: [localhost:3000][3].
+
+> NOTE: run `npm fetch` before running `npm start` command for first time
```
EugeneHlushko: This change is good

.github/CONTRIBUTING.md
```diff
@@ -19,22 +17,24 @@ Once you are in the project directory, run the following commands:
...
+
 - `npm run fetch` to retrieve external documentation/data.
 
 The final command, `npm run fetch`, is optional as both `npm run build` and `npm start`
 will do this automatically. This step pulls in documentation for loaders/plugins hosted
 in separate repositories such as the ones found in the [webpack-contrib][4] organization.
 See the `package.json` for the full list of `scripts`.
 
-> Note that __Node 6.9.x__ or above is required for the build process to run properly.
+> Note that **Node 6.9.x** or above is required for the build process to run properly.
```
EugeneHlushko: Across the docs we use `__` vs `**` so please revert.

package.json
```diff
@@ -28,7 +28,7 @@
     "clean-printable": "rimraf src/content/**/printable.md",
     "preclean": "run-s clean-dist clean-printable",
     "clean": "rimraf src/content/**/_*.md src/**/_*.json",
-    "start": "npm run clean-dist && npm run fetch && cross-env NODE_ENV=development webpack-dev-server --config webpack.dev.js --env.dev",
+    "start": "npm run clean-dist && cross-env NODE_ENV=development webpack-dev-server --config webpack.dev.js --env.dev",
```
EugeneHlushko: Why is this still highlighted as a change vs `master`? can you checkout this file from master we don't want changes here with this PR right?

.github/CONTRIBUTING.md
anikethsaha: @EugeneHlushko this is the work of prettier + markdown lint😅

.github/CONTRIBUTING.md
anikethsaha: Same prettier thing. 
I did the fix command with markdown lint and but still those lint staged checks where failing so I had to use no verify git flag. 
May be because of that it took the prettier changes. 
I will revert it soon

package.json
anikethsaha: Yeah, I will do it. 
It is from the previous commit of this PR .

EugeneHlushko:
Thanks!

PR #3355, opened 2019-11-13
Change `: false` to `= false` in other options

lencioni:
This brings it in line with the writers guide. https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types


lencioni:
Oh, ok! Thanks for the clarification!

PR #3360, opened 2019-11-16
Fix: corrected the mentioned loader api adddependency

anikethsaha:
`html-loader` doesnt uses `addDependency` method, _Might be outdated with the changes_
I changed the example with `sass-loader` and `less-loader` which does use it

Follow up of https://github.com/webpack/webpack.js.org/pull/3353

_I messed up the reverting and somehow I couldnt reopen that PR even though this branch is ahead of master so had to make a new one_


montogeek:
Thanks!

PR #3364, opened 2019-11-19
chore: remove unused file

jamesgeorge007:
Removed the respective file that isn't used anymore.


montogeek:
Thanks for your contribution, unfortunately this file shouldn't even exist. But you can delete it :)

jamesgeorge007:
@montogeek if that's the case, should I close this one or delete the respective file and have it appear as the `diff` instead.

jamesgeorge007:
Also, there're instances of `var` still being used within `src/utilities`. Shall I shoot yet another PR updating them to be `let`?

jamesgeorge007:
\cc @montogeek @EugeneHlushko

montogeek:
Thanks!

PR #3365, opened 2019-11-19
docs(versions) adds a snippet of the current doc version and link to …

EugeneHlushko:
…other version(s)

v4.webpack.js.org:

Screen Shot 2019-11-19 at 11 33 48 AM

And for current version it will say 5

### DIFF
--- a/src/components/Sidebar/Sidebar.jsx
+++ b/src/components/Sidebar/Sidebar.jsx
@@ -20,7 +20,7 @@ const docs = [
   }
 ];
 
-const currentDocsVersion = 4;
+const currentDocsVersion = 5;
 
 // Create and export the component
 export default ({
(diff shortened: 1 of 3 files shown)


src/components/Sidebar/Sidebar.jsx
```diff
@@ -45,6 +58,14 @@ export default ({
             </div>
           );
         })}
+        <div className="sidebar__docs-version">
+          You are reading webpack {currentDocsVersion} documentation. Change here to:
+          <ul>
+            {docs.filter(item => item.version !== currentDocsVersion).map(item => <li key={`webpack-${item.version}-docs`}>
+              <a rel="nofollow" href={item.url}>webpack {item.version} documentation</a>
```
montogeek: Why `nofollow`?

EugeneHlushko: i am afraid of SEO sanctions for duplicated content, ideally we add robots.txt with disallow in v4 for all the pages except for homepage what you think?

montogeek: Let's see how it goes from Google and user search perspective

PR #3371, opened 2019-11-26
docs(mode): minor fix on mode options enum

nemesis:
Fix typo in mode enum, since production repeats twice.


nemesis:
[Whoops](https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types)

PR #3378, opened 2019-12-01
fix - 2728 : Adds 4 new hooks to Compilation which allow to customize/extend Stats

jeffin143:
Handles : #2728

It does not solve full issue :), But a small improvement


wizardofhogwarts (dismissed):

src/content/api/compilation-hooks.md
```diff
@@ -707,3 +707,52 @@ This hooks allows changing the references reported by dependencies.
...
+compilation.hooks.statsPreset.for('my-preset').tap('MyPlugin', options => {
+  if(options.all === undefined) options.all = true;
+});
+```
+
+### `statsNormalize`
+
+`SyncHook`
+
+It's called for a options object and should normalize to a format usable by the following hooks. It should also normalize missing options to the default value.
```
jamesgeorge007:
```suggestion
It's called for an options object and should normalize to a format usable by the following hooks. It should also normalize missing options to the default value.
```

src/content/api/compilation-hooks.md
```diff
@@ -707,3 +707,52 @@ This hooks allows changing the references reported by dependencies.
 - Callback Parameters: `depRef` `dependency` `module`
 
 W> The `module` parameter will be removed in v5.0.0
+
+### `statsPreset`
+
+`HookMap`
+
+It's a HookMap called for a preset (if used) with an options object. A plugin which handles a preset should set options in the options object. It should avoid overriding existing options.
```
montogeek: Please don't use the concept in its definition: https://en.wikipedia.org/wiki/Circular_definition

PR #3379, opened 2019-12-01
docs(plugin) : Mention explicitly that ProfilingPlugin accepts only absolute paths

jeffin143:
Handles : #3235 

Only handles a bit


jeffin143:
ping @montogeek

montogeek:
Thanks!

PR #3381, opened 2019-12-01
docs(migration) : resolve.concord no longer exists

jeffin143:
_describe your changes..._
Handles : #2676 


jeffin143:
@EugeneHlushko @montogeek 
There seem to be lot of issues which states that ` Following option should be added in migration docs that it has been depreciated`

So should I add a new Section as `##### Depreciated/Removed options` Similar to 

#### Update outdated options

instead of the present approach where I add a separate heading for one option

jeffin143:
@EugeneHlushko @montogeek 
There seem to be lot of issues which states that ` Following option should be added in migration docs that it has been depreciated`

So should I add a new Section as `##### Depreciated/Removed options` Similar to 

#### Update outdated options

instead of the present approach where I add a separate heading for one option

jeffin143:
Ping @montogeek

PR #3383, opened 2019-12-01
docs(format) recent contributions violated writers guide fixed

EugeneHlushko:
Just found a few cases where we let misformatting in, fixed:

https://webpack.js.org/contribute/writers-guide/#typesetting


montogeek:
Thanks!

PR #3387, opened 2019-12-02
Add description of the option `stats.preset`

myshov:
There is no description  of `stats.preset`. Because of this there is no entry in the left menu.


montogeek:
Thanks!

PR #3389, opened 2019-12-02
doc(plugins) : rename optimization.splitChunks.cacheGroups.vendors

jeffin143:
Handling (only first point) : #2324


jeffin143:
@EugeneHlushko 

How should I handle all the rename options ?, Let me know I will make apt changes

Another example 
Rename `ChunkGroup.getModuleIndex` to `ChunkGroup.getModulePreOrderIndex`

src/content/migrate/5.md
```diff
@@ -67,6 +67,7 @@ Update the following options to their new version (if used):
 - `NamedChunksPlugin` => `optimization.chunkIds: 'named'`
 - `HashedModulesPlugin` => `optimization.moduleIds: 'hashed'`
 - `optimization.occurrenceOrder: true` => `optimization: { chunkIds: 'total-size', moduleIds: 'size' }`
+- Rename `optimization.splitChunks.cacheGroups.vendors` => `optimization.splitChunks.cacheGroups.defaultVendors`
```
montogeek:
```suggestion
- Renamed `optimization.splitChunks.cacheGroups.vendors` => `optimization.splitChunks.cacheGroups.defaultVendors`
```

jeffin143: Thanks for the really quick review :), Made apt changes

src/content/migrate/5.md
```diff
@@ -67,6 +67,7 @@ Update the following options to their new version (if used):
 - `NamedChunksPlugin` => `optimization.chunkIds: 'named'`
 - `HashedModulesPlugin` => `optimization.moduleIds: 'hashed'`
 - `optimization.occurrenceOrder: true` => `optimization: { chunkIds: 'total-size', moduleIds: 'size' }`
+- Renamed `optimization.splitChunks.cacheGroups.vendors` => `optimization.splitChunks.cacheGroups.defaultVendors`
```
rishabh3112:
```suggestion
- `optimization.splitChunks.cacheGroups.vendors` => `optimization.splitChunks.cacheGroups.defaultVendors`
```
This looks more consistent to me.

jeffin143: suggestion @montogeek 
I don't think thats a good option because for others it is update of values but here it is renamed

rishabh3112: @jeffin143  I think all corresponds to changes ( either rename or update ) that needs to be done by an user thus all having same notation makes more sense to me.

EugeneHlushko: Yeah arrow says it in the others dont see why we should add word `Renamed` here exclusively, just remove it and its good

wizardofhogwarts: @jeffin143 I also agree with the suggestion by @rishabh3112.

anikethsaha: Either looks fine
but I would lean more towards having `rename`, its safe to have as its actually just renamed method and it might be helpful for other dev while migrating by assuring that -its the same old method but changed the name is changed, so it's not a breaking change-

jeffin143: > This looks more consistent to me.

Thanks, made changes :)

EugeneHlushko:
@jeffin143 please resolve conflict, the rest LGTM

jeffin143:
> @jeffin143 please resolve conflict, the rest LGTM

Done :)

montogeek:
Thanks!

PR #3390, opened 2019-12-02
docs(plugin):Add entryOnly option in DllPlugin, also type

jeffin143:
Handles : [2192](https://github.com/webpack/webpack.js.org/issues/2192)


montogeek:
Thanks!

PR #3396, opened 2019-12-08
docs(api) clarify module ID type and value origin, link related option

### LINKED ISSUE #3392: Is module.id still a number
# Bug report
I read here that a module ID should be a number:
https://webpack.js.org/api/module-methods/#requireresolve

> Module ID is a number in webpack (in contrast to NodeJS where it is a string -- the filename).

When I use module id in output.filename like so: `[id].js`
I get `<filename>.js` (same as if I was using `[name].js`)

I would expect `<number>.js`

webpack version:  4.6.0
Node.js version: 10.13.0
Operating System: MacOS Mojave 10.14
Additional tools:

EugeneHlushko:
Closes #3392

### DIFF
--- a/src/content/api/module-methods.md
+++ b/src/content/api/module-methods.md
@@ -177,9 +177,11 @@ W> Using it asynchronously may not have the expected effect.
 require.resolve(dependency: String);
 ```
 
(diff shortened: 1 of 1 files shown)


montogeek:
Thanks!

PR #3397, opened 2019-12-08
docs(api) cli shortcut updated

### LINKED ISSUE #3330: p option not doing what is said in the docs
If we read current docs, and `webpack --help` (version 4.41.2), we see this : 

"-p   shortcut for --optimize-minimize --define process.env.NODE_ENV="production"  "

This is false.

Running `webpack --optimize-minimize --define process.env.NODE_ENV="production"` have no effect.

In fact, `webpack -p` is equivalent to running `webpack --mode production`.  Now my code is minimized.

Seems that in recent version, webpack does not care about the NODE_ENV value. The "mode" takes over.

EugeneHlushko:
Closes #3330 

Added trailing slashes to links along the main issue


src/content/api/cli.md
```diff
@@ -314,7 +314,7 @@ Parameter         | Explanation                              | Usage
 Shortcut | Replaces
 ---------|----------------------------
 -d       | `--debug --devtool cheap-module-eval-source-map --output-pathinfo`
--p       | `--optimize-minimize --define process.env.NODE_ENV="production"`, see [building for production](/guides/production)
+-p       | `--mode production`, see [building for production](/guides/production/)
```
rishabh3112: If this is helpful, In `webpack-cli v4` beta, we have `--prod` flag for this.

EugeneHlushko: Does that mean `-p` is the same as `--prod` in webpack-cli 4+ ?

rishabh3112: Yes, They mean same.
For reference, @EugeneHlushko here is the whole list of commands and flags and their alias from `webpack-cli v4 beta`.
image

EugeneHlushko: Allright, we will have to update this whole list when its out of beta

PR #3401, opened 2019-12-11
docs(plugins) remove loaderoptions plugin which was used for migratin…

EugeneHlushko:
…g wp1 to wp2

As agreed before we are removing this plugin from the docs, it is still accessible on v3 tag and v4.webpack.js.org


montogeek:
Thanks!

PR #3405, opened 2019-12-13
docs(config) output formatting fix

EugeneHlushko:
singlequotes in js snippets


montogeek:
Maybe we should lint this

montogeek:
Thanks!

PR #3406, opened 2019-12-13
docs(config) stats.preset option extended according to codebase

EugeneHlushko:
Extend the option documentation that was added in #3387

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -607,7 +607,7 @@ module.exports = {
 };
 ```
 
-Set `stats.preset` to `false` is the same as using a [preset](/configuration/stats/#stats) `'none'`.
+Set `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).
 
 ### `stats.providedExports`
 
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -607,7 +607,7 @@ module.exports = {
 };
 ```
 
-Set `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).
+Setting value of `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).
 
 ### `stats.providedExports`
 
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -594,7 +594,7 @@ module.exports = {
 
 ### `stats.preset`
 
-`string`
+`string` `boolean: false`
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/stats.md
```diff
@@ -607,6 +607,8 @@ module.exports = {
 };
 ```
 
+Set `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).
```
montogeek: I am having a hard time parsing this oration, can't understand it at all, sorry

EugeneHlushko: Lets try to make it shine, when stats.preset = false, a preset `'none'` is used

EugeneHlushko: Any idea on how to improve?

rishabh3112:
```suggestion
Set `stats.preset` to `false` to use `'none'` [stats preset](/configuration/stats/#stats).
```

EugeneHlushko: removing `tells webpack` doesnt really add any clarity to the oration, any thoughts @montogeek

rishabh3112: @EugeneHlushko it certainly does to me. Starting word `Set` doesn't make sense in the proposed change. In suggested change, it makes sense to me. I am leaving another change if `tells webpack` needs to be there.
```suggestion
Setting value of `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).
```

EugeneHlushko: I was also thinking about that:
```suggestion
Setting value of `stats.preset` to `false` tells webpack to use [stats preset `'none'`](/configuration/stats/#stats).
```

rishabh3112: Looks good to me 👍,  Just one more suggestion:
The change in order of `'none'` in your suggestion seems counter intuitive to me. For example, we say `Red Car` not `Car red` to describe a car which is `red` in color. Similarly we should call  `'none' stats preset` to describe `stats preset` which is `'none'` in value.

EugeneHlushko: Yeah i was thinking it terms of how we set key value pairs, e.g. `stats.preset = 'none'`

EugeneHlushko: i'm pretty much happy with `Setting value of `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).` lets see if Fernando can contribute to this discussion, elsewise i will add that as is

montogeek: @EugeneHlushko Like you suggest is fine for me

EugeneHlushko:
```suggestion
Setting value of `stats.preset` to `false` tells webpack to use `'none'` [stats preset](/configuration/stats/#stats).
```

PR #3362, opened 2019-11-18
docs(configuration) : New stats option orphanModules, defaults to false

jeffin143:
_describe your changes..._
Add New stats option orphanModules, defaults to false , Handling : https://github.com/webpack/webpack.js.org/issues/2437

- [x] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [ ] Make sure your PR complies with the [writer's guide][2].
- [x] Review the diff carefully as sometimes this can reveal issues.
- __Remove these instructions from your PR as they are for your eyes only.__

[1]: https://cla.js.foundation/webpack/webpack.js.org
[2]: https://webpack.js.org/contribute/writers-guide/

PR #3407, opened 2019-12-13
docs(config) stats.orphanModules doc improved slightly

EugeneHlushko:
A small follow up to #3362 https://github.com/webpack/webpack.js.org/pull/3362/files#r357783625

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -320,7 +320,7 @@ module.exports = {
 
 `boolean = false`
 
-Tells `stats` whether to hide `orphan` modules. A module is an `orphan` if it is not included in any chunk. Orphans modules are now hidden by default in `Stats`.
+Tells `stats` whether to hide `orphan` modules. A module is an `orphan` if it is not included in any chunk. Orphan modules are hidden by default in `stats`.
 
 ```javascript
 module.exports = {


src/content/configuration/stats.md
```diff
@@ -320,7 +320,7 @@ module.exports = {
 
 `boolean = false`
 
-Tells `stats` whether to hide `orphan` modules. A module is an `orphan` if it is not included in any chunk. Orphans modules are now hidden by default in `Stats`.
+Tells `stats` whether to hide `orphan` modules. A module is an `orphan` if it is not included in any chunk. Orphan modules are hidden by default in `stats`.
```
montogeek:
```suggestion
Tells `stats` whether to hide `orphan` modules. A module is `orphan` if it is not included in any chunk. Orphan modules are hidden by default in `stats`.
```

jeffin143: Should we mention , from webpack 5

Now could be replaced by webpack 5 ig

EugeneHlushko: I think `an` is good, just ran it through my native speaker mate.

@jeffin143 It says webpack 5 docs in the left hand side, so we should stop doing these mentions probably

jeffin143:
Just for reference #2437

PR #3413, opened 2019-12-19
docs(webpack5): add new cache options

rishabh3112:
_describe your changes..._

Part of #2774
Adds documentations for:
1. `cache.idleTimeout` option
2. `cache.idleTimeoutForInitialStore` option

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -10,6 +10,7 @@ contributors:
   - vansosnin
   - EugeneHlushko
   - skovy
+  - rishabh3112
 related:
   - title: Using Records
     url: https://survivejs.com/webpack/optimizing/separating-manifest/#using-records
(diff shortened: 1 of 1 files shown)


src/content/configuration/other-options.md
```diff
@@ -210,6 +211,43 @@ module.exports = {
...
+```javascript
+module.exports = {
+  //..
+  cache: {
+    idleTimeout: 10000
+  }
+};
+```
+
+W> `cache.idleTimeout` is only available for when [`cache.store`](#cachestore) is set to either `'pack'` or `'idle'`
```
EugeneHlushko: Lets use simpler form
```suggestion
W> `cache.idleTimeout` is only available when [`cache.store`](#cachestore) is set to either `'pack'` or `'idle'`
```

src/content/configuration/other-options.md
```diff
@@ -210,6 +211,43 @@ module.exports = {
 
 W> Don't share the cache between calls with different options.
 
+### `cache.idleTimeout`
+
+`number = 10000`
+
+Time in `ms`. `cache.idleTimeout` denotes the time period after which the cache storing should happen. Defaults to `10000` which is equivalent to `10s`
```
EugeneHlushko: No need to elaborate when it already says milliseconds
```suggestion
Time in milliseconds. `cache.idleTimeout` denotes the time period after which the cache storing should happen.
```

src/content/configuration/other-options.md
```diff
@@ -210,6 +211,43 @@ module.exports = {
...
+};
+```
+
+W> `cache.idleTimeout` is only available for when [`cache.store`](#cachestore) is set to either `'pack'` or `'idle'`
+
+### `cache.idleTimeoutForInitialStore`
+
+`number = 0`
+
+Time in `ms`. `cache.idleTimeoutForInitialStore` is the time period after which the initial cache storing should happen. It defaults to `0` which is equivalent to `0s`.
```
EugeneHlushko:
```suggestion
Time in milliseconds. `cache.idleTimeoutForInitialStore` is the time period after which the initial cache storing should happen.
```

src/content/configuration/other-options.md
```diff
@@ -210,6 +211,43 @@ module.exports = {
...
+```javascript
+module.exports = {
+  //..
+  cache: {
+    idleTimoutForInitialStore: 0
+  }
+};
+```
+
+W> `cache.idleTimeoutForInitialStore` is only available for when [`cache.store`](#cachestore) is set to either `'pack'` or `'idle'`
```
EugeneHlushko:
```suggestion
W> `cache.idleTimeoutForInitialStore` is only available when [`cache.store`](#cachestore) is set to either `'pack'` or `'idle'`
```

rishabh3112:
Thanks @EugeneHlushko, Have applied your suggestions and it looks better now to me :)

EugeneHlushko:
Thanks! Looks good

PR #3414, opened 2019-12-20
docs(module): add more information about nested rules

opl-:
This change explains better the relationship between `Rule.rules` and `Rule.oneOf`.

It also makes the two more connected, by pointing to the Nested rules section, which is motivated by both [webpack-chain](https://github.com/neutrinojs/webpack-chain/pull/220) and [vue-loader](https://github.com/vuejs/vue-loader/blob/cf00a45cde52f3961b36b37265761b0d92affb0c/lib/plugin-webpack4.js#L149) developers having missed the existence of `Rule.rules`

EDIT: ~~Just realized I forgot to pull master before adding my changes. Fixing it right now.~~ Done. :+1:

### DIFF
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -16,6 +16,7 @@ contributors:
   - lukasgeiter
   - skovy
   - smelukov
+  - opl-
 ---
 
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Thanks!

PR #3423, opened 2019-12-27
docs: improve compilation-hooks

wizardofhogwarts:


src/content/api/compilation-hooks.md
```diff
@@ -161,7 +161,7 @@ W> This hook will be removed in v5.0.0
 
 `SyncBailHook`
 
-Called at the beginning of the modules optimization phase. A plugin can tap into this hook to perform modules optimizations.
+Called at the beginning of the modules optimization phase. A plugin can tap into this hook to perform module optimizations.
```
EugeneHlushko: I think its actually `modules` because in the callback you receive many of them

src/content/api/compilation-hooks.md
```diff
@@ -197,7 +197,7 @@ W> This hook will be removed in v5.0.0
 
 `SyncBailHook`
 
-Called at the beginning of the chunks optimizion phase. A plugin can tap into this hook to perform chunks optimizations.
+Called at the beginning of the chunks optimization phase. A plugin can tap into this hook to perform chunk optimizations.
```
EugeneHlushko: Same as above, multiple chunks coming in (0 to infinity)

src/content/api/compilation-hooks.md
```diff
@@ -269,7 +269,7 @@ W> This hook will be removed in v5.0.0
 
 `SyncHook`
 
-Called after the chunkmodules optimization has successfully completed.
+Called after the chunkmodules optimization has completed with success.
```
EugeneHlushko:
```suggestion
Called after the chunkmodules optimization has completed successfully.
```

src/content/api/compilation-hooks.md
rishabh3112: Optimization is specific to an individual chunk (so, current version in docs is also incorrect), thus I left a suggestion I feel that should be here.
```suggestion
Called at the beginning of the chunk optimization phase. A plugin can tap into this hook to perform optimizations on chunks.
```

src/content/api/compilation-hooks.md
rishabh3112: Similar as below.
```suggestion
Called at the beginning of the module optimization phase. A plugin can tap into this hook to perform optimizations on modules.
```

EugeneHlushko:
Thanks!

PR #3434, opened 2019-12-29
docs(configuration): fixed outdated reference to production mode resource

jamesgeorge007:
The respective relative links where kinda outdated eventhough they got resolved to the updated resource. Made them link up to the intended resource for the sake of consistency.


montogeek:
Thanks!

PR #3439, opened 2020-01-01
documentation for separate app and vendor entries

niravasher:
Screenshot from 2020-01-01 16-20-39

Closes https://github.com/webpack/webpack.js.org/issues/3244.

### DIFF
--- a/src/content/concepts/entry-points.md
+++ b/src/content/concepts/entry-points.md
@@ -68,6 +68,41 @@ Below is a list of entry configurations and their real-world use cases:
 
 ### Separate App and Vendor Entries
 
+__webpack.config.js__
+
+```javascript
+module.exports = {
+  entry: {
+    main: './src/app.js',
+    vendor: './src/vendor.js'
+  }
+};
+```
+
+__webpack.prod.js__
+
+```javascript
+module.exports = {
+  output: {
+    filename: [name].[contentHash].bundle.js
+  }
+};
+```
+
+__webpack.dev.js__
+
+```javascript
+module.exports = {
+  entry: {
+    filename: [name].bundle.js
+  }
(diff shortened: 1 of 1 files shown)


niravasher:
@jamesgeorge007 @montogeek @EugeneHlushko what else is required to merge the PR? Please suggest required changes

src/content/concepts/entry-points.md
```diff
@@ -68,6 +68,41 @@ Below is a list of entry configurations and their real-world use cases:
...
+module.exports = {
+  entry: {
+    filename: [name].bundle.js
+  }
+};
+```
+
+__What does this do?__ We are telling webpack that we would like 2 separate entry points (like the above example).
+
+__Why?__ The utility of this is that you can import required libraries or files that aren't modified (ex bootstrap, jquery, images, etc) inside vendor.js and they will be bundled together into their own HTML. Content Hash remains the same, which allows the browser to cache separately reducing load time.
```
jamesgeorge007:
```suggestion
__Why?__ With this you can import required libraries or files that aren't modified (ex bootstrap, jquery, images, etc) inside `vendor.js` and they will be bundled together into their own HTML. Content Hash remains the same, which allows the browser to cache separately thereby reducing load time.
```

src/content/concepts/entry-points.md
```diff
@@ -68,6 +68,41 @@ Below is a list of entry configurations and their real-world use cases:
...
+  }
+};
+```
+
+__webpack.prod.js__
+
+```javascript
+module.exports = {
+  output: {
+    filename: [name].[contentHash].bundle.js
```
montogeek: Please use valid JavaScript

PR #3440, opened 2020-01-01
Docs(migration): Add info about webpack.util.serialization

jeffin143:
fix : #2741


montogeek:
It is still exposed in webpack 5?

jeffin143:
> It is still exposed in webpack 5?

yes I suppose, because the change still persist.

Can you confirm @sokra ?

sokra:
Yes this is exposed, but I don't think it make sense to mention in the migration guide. That's something for the changelog...

jeffin143:
Since this is not needed anymore, let's close this

PR #3441, opened 2020-01-01
Docs(api) : remove Compilation.waitForBuildingFinished

jeffin143:
fix - [#2510](https://github.com/webpack/webpack.js.org/issues/2510)


jeffin143:
> Hi @jeffin143
> Actually there is one point missing to fix the #2510
> `add Module.invalidateBuild` method, from the linked issue should be added

I am not sure about its functionalities and hence I refrained myself from writing it, May be we could just finish the issue patch by patch

snitin315:
Fixed in master.

PR #3340, opened 2019-10-31
Update parser.md

prateekbh:
Fixed some typo

- [x] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [x] Make sure your PR complies with the [writer's guide][2].
- [x] Review the diff carefully as sometimes this can reveal issues.

PR #3460, opened 2020-01-04
docs(parser): fix expression and expressionAnyMember examples

henryqdineen:
This is something I noticed when reading the [JavascriptParser Hooks](https://webpack.js.org/api/parser) documentation. It looks like `parser.hooks.new` was mistakenly used for `expression` and `expressionAnyMember`


henryqdineen:
Realized that #3340 already exists. Sorry.

PR #3463, opened 2020-01-04
docs(api) minor formatting fixes, use singlequotes for js strings

EugeneHlushko:
- single quotes for js (including strings and their example/possible values)


montogeek:
Thanks!

PR #3467, opened 2020-01-07
docs(writing-a-loader): changed the depreciated memory-fs in example

anikethsaha:
memory-fs is depreciated and its recommended to migrate to memfs


anikethsaha:
> Update all the instances were `memory-fs` is referenced.

Most of them are in the loader and plugins's docs...I guess

anikethsaha:
cc @montogeek @EugeneHlushko

EugeneHlushko:
Thanks!

PR #3470, opened 2020-01-07
docs(migrate): replace => with ↦

simon04:


montogeek:
Thanks @simon04, Could you please elaborate why?
It also looks like an arrow function for me :)

simon04:
Sorry, sure: Use the power of Unicode for a nice typesetting of the arrows. I mean, `=>` is no arrow, but equals sign followed by greater-than sign.

montogeek:
Thanks!

PR #3474, opened 2020-01-08
docs(configuration): config Rule.resolve with an object in module

Mistyyyy:
[https://webpack.js.org/configuration/module/#ruleresolve](https://webpack.js.org/configuration/module/#ruleresolve)

The config Rule.resolve in module would cause a syntax error.


montogeek:
Thanks!

PR #3481, opened 2020-01-11
docs(configuration): Correct the function usage in externals config

Mistyyyy:
https://webpack.js.org/configuration/externals/#function
correct the function usage in externals config


montogeek:
Thanks!

PR #3499, opened 2020-01-20
docs(config) externals page updated acc to writers guide

EugeneHlushko:
- we dont use blocks for legend
- regex -> RegExp


montogeek:
Thanks!

PR #3509, opened 2020-01-30
Add missing div tag with an id "div" in index.html

mubarak-basha:
Add missing div tag with an id value of "div" in index.html code.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -78,6 +78,7 @@ __index.html__
     <script src="https://unpkg.com/lodash@4.16.6"></script>
   </head>
   <body>
+    <div id="div"></div>
     <script src="./src/index.js"></script>
   </body>
 </html>


montogeek:
Why it is missing?

mubarak-basha:
@montogeek I am sorry I don't understand what you are trying to ask

mubarak-basha:
It is missing because the JavaScript code reference an element with and id of value.
If you follow along the guide it will build successfully but, when you run the index.html page in browser you can see that it throws cannot set property of innerHTML of null error in browser console.

montogeek:
The JS is creating it and append it to the body:
```js
function component() {
  const element = document.createElement('div');

  // Lodash, currently included via a script, is required for this line to work

  return element;
}

```

There is not need to add the HTML tag.

mubarak-basha:
My bad. I misunderstood createElement as getElementById

PR #3511, opened 2020-01-31
Add info about passing cli args to nodejs

smelukov:
It about https://github.com/webpack/webpack-cli/pull/1195

### DIFF
--- a/src/content/api/cli.md
+++ b/src/content/api/cli.md
@@ -367,12 +367,14 @@ webpack --progress --profile
 ⋮
 ```
 
-## Pass CLI arguments to NodeJS
+## Pass CLI arguments to Node.js
 
-In case you need to pass nodejs-specific options or flags, you can use the `--node-args`-option.
+To pass arguments directly to Node.js process, you can use the `--node-args` option. All other flags and options will be received by the webpack-cli.
 
-Everything you specify there will be passed to NodeJS process, and the rest to webpack-cli as the options.
 
-For example, exec `webpack --node-args="--max-old-space-size=4096"`  to increasing the memory limit of NodeJS to 4gb
+For example, to increasing the memory limit of Node.js process to 4 GB
+
+```bash
+webpack --node-args="--max-old-space-size=4096"
 
 You can pass multiple options to NodeJS: `webpack --node-args="--max-old-space-size=4096" --node-args="-r /path/to/preload/file.js"`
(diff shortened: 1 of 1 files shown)


src/content/api/cli.md
```diff
@@ -366,3 +366,13 @@ webpack --progress --profile
...
 ⋮
 ```
+
+## Pass CLI arguments to NodeJS
+
+In case you need to pass nodejs-specific options or flags, you can use the `--node-args`-option.
+
+Everything you specify there will be passed to NodeJS process, and the rest to webpack-cli as the options.
+
+For example, exec `webpack --node-args="--max-old-space-size=4096"`  to increasing the memory limit of NodeJS to 4gb
```
EugeneHlushko:
```suggestion
For example, to increasing the memory limit of Node.js process to 4 GB

```bash
webpack --node-args="--max-old-space-size=4096"
```

src/content/api/cli.md
```diff
@@ -366,3 +367,15 @@ webpack --progress --profile
 6ms emitting
 ⋮
 ```
+
+## Pass CLI arguments to Node.js
+
+To pass arguments directly to Node.js process, you can use the `--node-args` option. All other flags and options will be received by the webpack-cli.
+
+For example, to increasing the memory limit of Node.js process to 4 GB
```
jamesgeorge007:
```suggestion
For example, to increase the memory limit of Node.js process to 4 GB
```

src/content/api/cli.md
```diff
@@ -366,3 +367,15 @@ webpack --progress --profile
...
+
+To pass arguments directly to Node.js process, you can use the `--node-args` option. All other flags and options will be received by the webpack-cli.
+
+For example, to increase the memory limit of Node.js process to 4 GB
+
+```bash
+webpack --node-args="--max-old-space-size=4096"
+```
+
+You can pass multiple options to Node.js: `webpack --node-args="--max-old-space-size=4096" --node-args="-r /path/to/preload/file.js"`
```
EugeneHlushko: Can you please format this example as ```bash too? Also remove `:` to be consistent with the first example. Also phrasing could be `Syntax of passing multiple options to Node.js`. I think sounds more natural for documentation rather than telling user what he can do... WDYT?

smelukov: Fixed the example
> WDYT

In the docs, there are many mentions of a user, (e.g. `This option allow you to ...`) and I think that it's ok even for docs ☺️

EugeneHlushko: Yeah they were created without certain guidelines, i just think that moving forward, we can improve the tone of voice and how we add infotmation, indeed it doesnt seem to be worth it to refactor other occurrences, but while we are adding new information, it can be improved

EugeneHlushko:
Thanks!

PR #3512, opened 2020-01-31
docs(config): cleanup default values

Neob91:
Some definitions of accepted/default config values didn't adhere to the writer's guide - I fixed them.

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -11,6 +11,7 @@ contributors:
   - EugeneHlushko
   - skovy
   - rishabh3112
+  - Neob91
 related:
   - title: Using Records
     url: https://survivejs.com/webpack/optimizing/separating-manifest/#using-records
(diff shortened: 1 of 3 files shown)


src/content/configuration/other-options.md
```diff
@@ -167,9 +168,9 @@ module.exports = {
 
 ### `cache.store`
 
-`string: 'pack'`
+`string = 'pack': 'pack'`
```
EugeneHlushko: do i understand correctly that schema only allows `'pack'` or maybe also `false` or any other values?

Neob91: I checked just now, to be sure, and it looks like the only supported value (and hence, the default as well).
Also, that's what the docs explicitly say below in ```pack is the only supported mode since webpack 5.0.x```

Neob91:
@EugeneHlushko I just rebased with `master` 🙂

EugeneHlushko:
@Neob91 no need, only do it in case we have conflicts or master has updated lint rules or similar

src/content/configuration/other-options.md
EugeneHlushko: yeah looks like it. but best to be on the safe side and double check :) https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L264

EugeneHlushko:
Thanks!

PR #3513, opened 2020-02-01
scroll to top when change topic from the sidebar

muhmushtaha:
I faced a problem when I was reading the docs which is if I was at the bottom of the page and moved to another section from the sidebar it will stay at the bottom of the page, so I added a scroll to top when I change the topic from the sidebar.

**Problem**
problem

**Solution**
solution


montogeek:
Thanks!

PR #3516, opened 2020-02-03
fix: corrected the path of the logger from webpack

anikethsaha:
_describe your changes..._

The path to webpack logger in the docs is fixed with the correct path


EugeneHlushko:
Critical! thanks

PR #3520, opened 2020-02-05
Improve progress plugin docs

smelukov:
Closes #3519

PR #3523, opened 2020-02-07
docs(comntrib) typed shortlists

EugeneHlushko:
We didn't have this in the guide, only on certain pages of the docs. Now its more obvious how to treat such lists

ref #3520


montogeek:
Please fix proselint error

PR #3537, opened 2020-02-14
docs(api): Update example link

chenxsan:
Update example link which is broken.

### DIFF
--- a/src/content/api/node.mdx
+++ b/src/content/api/node.mdx
@@ -9,6 +9,7 @@ contributors:
   - EugeneHlushko
   - lukasgeiter
   - toshihidetagami
+  - chenxsan
 ---
 
 webpack provides a Node.js API which can be used directly in Node.js runtime.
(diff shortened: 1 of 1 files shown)


src/content/api/node.mdx
```diff
@@ -223,7 +224,7 @@ stats.toJson({
 All available options and presets are described in the stats [documentation](/configuration/stats).
 
 > Here’s an [example]
-(https://github.com/webpack/analyse/blob/master/app/pages/upload/example.json)
+(https://github.com/webpack/analyse/blob/master/app/pages/upload/example1.json)
```
EugeneHlushko:
```suggestion
(https://raw.githubusercontent.com/webpack/analyse/master/app/pages/upload/example1.json)
```

chenxsan: With link to raw content you can't go anywhere but the json file, maybe you wanna check around then? I can update it if you insist :)

montogeek: File is not displayed by GitHub, it is better to show it by default:
image

EugeneHlushko:
Thanks!

PR #3292, opened 2019-09-19
UX: Make Content links visible (not hidden behind site header)

brandonkal:
Fixes the issue where one must constantly scroll up a bit after clicking a sidebar link (because content is hidden behind the site header).

The current documention site hides headers behind the site header. Clicking on a sidebar link, you should notice that you have to scroll up to see the title. This improves the experience.

This doesn't account for the additional notification banner that pushes the header down.

You can experiment with this proposed fix by adding this to the live site in dev tools:
```css
.markdown h3[id] {
    padding-top: 100px;
    margin-top: calc(-100px + 1.5em);
}
```

PR #3541, opened 2020-02-16
fix: #3540 (Links don't scroll to correct position)

### LINKED ISSUE #3540: Links don't scroll to correct position
When clicking the anchor links in the sidebar, it doesn't scroll to the correct position

## Current Behaviour 
It scrolls past the heading and past the description (actually it doesn't but it looks like it because the fixed topbar is covering the heading and the description).
webpack

## Expected Behaviour
It scrolls to the section so that the heading and the description is visible.
expected

## Possible Solution
According to [Stackoverflow](https://stackoverflow.com/questions/10732690/offsetting-an-html-anchor-to-adjust-for-fixed-header) a possible solution would be to apply something like `padding-top: 100px; margin-top: -100px;` to every heading. I have found that `100px` works well with a normal topbar and `200px` with a topbar with annoucement.

SimonSiefke:
Fix #3540

It isn't *perfect* because it doesn't cover all the possible heights the topbar can be in (mobile: 56px, mobile with banner: 56px + 58px, desktop: 56px + 27.3px, desktop with banner: 56px + 27.3px + 56px). 

I've tried to optimize for (probably) the most common use case desktop (56px + 27.3px)

Here is what it looks like on desktop now:
desktop-normal

<br>

On desktop with banner (heading is covered):
desktop_banner

<br>

On mobile (space at the top is a bit larger):
mobile

### DIFF
--- a/src/components/Markdown/Markdown.scss
+++ b/src/components/Markdown/Markdown.scss
@@ -2,6 +2,14 @@
 @import 'functions';
 @import 'prism-theme';
 
+$bannerHeight: 56px;
+$navigationHeight: 56px;
+$navigationSubHeight: 28px;
+$someExtraHeight: 10px;
+
+$topHeightDesktop: $navigationHeight + $navigationSubHeight + $someExtraHeight;
+$topHeightDesktopWithBanner: $bannerHeight + $topHeightDesktop;
+
 .markdown {
   line-height: 1.5em;
 
@@ -38,6 +46,7 @@
     margin: 1.5em 0 0.25em;
     color: getColor(fiord);
     word-break: break-all;
+    scroll-margin-top: $topHeightDesktop;
 
     tt,
     code {
@@ -380,5 +389,11 @@
     tt {
       white-space: pre-line;
     }
+  }
+}
+
+.notification-bar-visible .markdown {
+  h1,h2,h3,h4,h5,h6{
+    scroll-margin-top: $topHeightDesktopWithBanner;
   }
 }
--- a/src/components/NotificationBar/NotificationBar.jsx
+++ b/src/components/NotificationBar/NotificationBar.jsx
@@ -52,10 +52,16 @@ export default class NotificationBar extends React.Component {
     this.state = {
       dismissed: barDismissed()
     };
+    if (!this.state.dismissed && typeof document !== 'undefined') {
+      document.body.classList.add('notification-bar-visible');
+    }
   }
 
   onClose() {
     this.setState(state => {
+      if (!state.dismissed && typeof document !== 'undefined') {
+        document.body.classList.remove('notification-bar-visible');
+      }
       return {
         dismissed: !state.dismissed
       };


montogeek:
Related: https://github.com/webpack/webpack.js.org/pull/3292

EugeneHlushko:
@SimonSiefke thank you for trying to solve this, its been on our plate for a while. Can you add a special class name to body or any other container when the sponsorship banner is visible so that we can also have a modifier for it

SimonSiefke:
I've updated it so that it adds the class `notification-bar-visible` to the body when the notification bar is visible. That fixes that the heading is covered on desktop with banner.

SimonSiefke:
Apparently there is now an even better solution for this problem with `scroll-margin-top` (https://css-tricks.com/fixed-headers-and-jump-links-the-solution-is-scroll-margin-top/), which works for most browsers (https://caniuse.com/#search=scroll-margin-top)

If you want, I can refactor it.

src/components/Markdown/Markdown.scss
```diff
@@ -30,14 +38,19 @@
...
   h6 {
-    display: flex;
-    align-items: center;
     font-family: $font-stack-heading;
     font-weight: 600;
     line-height: 1.4;
-    margin: 1.5em 0 0.25em;
     color: getColor(fiord);
     word-break: break-all;
+    margin-bottom: 0.25em;
```
EugeneHlushko: can we use `scroll-margin-top` instead of all of these? https://css-tricks.com/fixed-headers-and-jump-links-the-solution-is-scroll-margin-top/

src/components/NotificationBar/NotificationBar.jsx
```diff
@@ -19,8 +19,10 @@ class MessageBar extends React.Component {
       <div className="notification-bar">
         <Container className="notification-bar__inner">
           <p>
-            Sponsor webpack and get apparel from the <a href="https://webpack.threadless.com">official shop</a>! All
-            proceeds go to our <a href="https://opencollective.com/webpack">open collective</a>!
+            Sponsor webpack and get apparel from the{' '}
```
EugeneHlushko: why formatting changes?

src/components/NotificationBar/NotificationBar.jsx
```diff
@@ -50,21 +52,31 @@ export default class NotificationBar extends React.Component {
...
       };
     });
   }
 
   render() {
     const { dismissed } = this.state;
 
-    return <React.Fragment>{!dismissed ? <MessageBar onClose={this.onClose} /> : null}</React.Fragment>;
+    return (
+      <React.Fragment>
```
EugeneHlushko: duplicate question, why formatting changes?

EugeneHlushko:
RIght, just saw your last comment. please do

montogeek:
We can also use this https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top

SimonSiefke:
Updated to use `scroll-margin-top`. Code looks a lot cleaner now. Also disabled prettier for this workspace, so it doesn't do this `{' '}` thing.

montogeek:
Preview thing would be really helpful here :(

EugeneHlushko:
Thanks!

PR #3547, opened 2020-02-16
docs(format) single quotes in js spotted, minor grammar

EugeneHlushko:
While reviewing recent PR i spotted mis- formatting on api pages, just cleared the occurrences and found another grammar improvement

### DIFF
--- a/src/content/api/hot-module-replacement.md
+++ b/src/content/api/hot-module-replacement.md
@@ -164,10 +164,10 @@ The `info` parameter will be an object containing some of the following values:
 
 ```js
 {
-  type: "self-declined" | "declined" |
-        "unaccepted" | "accepted" |
-        "disposed" | "accept-errored" |
-        "self-accept-errored" | "self-accept-error-handler-errored",
+  type: 'self-declined' | 'declined' |
+        'unaccepted' | 'accepted' |
+        'disposed' | 'accept-errored' |
+        'self-accept-errored' | 'self-accept-error-handler-errored',
   moduleId: 4, // The module in question.
   dependencyId: 3, // For errors: the module id owning the accept handler.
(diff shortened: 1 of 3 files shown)


anikethsaha:
can we add `.vscode` config to work with eslint and prettier and remove it from `.gitignore` ?

anshumanv:
> can we add .vscode config to work with eslint and prettier and remove it from .gitignore ?

I don't think that's a good idea given not everyone uses the same lint/format config. Having a single config in the project should be sufficient.

anikethsaha:
> I don't think that's a good idea given not everyone uses the same lint/format config. Having a single config in the project should be sufficient.

I think vscode wont look for global config if `.vscode` is present in the project. so in that way, we all will be using same vscode config and wont get formatting issues.
just that we need to configure prettier well with linters in the project config,

EugeneHlushko:
Yes linter should've caught this but its tricky in .md and .mdx files in different snippets. Therefore possible inconsistencies. Keep eye out for them since all the PRs require review

anikethsaha:
eslint + prettier works like a charm 🚀 

> its tricky in .md and .mdx files in different snippets.

Yea, this might be an issue, but there is an [`eslint plugin for mdx`](https://github.com/mdx-js/eslint-mdx) and for [`.md`](https://github.com/eslint/eslint-plugin-markdown)  but there is no rules and support for `.mdx` from `markdownlint`'s side. 

So if we are considering consistent formatting using prettier, then I think we may need to remove `markdownlint` and use` eslint-plugin-markdown` and `eslint-mdx` instead. 

Even removing `markdownlint` right now isnt a good idea either 


>  Keep eye out for them since all the PRs require review

Yes .

EugeneHlushko:
Another thing is that there might be a need for double quotes in those files e.g. JSON and regular quotes in plain text, so yeah its tricky

PR #3553, opened 2020-02-17
Entry description, entry.filename and entry.dependOn

smelukov:
Add info about entry description, `entry.filename` and `entry.dependOn`

### DIFF
--- a/src/content/configuration/entry-context.md
+++ b/src/content/configuration/entry-context.md
@@ -8,6 +8,7 @@ contributors:
   - byzyk
   - madhavarshney
   - EugeneHlushko
+  - smelukov
 ---
 
 The entry object is where webpack looks to start building the bundle. The context is an absolute string to the directory that contains the entry files.
@@ -33,7 +34,7 @@ By default the current directory is used, but it's recommended to pass a value i
 
 ## `entry`
 
-`string` `[string]` `object = { <key> string | [string] }` `(function() => string | [string] | object = { <key> string | [string] })`
+`string` `[string]` `object = { <key> string | [string] | object = { import string | [string], dependOn string | [string], filename string }}` `(function() => string | [string] | object = { <key> string | [string] } | object = { import string | [string], dependOn string | [string], filename string })`
 
 The point or points where to start the application bundling process. If an array is passed then all items will be processed.
 
@@ -56,6 +57,58 @@ module.exports = {
 ### Naming
 
 If a string or array of strings is passed, the chunk is named `main`. If an object is passed, each key is the name of a chunk, and the value describes the entry point for the chunk.
+
+### Entry descriptor
+
+If an object is passed the value might be a string, array of strings or descriptor:
+
+```js
+module.exports = {
+  //...
+  entry: {
+    home: './home.js',
+    about: ['./about.js'],
+    contact: { import: './contact.js' }
+  }
+};
+```
+
+Descriptor syntax might be used to pass additional options to an entry point.
+
+
+### Output filename
+
+By default, an output filename for the entry chunk is getting from [`output.filename`](/configuration/output/#outputfilename) but you can specify a custom output filename for a specific entry:
+
+```js
+module.exports = {
+  //...
+  entry: {
+    app: './app.js',
+    home: { import: './contact.js', filename: 'pages/[name][ext]' },
+    about: { import: './about.js', filename: 'pages/[name][ext]' }
+  }
+};
+```
+
+Descriptor syntax was used here to pass `filename`-option to the specific entry points.
+
+
+### Dependencies
+
+By default, every entry chunk stores all the modules that it uses. With `dependOn`-option you can share the modules from one entry chunk to another:
+
+```js
+module.exports = {
+  //...
+  entry: {
+    app: { import: './app.js', dependOn: 'react-vendors' },
+    'react-vendors': ['react', 'react-dom', 'prop-types']
+  }
+};
+```
+
+The `app` chunk will not contain the modules that `react-vendors` has.
 
 
 ### Dynamic entry
--- a/src/content/configuration/entry-context.md
+++ b/src/content/configuration/entry-context.md
@@ -60,15 +60,24 @@ If a string or array of strings is passed, the chunk is named `main`. If an obje
 
 ### Entry descriptor
 
-If an object is passed the value might be a string, array of strings or descriptor:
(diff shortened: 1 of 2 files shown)


src/content/configuration/entry-context.md
```diff
@@ -57,6 +58,58 @@ module.exports = {
 
 If a string or array of strings is passed, the chunk is named `main`. If an object is passed, each key is the name of a chunk, and the value describes the entry point for the chunk.
 
+### Entry descriptor
```
anikethsaha:
```suggestion
### Entry description
```

do you meant `description` ?

smelukov: Nope. `descriptor` )

anikethsaha: ok cool 👍

src/content/guides/code-splitting.md
```diff
@@ -116,6 +117,32 @@ The first of these two points is definitely an issue for our example, as `lodash
 
 ## Prevent Duplication
 
+### Entry dependencies
+
+The [`dependOn` option](https://webpack.js.org/configuration/entry-context/#dependencies) allows to share the modules between the chunks
```
jamesgeorge007:
```suggestion
The [`dependOn` option](/configuration/entry-context/#dependencies) allows to share the modules between the chunks
```

smelukov: thanks

src/content/configuration/entry-context.md
```diff
@@ -33,7 +34,7 @@ By default the current directory is used, but it's recommended to pass a value i
 
 ## `entry`
 
-`string` `[string]` `object = { <key> string | [string] }` `(function() => string | [string] | object = { <key> string | [string] })`
+`string` `[string]` `object = { <key> string | [string] | object = { import string | [string], dependOn string | [string], filename string }}` `(function() => string | [string] | object = { <key> string | [string] } | object = { import string | [string], dependOn string | [string], filename string })`
```
montogeek: @EugeneHlushko This might grow out of control, we should think in a different way to handle this.

src/content/configuration/entry-context.md
```diff
@@ -57,6 +58,58 @@ module.exports = {
 
 If a string or array of strings is passed, the chunk is named `main`. If an object is passed, each key is the name of a chunk, and the value describes the entry point for the chunk.
 
+### Entry descriptor
+
+If an object is passed the value might be a string, array of strings or descriptor:
```
montogeek: What is a descriptor?

src/content/configuration/entry-context.md
```diff
@@ -57,6 +58,58 @@ module.exports = {
...
+  }
+};
+```
+
+Descriptor syntax might be used to pass additional options to an entry point.
+
+
+### Output filename
+
+By default, an output filename for the entry chunk is getting from [`output.filename`](/configuration/output/#outputfilename) but you can specify a custom output filename for a specific entry:
```
montogeek:
```suggestion
By default, the output filename for the entry chunk is extracted from [`output.filename`](/configuration/output/#outputfilename) but you can specify a custom output filename for a specific entry:
```

src/content/configuration/entry-context.md
```diff
@@ -57,6 +58,58 @@ module.exports = {
...
+  //...
+  entry: {
+    app: './app.js',
+    home: { import: './contact.js', filename: 'pages/[name][ext]' },
+    about: { import: './about.js', filename: 'pages/[name][ext]' }
+  }
+};
+```
+
+Descriptor syntax was used here to pass `filename`-option to the specific entry points.
```
montogeek: We definitely need to show an example showing all the keys that descriptors receive.

smelukov: I’ve already did it in filename and deprndOn section. Or do you mean some example with all the fields in one place?

montogeek: > Or do you mean some example with all the fields in one place?

Yes, that would be great.

src/content/configuration/entry-context.md
smelukov: Descriptor is a structure or expression that describes something

src/content/configuration/entry-context.md
smelukov: Why we don’t use ts types?

EugeneHlushko: Originally we wanted to keep webpack docs typings short which were somewhat similar to what you see now, what we've done in the past is making them consistent and documenting them in contribute section. But later as we started being precise on types (previously this wasnt the case for the whole documentation), types started to look "longer" and "even longer" :D

There was a suggestion to convert it to TS but its still not clear if TS would look shorter, i think TS or this both has disadvantages, you dont say that entry level developer is going to understand typescript? This is also not the best approach in terms of accessibility of the docs. In the end, to use the tool, you dont need to understand its internals

anikethsaha: @EugeneHlushko I think types should be provided and currently how webpack docs displays type seems perfect. I wouldn't suggest to convert it to TS cause I **think** it will be much longer cause of external types and interfaces

Current type placements are perfect as well cause developers usually don't want to navigate or open a new tab just to see what kind of information it takes. 

Perhaps ! can we create a separate section of the page for types to link long types like this one there? (*again navigating point will oppose this*)

What I think is this is good. cause longer types are handy as long as it is distance itself from the main content.

What do you think ?

montogeek: I have an idea:
Extract, generate and export webpack source TS types to an static page, then link it from the docs.

Another one:
Extract types information into a JSON format, then embed it into the docs through React.

Everything automated.

anikethsaha: > Perhaps ! can we create a separate section of the page for types to link long types like this one there? (again navigating point will oppose this)

like this ?

> Perhaps ! can we create a separate section of the page for types to link long types like this one there? (again navigating point will oppose this)

anikethsaha: @montogeek webpack 's source has jsdocs types comment 

using [`this`](https://jsdoc.app/about-getting-started.html#generating-a-website)

we can add something like `type.webpack.js.org` ?

smelukov:
Can we merge it?

montogeek:
@smelukov I don't merge immediately because I wait for more contributors to take a look.

PR #3560, opened 2020-02-18
chore: og:image

38elements:
Since SVG may not be supported at Twitter card,
This uses [icon-pwa-512x512.png](https://github.com/webpack/webpack.js.org/blob/master/src/assets/icon-pwa-512x512.png).


EugeneHlushko:
Thanks! hope it finally works

38elements:
Currently, Logo is displayed at Twitter card.

PR #3562, opened 2020-02-18
docs(api): remove redundant new keyword

jamesgeorge007:
Follow up of #3545 

`createFsFromVolume()` is a function and should be invoked directly.

### DIFF
--- a/src/content/contribute/writing-a-loader.md
+++ b/src/content/contribute/writing-a-loader.md
@@ -6,6 +6,7 @@ contributors:
   - michael-ciniawsky
   - byzyk
   - anikethsaha
+  - jamesgeorge007
 ---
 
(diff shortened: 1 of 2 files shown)


src/content/contribute/writing-a-loader.md
```diff
@@ -6,6 +6,7 @@ contributors:
   - michael-ciniawsky
   - byzyk
   - anikethsaha
+  - jamesgeorge007
```
montogeek: Why this change?

jamesgeorge007: In regard to https://github.com/webpack/webpack.js.org/pull/3493. The list wasn't updated as part of that PR :sweat_smile:

montogeek:
Thanks!

PR #3571, opened 2020-02-20
doc(api): loaders doc add function remove prompt

lcy930323:
loaders doc add function remove prompt


montogeek:
@EugeneHlushko How can we transfer this to v4 docs? I think it should also be there.

anikethsaha:
> @EugeneHlushko How can we transfer this to v4 docs? I think it should also be there.

yes it should be in v4

PR #3576, opened 2020-02-22
docs: Remove redundant word

wizardofhogwarts:
'group' itself is explanatory. No need to mention 'together' explicitly.


montogeek:
Thanks!

PR #3586, opened 2020-02-26
docs(concepts) a follow up on page creation

EugeneHlushko:
- a follow up to do small grammar fixes
- added a link to template strings in output
- fix name of SplitChunksPlugin
- shortened info in `initial` and `non-initial`, use consistently

### DIFF
--- a/src/content/concepts/under-the-hood.md
+++ b/src/content/concepts/under-the-hood.md
@@ -1,11 +1,12 @@
 ---
 title: Under The Hood
-sort: 1
+sort: 13
 contributors:
   - smelukov
+  - EugeneHlushko
 ---
 
-> This section describes webpack internals and useful for plugin developers
+> This section describes webpack internals and can be useful for plugin developers
 
 The bundling is a function that takes some files and emits others.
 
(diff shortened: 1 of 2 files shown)


src/content/concepts/under-the-hood.md
```diff
@@ -27,7 +28,7 @@ __./app.js__
 export default 'the app';
 ```
 
-By using each other, the modules form a graph (`ModuleGraph`). 
+By using each other, the modules form a graph (`ModuleGraph`).
 
 Usually, modules don't exist by themselves and combine into chunks.
```
EugeneHlushko: combine into chunks doesnt make sense without changing to `are getting combined`, but i also feel we could add context and explicitly say that `When outputted by webpack, modules are usually combined into chunks.`

```suggestion
When outputted by webpack, modules are usually combined into chunks.
```

smelukov: Maybe `During bundling, modules are combined into chunks.`?

EugeneHlushko: makes sense

EugeneHlushko: WDYT?
```suggestion
During the bundling process, modules are combined into chunks.
```

smelukov: 👍

smelukov:
@EugeneHlushko thanks!

PR #3593, opened 2020-03-02
docs(config/resolve): Fixing a typo

bigomega:
In every example in resolve docs table, `abc` is the (assumed) name of the project. Making this path `/abc` consistent since a couple of examples were `/abs` even though the result was `/abc`.


montogeek:
Thanks!

PR #3594, opened 2020-03-02
ui(version) show the actual webpack version from master

### LINKED ISSUE #3592: Code Splitting: multiple entry points object schema not supported
# Bug report

> Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
>  - configuration.entry should be one of these:
>    function | object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string]

- Setup multiple entry points with object schema
`object = { import string | [string], dependOn string | [string], filename string }`
- Run webpack command

Support [documented schema](https://webpack.js.org/guides/code-splitting/).

Webpack version:  4.41.6
Operating System: MacOS Mojave, Ubuntu 1.18 (Linux 4.15)

EugeneHlushko:
Closes #3592 

result

### DIFF
--- a/src/components/Sidebar/Sidebar.jsx
+++ b/src/components/Sidebar/Sidebar.jsx
@@ -35,7 +35,7 @@ export default ({
       <div className="sidebar__inner">
         <div className="sidebar__shields">
           <a href="https://github.com/webpack/webpack/releases">
-            <Shield content="npm/v/webpack" label="webpack" />
+            <Shield content="github/package-json/v/webpack/webpack" label="webpack" />
           </a>
         </div>
         <Print url={currentPage} />


montogeek:
Uhmm, Do you think that it is going to be strictly the same version all the time? Sometimes webpack does the release, but some documentation haven't been finished.

src/components/Sidebar/Sidebar.jsx
```diff
@@ -35,7 +35,7 @@ export default ({
       <div className="sidebar__inner">
         <div className="sidebar__shields">
           <a href="https://github.com/webpack/webpack/releases">
-            <Shield content="npm/v/webpack" label="webpack" />
+            <Shield content="github/package-json/v/webpack/webpack" label="webpack" />
```
anikethsaha:
```suggestion
            <Shield content="github/package-json/v/webpack/webpack" label="webpack | master" />
```

I would suggest to add something similar ! WDYT ?

Also, keeping this as well 

```js
<Shield content="npm/v/webpack" label="webpack | latest" />
```

EugeneHlushko: master has only released version in package json no need to make shield bigger, if its in there, its released

EugeneHlushko: its a shame you cannot query the npm shield with betas, there is no such flag. Maybe we can ask for this feature in shields repo

EugeneHlushko:
@montogeek yeah thats how it currently worked, @sokra released something and issues with doc request were still in here or related PRs were in progress. So this doesnt change that behaviour, it only shows actual latest release.

src/components/Sidebar/Sidebar.jsx
anikethsaha: > its a shame you cannot query the npm shield with betas, there is no such flag. Maybe we can ask for this feature in shields repo

Didnt know about that. Yes I guess its good to have this enhancement from shield

montogeek:
@EugeneHlushko Isn't that the entire problem? People read that badge version and think that the current docs they are reading matches that webpack release.

Mookiepiece:
same problem on another page like 3592. I hope the version will be corrected soon

EugeneHlushko:
@montogeek we should aim to have docs in sync with the badge imo, right now its very close

EugeneHlushko:
I am merging this for now until we get a better solution, not showing it at all looks worse than showing a beta/alpha 5

EugeneHlushko:
Thanks for your participation everyone, we are looking forward to a better solution

PR #3596, opened 2020-03-02
docs(config) document output.chunkCallbackName option

EugeneHlushko:
- document `output.chunkCallbackName` configuration option


montogeek:
Thanks!

PR #3597, opened 2020-03-03
docs(config) follow up to link from list to recently merged option docs

EugeneHlushko:
Add missing link after #3596 got merged


montogeek:
Thanks!

PR #3610, opened 2020-03-10
fix(component): fix style bugs in Footer component

ricosmall:
Fix style bugs in `Footer` component which caused body's max width wider than screen width in mobile phone browser.

Before style change:

body's width become wider than screen width

After style change:

normal


EugeneHlushko:
Thanks great catch!

PR #3620, opened 2020-03-15
Docs: IgnorePlugin: Clarify the “deprecated” usage

iamakulov:
“Deprecated” means “it works but will be deleted soon”, which is not the case in webpack 5. Webpack 5 drop support for the old approach completely.


montogeek:
Thanks!

PR #3633, opened 2020-03-20
chore: integrate prettier ( js, jsx, scss )

snitin315:
_describe your changes..._

Integrating prettier. (js, jsx, scss)


snitin315:
> I think we should do it using two PRs, for example -
> 
> 1. JS, CSS and etc
> 2. Markdown

Sure :+1:

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "singleQuote": true,
+    "trailingComma": "all",
```
EugeneHlushko: this is totally opinionated, same as arrow parenthesis for single arg, why are you adding them?

snitin315: It's the general configuration used in many projects.
what's your opinion? What should be done?

montogeek: Without, please keep in mind current style.

chenxsan:
Closed in favor of https://github.com/webpack/webpack.js.org/pull/4518. Thanks anyway.

PR #3634, opened 2020-03-20
chore: use yarn to run scripts

snitin315:
_describe your changes..._

Using `yarn` to run scripts brings uniformity as we are using `yarn` for the project. 
 
### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "clean-printable": "rimraf src/content/**/printable.md",
     "preclean": "run-s clean-dist clean-printable",
     "clean": "rimraf src/content/**/_*.md src/**/_*.json",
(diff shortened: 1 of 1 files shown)


snitin315:
I just noticed something. We are using `yarn.lock` , Hence the contributor should install dependencies with `yarn` to avoid any conflicts. But the `CONTRIBUTING.md` is pointing to set up the project with `npm` as a priority.(with a side note of installation with `yarn`). 
Screenshot from 2020-03-20 18-03-17

**I think this should be opposite**, how about I change this as well in this PR along with the current change. This will be more appropriate. 

This makes more sense to me. What do you say?
@EugeneHlushko @evilebottnawi @montogeek

himanshu010:
this doesn't have any major effect on the project as it is doing the same thing in either case

snitin315:
suppressed by #3633

PR #3636, opened 2020-03-21
docs(api): improve module-methods.md

wizardofhogwarts:


montogeek:
Why do you find this necessary?

wizardofhogwarts:
> Why do you find this necessary?

When I was going through this first time, it was not appearing to be implicit. That's why to improve readability, I think this change is required.

EugeneHlushko:
Hmm the context seems there, like pretty much highlighted. I think this addition makes it a bit weird. Any more thoughts?

snitin315:
Closing due to age.

PR #3645, opened 2020-03-25
docs(loader): remove resolveSync()

### LINKED ISSUE #3644: resolveSync not defined in loader context
# Bug report

Loaders should be able to call `this.resolveSync()` for a synchronous module resolution as documented.

See issue webpack/webpack#1634 that was closed without resolution for context.

If this API has been removed and the webpack team does not care to restore it, it should be removed from the documentation.

This is important for loaders that require synchronous resolve. It enables plugins to avoid creating a separate enhanced-resolve instance and attempting to make its behavior match the one used by the webpack compiler.

See:

https://github.com/brandonkal/linaria/blob/27888fa8a3e15cedbfdcf779b18d81ca04af77ee/src/loader.ts#L277-L280

For a runnable example, you can clone that repository and run vscode-jest-tests in the VSCode debugger. You will see that the resolveSync is not defined in the loader's context, so the fallback code is executed.

webpack version:  4.40.2
Node.js version: v10.15.3
Operating System: MacOS 10.12.6
Additional tools: webpack-dev-server

snitin315:
Fixes  #3644

Removed `this.resolveSync`  from the depreciated list as it has been really removed.

### DIFF
--- a/src/content/api/loaders.md
+++ b/src/content/api/loaders.md
@@ -11,6 +11,7 @@ contributors:
   - jantimon
   - superburrito
   - wizardofhogwarts
+  - snitin315
 ---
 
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Thanks!

PR #3649, opened 2020-03-25
docs(examples) fix examples missing a require

EugeneHlushko:
In order to make all of our examples accessible i've added the missing require call to `path` as we already had it in almost all examples using `path` directly


EugeneHlushko:
Hmm travis hanging second time, re-triggering it

montogeek:
Uhdmm, what about adding ESLint to it? This has been talked in the past.

EugeneHlushko:
yeah that would be useful

PR #3651, opened 2020-03-26
docs(config): deterministic in optimization.chunkIds

anikethsaha:
added deterministic in optimization.chunkIds 

- ref https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L1111
- ref https://github.com/webpack/webpack/pull/8477


montogeek:
What does it mean it is dertiministic?

anikethsaha:
I think it means that chunk ids are determine based on previous or initial instead of random . 
Not sure though

montogeek:
Yeah, hard to know unless we read the code, but from an user perspective, reading just "deterministic" doesn't tell you anything, it is determined by what?

anikethsaha:
> Yeah, hard to know unless we read the code, but from an user perspective, reading just "deterministic" doesn't tell you anything, it is determined by what?

I am not 100% sure about that that's why didn't add that. 
Is there a way we can get to know that ?

montogeek:
Reading the code

sokra:
deterministic ids is like hashed ids, but using a numeric id.

So: deterministic means a short numeric id that is based on module/chunk name, so it's unlikely to change between compilations which is useful for long term caching.

anikethsaha:
thanks @sokra for the clarification. 

I will change this soon 👍

anikethsaha:
cc @sokra @montogeek . is the current one fine ?

src/content/configuration/optimization.md
```diff
@@ -259,9 +260,11 @@ Option                  | Description
 ----------------------- | -----------------------
 `'natural'`             | Numeric ids in order of usage.
 `'named'`               | Readable ids for better debugging.
+`'deterministic'`       | Short numeric ids which will not be changing between compilation. Good for long term caching. Enable by default for production mode
```
montogeek, line 263:
```suggestion
`'deterministic'`       | Short numeric ids which will not be changing between compilation. Good for long term caching. Enable by default for production mode.
```

montogeek:
Thanks!

PR #3655, opened 2020-03-28
output(config) : ADD output.enabledLibraryTypes

jeffin143:
In reference with :

https://github.com/webpack/webpack/blob/d00e65529a4c5a4e91843f2d4178c66eff3784ae/schemas/WebpackOptions.json#L1663

https://github.com/webpack/webpack/blob/d00e65529a4c5a4e91843f2d4178c66eff3784ae/lib/WebpackOptionsApply.js#L224-L227


src/content/configuration/output.md
```diff
@@ -1085,6 +1085,21 @@ module.exports = {
 };
 ```
 
+## `output.enabledLibraryTypes`
+
+`array`
```
EugeneHlushko, line 1090: array of strings?

src/content/configuration/output.md
```diff
@@ -1085,6 +1085,21 @@ module.exports = {
...
+
+List of library types enabled for use by entry points.
```
EugeneHlushko, line 1092: I think this needs some explanation, what happens when they require modules that arent in enabled types? What is the use case of the option

src/content/configuration/output.md
jeffin143: yes

How should , I add it
`array of strings` ?

src/content/configuration/output.md
jeffin143: I am not sure about the use case, but going through the code, it describes this as

If you are using a function as entry which sets "library", you need to add all potential library types to `output.enabledLibraryTypes`.

src/content/configuration/output.md
snitin315: [string]

EugeneHlushko:
Thanks!

PR #3656, opened 2020-03-28
Fix scroll bug for fragment

chenxsan:
When you open https://webpack.js.org/concepts/loaders/#resolving-loaders, it will always scroll to top which is unfortunately not what we want. It's a bug introduced in https://github.com/webpack/webpack.js.org/pull/3513.

This pull request should hopefully (I'm not 100% sure because I can't test the search function locally when docsearch only works on live site as far as I see) fix it with e2e tests written in cypress included.


package.json
```diff
@@ -52,7 +52,8 @@
     "deploy": "gh-pages -d dist",
     "preprintable": "npm run clean-printable",
     "printable": "node ./src/scripts/concatenate-docs.js",
-    "jest": "jest"
+    "jest": "jest",
+    "cypress:open": "cypress open"
```
anikethsaha, line 56:
```suggestion
    "cypress:open": "cypress run"
```
I think it has to be `run` as `open` opens a visual browser instead of headless and for CI use (which is I think will be an option for future if this repo is having tests) it is better as headless.

Although I am not completely sure about it. let me know 👍

cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+  "baseUrl": "http://localhost:3000"
```
anikethsaha, line 2: will it run a local server while running the tests ? if so , before/after wds ?

I guess using [this](https://github.com/bahmutov/start-server-and-test) might be helpful in case of testing the app running in local using wds .

chenxsan: Nope, we have to start webpack dev server before running cypress tests. But yes, if we consider integrating cypress into CI, we could automate it according to https://docs.cypress.io/guides/guides/continuous-integration.html#Boot-your-server

package.json
chenxsan: Sometimes we need visual browser for debugging :) Maybe we can add another `script` like `cypress:run` for CI usage?

anikethsaha: > Sometimes we need visual browser for debugging :) Maybe we can add another script like cypress:run for CI usage?

yes, both should be there if we are going to setup CI. for this PR , the current seems fine as I think adding CI might be in a separate PR. lets hear from others.

EugeneHlushko:
Yay, i will check on preview link

chenxsan:
@EugeneHlushko It's possible to run cypress tests [on server side](https://docs.cypress.io/guides/guides/continuous-integration.html#Boot-your-server):

image

Do you want me to add support for running cypress server side or prefer excluding them from git?

EugeneHlushko:
Can it also run on travis? Maybe you can make it work by tweaking `travis.yml` scripts

chenxsan:
@EugeneHlushko I'll check it, will let you know whether it works on travis.

chenxsan:
@EugeneHlushko It did work on travis https://travis-ci.org/github/chenxsan/webpack.js.org/jobs/681974601#L2835.

Btw, I've rebased my branch to the latest master branch.

EugeneHlushko:
@chenxsan interesting, it failed after another update with master. Want to take a look at the error message? https://travis-ci.org/github/webpack/webpack.js.org/jobs/688529595

chenxsan:
@EugeneHlushko Seems cypress binary is missing for unknown reason, I've updated the `travis.yml` to fix it according to [this example](https://github.com/bahmutov/yarn-cypress-cache/blob/master/.travis.yml#L23)

EugeneHlushko:
Thanks!

PR #3665, opened 2020-04-02
docs(dllPlugin): add missing options

EslamHiko:
adds `format` for dll plugin and `extentions` for DllReferencePlugin

https://github.com/webpack/webpack/blob/d00e65529a4c5a4e91843f2d4178c66eff3784ae/schemas/plugins/DllPlugin.json#L15
https://github.com/webpack/webpack/blob/d00e65529a4c5a4e91843f2d4178c66eff3784ae/schemas/plugins/DllReferencePlugin.json#L109


montogeek:
Thanks!

PR #3666, opened 2020-04-02
docs(HashedModuleIdsPlugin): add missing context option

EslamHiko:
add `context` option to HashedModuleIdsPlugin.

https://github.com/webpack/webpack/blob/d00e65529a4c5a4e91843f2d4178c66eff3784ae/schemas/plugins/HashedModuleIdsPlugin.json#L6


src/content/plugins/hashed-module-ids-plugin.md
```diff
@@ -18,6 +19,7 @@ new webpack.HashedModuleIdsPlugin({
 
 This plugin supports the following options:
 
+- `context`: The context directory for creating names.
```
montogeek, line 22: Reading webpack source code it seems it has to be an absolute path. Could you please add a note about it?

montogeek:
Thanks!

PR #3669, opened 2020-04-02
Docs(plugin): schema-utils api changed

izayl:
schema-utils api has change the third param to type below

```ts
export type ValidationErrorConfiguration = {
  name?: string | undefined;
  baseDataPath?: string | undefined;
  postFormatter?: PostFormatter | undefined;
};
```

see: [https://github.com/webpack/schema-utils/pull/41](https://github.com/webpack/schema-utils/pull/41)

from the guide, the developer may use the latest version of schema-utils that not matched the version that webpack used.

So I add a notice to warning this.

### DIFF
--- a/src/content/contribute/writing-a-plugin.md
+++ b/src/content/contribute/writing-a-plugin.md
@@ -76,7 +76,7 @@ module.exports = {
 ```
 
 
-Use [`schema-utils`](https://github.com/webpack/schema-utils) in order to validate the options being passed through the plugin options.(Notice that [`schema-utils`](https://github.com/webpack/schema-utils) api has changed, but webpack use the v1.0.0 version, so you should the same version with webpack.) Here is an example:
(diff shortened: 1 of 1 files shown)


src/content/contribute/writing-a-plugin.md
```diff
@@ -76,7 +76,7 @@ module.exports = {
 ```
 
 
-Use [`schema-utils`](https://github.com/webpack/schema-utils) in order to validate the options being passed through the plugin options. Here is an example:
+Use [`schema-utils`](https://github.com/webpack/schema-utils) in order to validate the options being passed through the plugin options.(Notice that [`schema-utils`](https://github.com/webpack/schema-utils) api has changed, but webpack use the v1.0.0 version, so you should the same version with webpack.) Here is an example:
```
montogeek, line 79:
```suggestion
Use [`schema-utils`](https://github.com/webpack/schema-utils) in order to validate the options being passed through the plugin options.(Notice that [`schema-utils`](https://github.com/webpack/schema-utils) API has changed, but webpack use the v1.0.0 version, so you should use the same version with webpack.) Here is an example:
```

izayl:
@EugeneHlushko Is there still has any problem with this PR that I need to modify?

EugeneHlushko:
@izayl i think its good, thank you!

PR #3671, opened 2020-04-02
docs(devserver): webpack-dev-server open new option type

EslamHiko:
Documentation for this PR : https://github.com/webpack/webpack-dev-server/pull/2492 in webpack dev server


src/content/configuration/dev-server.md
```diff
@@ -844,6 +844,21 @@ module.exports = {
 };
 ```
 
+Sometimes you may want to use some options when opening the browser like opening an incognito window. In order to do that you can set `open` to an object. The object must have `app` property as an array. The first element in the array must be the browser name and the other following elements are the options you want. For example:
```
EugeneHlushko, line 847: This is very abstract in the beginning. Need to make it more of a documentation level e.g. remove `Sometimes...` Say clearly what you can achieve or what is allowed when passing object to `devServer.open`.

Is `app` the only key allowed in this object?

src/content/configuration/dev-server.md
```diff
@@ -816,7 +816,7 @@ module.exports = {
 
 ## `devServer.open`
 
-`boolean = false` `string`
+`boolean = false` `string` `object`
```
EugeneHlushko, line 819: https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types
```suggestion
`boolean = false` `string` `object = { app [string] }`
```

src/content/configuration/dev-server.md
EslamHiko: what about : `If you want to use some switches when opening the browser like opening an incognito window (--incognito switch), you can set `open` to an object. The object must have `app` property as an array. The first element in the array must be the browser name and the other following elements are the switches you want. For example:`

No, all options in https://www.npmjs.com/package/open will work.

EugeneHlushko: Yes this sounds much better! Looks good, can you commit this?

EslamHiko: @EugeneHlushko Check my last commit 😄

EugeneHlushko:
Thanks!

PR #3676, opened 2020-04-04
docs(config) : Add number type to watchoption 

jeffin143:
https://github.com/webpack/webpack/blob/d00e65529a4c5a4e91843f2d4178c66eff3784ae/lib/Watching.js#L36-L39

The above few lines suggest we can give number to watchoption 

```
		if (typeof watchOptions === "number") {
			this.watchOptions = {
				aggregateTimeout: watchOptions
			};
```


EslamHiko:
@jeffin143 it has effects only in CLI, also it's documented here: https://webpack.js.org/configuration/dev-server/#devserverstdin---cli-only

jeffin143:
> @jeffin143 it has effect only in CLI, also it's documented here: https://webpack.js.org/configuration/dev-server/#devserverstdin---cli-only

Thanks :) , for pointing out the location :)

jeffin143:
> Maybe i didnt understand this the right way, but why is example given for another option?

That was a mistake on my part, I will refactor

jeffin143:
@EslamHiko @EugeneHlushko 

I am not sure if this correct, But the condition states that, I am not sure that if we are internally assigning number and that is been handled there with the condtion

src/content/configuration/watch.md
```diff
@@ -50,6 +50,15 @@ module.exports = {
 };
 ```
 
+__webpack.config.js__
```
EugeneHlushko, line 53: this example without explanation doesn't make much sense to me, do you want to explain what happens when you pass in a number?

src/content/configuration/watch.md
```diff
@@ -50,6 +50,17 @@ module.exports = {
 };
 ```
 
+When number is provided instead of an object, `watchOptions.aggregateTimeout` is set to the argument passed.
```
EugeneHlushko, line 53: This is uncommon to webpack docs (saying doing this instead of that), usually we just say `set to x and get these results... __example`
```suggestion
Providing a number to the `watchOptions` sets `watchOptions.aggregateTimeout` to the given number.
```

This way its shorter and clearer to the reader

montogeek:
It is safer :D

EugeneHlushko:
Thanks!

PR #3679, opened 2020-04-06
Break flex items into multiple lines

chenxsan:
Headings look weird currently, take this heading on page https://webpack.js.org/migrate/3/#automatic--loader-module-name-extension-removed for example:

image

Applying `flex-wrap: wrap` will make it look better:

compare


montogeek:
Thanks!

PR #3680, opened 2020-04-07
chore(app): Fix warning for react-banner

chenxsan:
Currently there's error related to react-banner in `development` mode when we input anything in the search box:

image

It exists in `production` too:

image

It is caused by https://github.com/skipjack/react-banner/blob/master/src/banner/banner-search.jsx#L55 calling `onSearch` prop which we do not provide.

This pull request set `onSearch` to an empty function to fix it.


montogeek:
Thanks!

PR #3657, opened 2020-03-29
docs(configuration): fix deprecated apply usage

crimx:
Calling `apply` on Tappable is marked deprecated.

PR #3681, opened 2020-04-08
docs(config) target examples standartized

EugeneHlushko:
- standardized examples of setting a configuration option
- noticed after #3657


montogeek:
Thanks!

PR #3682, opened 2020-04-09
fix(site): update search result style on mobile

chenxsan:
Search results on mobile are almost unstyled, this pull request would improve it:

search-result@3x

And here's a preview https://webpack-7gmqohurp.now.sh/

### DIFF
--- a/src/components/Navigation/Search.scss
+++ b/src/components/Navigation/Search.scss
@@ -12,7 +12,11 @@
 }
 
 // when notification is shown
-.notification-bar + header .navigation-search .algolia-autocomplete .ds-dropdown-menu {
+.notification-bar
+  + header
+  .navigation-search
+  .algolia-autocomplete
+  .ds-dropdown-menu {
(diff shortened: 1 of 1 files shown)


src/components/Navigation/Search.scss
```diff
@@ -50,11 +54,36 @@
 
   .algolia-docsearch-suggestion {
     padding: 0;
+    .algolia-docsearch-suggestion--subcategory-column {
```
montogeek, line 57: Wonder if the notify when they change these classes.

chenxsan: I'm afraid they won't since this seems impossible to be done. But if they stick to semver, we might be safe.

EugeneHlushko:
@montogeek since preview is off we'd sign this off manually, did you try it locally?

montogeek:
@EugeneHlushko Yep!

EugeneHlushko:
Thanks!

PR #3685, opened 2020-04-11
docs(node): added note for provideplugin in node global

anikethsaha:
Added a note for covering query about global variables required by external modules 

ref https://github.com/webpack/webpack/issues/10668

### DIFF
--- a/src/content/configuration/node.md
+++ b/src/content/configuration/node.md
@@ -8,6 +8,7 @@ contributors:
   - Rob--W
   - byzyk
   - EugeneHlushko
+  - anikethsaha
 ---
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/node.md
```diff
@@ -28,6 +29,8 @@ This is an object where each property is the name of a Node global or module and
 
 W> Not every Node global supports all four options. The compiler will throw an error for property-value combinations that aren't supported (e.g. `global: 'empty'`). See the sections below for more details.
 
+T> If you are using a module which needs global variables in it, use `ProvidePlugin` instead of `global`
```
montogeek, line 32:
```suggestion
T> If you are using a module which needs global variables in it, use `ProvidePlugin` instead of `global`.
```

montogeek:
Thanks!

Issue #3686, opened 2020-04-14
Invisible search results


chenxsan:
Search results have some unexpected invisible texts:

![image](https://user-images.githubusercontent.com/1091472/79229146-212c4e00-7e95-11ea-93d7-9305a579e778.png)

I can submit a pull request after https://github.com/webpack/webpack.js.org/pull/3682 is merged, or maybe you want me to push fix into that pull request?

montogeek:
Please create a new PR.

PR #3687, opened 2020-04-14
fix(site): fix style for search result

chenxsan:
This pull request contains two commits, first commit would fix issue https://github.com/webpack/webpack.js.org/issues/3686

compare@3x

While second commit adding borders would visually group suggestions together, 

image

image

I'm not quite sure about the second one which is why I break it into another commit. I can squash them if anyone think it's a good idea to merge it.

### DIFF
--- a/src/components/Navigation/Search.scss
+++ b/src/components/Navigation/Search.scss
@@ -90,15 +90,9 @@
     color: getColor(dusty-grey);
 
     .algolia-docsearch-suggestion--highlight {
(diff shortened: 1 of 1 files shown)


jeffin143:
Great work :tada:

montogeek:
It is fine like this :)

EugeneHlushko:
Yup looks good, you have an eye for ui ;]

EugeneHlushko:
Thanks!

PR #3688, opened 2020-04-14
ui(vote) fix markdown styles squeezing vote layout

EugeneHlushko:
Noticed markdown styles squeezing vote layout, fixed.

Before
Screenshot_2020-04-14 webpack

After
Screenshot_2020-04-14 Vote webpack


anshumanv:
Won't it make more sense to fix this directly in the voting app?

EugeneHlushko:
@anshumanv it is "our side" that adds that padding, webpack.js.org as a consumer breaks it with global padding, therefore fix is on this side as well

anshumanv:
That makes sense, thanks!

PR #3693, opened 2020-04-15
Add more info about options of banner plugin

jeffin143:
https://github.com/webpack/webpack/blob/2346f6b245a92f317535018129260d7c015734c9/declarations/plugins/BannerPlugin.d.ts#L34-L49


src/content/plugins/banner-plugin.md
```diff
@@ -28,9 +28,9 @@ new webpack.BannerPlugin(options);
   banner: string | function, // the banner as string or function, it will be wrapped in a comment
   raw: boolean, // if true, banner will not be wrapped in a comment
   entryOnly: boolean, // if true, the banner will only be added to the entry chunks
-  test: string | RegExp | Array,
-  include: string | RegExp | Array,
-  exclude: string | RegExp | Array,
+  test: string | RegExp | Array, // Include all modules that pass test assertion.
```
EugeneHlushko, line 31: The mechanism is the same, why different phrasing? https://github.com/webpack/webpack/blob/2346f6b245a92f317535018129260d7c015734c9/lib/ModuleFilenameHelpers.js#L164

jeffin143: IMO , only 2 options were defined and thus I wrote for other three .

I agree they do follow the same standard every where
So `include` option every where will mean same thing more or less 

But I guess there is no harm in adding them

EugeneHlushko:
Thanks!

PR #3699, opened 2020-04-17
Guides(Shimming): Correct the resolve file path of rule test

vinzid:
The index.js and globals.js are in the src directory, so the `require.resolve('index.js')` should be `require.resolve('./src/index.js')`, or else it will throw `error: Cannot find module 'index.js'` when built, and so does globals.js.


EugeneHlushko:
Thanks!

Issue #3333, opened 2019-10-25
Writer's guide - Types notation


montogeek:
Some time ago a writes guide was created that specifies how to add types notations to our documentation, but later some users found it confusing, an alternative is to use TypeScript notation which has better standardization process and it is more known by the community.

Another alternative is to use TypeScript documentation from webpack source code directly (it would be a fun AST ride, maybe).

montogeek:
Related: https://github.com/webpack/webpack.js.org/pull/3323

EugeneHlushko:
If this doesnt get traction, i suggest adding a small outline to the types in the docs somewhere or as i suggested earlier, a tooltip when hover types

rishabh3112:
@EugeneHlushko Typescript notations are less confusing and should be used in my opinion.

> Another alternative is to use TypeScript documentation from webpack source code directly (it would be a fun AST ride, maybe).

Major part of the code in the webpack source code in written in javascript with commented types, thus extracting the types from there wouldn't be a nice idea.

chenxsan:
I agree with @rishabh3112 here. We can see the two version here https://github.com/webpack/webpack.js.org/pull/3706#issuecomment-619119074, the TypeScript one is much better. Also, when user reading docs, they won't read the writers guide first to figure out the difference. So it would be better to choose the popular notations.

We might use [`details`/`summary`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) here:

<details>
  <summary>Click to see Options types</summary>
<pre>
type Entry =
  | string
  | [string]
  | {
      key:
        | string
        | [string]
        | {
            import: string | [string]
            dependOn: string | [string]
            filename: string
          }
    }
  | (() =>
      | string
      | [string]
      | {
          key:
            | string
            | [string]
            | {
                import: string | [string]
                dependOn: string | [string]
                filename: string
              }
        })
</pre>
</details>

PR #3706, opened 2020-04-19
docs(plugins): update option type

chenxsan:
I think `(string | RegExp)[]` is accurate than `Array`.

### DIFF
--- a/src/content/plugins/banner-plugin.md
+++ b/src/content/plugins/banner-plugin.md
@@ -3,6 +3,7 @@ title: BannerPlugin
 contributors:
   - simon04
   - byzyk
+  - chenxsan
 related:
   - title: banner-plugin-hashing test
     url: https://github.com/webpack/webpack/blob/master/test/configCases/plugins/banner-plugin-hashing/webpack.config.js
@@ -28,9 +29,9 @@ new webpack.BannerPlugin(options);
   banner: string | function, // the banner as string or function, it will be wrapped in a comment
   raw: boolean, // if true, banner will not be wrapped in a comment
   entryOnly: boolean, // if true, the banner will only be added to the entry chunks
-  test: string | RegExp | Array,
-  include: string | RegExp | Array,
-  exclude: string | RegExp | Array,
+  test: string | RegExp | (string | RegExp)[],
+  include: string | RegExp | (string | RegExp)[],
+  exclude: string | RegExp | (string | RegExp)[],
 }
 ```
 
--- a/src/content/plugins/banner-plugin.md
+++ b/src/content/plugins/banner-plugin.md
@@ -29,9 +29,9 @@ new webpack.BannerPlugin(options);
   banner: string | function, // the banner as string or function, it will be wrapped in a comment
   raw: boolean, // if true, banner will not be wrapped in a comment
   entryOnly: boolean, // if true, the banner will only be added to the entry chunks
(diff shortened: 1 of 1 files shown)


src/content/plugins/banner-plugin.md
```diff
@@ -28,9 +29,9 @@ new webpack.BannerPlugin(options);
   banner: string | function, // the banner as string or function, it will be wrapped in a comment
   raw: boolean, // if true, banner will not be wrapped in a comment
   entryOnly: boolean, // if true, the banner will only be added to the entry chunks
-  test: string | RegExp | Array,
-  include: string | RegExp | Array,
-  exclude: string | RegExp | Array,
+  test: string | RegExp | (string | RegExp)[],
+  include: string | RegExp | (string | RegExp)[],
+  exclude: string | RegExp | (string | RegExp)[],
```
snitin315, line 34:
```suggestion
  test: string | RegExp | [string, RegExp],
  include: string | RegExp | [string, RegExp],
  exclude: string | RegExp | [srting, RegExp],
```

chenxsan: Thanks for the link. But here's my reason, although both are valid in typescript/flow, `[string, RegExp]` means totally different thing from `(string | RegExp)[]`. Now that webpack is going to include [types for typescript](https://github.com/webpack/webpack/pull/10599), I think it's better to follow typescript's conventions.

Maybe we shall update the writers' guide?

jeffin143:
Should we merge #3693 , so that both could be reviewed in one go

I could fetch your branch make changes there , or you could fetch mine 
Feel free to let me know :)

src/content/plugins/banner-plugin.md
montogeek: Interesting, would be great to match source types agaisnt the docs, automatically.

chenxsan:
@jeffin143 Thanks for suggestion, but these two seem to be totally different things although they target the same file, maybe better keep them separated.

chenxsan:
Found some discussions about how to present types in doc here https://github.com/webpack/webpack.js.org/pull/3553/files/7c05c8994023d71fde997dee218774bc65952665#r380507188, and [from that example](https://webpack.js.org/configuration/entry-context/#entry), I would say the current guides https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types is kind of broken when dealing with complex cases:

image

I don't think anyone can figure out this easily.

Any here's the typescript version formatted with prettier:

```typescript
type Entry =
  | string
  | [string]
  | {
      key:
        | string
        | [string]
        | {
            import: string | [string]
            dependOn: string | [string]
            filename: string
          }
    }
  | (() =>
      | string
      | [string]
      | {
          key:
            | string
            | [string]
            | {
                import: string | [string]
                dependOn: string | [string]
                filename: string
              }
        })
```
It's much longer, but it's clean, and much easy to read.

src/content/plugins/banner-plugin.md
EugeneHlushko: Hi @chenxsan 
Its great that you suggest improvements, but currently to express an array of items limited to strings and RegExps we use `[string, RegExp]` in the documentation. We could continue the discussion of changing the types across the docs to be in TypeScript in the existing issue https://github.com/webpack/webpack.js.org/issues/3333

EugeneHlushko:
Thank you!

Issue #3697, opened 2020-04-17
Update Experiments page


hiroppy:
__Feature to document__

Currently, this page is too old. For example, we should add `asset`.

```js
experiments: {
  asset: true
},
```

We need to organize some experimental features and update this documentation for users because it's hard to realize what a feature is experimental.

https://webpack.js.org/configuration/experiments/

__Author(s)__

N/A

__Additional information__

[ ] I am willing to work on this issue and submit a pull request.

PR #3707, opened 2020-04-19
docs(configuration): add more experiments' options

chenxsan:
Add more experiments' options according to https://github.com/webpack/changelog-v5#experiments

closes: https://github.com/webpack/webpack.js.org/issues/3697

### DIFF
--- a/src/content/configuration/experiments.md
+++ b/src/content/configuration/experiments.md
@@ -19,7 +19,7 @@ Available options:
 - `mjs`: Support `.mjs` files as a way to define [EcmaScript modules](https://nodejs.org/api/esm.html#esm_ecmascript_modules)
 - `syncWebAssembly`: Support the old WebAssembly like in webpack 4
 - `asyncWebAssembly`: Support the new WebAssembly according to the [updated specification](https://github.com/WebAssembly/esm-integration), it makes a WebAssembly module an async module
-- `topLevelAwait`: Support [Top Level Await Stage 3 proposal](https://github.com/tc39/proposal-top-level-await), makes the module an async module when `await` is used on top-level
+- `topLevelAwait`: Support the [Top Level Await Stage 3 proposal](https://github.com/tc39/proposal-top-level-await), it makes the module an async module when `await` is used on the top-level
 - `importAsync`: import async modules with `import`
 - `importAwait`: import async modules with `import await`
 - `asset`: a type of module that allows to use asset files (fonts, icons, etc) without configuring additional loaders, similar to `file-loader`|`url-loader`|`raw-loader`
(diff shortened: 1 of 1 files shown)


src/content/configuration/experiments.md
```diff
@@ -17,6 +17,12 @@ W> Because experimental features have relaxed semantic versioning and might cont
 Available options:
 
 - `mjs`: Support `.mjs` files as a way to define [EcmaScript modules](https://nodejs.org/api/esm.html#esm_ecmascript_modules)
+- `syncWebAssembly`: Support old WebAssembly like in webpack 4
```
EugeneHlushko, line 20:
```suggestion
- `syncWebAssembly`: Support the old WebAssembly like in webpack 4
```

src/content/configuration/experiments.md
```diff
@@ -17,6 +17,12 @@ W> Because experimental features have relaxed semantic versioning and might cont
...
+- `asyncWebAssembly`: Support new WebAssembly according to the [updated spec](https://github.com/WebAssembly/esm-integration), makes a WebAssembly module an async module
```
EugeneHlushko, line 21:
```suggestion
- `asyncWebAssembly`: Support the new WebAssembly according to the [updated specification](https://github.com/WebAssembly/esm-integration), it makes a WebAssembly module an async module
```

src/content/configuration/experiments.md
```diff
@@ -17,6 +17,12 @@ W> Because experimental features have relaxed semantic versioning and might cont
...
+- `topLevelAwait`: Support [Top Level Await Stage 3 proposal](https://github.com/tc39/proposal-top-level-await), makes the module an async module when `await` is used on top-level
```
EugeneHlushko, line 22:
```suggestion
- `topLevelAwait`: Support the [Top Level Await Stage 3 proposal](https://github.com/tc39/proposal-top-level-await), it makes the module an async module when `await` is used on the top-level
```

src/content/configuration/experiments.md
```diff
@@ -17,6 +17,12 @@ W> Because experimental features have relaxed semantic versioning and might cont
...
+- `importAsync`: import async modules with `import`
+- `importAwait`: import async modules with `import await`
+- `asset`: a type of module that allows to use asset files (fonts, icons, etc) without configuring additional loaders, similar to `file-loader`|`url-loader`|`raw-loader`
```
EugeneHlushko, line 25: I think icons is the use case of images, better to list actual resource type. Also added spaces near pipes to make it less cryptic for the crawlers
```suggestion
- `asset`: a type of module that allows to use asset files (fonts, images, etc) without configuring loaders to handle their importing, similar to `file-loader` | `url-loader` | `raw-loader`
```

EugeneHlushko:
Thanks!

PR #3711, opened 2020-04-22
docs(migrate): fix webpack version

chenxsan:
The latest webpack 5 is published under `beta` not `next` as far as I see here https://www.npmjs.com/package/webpack?activeTab=versions:

image


chenxsan:
@EugeneHlushko Make sense now.

PR #3713, opened 2020-04-24
Update links behind subscription wall

marcofugaro:
Update the polling link and cache link to the wikipedia ones since the currently linked articles are behind a subscription wall

image


montogeek:
Thanks!

PR #3714, opened 2020-04-24
Docs(config) : add module.unsafeCache

jeffin143:
add module.unsafeCache

https://github.com/webpack/webpack.js.org/issues/2539
https://github.com/webpack/webpack/issues/10771

cc @salemhilal

### DIFF
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -51,7 +51,7 @@ module.exports = {
 
 ## `module.unsafeCache`
 
-`boolean` `function ()`
+`boolean` `function (module)`
 
 Cache the resolution of module requests. There are couple of defaults for `module.unsafeCache`:
 
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -51,9 +51,12 @@ module.exports = {
 
 ## `module.unsafeCache`
 
-`boolean: false` `function()`
+`boolean` `function()`
 
-Cache the resolving of module requests.
+Cache the resolving of module requests. There are couple of defaults for `module.unsafeCache`:
+
+- `false` if cache is disabled.
+- `true` if cache is enabled and the module appears to come from node_modules otherwise `false`.
 
 __webpack.config.js__
 
@@ -65,6 +68,8 @@ module.exports = {
   }
 };
 ```
+
+W> The default value is true in webpack 4.
 
 ## `module.rules`
 
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -56,7 +56,7 @@ module.exports = {
 Cache the resolution of module requests. There are couple of defaults for `module.unsafeCache`:
 
 - `false` if [`cache`](/configuration/other-options/#cache) is disabled.
-- `true` if [`cache`](/configuration/other-options/#cache) is enabled and the module appears to come from node_modules otherwise `false`.
+- `true` if [`cache`](/configuration/other-options/#cache) is enabled and the module appears to come from node modules, `false` otherwise.
 
 __webpack.config.js__
 
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -51,12 +51,12 @@ module.exports = {
 
 ## `module.unsafeCache`
 
-`boolean` `function()`
+`boolean` `function ()`
 
-Cache the resolving of module requests. There are couple of defaults for `module.unsafeCache`:
+Cache the resolution of module requests. There are couple of defaults for `module.unsafeCache`:
 
-- `false` if cache is disabled.
-- `true` if cache is enabled and the module appears to come from node_modules otherwise `false`.
+- `false` if [`cache`](/configuration/other-options/#cache) is disabled.
+- `true` if [`cache`](/configuration/other-options/#cache) is enabled and the module appears to come from node_modules otherwise `false`.
 
 __webpack.config.js__
 
@@ -69,7 +69,7 @@ module.exports = {
 };
 ```
 
-W> The default value is true in webpack 4.
+W> The default value is `true` in webpack 4.
 
 ## `module.rules`
 
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -49,6 +49,22 @@ module.exports = {
 };
 ```
 
+## `module.unsafeCache`
+
+`boolean: false` `function()`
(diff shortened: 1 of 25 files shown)


salemhilal:
That was fast! Thank you. I'm a little unclear about how docs for v4 and v5 are updated separately, but can you mention that:
1. v4's default is `true`, and
2. v5's default is false [unless the module appears to come from node_modules](https://github.com/webpack/webpack/blob/3f9e6d5/lib/config/defaults.js#L287-L294) (and is disabled if `cache` is disabled)?

jeffin143:
> That was fast! Thank you. I'm a little unclear about how docs for v4 and v5 are updated separately, but can you mention that:
> 
> 1. v4's default is `true`, and
> 2. v5's default is false [unless the module appears to come from node_modules](https://github.com/webpack/webpack/blob/3f9e6d5/lib/config/defaults.js#L287-L294) (and is disabled if `cache` is disabled)?

Just had a glance at the source code for deafult option and you are right it is false only if cache is disabled I will updated that

Also we need to raise a pr against v4 branch for the docs update for v4 , if you are ok with me raising the same , I can do it also if you want to handle it , no issues with it :) 

Thanks for digging and coming up with the issue

salemhilal:
It's up to you! I'd be happy to update the v4 docs if you're not eager about opening another PR, but otherwise you're welcome to it.

jeffin143:
> It's up to you! I'd be happy to update the v4 docs if you're not eager about opening another PR, but otherwise you're welcome to it.

Please feel free to open one :)

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,22 @@ module.exports = {
 };
 ```
 
+## `module.unsafeCache`
+
+`boolean: false` `function()`
```
EugeneHlushko, line 54: Formatting is a bit off, also did you mean `false` by default or only `false` is allowed when boolean given?
```suggestion
`boolean: false` `function ()`
```

jeffin143: I meant false by default

EugeneHlushko: ok in that case it should be:
```suggestion
`boolean = false` `function ()`
```

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,27 @@ module.exports = {
...
+```javascript
+module.exports = {
+  //...
+  module: {
+    unsafeCache: false,
+  }
+};
+```
+
+W> The default value is true in webpack 4.
```
EugeneHlushko, line 72:
```suggestion
W> The default value is `true` in webpack 4.
```

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,27 @@ module.exports = {
...
 ```
 
+## `module.unsafeCache`
+
+`boolean` `function()`
+
+Cache the resolving of module requests. There are couple of defaults for `module.unsafeCache`:
+
+- `false` if cache is disabled.
+- `true` if cache is enabled and the module appears to come from node_modules otherwise `false`.
```
EugeneHlushko, line 59: Can you please link the `cache` to a config property in docs

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,27 @@ module.exports = {
 };
 ```
 
+## `module.unsafeCache`
+
+`boolean` `function()`
```
EugeneHlushko, line 54:
```suggestion
`boolean` `function ()`
```

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,27 @@ module.exports = {
...
+
+Cache the resolving of module requests. There are couple of defaults for `module.unsafeCache`:
```
EugeneHlushko, line 56: Should it be `the resolution of` vs `resolving of`, to me, i read like resolving is continuous process, and if you say `the resolution of`, would mean it was resolved and given resolution was cached?

jeffin143: resolution of sounds better

salemhilal:
@EugeneHlushko two thoughts:
1. I'm ok with "the resolution of" personally.
2. Are other things besides the resolution cached, like dependencies?
3. Is it worth mentioning why it's unsafe in the first place?

Perhaps something like this might be a good idea:
> Cache the resolution of module requests, as well as the module's list of dependencies. Note that this feature can cause problems if the module is moved and should resolve to a different location.

salemhilal:
One more nit: should this section go below `noParse` to maintain alphabetical order?

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,27 @@ module.exports = {
 };
 ```
 
+## `module.unsafeCache`
+
+`boolean` `function ()`
```
salemhilal, line 54: I think this can be `function (module)` since `module.unsafeCache` becomes the cache predicate if it's set:
https://github.com/webpack/webpack/blob/e607681/lib/NormalModuleFactory.js#L163-L166

which in turn gets passed a copy of the module: 
https://github.com/webpack/webpack/blob/e607681/lib/NormalModuleFactory.js#L561

src/content/configuration/module.mdx
```diff
@@ -49,6 +49,27 @@ module.exports = {
...
+
+Cache the resolution of module requests. There are couple of defaults for `module.unsafeCache`:
+
+- `false` if [`cache`](/configuration/other-options/#cache) is disabled.
+- `true` if [`cache`](/configuration/other-options/#cache) is enabled and the module appears to come from node_modules otherwise `false`.
```
salemhilal, line 59: Nit: do you mind changing this to:
> ...come from node modules,  **`false` otherwise.**

salemhilal:
@jeffcooper86 I left a bunch of comments and opened https://github.com/webpack/webpack.js.org/pull/3728 to mirror these changes to the v4 docs.

src/content/configuration/module.mdx
EugeneHlushko: Yes good point! 
```suggestion
`boolean` `function (module)`
```

src/content/configuration/module.mdx
EugeneHlushko:
```suggestion
- `true` if [`cache`](/configuration/other-options/#cache) is enabled and the module appears to come from node modules, `false` otherwise.
```

EugeneHlushko:
Thanks!

PR #3724, opened 2020-05-02
fix(site): vertically align notification bar

chenxsan:
Notification bar is not vertically aligned on desktop:

align-center-vertically


EugeneHlushko:
Thanks!

PR #3725, opened 2020-05-02
docs: update argument name

chenxsan:
Not an issue at all, but I think would be an improvement if we can use the same name as types defined for typescript https://github.com/webpack/webpack/blob/082fce5000a5bdf0fbef99acc604d1c57eef511d/declarations/WebpackOptions.d.ts#L61

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -239,7 +239,7 @@ For example, if you have 2 libraries, with namespaces `library1` and `library2`,
 
 ## `output.filename`
 
-`string` `function (pathData) => string`
+`string` `function (pathData, assetInfo) => string`
 
 This option determines the name of each output bundle. The bundle is written to the directory specified by the [`output.path`](#outputpath) option.
 
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -22,6 +22,7 @@ contributors:
   - anikethsaha
   - jamesgeorge007
   - hiroppy
+  - chenxsan
 ---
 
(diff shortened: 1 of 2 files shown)


src/content/configuration/output.md
```diff
@@ -238,7 +239,7 @@ For example, if you have 2 libraries, with namespaces `library1` and `library2`,
 
 ## `output.filename`
 
-`string` `function (chunkData) => string`
+`string` `function (pathData) => string`
```
montogeek, line 242: This is missing `assetInfo` second parameter, could you please add it?

src/content/plugins/split-chunks-plugin.md
```diff
@@ -401,7 +402,7 @@ module.exports = {
 
 #### `splitChunks.cacheGroups.{cacheGroup}.filename`
 
-`string` `function (chunkData): string`
+`string` `function (pathData): string`
```
montogeek, line 405: Here too

src/content/plugins/split-chunks-plugin.md
```diff
@@ -401,7 +402,7 @@ module.exports = {
 
 #### `splitChunks.cacheGroups.{cacheGroup}.filename`
 
-`string` `function (chunkData): string`
+`string` `function (pathData, assetInfo): string`
```
EugeneHlushko, line 405:
```suggestion
`string` `function (pathData, assetInfo) => string`
```

EugeneHlushko:
Thanks!

PR #3728, opened 2020-05-03
Docs(config) : add module.unsafeCache to v4 docs

salemhilal:
This PR mirrors #3714 for the v4 documentation.
It addresses the following issues:
- #2539
- webpack/webpack#10771

### DIFF
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -14,6 +14,7 @@ contributors:
   - EugeneHlushko
   - superburrito
   - lukasgeiter
+  - jeffin143
   - salemhilal
 ---
 
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -14,6 +14,7 @@ contributors:
   - EugeneHlushko
   - superburrito
   - lukasgeiter
+  - salemhilal
 ---
 
 These options determine how the [different types of modules](/concepts/modules) within a project will be treated.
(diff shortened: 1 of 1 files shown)


src/content/configuration/module.mdx
```diff
@@ -53,6 +55,28 @@ module.exports = {
...
+```javascript
+module.exports = {
+  //...
+  module: {
+    unsafeCache: false,
+  }
+};
+```
+
+W> In Webpack 5, the default value is `true` if the `cache` option is enabled _and_ the module appears to come from the node_modules directory.
```
montogeek, line 75:
```suggestion
W> In webpack 5, the default value is `true` if the `cache` option is enabled _and_ the module appears to come from the node_modules directory.
```

Although this documentation is for webpack 5 by default, so the message should be for v4 instead.

salemhilal: You had me worried!

src/content/configuration/module.mdx
```diff
@@ -53,6 +55,28 @@ module.exports = {
...
+  module: {
+    unsafeCache: false,
+  }
+};
+```
+
+W> In webpack 5, the default value is `true` if the `cache` option is enabled _and_ the module appears to come from the node_modules directory.
+
+
+
```
EugeneHlushko, line 78: need to clean up unnecessary rows

salemhilal: Done

EugeneHlushko:
Thanks!

PR #3729, opened 2020-05-04
chore(site): Add Vercel logo to footer

montogeek:
Vercel has given us a Pro account that we have used since ~3 weeks ago, now we add a small banner to our footer that link back to them, it is also added to the README file.


EugeneHlushko:
Still wondering if we should do the footer link or just readme link (since actual hosting for prod is GH pages) @sokra wdyt?

montogeek:
Yeah, not sure, easier to remove code than add it :)

sokra:
I think we should not falsy claim that it's powered by vercel, for the pages that are really powered by gh-pages. This might gives negative PR to vercel, e. g. in causes when GitHub performs worse.

So only add that logo when it's true.

src/components/Footer/Footer.jsx
```diff
@@ -4,6 +4,7 @@ import Container from '../Container/Container';
 import Icon from '../../assets/icon-square-small.svg';
 import CC from '../../assets/cc.svg';
 import BY from '../../assets/by.svg';
+import Vercel from '../../assets/powered-by-vercel.svg';
```
EugeneHlushko, line 7:
```suggestion
```

src/components/Footer/Footer.jsx
```diff
@@ -36,6 +37,9 @@ const Footer = () => (
             alt="Creative Commons License"
             src={ BY } />
         </Link>
+        <Link className="footer__link footer__license" to="https://vercel.com/?utm_source=webpackdocs">
+          <img src={ Vercel } alt="Powered by Vercel" />
+        </Link>
```
EugeneHlushko, line 42:
```suggestion
```

EugeneHlushko:
Thanks!

PR #3732, opened 2020-05-05
docs(concepts): add module federation

chenxsan:
Close https://github.com/webpack/webpack.js.org/issues/3731

### DIFF
--- a/src/content/concepts/module-federation.md
+++ b/src/content/concepts/module-federation.md
@@ -16,43 +16,43 @@ This is often known as Micro-Frontends, but is not limited to that.
 
 ## Low-level concepts
 
-We distinguish between local and remote modules. Local modules are normal modules which are part of the current build. Remote modules are modules that are not part of the current build and loaded from a so called container at runtime.
+We distinguish between local and remote modules. Local modules are normal modules which are part of the current build. Remote modules are modules that are not part of the current build and loaded from a so-called container at the runtime.
 
 Loading remote modules is considered as async operation. When using a remote module these async operation will be placed in the next chunk loading operation(s) that is between the remote module and the entrypoint. It's not possible to use a remote module without a chunk loading operation.
 
-A chunk loading operation is usually an `import()`, but older constructs like `require.ensure` or `require([...])` are possible as well.
+A chunk loading operation is usually an `import()` call, but older constructs like `require.ensure` or `require([...])` are supported as well.
 
-A container is created through a container entry, which exposes async access to specified modules. The exposed access is separated into two steps:
+A container is created through a container entry, which exposes asynchronous access to the specified modules. The exposed access is separated into two steps:
 
-1. loading the module (async)
-2. evaluating the module (sync).
+1. loading the module (asynchronous)
+2. evaluating the module (synchronous).
 
-Step 1 will be done during chunk loading. Step 2 will be done during module evaluation interleaved with other (local and remote) modules. This way evaluation order is unaffected by converting a module from local to remote or the other way around.
+Step 1 will be done during the chunk loading. Step 2 will be done during the module evaluation interleaved with other (local and remote) modules. This way, evaluation order is unaffected by converting a module from local to remote or the other way around.
 
-Nesting container is possible. Containers can use modules from other containers. Circular dependencies between container are also possible.
+It is possible to nest a container. Containers can use modules from other containers. Circular dependencies between container are also possible.
 
 ### Overriding
 
 A container is able to flag selected local modules as "overridable". A consumer of the container is able to provide "overrides", which are modules that replace one of the overridable module of the container. All modules of the container will use the replacement module instead of the local module when the consumer provides one. When the consumer doesn't provide a replacement module, all modules of the container will use the local one.
 
-The container will organize overridable modules in a way that they do not need to be downloaded when they has been overriding by the consumer. This usually happens by placing them into separate chunks.
+The container will organize overridable modules in a way that they do not need to be downloaded when they have been overridden by the consumer. This usually happens by placing them into separate chunks.
 
-One the other hand the provider of the replacement modules will only provide async loading functions which allow the container to load replacement modules only when they are needed. The provider will organize replacement modules in a way that they do not need to be downloaded when they are not requested by the container at all. This usually happens by placing them into separate chunks.
+On the other hand, the provider of the replacement modules, will only provide asynchronous loading functions. It allows the container to load replacement modules only when they are needed. The provider will organize replacement modules in a way that they do not need to be downloaded at all when they are not requested by the container. This usually happens by placing them into separate chunks.
 
 A "name" is used to identify overridable modules from the container.
 
 Overrides are provided in a similar way as the container exposes modules, separated into two steps:
 
-1. Loading (async)
-2. evaluating (sync).
+1. Loading (asynchronous)
+2. evaluating (asynchronous)
 
-In nested scenarios, where a container uses other containers to load modules from, is overriding a transitive operation. Providing overrides to one container will also override these modules in nested container. This means that using a container will automatically make all "names" of the overridable modules, part of the overriding interface of the parent container.
+W> When nesting is used, providing overrides to one container will automatically override the modules with the same "name" in the nested container(s).
 
-Overrides must be provided before modules of the container are loaded, otherwise behavior is undefined.
+Overrides must be provided before the modules of the container are loaded, otherwise behavior is undefined.
 
 ## High-level concepts
 
-Each build acts as container and also consumes other build as containers. This way each build is able to access any other exposed module by loading it from its container.
+Each build acts as container and also consumes other build as container. This way each build is able to access any other exposed module by loading it from its container.
 
 Shared modules are modules that are both overridable and provided as overrides to nested container. They usually point to the same module in each build, e.g. the same library.
 
@@ -68,7 +68,7 @@ This plugin creates an additional container entry with the specified exposed mod
 
 ### `ContainerReferencePlugin` (low level)
 
-This plugin adds specified references to containers as externals and allows to import remote modules from these containers. It also calls the `override` API of these containers to provide overrides to them. Local overrides (via `__webpack_override__` or `override` API when build is also a container) plus specified overrides are provided to all referenced containers.
+This plugin adds specified references to containers as externals and allows to import remote modules from these containers. It also calls the `override` API of these containers to provide overrides to them. Local overrides (via `__webpack_override__` or `override` API when build is also a container) and specified overrides are provided to all referenced containers.
 
 ### `ModuleFederationPlugin` (high level)
 
@@ -76,13 +76,13 @@ This plugin combines `ContainerPlugin` and `ContainerReferencePlugin`. Overrides
 
 ## Concept goals
 
-- Any module type supported by webpack should be possible to expose and use.
-- Chunk loading should load everything needed in parallel. (Web: single round-trip to server)
+- It should be possible to expose and use any module type that webpack supports
+- Chunk loading should load everything needed in parallel (web: single round-trip to server)
 - Control from consumer to container
     - Overriding modules is a one-directional operation
     - Sibling containers cannot override each other's modules
 - Concept should be environment-independent
-    - Usable in web, node.js, etc.
+    - Usable in web, Node.js, etc
 
 ## Use cases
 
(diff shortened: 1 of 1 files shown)


src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Motivation
+
+Multiple separate builds should form a single application. These separate builds should not have build dependencies between each other, so they can be developed and deployed individually.
+
+This is often known as Micro-Frontends, but is not limited to that.
+
+## Low-level concepts
+
+We distinguish between local and remote modules. Local modules are normal modules which are part of the current build. Remote modules are modules that are not part of the current build and loaded from a so called container at runtime.
```
EugeneHlushko, line 19:
```suggestion
We distinguish between local and remote modules. Local modules are normal modules which are part of the current build. Remote modules are modules that are not part of the current build and loaded from a so-called container at the runtime.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+Loading remote modules is considered as async operation. When using a remote module these async operation will be placed in the next chunk loading operation(s) that is between the remote module and the entrypoint. It's not possible to use a remote module without a chunk loading operation.
```
EugeneHlushko, line 21:
```suggestion
Loading remote modules is considered an asynchronous operation. When using a remote module, these asynchronous operations, will be placed in the next chunks loading operation(s). That chunk is between the remote module and the entrypoint. It's not possible to use a remote module without a chunk loading operation.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+A chunk loading operation is usually an `import()`, but older constructs like `require.ensure` or `require([...])` are possible as well.
```
EugeneHlushko, line 23:
```suggestion
A chunk loading operation is usually an `import()` call, but older constructs like `require.ensure` or `require([...])` are supported as well.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+A container is created through a container entry, which exposes async access to specified modules. The exposed access is separated into two steps:
```
EugeneHlushko, line 25:
```suggestion
A container is created through a container entry, which exposes asynchronous access to the specified modules. The exposed access is separated into two steps:
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. loading the module (async)
```
EugeneHlushko, line 27:
```suggestion
1. loading the module (asynchronous)
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+2. evaluating the module (sync).
```
EugeneHlushko, line 28:
```suggestion
2. evaluating the module (synchronous).
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+Step 1 will be done during chunk loading. Step 2 will be done during module evaluation interleaved with other (local and remote) modules. This way evaluation order is unaffected by converting a module from local to remote or the other way around.
```
EugeneHlushko, line 30:
```suggestion
Step 1 will be done during the chunk loading. Step 2 will be done during the module evaluation interleaved with other (local and remote) modules. This way, evaluation order is unaffected by converting a module from local to remote or the other way around.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+Nesting container is possible. Containers can use modules from other containers. Circular dependencies between container are also possible.
```
EugeneHlushko, line 32:
```suggestion
It is possible to nest a container. Containers can use modules from other containers. Circular dependencies between container are also possible.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Overriding
+
+A container is able to flag selected local modules as "overridable". A consumer of the container is able to provide "overrides", which are modules that replace one of the overridable module of the container. All modules of the container will use the replacement module instead of the local module when the consumer provides one. When the consumer doesn't provide a replacement module, all modules of the container will use the local one.
+
+The container will organize overridable modules in a way that they do not need to be downloaded when they has been overriding by the consumer. This usually happens by placing them into separate chunks.
```
EugeneHlushko, line 38:
```suggestion
The container will organize overridable modules in a way that they do not need to be downloaded when they have been overridden by the consumer. This usually happens by placing them into separate chunks.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+One the other hand the provider of the replacement modules will only provide async loading functions which allow the container to load replacement modules only when they are needed. The provider will organize replacement modules in a way that they do not need to be downloaded when they are not requested by the container at all. This usually happens by placing them into separate chunks.
```
EugeneHlushko, line 40:
```suggestion
On the other hand, the provider of the replacement modules, will only provide asynchronous loading functions. It allows the container to load replacement modules only when they are needed. The provider will organize replacement modules in a way that they do not need to be downloaded at all when they are not requested by the container. This usually happens by placing them into separate chunks.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+A "name" is used to identify overridable modules from the container.
+
+Overrides are provided in a similar way as the container exposes modules, separated into two steps:
+
+1. Loading (async)
```
EugeneHlushko, line 46:
```suggestion
1. Loading (asynchronous)
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+2. evaluating (sync).
```
EugeneHlushko, line 47:
```suggestion
2. evaluating (asynchronous)
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+In nested scenarios, where a container uses other containers to load modules from, is overriding a transitive operation. Providing overrides to one container will also override these modules in nested container. This means that using a container will automatically make all "names" of the overridable modules, part of the overriding interface of the parent container.
```
EugeneHlushko, line 49: This paragraph is pretty complex. Can we make this much much easier to understand?
```suggestion
W> When nesting is used, providing overrides to one container will automatically override the modules with the same "name" in the nested container(s).
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+Overrides must be provided before modules of the container are loaded, otherwise behavior is undefined.
```
EugeneHlushko, line 51: Although, what do you mean "behaviour is undefined"
```suggestion
Overrides must be provided before the modules of the container are loaded, otherwise behavior is undefined.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+## High-level concepts
+
+Each build acts as container and also consumes other build as containers. This way each build is able to access any other exposed module by loading it from its container.
```
EugeneHlushko, line 55:
```suggestion
Each build acts as container and also consumes other build as container. This way each build is able to access any other exposed module by loading it from its container.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+Shared modules are modules that are both overridable and provided as overrides to nested container. They usually point to the same module in each build, e.g. the same library.
+
+## Building blocks
+
+### `OverridablesPlugin` (low level)
+
+This plugin makes specified modules "overridable". A local API (`__webpack_override__`) allows to provide overrides.
```
EugeneHlushko, line 63: This needs an example, we can skip it in this PR but nice to have.

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+### `ContainerPlugin` (low level)
+
+This plugin creates an additional container entry with the specified exposed modules. It also uses the `OverridablesPlugin` internally and exposes the `override` API to consumer of the container.
+
+### `ContainerReferencePlugin` (low level)
+
+This plugin adds specified references to containers as externals and allows to import remote modules from these containers. It also calls the `override` API of these containers to provide overrides to them. Local overrides (via `__webpack_override__` or `override` API when build is also a container) plus specified overrides are provided to all referenced containers.
```
EugeneHlushko, line 71:
```suggestion
This plugin adds specified references to containers as externals and allows to import remote modules from these containers. It also calls the `override` API of these containers to provide overrides to them. Local overrides (via `__webpack_override__` or `override` API when build is also a container) and specified overrides are provided to all referenced containers.
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+### `ModuleFederationPlugin` (high level)
+
+This plugin combines `ContainerPlugin` and `ContainerReferencePlugin`. Overrides and overridables are combined into a single list of specified shared modules.
+
+## Concept goals
+
+- Any module type supported by webpack should be possible to expose and use.
```
EugeneHlushko, line 79:
```suggestion
- It should be possible to expose and use any module type that webpack supports
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+- Chunk loading should load everything needed in parallel. (Web: single round-trip to server)
```
EugeneHlushko, line 80:
```suggestion
- Chunk loading should load everything needed in parallel (web: single round-trip to server)
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+- Control from consumer to container
+    - Overriding modules is a one-directional operation
+    - Sibling containers cannot override each other's modules
+- Concept should be environment-independent
+    - Usable in web, node.js, etc.
```
EugeneHlushko, line 85:
```suggestion
    - Usable in web, Node.js, etc
```

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Use cases
+
+### Separate builds per page
+
+Each page of a Single Page Application is exposed from container build in a separate build. The application shell is also a separate build referencing all pages as remote modules. This way each page can be separately deployed. The application shell is deployed when routes are updated or new routes are added. The application shell defines commonly used libraries as shared modules to avoid duplication of them in the page builds.
+
+### Components library as container
+
+Many applications share a common components library which could be built as container with each component exposed. Each application consumes components from the components library container. Changes to the components library can be separately deployed without the need to re-deploy all applications. The application automatically use the up-to-date version of the components library.
```
EugeneHlushko, line 95:
```suggestion
Many applications share a common components library which could be built as a container with each component exposed. Each application consumes components from the components library container. Changes to the components library can be separately deployed without the need to re-deploy all applications. The application automatically use the up-to-date version of the components library.
```

EugeneHlushko:
@chenxsan i know its draft but was interesting to read upfront. This is a very import addition

src/content/concepts/module-federation.md
chenxsan: It's the **next chunks loading operation(s)** that is between the remote module and entrypoint, or **chunk** itself?

src/content/concepts/module-federation.md
```diff
@@ -0,0 +1,95 @@
...
+A "name" is used to identify overridable modules from the container.
+
+Overrides are provided in a similar way as the container exposes modules, separated into two steps:
+
+1. Loading (asynchronous)
+2. evaluating (asynchronous)
+
+W> When nesting is used, providing overrides to one container will automatically override the modules with the same "name" in the nested container(s).
+
+Overrides must be provided before the modules of the container are loaded, otherwise behavior is undefined.
```
chenxsan, line 51: Maybe we can ask sokra for help about this `undefined` thing here? Or just remove `otherwise behavior is undefined`.

src/content/concepts/module-federation.md
chenxsan: I've added one.

src/content/concepts/module-federation.md
EugeneHlushko: yes lets remove for now

EugeneHlushko:
Oh right, it lacks sort, ima follow up

PR #3733, opened 2020-05-07
docs(plugins): clean option

chenxsan:
The `requestTimeout` option was removed in this commit https://github.com/webpack/webpack/commit/9d918615922f8e7fb8ab1c9712135908cf1cadd7#diff-8941681d920ac4feda44f522b986c2d0L99. There're only `multiStep` and `fullBuildTimeout` left as we can see here https://github.com/webpack/webpack/blob/master/types.d.ts#L2599


EugeneHlushko:
Thanks!

PR #3742, opened 2020-05-14
Fix name of dev-tool name in description

chumbalum:
The used value "cheap-module-eval-source-map" doesn't exist (maybe anymore) and "eval-cheap-module-source-map" should be used instead

https://webpack.js.org/configuration/devtool/


EugeneHlushko:
Thanks!

PR #3744, opened 2020-05-17
Update entry-points.md

sagarsoni7:
Update Concept (documentation): We are telling webpack that we would like an output file with filename: [name].bundle.js


EugeneHlushko:
Thanks!

PR #3750, opened 2020-05-23
Update index.md

R45H:


R45H:
> There's guide here https://webpack.js.org/branding/ about the brand name. Seems to be done on purpose.

Oh, sorry. I didn't know that 🙈

PR #3751, opened 2020-05-23
fix(site): Vertical alignment of "code" and "tt" elements

R45H:


montogeek:
Before/after looks the same, that CSS is no necessary then?

chenxsan:
I made a gif to see it more clearly:

ezgif com-video-to-gif

The default value of `vertical-align` is `baseline` which is different from `middle`.

EugeneHlushko:
Looks good, thanks!

PR #3752, opened 2020-05-24
fix(site): add idle word to cache.store

wxy1995:


montogeek:
@wxy1995 Thanks for your contribution, Could you please link the source code for this change? Thanks!

PR #3767, opened 2020-06-03
feat(optimize) disable bundlesize to fix prod deploy

EugeneHlushko:
Could be related: https://github.com/siddharthkp/bundlesize/issues/93

Master doesn't get deployed, until we figure out proper fix lets just disable it

ref: https://travis-ci.org/github/webpack/webpack.js.org/jobs/694100391

```
$ bundlesize && npm run sitemap
 ERROR  Could not add github status.
        401: Bad credentials 

```


montogeek:
Yeah, we also don't add a lot of code anyways :)

PR #3768, opened 2020-06-03
UI: Fix word-break on headings

ThomasOrlita:
On smaller devices headings do not wrap correctly:
image

After setting `word-break: break-word`:
image


montogeek:
What viewport resolution specifically?

ThomasOrlita:
On my phone it's 360x560, when I checked it in devtools it starts wrapping when the width is <=408px

EugeneHlushko:
thanks!

PR #3793, opened 2020-06-19
add links to navigate between pages

chenxsan:
Sometimes I would read document on mobile phone and navigating between pages are not an easy task when you have to open sidebar menu first. This pull request adds links at the bottom of page for navigating to previous/next page, and I think it would improve users' experience.

Screenshot:
image

### DIFF
--- a/src/components/Page/AdjacentPages.jsx
+++ b/src/components/Page/AdjacentPages.jsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import Link from '../Link/Link';
+import './AdjacentPages.scss';
+import PropTypes from 'prop-types';
+AdjacentPages.propTypes = {
+  previous: PropTypes.shape({
+    url: PropTypes.string,
+    title: PropTypes.string,
+  }),
+  next: PropTypes.shape({
+    url: PropTypes.string,
+    title: PropTypes.string,
+  }),
+};
+
+export default function AdjacentPages({ previous, next }) {
+  return (
+    <div className="adjacent-links">
+      {previous && (
+        <div className="adjacent-links__prev">
+          <div>« Previous</div>
+          <Link className="adjacent-links__link" to={previous.url}>
+            {previous.title}
+          </Link>
+        </div>
+      )}
+      {next && (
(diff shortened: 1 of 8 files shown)


montogeek:
Great addition, please revert formatting changes, we haven't decide about it and it is just noise in this PR.

chenxsan:
@montogeek reverted.

src/components/Page/AdjacentPages.jsx
```diff
@@ -0,0 +1,37 @@
...
+  }),
+  next: PropTypes.shape({
+    url: PropTypes.string,
+    title: PropTypes.string,
+  }),
+};
+
+export default function AdjacentPages({ previous, next }) {
+  return (
+    <div className="adjacent-links">
```
EugeneHlushko, line 18: Please add an if here to return `null` if none previous or next is defined given this container will add empty space to the page if rendered empty:
```scss
.adjacent-links {
  display: flex;
  padding: 30px 10px;
  justify-content: space-between;
```

src/utilities/get-adjacent-pages/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = function getAdjacentPages(haystack, needle, by = 'url') {
```
EugeneHlushko, line 1:
```suggestion
export default function getAdjacentPages(haystack, needle, by = 'url') {
```

src/utilities/get-adjacent-pages/index.js
```diff
@@ -0,0 +1,13 @@
...
+  let previous = undefined,
+    next = undefined;
```
EugeneHlushko, line 3:
```suggestion
  let previous, next;
```

src/utilities/get-adjacent-pages/index.js
chenxsan: Updated. I had to update `.babelrc` file too for test.

EugeneHlushko:
Thanks! Looking forward to hear positive feedback from the readers :)

PR #3808, opened 2020-06-30
docs(guides): Removed unnecessary publicPath option

jeean:
In the Dynamic Imports section inside Code-Splitting Guide,
`output.publicPath` option isn't always required to use `output.chunkFilename`.


EugeneHlushko:
Thanks!

PR #3799, opened 2020-06-25
clean deprecated bundle-loader

chenxsan:
closes #3798

PR #3809, opened 2020-06-30
docs(concepts): remove the instruction that consumes bundle-loader

jamesgeorge007:
Follow up of #3799

### DIFF
--- a/src/content/concepts/module-federation.md
+++ b/src/content/concepts/module-federation.md
@@ -182,7 +182,7 @@ __bootstrap.js__
 + ReactDOM.render(<App />, document.getElementById('root'));
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/concepts/module-federation.md
```diff
@@ -182,7 +182,7 @@ __bootstrap.js__
 + ReactDOM.render(<App />, document.getElementById('root'));
 ```
 
-Methods mentioned below work, but can have some limits or drawbacks.
+The below mentioned method works fine, but can have some limitations or drawbacks.
```
EugeneHlushko, line 185:
```suggestion
This method works but can have limitations or drawbacks.
```

EugeneHlushko:
Thanks!

PR #3813, opened 2020-07-01
move build script after cypress

chenxsan:
`cypress:ci` would start the server with `yarn start` script which would remove `dist` folder causing `Post-build` stage to fail.


EugeneHlushko:
Thanks!

PR #3816, opened 2020-07-03
docs(guides): use tag `next`

chenxsan:
1. `webpack` beta was [published under tag `next`](https://www.npmjs.com/package/webpack?activeTab=versions).
2. We would like users to install `webpack` under `devDependencies` instead of `dependencies`.

### DIFF
--- a/src/content/guides/installation.md
+++ b/src/content/guides/installation.md
@@ -7,6 +7,7 @@ contributors:
   - simon04
   - EugeneHlushko
   - sibiraj-s
+  - chenxsan
 ---
 
 This guide goes through the various methods used to install webpack.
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
@chenxsan please elaborate on `--save-dev`

chenxsan:
@EugeneHlushko Do you have question about using `--save-dev` here or do you want me to add some comments for it?

EugeneHlushko:
Yes either way it is not clear right now why we are suggesting --save-dev by default. There is obviously a use case for both setups but if you think we should default to being a dev dependency should probably explain why

chenxsan:
@EugeneHlushko I've added a tip about `--save-dev`.

I think most people use `webpack` for app bundling although there're cases like [`next.js`](https://github.com/vercel/next.js/blob/eecc73702b5500ab1476faedacaadf5623fd50a8/packages/next/package.json#L110) saving `webpack` under `dependencies`.

Also we already have `--save-dev` at the beginning of this page:

image

It would be better to keep consistent with that.

EugeneHlushko:
Thanks!

PR #3602, opened 2020-03-06
Docs(config) : defaults to false in node.js targets

jeffin143:
fix - #3057 

- [x] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [x] Make sure your PR complies with the [writer's guide][2].
- [x] Review the diff carefully as sometimes this can reveal issues.
- [x] Do not abandon your Pull Request: [Stale Pull Requests][3].

PR #3817, opened 2020-07-03
Fix node defaults

jaylinski:
This PR fixes the documented node defaults.

See https://github.com/webpack/webpack/blob/v5.0.0-beta.20/lib/config/defaults.js#L478-L510 for the actual node default configs.

### Timeline

* Node defaults were restored in PR https://github.com/webpack/webpack/pull/8602
* Documentation was updated accordingly in https://github.com/webpack/webpack.js.org/pull/2753
* Node defaults were changed for targets  `node`, `async-node` and `electron-main` in https://github.com/webpack/webpack/pull/9172
* Documentation was wrongly updated for the non-default targets `node`, `async-node` and `electron-main`, but should reflect the default target `web` https://github.com/webpack/webpack.js.org/pull/3602

Relates to https://github.com/webpack/webpack/pull/11054.

I would much prefer to disable node for all targets, but @sokra explained why it is not done in https://github.com/webpack/webpack/pull/11054#issuecomment-645673452.

### DIFF
--- a/src/content/configuration/node.md
+++ b/src/content/configuration/node.md
@@ -50,14 +50,18 @@ Since webpack 3.0.0, the `node` option may be set to `false` to completely turn 
 
 ## `node.global`
 
-`boolean = false`
+`boolean = true`
+
+Defaults to `false` for [targets](https://webpack.js.org/configuration/target/) `node`, `async-node` and `electron-main`.
 
 See [the source](https://nodejs.org/api/globals.html) for the exact behavior of this object.
 
 
 ## `node.__filename`
 
-`string` `boolean = false`
+`boolean` `string = mock`
+
+Defaults to `false` for [targets](https://webpack.js.org/configuration/target/) `node`, `async-node` and `electron-main`.
 
 Options:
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/node.md
```diff
@@ -50,14 +50,18 @@ Since webpack 3.0.0, the `node` option may be set to `false` to completely turn
 
 ## `node.global`
 
-`boolean = false`
+`boolean = true`
+
+Is `false` for [targets](https://webpack.js.org/configuration/target/) `node`, `async-node` and `electron-main`.
```
EugeneHlushko, line 55: Is it defaulting to `false` for these targets, or is it just `false` and you cannot override? If its defaulting to but still can be overridden by setting it to a different value, we should say so explicitly for all three occasions like so:
```suggestion
Defaults to `false` for [targets](https://webpack.js.org/configuration/target/) `node`, `async-node` and `electron-main`.
```

jaylinski: It is defaulting to `false` and will be overridden when specified in the webpack-config. Your suggested wording is more explicit, I added it to the commit. :+1:

EugeneHlushko: thank you! i didnt have time to double check so just asked :D

EugeneHlushko:
Thanks!

PR #3832, opened 2020-07-11
docs(configuration): fix heading level

chenxsan:
`infrastructureLogging` should be heading 2 instead of heading 3.

## Before

image

## After

image

### DIFF
--- a/src/content/configuration/other-options.md
+++ b/src/content/configuration/other-options.md
@@ -399,7 +399,7 @@ Options for infrastructure level logging.
 
 `object = {}`
 
-### infrastructureLogging.level
+### level
 
 `string`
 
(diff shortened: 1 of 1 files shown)


montogeek:
Would be good to strip the first part of the sub sub heading if the name is too long, see for example in your second screenshot that you can't really read the real option name.

montogeek:
I can't ignore this CLA bot :D, it seems its resources are too low to process its state correctly or maybe it is using something old to check for statutes.

chenxsan:
@montogeek updated.

image

EugeneHlushko:
In this case its fine, but i assume there can be cases where we wont do this for SEO reasons

EugeneHlushko:
Thanks!

PR #3849, opened 2020-07-16
docs(configuration): mention jsonp

chenxsan:
https://github.com/webpack/webpack/blob/f4b5555691cef31127e7ab9d7a998ad4f547368a/declarations/WebpackOptions.d.ts#L134

 > * Specifies the default type of externals ('amd*', 'umd*', 'system' and 'jsonp' depend on output.libraryTarget set to the same value).


EugeneHlushko:
Thanks!

PR #3824, opened 2020-07-08
merge is not a function

makuzaverite:
### Since webpack merge package has been updated to use destructure

Doc example here https://webpack.js.org/guides/production/#setup

was giving the following error 

image

Solution was simply to edit and add destructure to merge package imports like this
solution

Hope it Helps!

PR #3854, opened 2020-07-17
Changed the way to load the merge function

jkneepkens:
By using a named import, the merge function will work. Now you get a `TypeError: merge is not a function” in webpack config`.


chenxsan:
FYI, there is a similar pull request here https://github.com/webpack/webpack.js.org/pull/3824

PR #3865, opened 2020-07-20
fix(config) Use relative URLs in links

montogeek:
Follow up of https://github.com/webpack/webpack.js.org/pull/3817


EugeneHlushko:
Thanks!

PR #3871, opened 2020-07-22
docs(config): Clarify `mode` option copy

mattsacks:
Just inserting a code diff doesn't effectively communicate what options are being configured. This PR updates the `mode` configuration option docs to better clarify what options get set via `mode` and how to override them.


montogeek:
Thanks for making it clear with comments, but we have found that users pay more attention to the code examples, this is why the entire documentation, specifically the guides shows what is different between steps using diffs.

mattsacks:
@montogeek Makes sense given it's what I've seen used elsewhere in the docs now that I've spent some more time with them. Thanks for giving this a look!

PR #3875, opened 2020-07-23
clearify which 'index.html' do we mean here

m98:
As there is another `index.html` in the project root, and another one in `dist`, and in the document, we ask the user to open `index.html` after running the `npx webpack`, so we want the user to open the `index.html` which is inside `dist` directory. This just clarify which `index.html` do we mean to open

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -202,7 +202,7 @@ You can also set it to 'none' to disable any default behavior. Learn more: https
 
 T> Your output may vary a bit, but if the build is successful then you are good to go. Also, don't worry about the warning, we'll tackle that later.
 
-Open `dist/index.html` in your browser and, if everything went right, you should see the following text: 'Hello webpack'.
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.md
```diff
@@ -202,7 +202,7 @@ You can also set it to 'none' to disable any default behavior. Learn more: https
 
 T> Your output may vary a bit, but if the build is successful then you are good to go. Also, don't worry about the warning, we'll tackle that later.
 
-Open `index.html` in your browser and, if everything went right, you should see the following text: 'Hello webpack'.
+Open `dist/index.html` in your browser and, if everything went right, you should see the following text: 'Hello webpack'.
```
EugeneHlushko, line 205: I think this will be more accessible by wider audience, what do you think?
```suggestion
Open `index.html` from the `dist` directory in your browser and, if everything went right, you should see the following text: 'Hello webpack'.
```

m98: Sure, you're totally right! That's better

EugeneHlushko:
Thanks!

PR #3876, opened 2020-07-23
Update production.md

cseas:
Fix import for `webpack-merge`.
The import used in the article is wrong as per [this answer](https://stackoverflow.com/a/62846324/7435656).


EugeneHlushko:
Hi @cseas its a duplicate of #3824

PR #3885, opened 2020-08-01
docs: add new options for Runtime based optimization

### LINKED ISSUE #3880: Document webpack change: Runtime based optimization

See pull request: https://github.com/webpack/webpack/pull/11235

---

This allows webpack to optimize modules differently depending on in which runtime they are used.
By default each entrypoint is a runtime, here this means a scope for optimization.

Used exports are tracked per runtime. Modules are concatenated per runtime. Code Generation runs per runtime.
Chunks with equal modules that differ only in used exports are not merged by default. For `splitChunks` `usedExports: false` can be set globally or per cacheGroup to allow merging modules with different used exports together (tradeoff: better caching, larger initial download).

`import("module")` for two different locations in source code will lead to independent optimized chunks.
`import("module")` from one location in source code from a module that is included in multiple locations will lead to a single chunk with combined optimization. In future we may offer a magic comment to control this behavior.

I assume that this has a performance influence on larger builds with many entrypoints. `optimization.usedExports: "global"` allows to opt-out of per runtime optimization.

feature, refactoring

yes

yes

* `entry[x].runtime` allows to set the runtime chunk for an entrypoint
  * `optimization.runtimeChunk` allows to set it globally for unspecified entrypoints
* added `optimization.usedExports: "global"` to opt-out from used exports analysis per runtime
* added `splitChunks.usedExports` and `splitChunks.cacheGroups[x].usedExports` to enable/disable used exports comparing for modules
  * `splitChunks.usedExports` defaults to `true` in production

jamesgeorge007:
closes #3880

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -282,6 +282,16 @@ module.exports = {
 };
 ```
 
+### `splitChunks.usedExports`
+
+#### `splitChunks.cacheGroups{cacheGroup}.usedExports`
+
+`boolean = true`
+
+Figure out which exports are used by modules to mangle export names, omit unused exports and generate more efficient code.
+When it is `true`: analyse used exports for each runtime, when it is `"global"`: analyse exports globally for all runtimes combined).
+
+
 ### `splitChunks.cacheGroups`
 
 Cache groups can inherit and/or override any options from `splitChunks.*`; but `test`, `priority` and `reuseExistingChunk` can only be configured on cache group level. To disable any of the default cache groups, set them to `false`.
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -452,7 +452,7 @@ module.exports = {
 
 ## `optimization.usedExports`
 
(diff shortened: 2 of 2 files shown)


src/content/plugins/split-chunks-plugin.md
```diff
@@ -282,6 +282,15 @@ module.exports = {
 };
 ```
 
+### `splitChunks.usedExports`
+
+#### `splitChunks.cacheGroups{cacheGroup}.usedExports`
+
+`boolean = true`
+
+Figure out which exports are used by modules to mangle export names, omit unused exports and generate more efficient code (true: analyse used exports for each runtime, "global": analyse exports globally for all runtimes combined).
```
montogeek, line 291:
```suggestion
Figure out which exports are used by modules to mangle export names, omit unused exports and generate more efficient code.
When it is `true`: analyse used exports for each runtime, when it is `"global"`: analyse exports globally for all runtimes combined).
```

montogeek (approved):

montogeek:
Thanks!

Issue #3886, opened 2020-08-03
Tree Shaking localization in Chinese has a grammar issue


sy0uta1k0n:
__Feature to document__

tree shaking: https://webpack.docschina.org/guides/tree-shaking/

multiple '代码' provided
It should be <这样会允许去掉这
'+段'
'-代码'
代码，但仍然会有一些导入的问题需要被包括/评估，因为它们包含了副作用。>

__Additional information__
<img width="1115" alt="tree shaking" src="https://user-images.githubusercontent.com/13308844/89142828-0fbd9d80-d57b-11ea-86e3-7741d6db7aff.png">

[Y] I am willing to work on this issue and submit a pull request.

chenxsan:
Please open an issue on https://github.com/docschina/webpack.js.org/issues, it can't be fixed in this repository.

sy0uta1k0n:
Got it and I had open a new issue in the mentioned repo

Issue #3834, opened 2020-07-13
Display Problem on Chinese AssetManagement Website


Ji4ming:
https://www.webpackjs.com/guides/asset-management/

The change from main.js to bundle.js(line 35-36) is not truly highlighted on the webpage which may lead to certain misunderstandings or omissions.
This change is supposed to be highlighted as it is crucial for all the changes we will make in latter sections.

chenxsan:
Not a problem can be fixed in this repository. You might want to create an issue on the Chinese translation site https://github.com/docschina/webpack.js.org/issues

PR #3887, opened 2020-08-03
update issue_template

chenxsan:
I found there're some issues related to the Chinese translation which can't be fixed in this repository, e.g.:

1. https://github.com/webpack/webpack.js.org/issues/3834
2. https://github.com/webpack/webpack.js.org/issues/3886

Maybe we can add a note in the issue template to reduce such issues in future?


EugeneHlushko:
Thanks!

PR #3889, opened 2020-08-03
Fix wrong proxy type

BennyAlex:
Just change proxy from array to object

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -1136,10 +1136,10 @@ __webpack.config.js__
 module.exports = {
   //...
   devServer: {
-    proxy: [{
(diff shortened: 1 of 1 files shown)


chenxsan:
From the definition here https://github.com/webpack/webpack-dev-server/blob/adeb92e1e37551a6cbf3063942d6c2c7efbdff10/lib/options.json#L292, array is valid as well.

BennyAlex:
@chenxsan 
ive got this error: When `proxy` in package.json is an object, each `context` object must have a `target` property specified as a url string


So I thought its not working, but it looks like this is an problem with vue-cli and not webpack itself, see https://github.com/vuejs/vue-cli/issues/5740

PR #3898, opened 2020-08-06
docs(configuration): document external script

### LINKED ISSUE #3806: Document webpack change: avoid collision in `data-webpack` by adding the uniqueName

See pull request: https://github.com/webpack/webpack/pull/11065

---

bugfix

yes

yes

* `data-webpack` has `output.uniqueName` + `:` as prefix

chenxsan:
closes #3806

Related issue #3772

### DIFF
--- a/src/content/configuration/externals.md
+++ b/src/content/configuration/externals.md
@@ -321,7 +321,7 @@ Supported types:
 - `'system'`
 - `'promise'` - same as `'var'` but awaits the result (async module)
 - `'import'` - uses `import()` to load a native EcmaScript module (async module)
-- `'script'` - load script with HTML `<script>` element and use the global variable defined.
+- `'script'` - load script exposing predefined global variables with HTML `<script>` element
 
 __webpack.config.js__
 
@@ -334,14 +334,33 @@ module.exports = {
 
 ### `script`
 
-External script can be loaded from any URL when [`externalsType`](#externalstype) is set to `'script'`. The `<script>` tag would be removed after it has been loaded.
+External script can be loaded from any URL when [`externalsType`](#externalstype) is set to `'script'`. The `<script>` tag would be removed after the script has been loaded.
 
-There're two syntaxes for loading external script in webpack configuration:
+#### Syntax
 
-1. `{externals: {packageName: ['http://example.com/script.js', 'global', 'property', 'property']}}` (properties are optional)
-2. `{externals: {packageName: 'global@http://example.com/script.js'}}`
+```javascript
+module.exports = {
+  externals: {
+    packageName: ['http://example.com/script.js', 'global', 'property', 'property'] // properties are optional
+  }
+};
+```
 
-For example, you can load lodash and jQuery from CDN:
+You can also use the shortcut syntax if you're not going to specify any properties:
+
+```javascript
+module.exports = {
+  externals: {
+    packageName: 'global@http://example.com/script.js' // no properties here
+  }
+};
+```
+
+T> [`output.publicPath`](/configuration/output/#outputpublicpath) won't be added to the provided URL.
+
+#### Example
+
+Let's load a `lodash` from CDN:
 
 __webpack.config.js__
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/externals.md
```diff
@@ -330,3 +331,60 @@ module.exports = {
...
+Then use them in your code:
+
+```js
+import $ from 'jquery';
+// `import 'jquery'` works too if you only want to use global variable
+import _ from 'lodash';
+// both local variable `$` and global variable `window.$` would be available here
+```
+
+T> `publicPath` won't be added to the url.
```
EugeneHlushko, line 368:
```suggestion
T> [`output.publicPath`](/configuration/output/#outputpublicpath) won't be added to the provided URL.
```

src/content/configuration/externals.md
```diff
@@ -330,3 +331,60 @@ module.exports = {
...
 };
 ```
+
+### `script`
+
+External script can be loaded from any url when `externalsType` is set to `script`.
+
+There're two syntaxes for loading external script in webpack configuration:
+
+1. `{externals: {packageName: ['http://example.com/script.js', 'global', 'property', 'property']}}` (properties are optional)
```
EugeneHlushko, line 341: These example snippets are less readable than for example just showing the syntax in the code snippet. Also properties are not described well. I suggest you do two full config examples for each of the syntaxes with more details on `property`, `property` use-cases and usage.

src/content/configuration/externals.md
```diff
@@ -330,3 +331,60 @@ module.exports = {
   externalsType: 'promise'
 };
 ```
+
+### `script`
+
+External script can be loaded from any url when `externalsType` is set to `script`.
+
+There're two syntaxes for loading external script in webpack configuration:
```
EugeneHlushko, line 339: Could rephrase this one when giving examples one by one

src/content/configuration/externals.md
```diff
@@ -330,3 +331,60 @@ module.exports = {
   externalsType: 'promise'
 };
 ```
+
+### `script`
+
+External script can be loaded from any url when `externalsType` is set to `script`.
```
EugeneHlushko, line 337:
```suggestion
External script can be loaded from any URL when [`externalsType`](#externalstype) is set to `'script'`. The `<script>` tag would be removed after it has been loaded.
```

src/content/configuration/externals.md
```diff
@@ -321,6 +321,7 @@ Supported types:
 - `'system'`
 - `'promise'` - same as `'var'` but awaits the result (async module)
 - `'import'` - uses `import()` to load a native EcmaScript module (async module)
+- `'script'` - load script with HTML `<script>` element and use the global variable defined.
```
EugeneHlushko, line 324: Is this more accessible than `global variable defined`? I think it assumes you understand context well enough
```suggestion
- `'script'` - load script with HTML `<script>` element and store the result in a selected global variable
```

src/content/configuration/externals.md
```diff
@@ -330,3 +331,60 @@ module.exports = {
...
+
+Both `head` and `window._` are exposed now when you `import` `lodash`:
+
+```js
+import head from 'lodash';
+console.log(head([1, 2, 3])); // logs 1 here
+console.log(window._.head(['a', 'b'])); // logs a here
+```
+
+T> When loading code with `script` tags, the runtime will try to find an existing `script` tag that matches the `src` attribute or has a specific `data-webpack` attribute. For chunk loading `data-webpack` attribute would have value of `"[output.uniqueName]:chunk-[chunkId]"` while external script has value of `"[output.uniqueName]:[global]"`. The script tag would be removed once the script is loaded. Options like `output.chunkLoadTimeout`, `output.crossOriginLoading` and `output.scriptType` apply to these external scripts too.
```
EugeneHlushko, line 390: I think these are two separate blocks, also i think we need to move this to the very top: 'The `<script>` tag would be removed after it has been loaded."
```suggestion
T> When loading code with HTML `<script>` tags, the webpack runtime will try to find an existing `<script>` tag that matches the `src` attribute or has a specific `data-webpack` attribute. For chunk loading `data-webpack` attribute would have value of `'[output.uniqueName]:chunk-[chunkId]'` while external script has value of `'[output.uniqueName]:[global]'`. 

T> Options like `output.chunkLoadTimeout`, `output.crossOriginLoading` and `output.scriptType` will also have effect on the external scripts loaded this way.
```

src/content/configuration/externals.md
chenxsan: How about?:

> load script exposing predefined global variables with HTML `<script>` element

From what I learned so far, we can only load external scripts here with `<script>` element when they expose predefined global variables like jquery did with global `window.jQuery` and `window.$`, otherwise it'll fail. Also we can't use random variable name but those predefined ones in `global` placeholder.

EugeneHlushko: Yeah reads much better and also makes sense, lets get Tobias to comment if this is right tho

EugeneHlushko (approved):

EugeneHlushko:
Thanks!

PR #3901, opened 2020-08-07
fix header visibility on safari

chenxsan:
The [`scroll-margin-top`](https://github.com/webpack/webpack.js.org/blob/master/src/components/Markdown/Markdown.scss#L50) we use is [not applied for scrolls to fragment target on safari](https://bugs.webkit.org/show_bug.cgi?id=189265) which causes problem for Safari users.

For example, when we open https://webpack.js.org/concepts/#entry on safari:

image

While on Chrome we have visible header:

image

This pull request should fix the problem on safari.


EugeneHlushko:
Thanks!

PR #3911, opened 2020-08-11
docs(configuration): update link

chenxsan:
The defaults part was moved to https://github.com/webpack/webpack/blob/master/lib/config/defaults.js in this commit https://github.com/webpack/webpack/commit/6477ca56f8046e4642d983f68a68805b83982537#diff-7473dd12403a310ad8cf18d0ab9f10b9


EugeneHlushko:
Thanks!

Issue #3578, opened 2020-02-23
don't make plus-signs selectable


SimonSiefke:
### Expected behavior: 
I can copy and paste the code, the plus-signs are not copied with the code.

### Current behavior:
When I copy and paste the code, I manually need to remove each `+` because the the plus-signs are copied along with the code.

![webpack-select](https://user-images.githubusercontent.com/23744935/75109201-8300d200-5621-11ea-9ec4-cc61dd659b9c.png)

anikethsaha:
Those `+` and `-` signs are required in order to make it better to understand how to change the code for migration or how to replace any deprecated APIs with the new one. 

these are not usually long snippets so manually deleting should not be a much of work. 👍

SimonSiefke:
I was thinking that maybe 
```html
<p><span style="user-select:none">+</span>document.body.removeChild(element);</p>
<p><span style="user-select:none">+</span>element = component();</p>
<p><span style="user-select:none">+</span>document.body.appendChild(element);</p>
```

would be a good solution. 

I agree that it is not much work to manually remove the plus-signs but I think it is unnecessary and that the guides would be improved when the code can just be copied and pasted.

chenxsan:
@SimonSiefke  It could be possible if we bring Prismjs plugin https://prismjs.com/plugins/diff-highlight/ support into [Refactor](https://github.com/wooorm/refractor)  which this doc site relies one.

DanielRuf:
Or render `:before` pseudo elements.

I think this issue can be closed as https://github.com/PrismJS/prism/pull/2281 is merged?

chenxsan:
@DanielRuf That commit is not released in a new version PrismJS yet. Maybe we can keep this issue open if we do want to bring this feature into doc site.

chenxsan:
- [x] Wait for new PrismJS to be released ([1.21.0](https://github.com/PrismJS/prism/releases/tag/v1.21.0) is out)
- [ ] Wait for [refractor](https://github.com/wooorm/refractor) to update to PrismJS 1.21.0
- [ ] Wait for [remark-refractor](https://github.com/montogeek/remark-refractor) to update to then-latest refractor

chenxsan:
I've created a pull request here https://github.com/webpack/webpack.js.org/pull/3912 so anyone can check how it looks like now https://webpack-js-org-git-fork-chenxsan-feature-user-select-none-diff.webpack-docs.vercel.app/guides/asset-management/#loading-images.

PR #3912, opened 2020-08-12
optimize diff highlighting

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/3578

<del>Currently the dependency uses github branch of my forked repository, I'll update it after https://github.com/montogeek/remark-refractor/pull/1 is merged.</del>

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "remark-custom-blockquotes": "1.0.0",
     "remark-extract-anchors": "1.1.1",
     "remark-loader": "^0.3.0",
-    "remark-refractor": "git://github.com/chenxsan/remark-refractor.git#update-refractor",
+    "remark-refractor": "git://github.com/montogeek/remark-refractor.git",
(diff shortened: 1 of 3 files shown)


EugeneHlushko:
This still has an issue, you select less spaces than you should where such characters had select disabled

chenxsan:
@EugeneHlushko Indeed. But missing whitespace or indent doesn't break program in JavaScript. With tools like prettier, code without `+` `-` could be formatted while impossible if we have `+` `-` presented.

@SimonSiefke What do you think here? Do you mind the missing spaces?

SimonSiefke:
Since I use prettier, I don't mind the spaces :)

chenxsan:
I'll update the dependency when a new version of `remark-refractor` is published on npm.

EugeneHlushko (approved):

EugeneHlushko:
Thanks!

PR #3918, opened 2020-08-19
docs(configuration): document output.charset

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/3922

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -98,7 +98,7 @@ module.exports = {
 
 This option tells webpack to add `charset` attribute with value of `utf-8` to HTML `<script>` element.
 
-T> `charset` attribute for `<script>` element was [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes). It defaults to `true` for compatibility of non-modern browsers.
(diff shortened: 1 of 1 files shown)


jeffin143 (approved):
LGTM 🎉

src/content/configuration/output.md
```diff
@@ -92,6 +92,14 @@ module.exports = {
 ```
 
 
+## `output.charset`
+
+`boolean = true`
+
+This option tells webpack to add `charset` attribute with value of `utf-8` to HTML `<script>` element.
+
+T> Although `charset` attribute for `<script>` element was [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes), webpack still adds it by default for compatibility of non-modern browsers.
```
montogeek, line 101:
```suggestion
T> Although `charset` attribute for `<script>` element was [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Deprecated_attributes), webpack still adds it by default for compatibility with non-modern browsers.
```

EugeneHlushko (approved):

EugeneHlushko:
Thanks!

PR #3923, opened 2020-08-20
docs(concepts): add notice about webpack-sources

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/3921

### DIFF
--- a/src/content/concepts/plugins.md
+++ b/src/content/concepts/plugins.md
@@ -15,7 +15,7 @@ __Plugins__ are the [backbone](https://github.com/webpack/tapable) of webpack. w
 
 They also serve the purpose of doing __anything else__ that a [loader](/concepts/loaders) cannot do.
 
-T> When consuming [`webpack-sources`](https://github.com/webpack/webpack-sources) package in plugins, use `require("webpack").sources` instead of `require("webpack-sources")` to avoid version conflicts for persistent caching.
+T> When consuming [`webpack-sources`](https://github.com/webpack/webpack-sources) package in plugins, use `require('webpack').sources` instead of `require('webpack-sources')` to avoid version conflicts for persistent caching.
 
 
 ## Anatomy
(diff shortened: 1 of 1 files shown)


montogeek (approved):

rishabh3112 (approved):

jamesgeorge007 (approved):

src/content/concepts/plugins.md
```diff
@@ -8,12 +8,15 @@ contributors:
...
   - MisterDev
   - byzyk
+  - chenxsan
 ---
 
 __Plugins__ are the [backbone](https://github.com/webpack/tapable) of webpack. webpack itself is built on the __same plugin system__ that you use in your webpack configuration!
 
 They also serve the purpose of doing __anything else__ that a [loader](/concepts/loaders) cannot do.
 
+T> When consuming [`webpack-sources`](https://github.com/webpack/webpack-sources) package in plugins, use `require("webpack").sources` instead of `require("webpack-sources")` to avoid version conflicts for persistent caching.
```
EugeneHlushko, line 18:
```suggestion
T> When consuming [`webpack-sources`](https://github.com/webpack/webpack-sources) package in plugins, use `require('webpack').sources` instead of `require('webpack-sources')` to avoid version conflicts for persistent caching.
```

Besides, does "to avoid version conflicts for persistent caching" sound right to y'all? I think it meant to be "to avoid version conflicts (when using)(in) persistent caching"

EugeneHlushko (changes requested):

src/content/concepts/plugins.md
chenxsan: I'm not quite sure of it. Maybe we can remove `for persistent caching` since `avoid version conflicts` is enough in my opinion.

EugeneHlushko (approved):

EugeneHlushko:
Thanks!

PR #3935, opened 2020-08-25
docs(configuration): specify default value as part of type information

jamesgeorge007:
webpack defaults to `production` mode.

### DIFF
--- a/src/content/configuration/optimization.md
+++ b/src/content/configuration/optimization.md
@@ -23,11 +23,9 @@ Since version 4 webpack runs optimizations for you depending on the chosen  [`mo
 
 ## `optimization.minimize`
 
-`boolean`
+`boolean = true`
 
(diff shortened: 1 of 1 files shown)


montogeek (approved):

EugeneHlushko (changes requested):
if we are doing this, can you also change all similar occurances in one PR? There at least 5-6 places i've seen this

EugeneHlushko (approved):
will do a follow up to not hold this off

EugeneHlushko:
Thanks!

EugeneHlushko:
Hmm this is actually tricky, @montogeek there are places where we say 

By default `optimization.x` is enabled in `production` [mode](/configuration/mode/) and disabled elsewise.

i think it might be useful for people to understand that in dev mode it all might be different... maybe this should be reverted after all?

PR #3944, opened 2020-09-02
docs(guides): remove superfluous link

chenxsan:
Now we have links navigating to previous/next pages, I think there's no need to mention the next guide in the page again:

image

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -624,12 +624,6 @@ __src/index.js__
   document.body.appendChild(component());
 ```
 
-
-## Next guide
-
-Let's move on to [Output Management](https://webpack.js.org/guides/output-management/)
-
-
 ## Further Reading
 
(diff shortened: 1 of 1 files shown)


montogeek:
Where is do we have this "Next guide" subtitle?

chenxsan:
There won't be any `Next guide` subtitle in this page then :) My point is the link is kind of duplicated with what we have now in `previous/next` links in this specific case. Feel free to close it you think it's better to keep it.

snitin315:
I think he meant to remove such `Next Guide` subtitle from other places in the website too (if any).

chenxsan:
Then there's only this one standalone paragraph as I found.

EugeneHlushko:
i think there is difference between the current next guide and generic "next page" thing, maybe leave it as is for accessibility of the big variety of visitors of webpack.js.org, it did not harm our SEO so far

PR #3970, opened 2020-09-12
docs(guides): clean html-webpack-template link

chenxsan:
html-webpack-template is [not compatible with html-webpack-plugin v4](https://github.com/jaketrent/html-webpack-template/issues/84), that means we can't use it in in webpack 5, so there's no need to mention it in the guide.


evilebottnawi (approved):

EugeneHlushko (approved):

EugeneHlushko:
Thanks!

PR #4001, opened 2020-09-25
Add CSV example to `guides/asset-management`

billfienberg:
## Description

The [asset management guide](https://webpack.js.org/guides/asset-management/) mentions CSV, and includes an example config of how to load CSV, but doesn't actually include loading a real CSV file. This PR updates the [Loading Data](https://webpack.js.org/guides/asset-management/#loading-data) section to include an actual CSV example. 

>To import CSVs, TSVs, and XML you could use the csv-loader and xml-loader. Let's handle loading all three:

### DIFF
--- a/src/content/guides/asset-management.md
+++ b/src/content/guides/asset-management.md
@@ -657,6 +657,7 @@ __project__
     |- index.html
   |- /src
 -   |- data.xml
+-   |- data.csv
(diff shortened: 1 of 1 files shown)


EugeneHlushko (approved):

EugeneHlushko:
Thanks!

PR #4015, opened 2020-09-29
docs(configuration): document preferRelative

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4011


src/content/configuration/resolve.md
```diff
@@ -414,6 +414,35 @@ module.exports = {
 ```
 
 
+### `resolve.preferRelative`
+
+`boolean`
+
+When enabled, webpack would prefer to resolve module requests as relative requests than modules from `node_modules` directories.
```
montogeek, line 421:
```suggestion
When enabled, webpack would prefer to resolve module requests as relative requests instead of using modules from `node_modules` directories.
```

sokra (approved):

PR #4021, opened 2020-10-02
fix syntax error in splitChunks plugin exmple

Taimoor0217:
In the [splitChunks.name](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunksname) example of splitChunks. A colon is missing after the name attribute. 

Screen shot attached.

color_error


evilebottnawi (commented):
It is method, you don't need colon here

EugeneHlushko:
Closing as clarified

PR #4029, opened 2020-10-07
fix configuration page: add missing popover tags

sokra:
split details in multiple smaller popover


montogeek:
Glad that my frankenstein language is easy to change :)

sokra:
We need to reconsider these popovers in future. They are not accessible without a mouse and unreadable on mobile. Some kind of inline expanding code snippet would be better. And the comments need to be focusable.

PR #4033, opened 2020-10-07
improve build workflow

sokra:
cache repository data up to 48 hours
accept fetch errors when cached data is available
This tries to reduce the number of rate limit errors

improve stages in travis


EugeneHlushko (approved):
Looks like its working 👍

EugeneHlushko:
still to pass travis, but the website itself up and running on vercel :P

sokra:
> still to pass travis, but the website itself up and running on vercel :P

Running into rate limit. After this PR the API results would be cached and it would fallback to the cache when fetch fails...

PR #4059, opened 2020-10-10
Include optimization.noEmitOnErrors in migration guide

loilo:
The `optimization.noEmitOnErrors` option has been inverted to `optimization.emitOnErrors`, which I just noticed by accident. This step is currently missing from the migration guide.


sokra (approved):

sokra:
Thanks

PR #4067, opened 2020-10-13
docs(configuration): update rule test

chenxsan:
`/\.m?js/` could match `.json` too https://github.com/webpack/webpack/issues/11467#issuecomment-707722067, which is not what we want.


sokra (approved):

sokra:
Thanks

PR #4077, opened 2020-10-17
Update mode.md

tititiou36:
Fix the syntax of the `mode` configuration option

### DIFF
--- a/src/content/configuration/mode.md
+++ b/src/content/configuration/mode.md
@@ -14,7 +14,7 @@ related:
 
 Providing the `mode` configuration option tells webpack to use its built-in optimizations accordingly.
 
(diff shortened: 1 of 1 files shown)


chenxsan:
Actually it's [how](https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types) we define typing in documentation at this moment.

```
`string = 'production': 'none' | 'development' | 'production'`
```
means the type is `string`, and has a value of `'none'`, `'development'` or `'production'` while `'production'` is the default value.

We have an issue discussing the typing at https://github.com/webpack/webpack.js.org/issues/3333.

tititiou36:
Ok, sorry for the noise

PR #4078, opened 2020-10-18
docs(configuration): update node page for webpack 5

chenxsan:
The contents for `node` page changed a lot in webpack 5, so I refactor some.

For the sake of review, I've added some comments too.

Preview URL: https://webpack-js-org-git-bugfix-update-node.webpack-docs.vercel.app/configuration/node/

### DIFF
--- a/src/content/configuration/node.md
+++ b/src/content/configuration/node.md
@@ -9,29 +9,18 @@ contributors:
   - byzyk
   - EugeneHlushko
   - anikethsaha
+  - chenxsan
 ---
 
-The following Node.js options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html) and modules. This allows code originally written for the Node.js environment to run in other environments like the browser.
+The following Node.js options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html).
 
-This feature is provided by webpack's internal [`NodeStuffPlugin`](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js) plugin. If the target is "web" (default) or "webworker", the [`NodeSourcePlugin`](https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js) plugin is also activated.
+This feature is provided by webpack's internal [`NodeStuffPlugin`](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js) plugin.
 
+W> As of webpack 5, You can configure only `global`, `__filename` or `__dirname` under `node` option. If you're looking for how to polyfill `fs` alike in Node.js under webpack 5, please check [resolve.fallback](/configuration/resolve/#resolvefallback) for help.
 
 ## `node`
 
-`boolean = false` `object`
-
-This is an object where each property is the name of a Node global or module and each value may be one of the following...
-
-- `true`: Provide a polyfill.
-- `'mock'`: Provide a mock that implements the expected interface but has little or no functionality.
-- `'empty'`: Provide an empty object.
-- `false`: Provide nothing. Code that expects this object may crash with a `ReferenceError`. Code that attempts to import the module using `require('modulename')` may trigger a `Cannot find module "modulename"` error.
-
-W> Not every Node global supports all four options. The compiler will throw an error for property-value combinations that aren't supported (e.g. `global: 'empty'`). See the sections below for more details.
-
-T> If you are using a module which needs global variables in it, use `ProvidePlugin` instead of `global`.
-
-These are the defaults:
+`boolean: false` `object`
 
 __webpack.config.js__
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/node.md
```diff
@@ -9,29 +9,18 @@ contributors:
   - byzyk
   - EugeneHlushko
   - anikethsaha
+  - chenxsan
 ---
 
-The following Node.js options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html) and modules. This allows code originally written for the Node.js environment to run in other environments like the browser.
+The following Node.js options configure whether to polyfill or mock certain [Node.js globals](https://nodejs.org/docs/latest/api/globals.html).
```
chenxsan, line 15: `node` only cares about three Node.js globals now, so informations related to `modules` are removed.

chenxsan: `This allows code originally written for the Node.js environment to run in other environments like the browser.` was removed as it's not true anymore, we have to polyfill Node.js modules like `fs` etc too, which is described in `resolve.fallback` section.

src/content/configuration/node.md
```diff
@@ -9,29 +9,18 @@ contributors:
...
 
-This feature is provided by webpack's internal [`NodeStuffPlugin`](https://github.com/webpack/webpack/blob/master/lib/NodeStuffPlugin.js) plugin. If the target is "web" (default) or "webworker", the [`NodeSourcePlugin`](https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js) plugin is also activated.
```
chenxsan, line 16: `NodeSourcePlugin` is empty now https://github.com/webpack/webpack/blob/master/lib/node/NodeSourcePlugin.js, so no need to keep.

src/content/configuration/node.md
```diff
@@ -9,29 +9,18 @@ contributors:
...
-- `true`: Provide a polyfill.
-- `'mock'`: Provide a mock that implements the expected interface but has little or no functionality.
-- `'empty'`: Provide an empty object.
-- `false`: Provide nothing. Code that expects this object may crash with a `ReferenceError`. Code that attempts to import the module using `require('modulename')` may trigger a `Cannot find module "modulename"` error.
-
-W> Not every Node global supports all four options. The compiler will throw an error for property-value combinations that aren't supported (e.g. `global: 'empty'`). See the sections below for more details.
-
-T> If you are using a module which needs global variables in it, use `ProvidePlugin` instead of `global`.
-
-These are the defaults:
```
chenxsan, line 34: `empty` was removed in webpack 5 https://github.com/webpack/webpack/blob/60f7ce301df553d8ab0276ba2838ddb60ead1c94/declarations/WebpackOptions.d.ts#L1369

chenxsan: I've removed the whole part as I think they're already listed below, there's no need to duplicate them here. But I can revert it if you think it's better to keep.

src/content/configuration/node.md
```diff
@@ -46,38 +35,40 @@ module.exports = {
...
-Defaults to `false` for [targets](/configuration/target/) `node`, `async-node` and `electron-main`.
+`boolean` `string: 'mock' | 'eval-only'`
 
 Options:
 
 - `true`: The dirname of the __input__ file relative to the [`context` option](/configuration/entry-context/#context).
-- `false`: The regular Node.js `__dirname` behavior. The dirname of the __output__ file when run in a Node.js environment.
+- `false`:  Webpack won't touch it, which means you have the regular Node.js `__dirname` behavior. The dirname of the __output__ file when run in a Node.js environment.
 - `'mock'`: The fixed value `'/'`.
+- `'eval-only'`
```
chenxsan, line 74: I'm not sure of this value as there's no description in webpack repository.

src/content/configuration/node.md
```diff
@@ -46,38 +35,40 @@ module.exports = {
...
 
-Defaults to `false` for [targets](/configuration/target/) `node`, `async-node` and `electron-main`.
+`boolean` `string: 'mock' | 'eval-only'`
 
 Options:
 
 - `true`: The filename of the __input__ file relative to the [`context` option](/configuration/entry-context/#context).
-- `false`: The regular Node.js `__filename` behavior. The filename of the __output__ file when run in a Node.js environment.
-- `'mock'`: The fixed value `'index.js'`.
+- `false`: Webpack won't touch your `__filename` code, which means you have the regular Node.js `__filename` behavior. The filename of the __output__ file when run in a Node.js environment.
```
montogeek, line 60:
```suggestion
- `false`: webpack won't touch your `__filename` code, which means you have the regular Node.js `__filename` behavior. The filename of the __output__ file when run in a Node.js environment.
```

chenxsan: I think Capital `W` is allowed now when `webpack` occurs in the beginning https://webpack.js.org/blog/2020-10-10-webpack-5-release/#other-minor-changes :)

image

src/content/configuration/node.md
```diff
@@ -46,38 +35,40 @@ module.exports = {
...
 
-Since webpack 3.0.0, the `node` option may be set to `false` to completely turn off the `NodeStuffPlugin` and `NodeSourcePlugin` plugins.
+Since webpack 3.0.0, the `node` option may be set to `false` to completely turn off the `NodeStuffPlugin` plugin.
 
 ## `node.global`
 
-`boolean = true`
+`boolean`
 
-Defaults to `false` for [targets](/configuration/target/) `node`, `async-node` and `electron-main`.
```
alecmev, line 55: Why remove this? Is there a summary of [`TARGETS`](https://github.com/webpack/webpack/blob/32bd1a2ff53c6b7fbf99d0e473d6db98f9e3a3f2/lib/config/target.js#L75-L269) I'm missing, or...?

chenxsan: I don't remember the specific reason. There're more than three targets having `node.global` set to `false`, which could be the reason I removed this.

alecmev: The deleted info is outdated, indeed, but why not update it, instead of deleting it? How can I find out what it defaults to without digging into the code?

I'm concerned about this because I actually had an old todo, about improving this page, and by the time I got around to it, somebody already added the defaults, and when I went to look who did it, out of curiosity, I find that it got deleted :smile:

chenxsan: pr welcome.

alecmev: This doesn't encourage me to contribute in this particular case. As it stands, it'll just get deleted a few months later again because of being slightly inaccurate.

chenxsan: Yeah, it's difficult to align with all changes happening in webpack core project. There will always some stale contents.

Issue #4083, opened 2020-10-21
Print link isn't working


jvmazagao:
__Feature to document__
The print section isn't open a dialog to print the section os save the section.

![After clicking the print section in documentation](https://user-images.githubusercontent.com/22477816/96721675-d8dd7380-1382-11eb-9aa2-bfa6039a3a9f.png)

__Author(s)__

__Additional information__

Steps -
1. Open [Webpack page](https://webpack.js.org/)
2. Go to the [documentation section](https://webpack.js.org/concepts/)
3. Click on the Print Section link.

What I'm as a user expect -
I expect in the concepts part or in every part to print the current section, or, to print all the documentation section.

[ ] I am willing to work on this issue and submit a pull request.

chenxsan:
Ah, I see, you mean the "Print Section" in the sidebar:

![image](https://user-images.githubusercontent.com/1091472/96809371-52da2f00-144d-11eb-91b1-694403fcd997.png)

I've misunderstood you as I thought it would be `Print Document`:

![image](https://user-images.githubusercontent.com/1091472/96809448-743b1b00-144d-11eb-8ccc-d90b79355e30.png)

Indeed, it's broken.

chenxsan:
Hi, I've pushed a pull request to fix it, can you please try the [preview url](https://github.com/webpack/webpack.js.org/pull/4086#issuecomment-714201670) over there?

jvmazagao:
> Hi, I've pushed a pull request to fix it, can you please try the [preview url](https://github.com/webpack/webpack.js.org/pull/4086#issuecomment-714201670) over there?

Working.

PR #4086, opened 2020-10-22
fix bug for printable

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4083

When we open pages like https://webpack.js.org/concepts/printable/, it would fallback to `/app-shell/index.html` which is wrong as we already have those static html files under `/*/printable/index.html` to be served.

### DIFF
--- a/webpack.prod.js
+++ b/webpack.prod.js
@@ -63,6 +63,7 @@ module.exports = env => merge(common(env), {
         }))
       ],
       navigateFallback: '/app-shell/index.html',
+      navigateFallbackDenylist: [/printable/],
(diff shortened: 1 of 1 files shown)


chenxsan:
I'll see if I can add a e2e test preventing future regression.

jvmazagao (approved):

montogeek (approved):

snitin315 (approved):

jvmazagao:
Can we merge?

chenxsan:
@jvmazagao Merged, and thanks for reporting issue.

PR #4087, opened 2020-10-22
Add e2e test for Print Section

chenxsan:
A follow-up of https://github.com/webpack/webpack.js.org/pull/4086. CI would fail at the moment, but should pass when that PR get merged.


chenxsan:
Closed as we don't run `yarn printable` under `development` mode which means there's no way to test them.

Issue #4112, opened 2020-10-30
Update Loaders testing docs


ogonkov:
__Feature to document__

Loaders testing examples is not valid anymore with webpack@5

__Author(s)__

__Additional information__

Developers already confused by docs inconsistency, see for example this [SO question](https://stackoverflow.com/questions/64548728/module-source-is-undefined-following-webpack-5-loader-guide)

[ ] I am willing to work on this issue and submit a pull request.

PR #4114, opened 2020-11-01
Corrected Jest code for Loader Writer's Guide.

pglezen:
closes https://github.com/webpack/webpack.js.org/issues/4112

The [Jest testing example](https://webpack.js.org/contribute/writing-a-loader/#testing) of the Loader Writer's Guide needs the following for Webpack5.

* The `stats.toJson()` method now requires `{source: true}` as a parameter in order to populate the `source` property of the `module` object.
* If people create the `example.txt` with a line ending, then the expected output should also have a line ending.

I verified that with these changes the Jest test passes as coded in the document of my pull request.

### DIFF
--- a/src/content/contribute/writing-a-loader.md
+++ b/src/content/contribute/writing-a-loader.md
@@ -370,9 +370,9 @@ import compiler from './compiler.js';
 
 test('Inserts name and outputs JavaScript', async () => {
   const stats = await compiler('example.txt', { name: 'Alice' });
-  const output = stats.toJson().modules[0].source;
+  const output = stats.toJson({source: true}).modules[0].source;
 
-  expect(output).toBe('export default "Hey Alice!\"');
+  expect(output).toBe('export default "Hey Alice!\\n"');
 });
 ```
 


src/content/contribute/writing-a-loader.md
```diff
@@ -370,9 +370,9 @@ import compiler from './compiler.js';
 
 test('Inserts name and outputs JavaScript', async () => {
   const stats = await compiler('example.txt', { name: 'Alice' });
-  const output = stats.toJson().modules[0].source;
+  const output = stats.toJson({source: true}).modules[0].source;
 
-  expect(output).toBe('export default "Hey Alice!\"');
+  expect(output).toBe('export default "Hey Alice!\\n"');
```
chenxsan, line 375: Can you double check this one? I've tested locally and it's not needed:

image

Just in case you want to have a look, here's the [repository](https://github.com/chenxsan/webpack-write-a-loader) I used to test.

pglezen: It's hard to really tell.  Line endings are not visible in single-line code listings.  The listing in your screenshot might have a line ending; or it might not.  You can't tell with a single line file.  The vi editor, by default, will automatically create a line ending on the end, even though I don't press enter.  I can create a file with no line endings with the `echo -n` option:

```
echo -n 'Hey [name]!' > example.txt
```

Perhaps this, too, could be added to the instructions.  Then we could remove the `\n` on line 375 above.

chenxsan: How about adding an explicit comment to that example, for instance:

```
Hey [name]! // no line ending
```
And update the assertion to:

```
expect(output).toBe('export default "Hey Alice! // no line ending\"');
```

pglezen: Then you're factually correct.  I just worry about people with editors that automatically add a new-line without them knowing.  But, hey, if you're far enough along to be writing webpack loaders, you can figure out whether your editor adds a gratuitous new-line before end of file.  So I'm good with it.

chenxsan: I can see your worry here, as we're trying to cover two possible cases here with one assertion.

chenxsan (approved):
I'll merge as it is. If anyone does have a problem later, we can add a warning for the `example.txt` reminding them to add a line ending at the end.

chenxsan:
@pglezen Thank you for your contribution to webpack 👍

PR #4117, opened 2020-11-01
doc(plugins):  Fix splitChunks.name

hamzahamidi:

[SplitChunksPlugin.optimization.splitChunks.name: true ](https://webpack.js.org/plugins/split-chunks-plugin/#splitchunksname) is not allowed for webpack-5 (see changelog here https://webpack.js.org/blog/2020-10-10-webpack-5-release/#changes-to-the-structure:~:text=Automatic%20names%20are%20no%20longer%20supported).

No

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -213,15 +213,13 @@ The difference between `maxInitialSize` and `maxSize` is that `maxInitialSize` w
 
 Also available for each cacheGroup: `splitChunks.cacheGroups.{cacheGroup}.name`.
 
-The name of the split chunk. Providing `false` will keep the same name of the chunks so it doesn't change names unnecessarily.
+The name of the split chunk. Providing `false` will keep the same name of the chunks so it doesn't change names unnecessarily. It is the recommended value for production builds.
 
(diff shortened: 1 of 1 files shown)


src/content/plugins/split-chunks-plugin.md
```diff
@@ -209,11 +209,11 @@ The difference between `maxInitialSize` and `maxSize` is that `maxInitialSize` w
 
 ### `splitChunks.name`
 
-`boolean = true` `function (module, chunks, cacheGroupKey) => string` `string`
+`boolean = false` `function (module, chunks, cacheGroupKey) => string` `string`
 
 Also available for each cacheGroup: `splitChunks.cacheGroups.{cacheGroup}.name`.
 
-The name of the split chunk. Providing `true` will automatically generate a name based on chunks and cache group key.
+The name of the split chunk. Providing `false` will keep the same name of the chunks so it doesn't change names unnecessarily.
```
snitin315, line 216: Line 224
> T> It is recommended to set `splitChunks.name` to `false` for production builds so that it doesn't change names unnecessarily.

I think this tip is not needed anymore then.

hamzahamidi: @snitin315  I think it's still necessary to show that it's the recommended option for production.

snitin315: How about merging both sentences?

chenxsan (approved):

chenxsan:
@hamzahamidi Thank you for your first contribution to webpack documentation.

PR #4126, opened 2020-11-05
chore: update PR template

snitin315:
_describe your changes..._

Mark the instructions as a comment because they are for the author's eye only.


chenxsan:
I think it would be better to keep as it is.

Here's the current view:

image

New committer could click the link from `Preview` panel. If we comment them out, they have to copy and paste those links if they want to take a look, and that is inefficient in my opinion.

snitin315:
Makes sense 👍🏻

PR #4129, opened 2020-11-06
docs(concepts) entry-points, combine multiple entries to one bundle file by using Array syntax.

adyjs:
docs(concepts) entry-points 
- Combine multiple entries to one bundle file by using Array syntax.


src/content/concepts/entry-points.md
```diff
@@ -64,6 +65,28 @@ T> __"Scalable webpack configurations"__ are ones that can be reused and combine
 
 T> You can pass empty object `{}` to `entry` when you have only entrypoints generated by plugins.
 
+## Array Syntax
```
chenxsan, line 68: Thank you for your patience.

I think the `Single Entry Syntax` section already cover String Syntax (`string`) and Array Syntax (`[string]`):

image

I'm not sure it's a good idea to separate these them. We want to underline the different usages between Single Entry and Multi Entry, instead of the String/Object/Array syntax. How about adding this example to `Single Entry Syntax` section?

We can change `What happens when you pass an array to entry? ` tip into a paragraph and add the example afterward.

adyjs: Hi , chenxsan:

Thank you for your reply, 
your suggest looks great, 
Let's do this.

adyjs: Actually , this is my first time do Pull-Request to open source community.
so , I am not really know about what is the next step?

Uh... , Do I need to cancel or delete this pull request , modify the content , then do pull request again?
Or you just modify directly, then merge.
Thanks.

chenxsan: Just update the change in another commit and push to this branch. No need to cancel or delete this PR.

chenxsan (approved):

chenxsan:
@adyjs Thank you for your contribution to webpack documentation.

adyjs:
> @adyjs Thank you for your contribution to webpack documentation.

Thank you !!!

PR #4132, opened 2020-11-07
docs(cli): update --progress usage

snitin315:
_describe your changes..._

update `--progress` usage.


src/content/api/cli.md
```diff
@@ -272,6 +272,20 @@ webpack --analyze
 
 W> Make sure you have `webpack-bundle-analyzer` installed in your project else CLI will prompt you to install it.
 
+## Progress
+
+To check the progress of any webpack compilation you can use the `--progress` flag.
+
+```sh
```
chenxsan, line 279:
```suggestion
```bash
```
Let's use `bash` here as `sh` is not supported in PrismJS https://travis-ci.org/github/webpack/webpack.js.org/jobs/741912024#L570

src/content/api/cli.md
```diff
@@ -272,6 +272,20 @@ webpack --analyze
...
+webpack --progress
+```
+
+To collect profile data for progress steps you can pass `profile` as value to `--progress` flag.
+
+```sh
```
chenxsan, line 285:
```suggestion
```bash
```

chenxsan (changes requested):

src/content/api/cli.md
snitin315: then we probably need to update it at more places too.

chenxsan (approved):

PR #4099, opened 2020-10-26
docs(guides): update against webpack 5

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4090 fixes https://github.com/webpack/webpack.js.org/issues/4098

`Guide` section needs to be updated against webpack 5, I'll do it in this pull request.

PREVIEW URL: https://webpack-js-org-git-feature-update-guide-against-webpack5.webpack-docs.vercel.app/guides/getting-started/

- [x] update getting-started.md (code demo https://github.com/chenxsan/webpack-demo)
- [x] update asset-management.md (code demo https://github.com/chenxsan/webpack-demo/pull/2)
- [ ] update output-management.md (code demo https://github.com/chenxsan/webpack-demo/pull/3)

PR #4134, opened 2020-11-08
docs(guides): update getting-started against webpack 5

chenxsan:
I found it would be difficult to review all those changes in one pull request like https://github.com/webpack/webpack.js.org/pull/4099, thus I'm splitting it into small ones. Hopefully we can alleviate the pain of reviewing a lot.

Regarding those diff blocks, I'm using Prettier to format code first, and `git show --no-prefix -U1000 > patch.diff` to generate the diffs. It brought quite a lot of changes to the code blocks which is a pain for reviewer at the moment, but I think these changes are beneficial for future as contributors can share the same tools when updating guides with diff blocks. Before this I have to do it manually as there's no guide how to do it, which is inefficient and error-prone.

## Related code repository

https://github.com/chenxsan/webpack-demo

## PREVIEW

https://webpack-js-org-git-feature-update-getting-started.webpack-docs.vercel.app/guides/getting-started/

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -25,9 +25,9 @@ contributors:
   - anshumanv
 ---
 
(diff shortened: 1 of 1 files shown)


snitin315 (approved):
LGTM

PR #4136, opened 2020-11-08
Update asset-management.md

fey:
Fix rule type for yml files
https://webpack.js.org/guides/asset-management/#customize-parser-of-json-modules

```
      {
        test: /\.yaml$/,
        type: 'json',
        parser: {
          parse: yaml.parse
        }
      },
```


chenxsan:
We're customizing the parser of **JSON Module**, so the type is set to `json` instead of `yaml`.

fey:
Thanks!

Issue #4072, opened 2020-10-16
Allow to change site theme (light <-> dark)


Jokero:
Currently theme is forced without possibility to change it. I prefer to use dark theme in MacOS, but I'm not very happy with dark theme of webpack site (it's too dark for me), so I'll be glad to use the light version (and maybe I'm not alone in this desire).

Toggle can be added next to language switch:
![image](https://user-images.githubusercontent.com/434135/96294695-74728b00-0ff5-11eb-8aae-62099d7457b2.png)

[x] I am willing to work on this issue and submit a pull request.

chenxsan:
I second this, it would be great to give user an option.

Jokero:
@chenxsan Do you maybe know what icon can be used for that?

chenxsan:
@Jokero I saw people use icons like below (from https://dark-mode-class-toggle.glitch.me/):

![moon](https://googlechromelabs.github.io/dark-mode-toggle/demo/moon.png)

![sun](https://googlechromelabs.github.io/dark-mode-toggle/demo/sun.png)

PR #4137, opened 2020-11-08
Ability to change the site theme (device/light/dark)

Jokero:
See https://github.com/webpack/webpack.js.org/pull/4137#issuecomment-757264840 for latest changes.

---

* Quick solution for switching site theme. Fixes https://github.com/webpack/webpack.js.org/issues/4072. Complete solution should be implemented with help of CSS variables
* Icons and text links are aligned vertically a bit (not ideally but better than before)
* Now `<a>` takes up the entire space of the dropdown `<li>` item, so clickable area became bigger

In dark mode:
image

In light mode:
image


chenxsan:
1. Can we remove `device` option? I think a button to toggle between `dark` and `light` is enough to cover most of the use cases. It could simplify the code and the UI a lot.
2. Would be great if you can add either unit tests or e2e tests. We have Jest for unit tests and Cypress for e2e tests in this repository.

Jokero:
1. I actually like "device" option because it looks as the sane default. If user prefers the dark theme, he/she will see it on site. The same works for light option chosen on OS level. For example Youtube is using "device" theme by default too. 
I would like to get more opinions to decide whether to delete it or not
2. I will take a look

chenxsan:
@Jokero I've merged the master branch into it, hope this can alleviate some pain for you. By the way, please let me know if you need help.

Arichy:
What's the progress now?
My eyes are blind because of dark theme documentation😭

montogeek:
Friendly ping

montogeek:
> My eyes are blind because of dark theme documentation

It is automatic according to your operating system configuration.

Arichy:
> > My eyes are blind because of dark theme documentation
> 
> It is automatic according to your operating system configuration.

But I like the dark theme on MacOS except webpack documentation.

chenxsan:
I'm taking over this pull request as there's no respond from @Jokero so far. If you're not ok with it, let me know, @Jokero.

chenxsan:
## Summary of my changes to the original version

### No `device` option

In the original design, there're three options, i.e., `device`, `light` and `dark`. And it defaults to the `device`. I think it's a good practice to follow user's system preference with `prefers-color-scheme`. Unfortunately, the `dark` mode on this site is an afterthought, it works, but I don't think it works as good as the `light` mode. Thus I've removed the `device` option and `prefers-color-scheme`. Now it defaults to `light` mode, user can enable the `dark` mode with one click.

But I want to underline that I'm not against `device` option at all, we can bring it back in future when `dark` mode is ready.

### Use button instead of dropdown

Since we have only two options left now, there's no necessary to use dropdown now, button is much better for such a use case. And it simplifies the code a lot compared to the dropdown version.

montogeek:
Looks really nice, It is there a way to avoid the glitch in the navigation bar when going from light to dark?

chenxsan:
> Looks really nice, It is there a way to avoid the glitch in the navigation bar when going from light to dark?

That seems to be the notification bar hiding itself:

image

I believe we can improve it, but let's keep it in another pull request.

chenxsan:
@montogeek Btw, do you think you still have time to take care the prettier thing? As you can see in this pull request, I've reverted a lot of code formatting because of no prettier, it's sort of a waste of our time. I can take care of it if you don't have the time.

Jokero:
@chenxsan Sorry for disappearing 😄 It also looks nice with one button. Although yellow icon does not really fit navigation bar (attracts too much attention in my opinion).

In original PR I tried to make the site look the same as on production now in dark mode. With new changes there are differences (although this should not be critical).

For example navigation links:
Production
image
This branch
image

chenxsan:
> @chenxsan Sorry for disappearing 😄 It also looks nice with one button. Although yellow icon does not really fit navigation bar (attracts too much attention in my opinion).
> 
> In original PR I tried to make the site look the same as on production now in dark mode. With new changes there are differences (although this should not be critical).
> 
> For example navigation links:
> Production
> image
> This branch
> image

Yeah, the new `[data-theme="dark"]` introduced another CSS specificity overwriting the rule. There might be other places affected.

chenxsan (approved):

chenxsan:
@Jokero Thanks for your contribution!

proteriax:
@chenxsan The button is not accessible on mobile.

chenxsan:
> @chenxsan The button is not accessible on mobile.

It's not available at the moment, we will add it in near future when we figure out where to add.

PR #4139, opened 2020-11-08
docs: update --target type

snitin315:
_describe your changes..._

`--target` can accept multiple values.


snitin315:
CI is not responding 😞

src/content/api/cli.md
```diff
@@ -65,7 +65,7 @@ Note: These are the flags with webpack v4, starting v5 CLI also supports [core f
 | `--progress`        | boolean         | Print compilation progress during build                                                                        |
 | `--help`            | boolean         | Outputs list of supported flags and commands                                                                   |
 | `--output-path, -o` | string          | Output location of the file generated by webpack e.g. `./dist`                                                 |
-| `--target, -t`      | string          | Sets the build target                                                                                          |
+| `--target, -t`      | string[]        | Sets the build target                                                                                          |
```
chenxsan, line 68: Shouldn't it be `string, string[]`? I think `string[]` means array only? But `--target` can accept both types.

snitin315: Actually in the case of CLI, 
 `webpack --target node` it sets `configuration.target = [ 'node' ]` similarly for 
`webpack --target node --target async-node` -> `configuration.target - [ 'node', 'async-node' ]`

snitin315: Same with any flag that can accept multiple values. So it sets always an array of `string[]`

chenxsan: I see your point here, but it's how it works under the hood. Anyway, there're already many `string[]` in that page, seems no one complained about it.

chenxsan (approved):

PR #4141, opened 2020-11-08
docs(configuration): fix publicPath default value

mrzalyaul:
Set `publicPath` to auto

### DIFF
--- a/src/content/configuration/output.md
+++ b/src/content/configuration/output.md
@@ -1056,7 +1056,7 @@ T> It also adds some info about tree shaking to the generated bundle.
 
 ## `output.publicPath`
 
-`string = 'auto'` `function`
+`string` `function`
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.md
```diff
@@ -1055,7 +1056,7 @@ T> It also adds some info about tree shaking to the generated bundle.
 
 ## `output.publicPath`
 
-`string = ''` `function`
+`string = 'auto'` `function`
```
chenxsan, line 1059:
```suggestion
`string` `function`
```
It depends. As per https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.32:

> `output.publicPath` is now `"auto"` by default when supported by target

And here's the code setting default value for it https://github.com/webpack/webpack/blob/master/lib/config/defaults.js#L648, which means it defaults to `auto` in some cases while `''` in other cases.

chenxsan (approved):

chenxsan:
@mrzalyaul Thanks for your contribution to webpack documentation.

PR #4149, opened 2020-11-10
docs(config): add default value for devtool

snitin315:
add a default value for devtool, source code https://github.com/webpack/webpack/blob/master/lib/config/defaults.js#L151


chenxsan (approved):
Would be great if you can share the source code link next time as we can move faster with it then.

PR #4155, opened 2020-11-12
add postinstall script

chenxsan:
A lot of people (including me) don't like to read something like https://github.com/webpack/webpack.js.org/blob/master/.github/CONTRIBUTING.md before contributing, and telling them to RTFM won't do any help.

I can still remember the first time I started this project and got those `_supporters.json not found` errors , it just surprised me. Only after that I read the contributing guide and learned how to fix it.

But we can do better with `postinstall` hook to fetch those datum for us:

1. Files like `_supporters.json` or `_starter-kits` are kind of dependencies as those defined under `devDependencies` and `dependencies`. It makes sense to keep them up to date when we run `yarn install` to install dependencies.
2. We don't change `package.json` or `yarn.lock` a lot, so those `fetch` scripts won't be running frequently, I think the time it adds to the installation stage is acceptable.
3. It **improves the contributor experience** a lot as they can clone this repository and run `yarn install`, `yarn start`, no need to read the contributing guide to get this project running without errors.


snitin315 (dismissed):

chenxsan:
Seems to be a bad idea using `postinstall` hook https://yarnpkg.com/advanced/lifecycle-scripts#a-note-about-postinstall

chenxsan:
I'll see if I can find some other methods.

PR #4163, opened 2020-11-14
apply travis only for master branch

chenxsan:
I've enabled the github actions for Pull Requests here https://github.com/webpack/webpack.js.org/blob/master/.github/workflows/testing.yml, and we'll monitor it for a while before migrating those travis jobs running against pull requests to github actions.

As you can see in this PR that I'm currently keeping the job of master branch building to travis as it won't run frequently. Hopefully we can migrate it too in near future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,56 +10,10 @@ cache:
     - repositories
 
 stages:
-  - Lint
-  - Test
   - Build
 
 jobs:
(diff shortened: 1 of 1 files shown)


chenxsan:
It's been 7 days since we enabled Github Actions. So far so good. I think it's time to merge this PR.

PR #4166, opened 2020-11-14
update production.md

xp44mm:
block code : diff -> javascript


src/content/guides/production.md
```diff
@@ -228,6 +228,6 @@ It is crucial to minimize your CSS for production. Please see the [Minimizing fo
 
 ## CLI Alternatives
 
-Some of what has been described above can also be achieved by using the command line. For example, the `--optimize-minimize` flag will include the `TerserPlugin` behind the scenes. The `--define process.env.NODE_ENV="'production'"` will do the same for the `DefinePlugin` instance described above. And, `webpack -p` will automatically invoke both those flags and thus the plugins to be included.
```
xp44mm, line 231: i run webpack in cli, but:

```bash
# npx webpack --optimize-minimize
[webpack-cli] Unknown argument: --optimize-minimize
```
all those args is unknown.

chenxsan:
`diff` is ok here.

src/content/guides/production.md
chenxsan: If you think this is a bug, please report it on webpack-cli repository, thanks.

xp44mm: If the problem is not very easy to say, just leave it to the future.

montogeek:
As @chenxsan said, those code snippets should be formatted using `diff`.

Issue #4175, opened 2020-11-16
weird/strange problems with rendered docs


evilebottnawi:
__Feature to document__

Looking at the plugins for brotli compression here: https://webpack.js.org/plugins/compression-webpack-plugin/#using-brotli

The example is meant to set the `zlib.constants.BROTLI_PARAM_QUALITY` value to be `11`. When this has been processed by whatever workflow to get it ready for the site, that line has now become:

    [zlib.constants.BROTLI_PARAM_QUALITY]: https://github.com/webpack-contrib/compression-webpack-plugin/blob/master/11,

__Author(s)__

Any

__Additional information__

Ref: https://github.com/webpack-contrib/compression-webpack-plugin/issues/221

[ ] I am willing to work on this issue and submit a pull request.

chenxsan:
Here's what I found so far. 

First, It's expected  as the string `[zlib.constants.BROTLI_PARAM_QUALITY]: 11,` matches [`externalLinkRegex` ](https://github.com/webpack/webpack.js.org/blob/master/src/utilities/process-readme.js#L6):

![image](https://user-images.githubusercontent.com/1091472/99261270-6fecfc80-2857-11eb-89a2-de4e649e7489.png)

Second, It's valid to use the [below link syntax](https://github.github.com/gfm/#link-reference-definition) in github readme:

```
this is [webpack].

[webpack]: https://webpack.js.org
```

this is [webpack].

[webpack]: https://webpack.js.org

So I think we can improve the `externalLinkRegex`.

evilebottnawi:
Thanks for helping

chenxsan:
I just found another similar problem here https://webpack.js.org/plugins/css-minimizer-webpack-plugin/#using-custom-minifier-clean-css:

![image](https://user-images.githubusercontent.com/1091472/99277375-b39c3200-2868-11eb-8782-4d17ce576931.png)


Unfortunately, there're so many variants defined in https://github.github.com/gfm/#link-reference-definitions, it would be almost impossible to match all patterns with a regular expression.

The most reliable solution IMO would be leaving the link reference definition as it is at the moment(we can still add it back when we have enough reports to write some tests). I.e., remove this line https://github.com/webpack/webpack.js.org/blob/master/src/utilities/process-readme.js#L95.

cc @evilebottnawi What do you think here?

PR #4176, opened 2020-11-16
remove externalLinkRegex

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4175


chenxsan:
I'm merging this for the moment and we can add it back later when we have enough reports to support test.

PR #4183, opened 2020-11-17
docs(api): clarify `webpackExports`' function

WofWca:
1. "when using dynamic imports" – "dynamic" is assumed for `import()` isn't it?
2. You have to manually specify which imports are to be bundled, not just tell it not to bundle unused exports.

### DIFF
--- a/src/content/api/module-methods.md
+++ b/src/content/api/module-methods.md
@@ -12,6 +12,7 @@ contributors:
   - EugeneHlushko
   - chenxsan
   - jamesgeorge007
+  - WofWca
 related:
   - title: CommonJS Wikipedia
     url: https://en.wikipedia.org/wiki/CommonJS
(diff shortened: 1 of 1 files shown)


chenxsan:
`dynamic` means `import('./path/to/module').then()`. You can see this PR https://github.com/webpack/webpack/pull/10964 introducing this feature for more.

Regarding the second point, I think you're right.1

WofWca:
> `dynamic` means `import('./path/to/module').then()`

Yeah, the `import()` function. Dynamic.
While `import some from 'thing';` is static, and magic comments don't apply to it.
What am I missing?

chenxsan:
Nope, you are right here. Let's keep the 'dynamic' part.

src/content/api/module-methods.md
```diff
@@ -163,7 +164,7 @@ T> Note that all options can be combined like so `/* webpackMode: "lazy-once", w
 
 T> Note that `webpackInclude` and `webpackExclude` options do not interfere with the prefix. eg: `./locale`.
 
-`webpackExports`: tells webpack to only bundle the used exports of a module when using dynamic imports. It can decrease the output size of a chunk. Available since [webpack 5.0.0-beta.18](https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.18).
+`webpackExports`: tells webpack to only bundle the specified exports of a module. It can decrease the output size of a chunk. Available since [webpack 5.0.0-beta.18](https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.18).
```
chenxsan, line 167:
```suggestion
`webpackExports`: tells webpack to only bundle the specified exports of a dynamically `import()`ed module. It can decrease the output size of a chunk. Available since [webpack 5.0.0-beta.18](https://github.com/webpack/webpack/releases/tag/v5.0.0-beta.18).
```
How about changing to this?

WofWca: The whole section is about dynamic import, so I see this as redundant. But I won't insist. I think you have access to edit my branch, so you can apply the suggestion?

chenxsan: People might copy that paragraph and use it somewhere without the context of `dynamic import` section, hence I prefer to keep it in this case.

chenxsan (approved):

chenxsan:
@WofWca Thank you for your contribution to webpack documentation.

PR #4184, opened 2020-11-18
Remove `webpack-dev-middleware` note from stats documentation

grrizzly:
In previous versions of `webpack-dev-middleware`, the documentation was correct in expecting `stats` to be passed in the options.

Starting with 4.0.0, this is no longer true. Their [changelog](https://github.com/webpack/webpack-dev-middleware/releases/tag/v4.0.0-rc.0) indicates that the `stats` option is now honored from the main configuration. 

In fact, using their latest package will fail to compile in Typescript if you follow the official webpack docs. This PR opts to remove this advice in accordance with modern usage.

### DIFF
--- a/src/content/configuration/stats.md
+++ b/src/content/configuration/stats.md
@@ -16,6 +16,7 @@ contributors:
   - pixel-ray
   - snitin315
   - u01jmg3
+  - grrizzly
 ---
 
 `object` `string`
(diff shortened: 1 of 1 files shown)


evilebottnawi (approved):

chenxsan:
@grrizzly Thank you for your contribution to webpack documentation.

PR #4185, opened 2020-11-18
Fix processAssets stage example

liorgreenb:
The example uses:
`PROCESS_ASSETS_STAGE_ADDITIONS`

which doesn't exist, i suppose it was meant to use PROCESS_ASSETS_STAGE_ADDITIONAL

### DIFF
--- a/src/content/api/compilation-hooks.md
+++ b/src/content/api/compilation-hooks.md
@@ -9,6 +9,7 @@ contributors:
   - wizardofhogwarts
   - EugeneHlushko
   - chenxsan
+  - liorgreenb
 ---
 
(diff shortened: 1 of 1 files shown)


chenxsan:
Webpack does have this `PROCESS_ASSETS_STAGE_ADDITIONS ` https://github.com/webpack/webpack/blob/master/types.d.ts#L1533

liorgreenb:
> Webpack does have this `PROCESS_ASSETS_STAGE_ADDITIONS ` https://github.com/webpack/webpack/blob/master/types.d.ts#L1533

You are right, my mistake

chenxsan:
No worries.

PR #4188, opened 2020-11-18
docs: remove excess backslash escapes

textbook:
These had previously been introduced for *.mdx files (see #2917) but seem to no longer be needed. As a result, the rendered output includes redundant escapes that would break file matching, e.g.:

```js
module.exports = {
  module: {
    rules: [
      { test: /\\.css$/, use: 'css-loader' },
      { test: /\\.ts$/, use: 'ts-loader' }
    ]
  }
};
```

See e.g. [this Stack Overflow question](https://stackoverflow.com/q/64882726/3001761) for resulting confusion.

### DIFF
--- a/src/content/api/node.mdx
+++ b/src/content/api/node.mdx
@@ -11,6 +11,7 @@ contributors:
   - toshihidetagami
   - chenxsan
   - jamesgeorge007
+  - textbook
 ---
 
 webpack provides a Node.js API which can be used directly in Node.js runtime.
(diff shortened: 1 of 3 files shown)


chenxsan:
Thank you for taking your time, seems like mdx fixed the bug this year https://github.com/mdx-js/mdx/pull/991 and we failed to update those changes.

chenxsan (approved):

chenxsan:
@textbook Thanks for your contribution to webpack documentation.

PR #4189, opened 2020-11-20
feat: update Loader section

sarahdayan:
This adds a warning with some possibly unknown side-effects of `ts-loader`, and adds a note on how Babel can replace it altogether with the right preset or plugin.


chenxsan (approved):

chenxsan:
@sarahdayan Thanks for your contribution!

Issue #4162, opened 2020-11-14
Replace proselint with vale


chenxsan:
[Proselint](https://github.com/amperser/proselint) hasn't been updated for more than two years now. I did some research and found there's alternative to use, e.g., [vale](https://docs.errata.ai/vale/about/) . vale is used by github, homebrew, etc. as per https://docs.errata.ai/vale/about#open-source-configurations, and it [performs much better](https://github.com/errata-ai/vale#benchmarks) than proselint with a proselint compatible style https://github.com/errata-ai/proselint.

Therefore I think we can migrate proselint to vale.

PR #4195, opened 2020-11-22
Feature/replace proselint with vale

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4162

## TODO

- [x] replace proselint with vale
- [x] fix some errors in markdown files
- [x] remove Python requirements

Proselint job ran for 2m 46s before according to https://github.com/webpack/webpack.js.org/runs/1436990385?check_suite_focus=true, and now only 45s https://github.com/webpack/webpack.js.org/runs/1437588188?check_suite_focus=true


montogeek:
Does it need all this configuration?

PR #4203, opened 2020-11-24
use github token for CI

chenxsan:
Increases our API rate limit with authentication.


chenxsan:
We need to fix this 503 error from opencollective too, otherwise it would waste a lot of our time.

montogeek:
Nice!

PR #4210, opened 2020-11-26
Feature/bye travis the flash

chenxsan:
Time to get rid of travis from `master` branch as well.


jeffin143:
Nice name for the branch :)

package.json
```diff
@@ -51,7 +51,6 @@
     "lint:links": "hyperlink -c 8 --root dist -r dist/index.html --canonicalroot https://webpack.js.org/ --internal --skip /plugins/extract-text-webpack-plugin/ --skip /printable --skip https:// --skip http:// --skip sw.js > internal-links.tap; cat internal-links.tap | tap-spot",
     "sitemap": "cd dist && sitemap-static --ignore-file=../sitemap-ignore.json --pretty --prefix=https://webpack.js.org/ > sitemap.xml",
     "serve": "npm run build && sirv start ./dist --port 4000",
-    "deploy": "gh-pages -d dist",
```
montogeek, line 54: Uhmm, What if you need to deploy/force it?

chenxsan: In which case do you need to deploy/force it? It's a bad idea to deploy without running against CI. If you must, feel free to add it back.

montogeek: In the case the CI is not available.

PR #4213, opened 2020-11-27
fix header link

chenxsan:
The header link was missing when migrating icon fonts to inlined svg in this PR https://github.com/webpack/webpack.js.org/pull/4192.

image


chenxsan:
I don't know why Vercel deployment failed again.

Issue #4211, opened 2020-11-26
503 error when running `fetch:supporters` on CI


chenxsan:
We got a lot of 503 error from opencollective API when running `yarn fetch:supporters` on CI, and I have to rerun CI when failed. I've had a report on opencollective https://github.com/opencollective/opencollective/issues/3723.

We might need to improve `fetch-supporters.js`.

PR #4214, opened 2020-11-27
fix 503 error from opencollective

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4211

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,6 +42,7 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@3.7.1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OPENCOLLECTIVE_API_KEY: ${{ secrets.OPENCOLLECTIVE_API_KEY }}
           BRANCH: gh-pages
           FOLDER: dist
           CLEAN: true
--- a/src/utilities/fetch-supporters.js
+++ b/src/utilities/fetch-supporters.js
@@ -17,16 +17,11 @@ const absoluteFilename = path.resolve(
   filename
 );
 
-let graphqlEndpoint =
+const graphqlEndpoint =
   'https://api.opencollective.com/graphql/v2';
 
 // https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v2/query/TransactionsQuery.ts#L81
 const graphqlPageSize = 1000;
-
-if (process.env && process.env.CI && process.env.OPENCOLLECTIVE_API_KEY) {
-  // use api key when deploying to production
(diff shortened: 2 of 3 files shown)


src/utilities/fetch-supporters.js
```diff
@@ -9,9 +9,24 @@ const asyncWriteFile = promisify(fs.writeFile);
...
 
-const graphqlEndpoint = 'https://api.opencollective.com/graphql/v2';
+let graphqlEndpoint =
+  'https://api.opencollective.com/graphql/v2';
+
+// https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v2/query/TransactionsQuery.ts#L81
+const graphqlPageSize = 1000;
+
+if (process.env && process.env.CI && process.env.OPENCOLLECTIVE_API_KEY) {
+  // use api key when deploying to production
```
chenxsan, line 27: Currently I'm using a api key from my own opencollective account, you can replace it here https://github.com/webpack/webpack.js.org/settings/secrets/actions

src/utilities/fetch-supporters.js
```diff
@@ -32,10 +47,12 @@ const membersGraphqlQuery = `query account($limit: Int, $offset: Int) {
   }
 }`;
 
-const transactionsGraphqlQuery = `query account($limit: Int, $offset: Int) {
-  account(slug: "webpack") {
-    transactions(limit: $limit, offset: $offset, includeIncognitoTransactions: false) {
-      nodes {
+// only query transactions in last year
+const transactionsGraphqlQuery = `query transactions($dateFrom: ISODateTime, $limit: Int, $offset: Int) {
```
chenxsan, line 51: Now we query only data in last year.

src/utilities/fetch-supporters.js
sokra: Do we really need an API key for that. That should be public data and it worked without before too.

chenxsan: ok, i'll remove it and let's see how the improved version works first.

.github/workflows/deploy.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@3.7.1
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OPENCOLLECTIVE_API_KEY: ${{ secrets.OPENCOLLECTIVE_API_KEY }}
```
montogeek, line 45: Which account holds this key?

chenxsan: No longer used, could be removed. PR here https://github.com/webpack/webpack.js.org/pull/4390

PR #4221, opened 2020-11-29
fix: remove --info-verbosity option as it is not supported in the lat…

HosseinAgha:
Looks like the docs contain a removed cli option: `--info-verbosity`
Checkout https://github.com/webpack/webpack-cli/pull/1550


snitin315 (approved):
Thanks!

chenxsan (approved):

PR #4228, opened 2020-12-01
docs(api): add info regarding --no-watch and --no-devtool

jamesgeorge007:
Information regarding `--no-watch` and `--no-devtool` was missing from the CLI docs.


snitin315 (dismissed):

chenxsan:
Can we remove those `--no-xx` flags from `Flags` section as they're already presented in `Negated Flags` section? Otherwise we're duplicating those `--no-xx` flags.

chenxsan (approved):

PR #4231, opened 2020-12-03
fix lint

chenxsan:
Cleanup eslint rules.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,10 +14,7 @@ module.exports = {
     __DEV__: true,
   },
   rules: {
-    'no-undef': 2,
-    'no-unreachable': 2,
-    'no-unused-vars': 0,
-    'no-console': 0,
+    'no-console': 'off',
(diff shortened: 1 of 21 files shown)


.eslintrc.js
```diff
@@ -39,10 +36,11 @@ module.exports = {
       rules: {
         indent: ['error', 2],
         quotes: ['error', 'single'],
-        'no-undef': 0,
-        'no-constant-condition': 0,
-        'no-useless-escape': 0,
-        'no-dupe-keys': 0,
+        'no-undef': 'off',
```
chenxsan, line 39: number is not intuitive.

cypress/plugins/index.js
```diff
@@ -15,7 +15,7 @@
 /**
  * @type {Cypress.PluginConfig}
  */
-module.exports = (on, config) => {
```
chenxsan, line 18: Those are unused variables.

Issue #4138, opened 2020-11-08
Site does not work properly in IE11


Jokero:
Looks like it's broken after update to webpack 5 which transpiles its internals now to es2015 instead of es5. Is this an intentional decision to not support IE 11?

![image](https://user-images.githubusercontent.com/434135/98463383-ec366e80-21cb-11eb-922b-204b3d01e44a.png)

chenxsan:
Can you check this preview https://webpack-js-org-git-feature-use-bionic.webpack-docs.vercel.app/ under IE 11? It's a preview right before the upgrade of webpack 5.

Jokero:
Syntax error in another place
![image](https://user-images.githubusercontent.com/434135/98463741-a4651680-21ce-11eb-8944-a07157ca4ffb.png)

chenxsan:
Then it might not function well under IE 11 for some time. Feel free to send a PR.

Jokero:
I personally do not use IE11, just noticed it when tested theme switch. If no one complains we can close the issue

evilebottnawi:
@chenxsan Let's use `target: ['web', 'es5']` or adding IE11 to browserslist and update webpack to latest (we have bugs)

chenxsan:
I'll fix it soon.

chenxsan:
I've spent some time tinkering with IE 11 support here https://github.com/webpack/webpack.js.org/pull/4247 but I don't think it will worth it. There're so many packages built without IE 11 support on npm:

```
          path.resolve(__dirname, 'node_modules/punycode'),
          path.resolve(__dirname, 'node_modules/react-banner'),
          path.resolve(__dirname, 'node_modules/react-tiny-popover')
```
And there could be more, hence I suggest dropping support for IE 11.

chenxsan:
I'm closing this for two reasons:

1. Internet Explorer 11 desktop application will go out of support and be retired on June 15, 2022 as per https://techcommunity.microsoft.com/t5/windows-it-pro-blog/internet-explorer-11-desktop-app-retirement-faq/ba-p/2366549
2. We have 0.03% users in the last 30 days using IE

Anyway if anyone want to support it, feel free to file a pull request.

PR #4247, opened 2020-12-05
Bugfix/target ie 11

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4138

## TODO

1. [x] `punycode` (an indirect dependency of `docsearch.js`) containing `=>` in its build [breaks IE 11](https://github.com/bestiejs/punycode.js/issues/86) (compile it again with babel)

    image
2. [ ] `Symbol` is undefined, see https://github.com/webpack/webpack-dev-server/issues/2904


chenxsan:
Too many packages from npm needs transpiled again for IE 11. Hence I suggest dropping support for IE 11.  Anyway, If anyone is interested in support IE 11, you can continue with this PR or create a new one.

PR #4248, opened 2020-12-05
Feature/upgrade docsearch

chenxsan:
- [ ] we need to refactor `Navigation` first.


src/components/Navigation/Navigation.jsx
```diff
@@ -12,9 +12,9 @@ import Dropdown from '../Dropdown/Dropdown';
 import isClient from '../../utilities/is-client';
 
 // Load Styling
-import 'docsearch.js/dist/cdn/docsearch.css';
 import './Navigation.scss';
-import './Search.scss';
+import '@docsearch/css/dist/style.css';
+import 'prismjs/themes/prism-okaidia.css';
```
chenxsan, line 17: @alexander-akait Do you have any idea why `import '@docsearch/css/dist/style.css'` won't import styles while `import 'prismjs/themes/prism-okaidia.css'` does? Seems like the `@` character was causing the problem.

chenxsan:
I'm afraid we can't upgrade docsearch as its styles use css variables which are not supported in IE 11.

src/components/Navigation/Navigation.jsx
chenxsan: I know what's going on here, the `@docsearch/css` package has `sideEffects: false` in `package.json` causing the problem.

src/components/Navigation/Navigation.jsx
```diff
@@ -12,14 +13,12 @@ import Dropdown from '../Dropdown/Dropdown';
 import isClient from '../../utilities/is-client';
 
 // Load Styling
-import 'docsearch.js/dist/cdn/docsearch.css';
 import './Navigation.scss';
-import './Search.scss';
+import '@docsearch/css';
```
chenxsan, line 17: This won't work at the moment, see https://github.com/algolia/docsearch/pull/991

chenxsan:
Superseded by https://github.com/webpack/webpack.js.org/pull/5171

PR #4253, opened 2020-12-05
Feature/enable dependabot

chenxsan:
Time to enable the bot so we don't have to upgrade dependencies manually.


montogeek:
This won't work because the bot can't sign the CLA.

PR #4260, opened 2020-12-06
Feature/watch built size

chenxsan:
`bundlesize` seems to be [unmaintained now](https://github.com/siddharthkp/bundlesize/graphs/contributors), thus we're going to replace it with [`size-limit`](https://github.com/ai/size-limit). There're many alternatives https://github.com/GoogleChrome/lighthouse/issues/10472 but I think `size-limit` fits our usage quite good as the [github action it provides](https://github.com/andresz1/size-limit-action) requires very few configurations.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -126,3 +126,4 @@ jobs:
         uses: andresz1/size-limit-action@v1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          skip_step: build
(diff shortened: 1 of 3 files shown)


chenxsan:
`size-limit` needs to be run for twice to compare the results, one for pull request, one for master branch. Which means we might hit the API limit again.

montogeek:
Where can we see how the bundle changed?

chenxsan:
No, you can't. I've removed it because the action can't be applied on forked pull requests which render it useless.

montogeek:
Yeah, saw it later :(

PR #4276, opened 2020-12-08
Feature/upgrade deps

chenxsan:
Unfortunately the [dependabot can't sign CLA](https://github.com/webpack/webpack.js.org/pull/4275) which means we can only use its pr for notification.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "alex": "^5.1.0",
     "autoprefixer": "^10.0.4",
     "babel-loader": "^8.2.2",
-    "copy-webpack-plugin": "^6.3.2",
+    "copy-webpack-plugin": "^6.4.0",
     "cross-env": "^7.0.2",
     "css-loader": "^5.0.1",
     "css-minimizer-webpack-plugin": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "cross-env": "^7.0.2",
     "css-loader": "^5.0.1",
     "css-minimizer-webpack-plugin": "^1.1.3",
-    "cypress": "^6.0.1",
+    "cypress": "^6.1.0",
     "directory-tree": "^2.2.4",
     "directory-tree-webpack-plugin": "^1.0.2",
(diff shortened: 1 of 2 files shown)


montogeek:
These notifications are more spammy than useful. I would rather just upgrade dependencies every week.

chenxsan:
> These notifications are more spammy than useful. I would rather just upgrade dependencies every week.

Then you would have to read the changelogs spanning multiple versions, it's more painful for me, that's why I prefer those notifications.

snitin315:
We ignore CLA for dependabot in webpack and webpack-cli repo

chenxsan:
> We ignore CLA for dependabot in webpack and webpack-cli repo

I have known that for a while. The reason I don't ask sorka for help is because we can't let the bot merge naively, there's no CI job ensuring it will run fine in development environment.

Every time I upgrade them, I would test those upgrades against development environment locally.

montogeek:
> Then you would have to read the changelogs spanning multiple versions, it's more painful for me, that's why I prefer those notifications.

That's what I do every week in several repos. I can do it :)

montogeek:
> We ignore CLA for dependabot in webpack and webpack-cli repo

It is that ok from the JS Organization perspective?

chenxsan:
> > Then you would have to read the changelogs spanning multiple versions, it's more painful for me, that's why I prefer those notifications.
> 
> That's what I do every week in several repos. I can do it :)

Then feel free to remove https://github.com/webpack/webpack.js.org/blob/master/.github/dependabot.yml.

montogeek:
Let's wait for confirmation about the legal side of it.

snitin315:
> > We ignore CLA for dependabot in webpack and webpack-cli repo
> 
> It is that ok from the JS Organization perspective?

I'm not sure about it. 

/cc @sokra @alexander-akait

sokra:
You can ignore CLA for bots.

In the webpack/webpack repo we use the webpack-bot to override the CLA check so merging is not blocked by that.

PR #4277, opened 2020-12-08
fix custom blockquote

chenxsan:
Seems like 1.0.1 has bug causing issues like below:

image

By the way, we should consider eliminating those exotic packages as the code are not in control of this repository.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "redirect-webpack-plugin": "^1.0.0",
     "remark": "^13.0.0",
     "remark-autolink-headings": "^6.0.1",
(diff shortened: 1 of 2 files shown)


montogeek:
> By the way, we should consider eliminating those exotic packages as the code are not in control of this repository.

This is be design. They should live outside of this repo, I can also add you as contributor to those plugins. What do you think?

chenxsan:
I'm afraid I have no interest maintaining other plugins at the moment.

montogeek:
How is that different from maintaining it here?

montogeek:
In that case I can update the plugins and release new versions with the fixes.

PR #4278, opened 2020-12-08
lazy load NotificationBar

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4414

Currently the `NotificationBar` is included in precompiled HTML, which means browsers would render it before we hide it with js which caused a flash of unstyled content.

In this pull request,

1. we lazy load it which means it's eliminated from the precompiled HTML
2. we hide it by default so returning users who had it hidden won't see it at all, and they don't have to load the extra script as well, only new users would see a flash of unstyled content at the top. We can improve this for new users by moving the notification bar to the right bottom when we have the `UPDATE` button removed.

---

**Note that we'll fix the following issue in another pull request in near future.**

We have people complaining on twitter that the snippets of webpack search results show content of notification bar.

image

I've inspected the search results and did find something like below which is bad for user experience:

image

This pull request would lazy load the notification bar and stop it from included in html.


chenxsan:
This method just works around the issue, not fix it. We might need something like `react-helmet` to specify `<meta name='description' />` for every page.

jeffin143:
Although workaround, But I really like the improvement , and the thought 🚀

montogeek:
I would also prefer a proper solution, and return the correct HTML tags from the server.

PR #4287, opened 2020-12-09
Feature/add UI to activate service worker

chenxsan:
See https://mobile.twitter.com/wSokra/status/1336231899707084800

That issue was caused by the cache first strategy for precaching, see https://github.com/GoogleChrome/workbox/issues/1767 for more details.

## Screenshot

By default, a notification will pop up in the bottom right, and you can click the `UPDATE` button to activate new service work:

image

Of course you can ignore it and collapse that large notification into a small button:

image

This could solve the problem of outdated page, but I think this might be too aggressive  when we update the sites frequently.

Or we can remove service worker if users don't use offline at all, but this would be the last thing I want to do.


chenxsan:
Got a bug:

image

chenxsan:
It's been here for two weeks now, and I believe it's good to merge now.

package.json
```diff
@@ -153,11 +153,14 @@
     "react-dom": "^17.0.1",
     "react-g-analytics": "0.4.2",
     "react-router-dom": "^5.2.0",
+    "react-spring": "next",
```
montogeek, line 156: I guess you want to experiment with this library, but this is a huge library for a single and simple animation. Would be good to do it in pure CSS.

chenxsan: It's the best and fast animation library for React to accomplish what I need since I don't know how to do it in pure CSS. Regarding the size, it's acceptable to me:

image


Anyway, feel free to remove it if you don't like the idea here.

PR #4289, opened 2020-12-09
Feature/upgrade deps

chenxsan:


package.json
```diff
@@ -163,7 +163,7 @@
     "react-dom": "^17.0.1",
     "react-g-analytics": "0.4.2",
     "react-router-dom": "^5.2.0",
-    "react-tiny-popover": "5",
+    "react-tiny-popover": "^6.0.1",
```
chenxsan, line 166: We should consider getting rid of it as it doesn't work well on mobile:

montogeek: What do you suggest?

chenxsan: I'm considering remove the whole block, i.e., https://webpack.js.org/configuration/#options. It's almost not maintainable as far as I see, and it can become outdated quickly because of no maintenance.

montogeek: oh no, we need it, it functions as a single place to quickly explain all the options

PR #4290, opened 2020-12-10
remove gitter-sidecar

chenxsan:
Here's some reasons for removing gitter sidecar:

1. I bet a lot of people have no idea where that icon would lead them to before they click it. Gitter's icon is not that famous as twitter or StackOverflow.
2. I myself have never use that function, and it's kinda useless in my opinion as there are so many questions over there without an answer.
3. Questions or answers in gitter are not indexed by search engines.
4. We can save some bytes by removing it, see https://bundlephobia.com/result?p=gitter-sidecar@1.5.0
5. We can keep the space for something more useful like a notification banner added in https://github.com/webpack/webpack.js.org/pull/4287
6. We already have link to gitter in the footer and webpack core's readme


anshumanv:
I agree

anshumanv (approved):

montogeek:
> Questions or answers in gitter are not indexed by search engines.

They are.

montogeek:
Have we receive any complains about it?

chenxsan:
> > Questions or answers in gitter are not indexed by search engines.
> 
> They are.

I have never seen any gitter Questions or answers show up in the google search results.

I don't receive any complain so far.

montogeek:
I get them all the time, it evens scrolls up exactly to the message in the channel.


> I don't receive any complain so far.

Good for everybody then.

Issue #4200, opened 2020-11-24
Bug in remark-responsive-tables


chenxsan:
When there're more than 2 heads in table, https://github.com/montogeek/remark-responsive-tables won't work:

![image](https://user-images.githubusercontent.com/1091472/100047979-b8f01280-2e4e-11eb-9053-26f7e317a566.png)

![image](https://user-images.githubusercontent.com/1091472/101986184-899f2980-3cc7-11eb-8849-6cd27441b571.png)

PR #4311, opened 2020-12-14
make table responsive

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4200


chenxsan:
> ERROR in ./content/plugins/_eslint-webpack-plugin.md
Module build failed (from ../node_modules/remark-loader/dist/cjs.js):
NonErrorEmittedError: (Emitted value instead of an instance of Error) TypeError: Cannot read property 'value' of undefined

montogeek:
It is also a good practice to give credits to the original author, once again I reiterate that these changes have to be outside of the repo, in external dependencies.

PR #4323, opened 2020-12-18
docs(guides): explain why index.html is in dist/

Flimm:
In the getting started guide, I was confused why a file that was written manually would be included in the `dist/` directory. This pull request explains that eventually `dist/index.html` will be generated manually.

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -143,6 +143,9 @@ __project__
     |- index.js
 ```
 
+T> The observant among you may have noticed that `index.html` is created manually, even though it is now placed in the `dist` directory. Later on in this guide, we will generate `index.html` rather than edit it manually. Once this is done, it should be safe to empty the `dist` directory and to regenerate all the files within it.
+
+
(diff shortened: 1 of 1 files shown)


chenxsan (approved):

src/content/guides/getting-started.md
```diff
@@ -143,6 +143,9 @@ __project__
     |- index.js
 ```
 
+T> The observant among you may have noticed that `index.html` is created manually, even though it is now placed in the `dist` directory. Later on in this guide, we will generate `index.html` rather than edit it manually. Once this is done, it should be safe to empty the `dist` directory and to regenerate all the files within it.
```
montogeek, line 146: I don't like the tone of this oration, would rather have: `You may have noticed...`

Flimm: Fair enough. I did wonder about that wording, but the pull request had already been merged by the time I thought about it.

montogeek: Feel free to submit a correction :)

PR #4324, opened 2020-12-18
Migration guide should link to Asset Modules guide

thw0rted:
Link to new Asset Modules guide in appropriate places.  Per the guide,

> When using the old assets loaders (i.e. file-loader/url-loader/raw-loader) along with Asset Module in webpack 5, you might want to stop Asset Module from processing your assets again as that would result in asset duplication

This means that certain configurations can break under 4->5 migration if they aren't aware of the new Asset Module behavior.

### DIFF
--- a/src/content/migrate/5.md
+++ b/src/content/migrate/5.md
@@ -33,7 +33,7 @@ If you are using webpack version less than 4 please see the [webpack 4 migration
 
 Some Plugins and Loaders might have a beta version that has to be used in order to be compatible with webpack 5.
 
-W> Check related Plugins and Loaders migration guide when upgrading across major versions.  Several commonly-used loaders like `raw-loader`, `url-loader`, and `file-loader` should be replaced using the new built-in [Asset Modules](/guides/asset-modules/) feature.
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

src/content/migrate/5.md
```diff
@@ -33,7 +33,7 @@ If you are using webpack version less than 4 please see the [webpack 4 migration
 
 Some Plugins and Loaders might have a beta version that has to be used in order to be compatible with webpack 5.
 
-W> Check related Plugins and Loaders migration guide when upgrading across major versions.
+W> Check related Plugins and Loaders migration guide when upgrading across major versions.  Several commonly-used loaders like `raw-loader`, `url-loader`, and `file-loader` should be replaced using the new built-in [Asset Modules](/guides/asset-modules/) feature.
```
chenxsan, line 36:
```suggestion
W> Check related Plugins and Loaders migration guide when upgrading across major versions. Several commonly-used loaders like `raw-loader`, `url-loader`, and `file-loader` should be replaced using the new built-in [Asset Modules](/guides/asset-modules/) feature.
```

chenxsan (approved):

PR #4326, opened 2020-12-20
Feature/enable eslint-plugin-mdx

chenxsan:
We have some `mdx` files now, and it's likely that we will migrate all `.md` to `.mdx`, hence enabling `eslint-plugin-mdx` should help catching problems.

Thanks to @JounQin, all issues listed below are gone now.

<del>
There're some issues at the moment:

1. Those code blocks in `.mdx` files won't get linted as `eslint-plugin-mdx` doesn't support it.
2. `vscode-mdx` extension doesn't work as expected in my test.
</del>


montogeek:
>  and it's likely that we will migrate all .md to .mdx

WHy?

chenxsan:
> > and it's likely that we will migrate all .md to .mdx
> 
> Why?

Because the [`Badge`](https://github.com/webpack/webpack.js.org/tree/master/src/components/Badge) component might be used in other markdown files. And we're maintaining two different configuration for `md` and `mdx` https://github.com/webpack/webpack.js.org/blob/master/webpack.common.js#L57, it's not necessary if `mdx` works for both cases.

montogeek:
> Because the Badge component might be used in other markdown files. And we're maintaining two different configuration for md and mdx https://github.com/webpack/webpack.js.org/blob/master/webpack.common.js#L57, it's not necessary if mdx works for both cases.

Agree, but we shouldn't blindly switch everything when MDX features are only used in a few files of hundreds.
It will also impact the performance of the build.

chenxsan:
> > Because the Badge component might be used in other markdown files. And we're maintaining two different configuration for md and mdx https://github.com/webpack/webpack.js.org/blob/master/webpack.common.js#L57, it's not necessary if mdx works for both cases.
> 
> Agree, but we shouldn't blindly switch everything when MDX features are only used in a few files of hundreds.
> It will also impact the performance of the build.

Don't worry, I won't migrate all to `mdx` at once without any specific reason.

JounQin:
Hi, what do you mean by `vscode-mdx extension doesn't work as expected in my test.`?

Maybe related to https://github.com/mdx-js/vscode-mdx/pull/163?

.eslintrc.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
   plugins: ['markdown', 'cypress', 'react-hooks'],
   globals: {
     __DEV__: true,
+    Badge: true, // global MDX component
```
JounQin, line 15: I think it should be added at `md/mdx` section to affect these files only.

.eslintrc.js
```diff
@@ -26,15 +27,23 @@ module.exports = {
     react: {
       version: 'detect',
     },
+    'mdx/code-blocks': true,
```
JounQin, line 30: Same as above IMO.

chenxsan:
> Hi, what do you mean by `vscode-mdx extension doesn't work as expected in my test.`?
> 
> Maybe related to [mdx-js/vscode-mdx#163](https://github.com/mdx-js/vscode-mdx/pull/163)?

It won't highlight those errors:

image

here's a version of md:

image

.eslintrc.js
```diff
@@ -26,15 +27,23 @@ module.exports = {
...
     {
-      files: ['**/*.{md,mdx}'],
+      files: ['**/*.mdx'],
+      extends: ['plugin:mdx/recommended'],
+      rules: {
+        semi: ['off'],
+      },
+    },
+    {
+      files: ['**/*.md'],
```
JounQin, line 42: It can be linted by `eslint-plugin-mdx` as-is now, with `remark-lint` support.

JounQin:
> It won't highlight those errors:

Yes, try to add `eslint.validate` as https://github.com/mdx-js/vscode-mdx/pull/163/files

.eslintrc.js
chenxsan: You're right, I've updated it.

chenxsan:
> > It won't highlight those errors:
> 
> Yes, try to add `eslint.validate` as https://github.com/mdx-js/vscode-mdx/pull/163/files

Cool, did fix the problem.

.eslintrc.js
chenxsan: This repository use `markdownlint` at the moment so I'll leave it as it is.

chenxsan:
@JounQin I just notice a warning on my local computer:

```bash
/Users/sam/Documents/githubRepos/webpack.js.org/src/components/NotificationBar/Notification.mdx
  1:94   warning  `"` can be escaped with `&quot;`, `&ldquo;`, `&#34;`, `&rdquo;`  mdx/no-unescaped-entities
  1:95   warning  `>` can be escaped with `&gt;`                                   mdx/no-unescaped-entities
  1:111  warning  `>` can be escaped with `&gt;`                                   mdx/no-unescaped-entities
```
Here's the content of `Notification.mdx`:

```mdx
Webpack 5 has been officially released. Read our <a href="/blog/2020-10-10-webpack-5-release/">announcement</a>. Not ready yet? Read <a href="https://v4.webpack.js.org/">webpack 4 documentation here</a>.
```
Do you have any idea? I think it's valid syntax to use jsx inside markdown?

JounQin:
@chenxsan Please report it to `eslint-mdx`, I can reproduce it.

JounQin (approved):

PR #4111, opened 2020-10-30
Upgrade Dependencies (2020-11)

chenxsan:
Keeping dependencies up to date is beneficial to the health of this project, hence I suggest doing it every month.

- [x] replace [deprecated `request-promise`](https://github.com/request/request-promise#readme) with [alternative libraries](https://github.com/request/request/issues/3143)
- [x] upgrade react to v17 and enable [the new jsx transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) 
- [x] replace `node-sass` with `sass` as the former [was deprecated](https://github.com/sass/node-sass#node-sass) in favor of the latter
- [x] upgrade remark to 13 (breaking change)

PR #4355, opened 2021-01-01
replace Geomanist with Source Serif Pro

chenxsan:
Currently we have `Geomanist` for headings, `Source Sans Pro` for body texts, and `Source Code Pro` for code, all of the three fonts are sans-serif, which makes the page boring, and there's not enough contrast between body texts and headings.

### DIFF
--- a/src/components/Markdown/Markdown.scss
+++ b/src/components/Markdown/Markdown.scss
@@ -1,3 +1,4 @@
+@use "sass:color";
 @import 'vars';
 @import 'functions';
 @import 'prism-theme';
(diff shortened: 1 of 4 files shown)


montogeek:
Please request reviews, these kind of changes should be discussed.

chenxsan:
Sorry about that, I would be willing to request reviews if you did show up in this repository more frequently :)

If you think this is a bad idea, you can revert it now and we can discuss this.

BTW, do you own the Vercel Account? Can you invite me and @sokra into the team? Last time I wrote an email to @sokra because of https://github.com/webpack/webpack.js.org/pull/4111#issuecomment-731750610, he told me he had no idea who owns the Vercel Account.

PR #4373, opened 2021-01-04
chore: use yarn to run scripts

snitin315:
use `yarn` instead of `npm` to run scripts as the repository uses `yarn` and contributing docs says usage of `yarn` as well.


chenxsan:
I believe it's not necessary to keep consistency here. We recommend using `yarn` in `contributing.md` because there're some yarn only features used in this repository. There's no problem using `npm` in those npm run-scripts. And it would only cause problem if we remove yarn only features in future.

snitin315:
Agreed 👍🏻

PR #4375, opened 2021-01-04
replace :emoji: with utf-8 emoji

chenxsan:
It seems to me that github readme [supports those `:emoji:` syntax](https://github.com/webpack-contrib/eslint-webpack-plugin/blob/master/README.md):

image

While webpack.js.org doesn't:

image

This pull request should show those emojis. As it's run in the build stage, `remark-emoji` won't be included in the client side.

## Preview

https://webpack-js-org-git-fork-chenxsan-feature-enable-emoji-support.webpack-docs.vercel.app/plugins/eslint-webpack-plugin/

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "remark": "^13.0.0",
     "remark-autolink-headings": "^6.0.1",
     "remark-custom-blockquotes": "1.0.0",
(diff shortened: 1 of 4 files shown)


montogeek (approved):

montogeek:
Thanks!

PR #4380, opened 2021-01-04
Update entry-points.md

MananTank:
"Multi-Entry (Shorthand) Syntax" should not be mentioned under "Single Entry (Shorthand) Syntax"

Multi-Entry (Shorthand) Syntax should be mentioned in its own separate section so that reader can understand that there are  shorthand syntax is not only limited to a single entry

### DIFF
--- a/src/content/concepts/entry-points.md
+++ b/src/content/concepts/entry-points.md
@@ -10,6 +10,7 @@ contributors:
   - Zearin
   - chenxsan
   - adyjs
+  - MananTank
 ---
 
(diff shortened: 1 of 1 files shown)


chenxsan:
As you can see in the source, it's `multi-main entry`, not `multi-entry`.

I believe this code:

```js
entry: [a, b]
```
is alternative to:

```js
entry: {
  main: [a, b]
}
```
There's only one single `main` entry here, no multiple entries.

PR #4382, opened 2021-01-05
docs(migrate): improve webpack 5 migration guide

chenxsan:
The current migration guide for webpack 5 is sort of intimidating:

image

 I've made some improvements to it in this pull request:

image

## Preview

https://webpack-js-org-git-fork-chenxsan-feature-migrate-5.webpack-docs.vercel.app/migrate/5/

### DIFF
--- a/src/content/migrate/5.md
+++ b/src/content/migrate/5.md
@@ -16,68 +16,55 @@ This guide aims to help you migrating to webpack 5 when using webpack directly. 
 
 ## Preparations
 
-webpack 5 requires at least Node.js 10.13.0 (LTS).
+Webpack 5 requires at least Node.js 10.13.0 (LTS), so make sure you upgrade your Node.js if you're still running an older one.
 
-T> Using newer Node.js version can improve build performance.
+## Upgrade webpack 4 and its plugins/loaders
 
-### Upgrade webpack and its dependencies
+1. Upgrade `webpack` 4 to the latest available version.
 
-#### Upgrade webpack 4 to the latest available version
+    - When using webpack >= 4, upgrading to the latest webpack 4 version should not require additional guidance.
+    
+    - If you are using webpack version less than 4 please see the [webpack 4 migration guide](/migrate/4/).
 
-When using webpack >= 4, upgrading to the latest webpack 4 version should not require additional guidance.
-If you are using webpack version less than 4 please see the [webpack 4 migration guide](/migrate/4/).
+2. Upgrade `webpack-cli` to the latest available version (when used)
 
-#### Upgrade webpack-cli to the latest available version (when used)
+3. Upgrade all used Plugins and Loaders to the latest available version
 
-#### Upgrade all used Plugins and Loaders to the latest available version
+    Some Plugins and Loaders might have a beta version that has to be used in order to be compatible with webpack 5.
 
-Some Plugins and Loaders might have a beta version that has to be used in order to be compatible with webpack 5.
+### Make sure your build has no errors or warnings
 
-W> Check related Plugins and Loaders migration guide when upgrading across major versions. Several commonly-used loaders like `raw-loader`, `url-loader`, and `file-loader` should be replaced using the new built-in [Asset Modules](/guides/asset-modules/) feature.
+There might be new errors or warnings because of the upgraded versions of `webpack`, `webpack-cli`, Plugins and Loaders. Keep an eye for deprecation warnings during the build.
 
-W>  ExtendedAPIPlugin is removed and the logic is merged into [`APIPlugin`](https://github.com/webpack/webpack/blob/master/lib/APIPlugin.js).
-
-#### Make sure your build has no errors or warnings
-
-There might be new errors or warnings because of the upgraded versions of webpack, webpack-cli, Plugins and Loaders. Keep an eye for deprecation warnings during the build.
-
-You can invoke webpack this way to get stack traces for deprecation warnings to figure out which Plugins and Loaders are responsible.
+You can invoke `webpack` this way to get stack traces for deprecation warnings to figure out which Plugins and Loaders are responsible.
 
 ```bash
 node --trace-deprecation node_modules/webpack/bin/webpack.js
 ```
 
-W> webpack 5 removes all deprecated features. In order to proceed, there should be no webpack deprecation warnings during the build.
+As webpack 5 removes all deprecated features, make sure there's no webpack deprecation warnings during the build in order to proceed.
 
-#### Make sure you are using entry point information from stats
+### Make sure to use `mode`
 
-T> If you are using [HtmlWebpackPlugin](/plugins/html-webpack-plugin/) skip this step.
+Set `mode` to either [`production`](/configuration/mode/#mode-production) or [`development`](/configuration/mode/#mode-development) to make sure that corresponding defaults are set.
 
-When using static HTML or creating HTML in other way, make sure to use entry points from stats JSON to generate `<script>`, `<style>` and `<link>` tags.
-
-If this is not possible, avoid setting `splitChunks.chunks: 'all'` and `splitChunks.maxSize` later in this guide. Note that this is sub-optimal and a workaround.
-
-#### Make sure to use `mode`
-
-Set mode to either [`production`](/configuration/mode/#mode-production) or [`development`](/configuration/mode/#mode-development) to make sure that corresponding defaults are set.
-
-#### Update outdated options
+### Update outdated options
 
 Update the following options to their new version (if used):
 
-- `optimization.hashedModuleIds: true` ↦ `optimization.moduleIds: 'deterministic'`
-- `optimization.namedChunks: true` ↦ `optimization.chunkIds: 'named'`
-- `optimization.namedModules: true` ↦ `optimization.moduleIds: 'named'`
-- `NamedModulesPlugin` ↦ `optimization.moduleIds: 'named'`
-- `NamedChunksPlugin` ↦ `optimization.chunkIds: 'named'`
-- `HashedModuleIdsPlugin` ↦ `optimization.moduleIds: 'deterministic'`
-- `optimization.noEmitOnErrors: false` ↦ `optimization.emitOnErrors: true`
-- `optimization.occurrenceOrder: true` ↦ `optimization: { chunkIds: 'total-size', moduleIds: 'size' }`
-- `optimization.splitChunks.cacheGroups.vendors` ↦ `optimization.splitChunks.cacheGroups.defaultVendors`
-- `Compilation.entries` ↦ `Compilation.entryDependencies`
-- `serve` ↦ `serve` is removed in favor of [`DevServer`](/configuration/dev-server/)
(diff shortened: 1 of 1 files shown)


src/content/migrate/5.md
```diff
@@ -16,68 +16,55 @@ This guide aims to help you migrating to webpack 5 when using webpack directly.
...
 
-#### Upgrade webpack 4 to the latest available version
-
-When using webpack >= 4, upgrading to the latest webpack 4 version should not require additional guidance.
-If you are using webpack version less than 4 please see the [webpack 4 migration guide](/migrate/4/).
-
-#### Upgrade webpack-cli to the latest available version (when used)
-
-#### Upgrade all used Plugins and Loaders to the latest available version
+    - When using webpack >= 4, upgrading to the latest webpack 4 version should not require additional guidance.
```
montogeek, line 25: Might be useful to add a command for it? `npm upgrade webpack@4.44.2`?

src/content/migrate/5.md
```diff
@@ -91,25 +78,27 @@ module.exports = {
...
 
-#### Clean up configuration
+### Clean up configuration
 
 - Consider removing `optimization.moduleIds` and `optimization.chunkIds` from your webpack configuration. The defaults could be better, because they support long term caching in [`production mode`](/configuration/mode/#mode-production) and debugging in [`development` mode](/configuration/mode/#mode-development).
 - When using `[hash]` placeholder in webpack configuration, consider changing it to `[contenthash]`. It is not the same, but proven to be more effective.
 - If you are using Yarn's PnP and the `pnp-webpack-plugin`, we have good news: it is supported by default now. You have to remove it from the configuration.
 - If you are using `IgnorePlugin` with a regular expression as argument, it takes an `options` object now: `new IgnorePlugin({ resourceRegExp: /regExp/ })`.
-- If you are using `node.something: 'empty'` replace it with `resolve.fallback.something: false`.
+- If you are using something like `node.fs: 'empty'` replace it with `resolve.fallback.fs: false`.
```
montogeek, line 99: What about

```suggestion
- If you are using something like `node.{fs,path}: 'empty'` replace it with `resolve.fallback.{fs,path}: false`.
```

?

src/content/migrate/5.md
```diff
@@ -144,58 +133,95 @@ Consider removing defaults:
 - Using `output.path: path.resolve(__dirname, 'dist')`: you can omit it, that's the default.
 - Using `output.filename: '[name].js'`: you can omit it, that's the default.
 
-Need to support an older browser?
+### Need to support an older browser like IE 11?
 
-- By default, webpack will use your browserslist config to decide which code style to emit.
-- Without a browserslist it would emit ES6 style. You can use `target: ["web", "es5"]` to change it to ES5.
+- By default, webpack 5 will use your `browserslist` config to decide which code style to emit for the runtime code.
+- Without a `browserslist` webpack's runtime code uses ES2015 syntax (e.g., arrow function) to build smaller bundles. If your build targets environments that don't support ES2015 syntax (like IE11), you'll need to set `target: ['web', 'es5']` to revert to ES5 syntax.
```
montogeek, line 139:
```suggestion
- Without a `browserslist` webpack's runtime code uses ES2015 syntax (e.g., arrow function) to build smaller bundles. If your build targets environments that don't support ES2015 syntax (like IE11), you'll need to set `target: ['web', 'es5']` to use ES5 syntax.
```

It is not actually a revert, it just depends in your needs.

src/content/migrate/5.md
```diff
@@ -144,58 +133,95 @@ Consider removing defaults:
...
+- Webpack will use it to name files in production and development modes.
+- Webpack 5 will automatically assign useful file names in `development` mode even when not using `webpackChunkName`.
+
+#### Using named exports from JSON modules
+
+This is not supported by the new specification and you will get a warning. Instead of:
 
-Using named exports from JSON modules: this is not supported by the new specification and you will get a warning. Instead of `import { version } from './package.json'; console.log(version);` use `import package from './package.json'; console.log(package.version);`
+```js
+import { version } from './package.json';
```
montogeek, line 158: Very nice!

src/content/migrate/5.md
```diff
@@ -144,58 +133,95 @@ Consider removing defaults:
...
+    The warning message should tell you what can be improved.
+
+- Level 5: __Runtime Errors__.
+
+    This is tricky. You probably have to debug to find the problem. A general advise is difficult here. But we do list some common advices below regarding Runtime Errors:
 
-- Regarding Runtime Errors:
     - `process` is not defined.
         - webpack 5 does no longer include a polyfill for this Node.js variable. Avoid using it in the frontend code.
         - Want to support frontend and browser usage? Use the `exports` or `imports` package.json field to use different code depending on the environment.
```
montogeek, line 203: > frontend and browser usage

It should be server and browser, right?

src/content/migrate/5.md
chenxsan: Not necessary, users should already know how to do it. Otherwise we would teach them to upgrade `webpack-cli` too.

src/content/migrate/5.md
chenxsan: I bet user would search `node.fs`.

src/content/migrate/5.md
chenxsan: Actually it does mean `frontend and browser`, there won't be any problem using `process` server side.

src/content/migrate/5.md
montogeek: Why? It is that common?

src/content/migrate/5.md
montogeek: Yeah, I see it now. Then I would ask what is the difference between frontend and browser, for me browser usage is enough.

Or given the context, which is using server, we could word it differente: client usage.

src/content/migrate/5.md
chenxsan: What I mean here is users would rather search `node.fs` than `node.{fs`. I don't expect google or algolia to find `node.{fs` for us when users type `node.fs`.

src/content/migrate/5.md
chenxsan: I don't agree with it. If users are already using webpack 4 and trying to upgrade webpack 5, they should obviously know how to upgrade to the latest webpack 4. I don't think people don't even know how to upgrade to the latest webpack 4 would read this migration.

src/content/migrate/5.md
montogeek: oh yeah

src/content/migrate/5.md
montogeek: No, we don't know what is the knowledge in teams.

In any case, there is already a link to the v4 migration, so both versions are covered.

src/content/migrate/5.md
chenxsan: Yeah, I think `browser usage` should be enough. It's not me who wrote those texts, so I don't really know the context here.

src/content/migrate/5.md
```diff
@@ -144,58 +133,95 @@ Consider removing defaults:
...
```
chenxsan, line 203:
```suggestion
        - Want to support browser usage? Use the `exports` or `imports` package.json field to use different code depending on the environment.
```

src/content/migrate/5.md
montogeek: yeah, I know, just wanted to clarify it :)

src/content/migrate/5.md
montogeek: bummer, suggestion doesn't work

montogeek (approved):

src/content/migrate/5.md
```diff
@@ -239,6 +261,6 @@ The changes to webpack internals such as: adding types, refactoring code and met
 webpack 5 ships with built-in [`this.getOptions`](/api/loaders/#thisgetoptionsschema) method available in loader context. This is a breaking change for loaders that had been using `getOptions` method from previously preferred [schema-utils](https://github.com/webpack/schema-utils):
 
 - `this.getOptions` is available since webpack 5
-- Instead of JSON5 it supports JSON as a query string: `?{arg:true}` ↦ `?{"arg":true}`. Using JSON5 should be considered and documented as deprecated in favor of JSON in respective Loader's documentation.
+- Instead of JSON5 it supports JSON as a query string: `?{arg:true}` → `?{"arg":true}`. Using JSON5 should be considered and documented as deprecated in favor of JSON in respective Loader's documentation.
```
jamesgeorge007, line 264:
```suggestion
- Instead of JSON5 it supports JSON as a query string: `?{arg:true}` → `?{"arg":true}`. Using JSON5 should be considered and documented as deprecated in favor of JSON in the respective Loader's documentation.
```

src/content/migrate/5.md
```diff
@@ -144,58 +133,95 @@ Consider removing defaults:
...
+
+    You probably get a lot of deprecation warnings. This is not directly a problem. Plugins need time to catch up with core changes. Please report these deprecations to the plugins. These deprecations are only warnings and the build will still work with only minor drawbacks (like less performance).
+
     - You can hide deprecation warnings by running node with `--no-deprecation` flag, e.g.: `node --no-deprecation node_modules/webpack/bin/webpack.js`. This should only be a temporary workaround.
     - Plugins and Loaders contributors can follow the advises in the deprecation messages to improve the code.
-- Regarding Performance issues:
+
+- Level 7: __Performance issues__.
+
+    Usually performance should improve with webpack 5, but there are also a few cases where performance get worse.
```
jamesgeorge007, line 221:
```suggestion
    Usually, performance should improve with webpack 5, but there are also a few cases where performance gets worse.
```

src/content/migrate/5.md
```diff
@@ -144,58 +133,95 @@ Consider removing defaults:
...
-- Level 2: webpack exits with an error. The error message should tell you what needs to be changed.
-- Level 3: Build Errors. The error message should have a `BREAKING CHANGE:` note.
-- Level 4: Build Warnings. The warning message should tell you what can be improved.
-- Level 5: Runtime Errors. This is tricky. You probably have to debug to find the problem. A general advise is difficult here.
-- Level 6: Deprecation Warnings. You probably get a lot of deprecation warnings. This is not directly a problem. Plugins need time to catch up with core changes. Please report these deprecations to the plugins. These deprecations are only warnings and the build will still work with only minor drawbacks (like less performance).
-- Level 7: Performance issues. Usually performance should improve with webpack 5, but there are also a few cases where performance get worse.
+- Level 1: __Schema validation fails__.
+    
+    Configuration options have changed. There should be a validation error with a `BREAKING CHANGE:` note, or a hint which option should be used instead.
```
jamesgeorge007, line 183:
```suggestion
    Configuration options have changed. There should be a validation error with a `BREAKING CHANGE:` note, or a hint of which option should be used instead.
```

jmetev1:
These docs suggest `Upgrade all used Plugins and Loaders to the latest available version`. Is that the best way to go? Many of the latest loaders and plugins have peer dependencies of webpack 5 so npm will try and prevent this. @chenxsan ?

PR #4386, opened 2021-01-05
docs(api): add --watch-options-stdin and --no-watch-options-stdin

snitin315:
add `--watch-options-stdin` and `--no-watch-options-stdin`


src/content/api/cli.md
```diff
@@ -44,47 +44,49 @@ webpack-cli offers a variety of commands to make working with webpack easy. By d
...
+| `--config-name`         | string[]        | Name of the configuration to use                                                                               |
+| `--name`                | string          | Name of the configuration. Used when loading multiple configurations                                           |
+| `--color`               | boolean         | Enable colors on console                                                                                       |
+| `--merge, -m`           | boolean         | Merge two or more configurations using webpack-merge e.g. `-c ./webpack.config.js -c ./webpack.test.config.js` |
+| `--env`                 | string[]        | Environment passed to the configuration when it is a function                                                  |
+| `--progress`            | boolean, string | Print compilation progress during build                                                                        |
+| `--help`                | boolean         | Outputs list of supported flags and commands                                                                   |
```
chenxsan, line 66:
```suggestion
| `--help`                | boolean         | Output list of supported flags and commands                                                                   |
```
Let's keep consistency. There're many other similar problems, please update them as well.

src/content/api/cli.md
```diff
@@ -44,47 +44,49 @@ webpack-cli offers a variety of commands to make working with webpack easy. By d
...
+| `--output-path, -o`     | string          | Output location of the file generated by webpack e.g. `./dist`                                                 |
+| `--target, -t`          | string[]        | Sets the build target                                                                                          |
```
chenxsan, line 68:
```suggestion
| `--target, -t`          | string[]        | Set the build target                                                                                          |
```

E.g., this one.

src/content/api/cli.md
snitin315: hmm updating in cli too https://github.com/webpack/webpack-cli/pull/2307

chenxsan:
Do we reach an agreement on adding `s` or not? From what I can tell, webpack core contains both forms:

image

I would prefer the no `s` version because it's simple, and it's correct as well.

snitin315:
> I would prefer the no s version because it's simple, and it's correct as well.

Not yet hence I'm here only adding `--watch-options-stdin` and `--no-watch-options-stdin`.

Can you create a specific issue about it?

chenxsan (approved):

PR #4409, opened 2021-01-08
fix(docs): Allow the scrolling of the Sidebar

smroe:
Removed the CSS styling that prevents the scrolling of the Sidebar component. 

When long content is displayed in the Sidebar component, for example https://webpack.js.org/guides/tree-shaking/ with some of the tree options opened, the end of the Sidebar menu is inaccessible as it doesn't scroll with the page.

### DIFF
--- a/src/components/Sidebar/Sidebar.scss
+++ b/src/components/Sidebar/Sidebar.scss
@@ -16,9 +16,7 @@
 .sidebar__inner {
   padding: 1.5em;
   position: sticky;
-  top: -1px;
   overflow: scroll;
-  max-height: 100vh;
   // To hide scrollbar in Firefox
   scrollbar-width: none;
   // To hide scrollbar in Chrome


chenxsan:
You have to scroll inside the sidebar :)

If we remove those css rules, we would have something like this:

image

I'm afraid it's a bad idea.

smroe:
But leaving it as is there's no way to see or access the bottom of the menu, which is a worst idea! What's the point of having a menu that length if you can't access the options? I guess search engines will see them, but currently it's a case of styling over usability.

chenxsan:
> But leaving it as is there's no way to see or access the bottom of the menu, which is a worst idea! What's the point of having a menu that length if you can't access the options? I guess search engines will see them, but currently it's a case of styling over usability.

Just like I said, you have to move your mouse into the sidebar area to scroll.

chenxsan:
Here's a video you can check: 

https://user-images.githubusercontent.com/1091472/104013060-df45f380-51eb-11eb-94e0-53c5bf5bbb5a.mp4

smroe:
> Here's a video you can check:

How are you scrolling that? I don't get any scrollbars or any other means to move the panel!

chenxsan:
> > Here's a video you can check:
> 
> How are you scrolling that? I don't get any scrollbars or any other means to move the panel!

Can you share which browser and its version are you running?

smroe:
I'm using Firefox V84.0.2 (64-bit), but the same issue occurs on Chrome Version 87.0.4280.88 (Official Build) (64-bit).

chenxsan:
I've tested the site against the latest Firefox and the latest Chrome on macOS, sorry but I can't reproduce it.


https://user-images.githubusercontent.com/1091472/104015064-1538a700-51ef-11eb-9317-a7d023dc036e.mp4


https://user-images.githubusercontent.com/1091472/104015072-179b0100-51ef-11eb-8d0b-6622fdfa7cab.mp4

smroe:
How are you scrolling the view, are you using a scroll wheel on your mouse? Looking at your movies it would appear so. If the user doesn't have a mouse with a scroll wheel, how do they scroll the panel?

chenxsan:
Yes, I was using a scroll wheel on my mouse. Let's say you have only keyboard, it's possible to use the TAB key to get into the sidebar area and using UP/DOWN keys to scroll, it takes time, but it should work.

If you take a look at sites like next/vue/etc. with sidebar, they all share this same problem.

smroe:
> If you take a look at sites like next/vue/etc. with sidebar, they all share this same problem.

It's good that you agree that it's a problem. Just because everyone else does it doesn't mean that it's correct. Visitors to a site shouldn't be penalized if they don't have a scroll wheel mouse. 

The problem is larger than just the problems scrolling the menu. As the menu starts to scroll with the window then stops it appears that you have reached the end of the menu when you haven't. In my case, at https://webpack.js.org/guides/installation/, the menu scrolls with the page until the Public Path item and then stops. As there is no prompt to suggest that there are any more other menu options and the menu has moved to a point then stopped, it appears that you have reached the end of the menu, but there are unseen menu options. Close the selected menu and additional menu items appear. 

I guess an alternative would be for the menu panel to have it's own scrollbar.

chenxsan:
I agree, a scrollbar would be nice. Do you want to update your pull request so we can see how it works with a sidebar scrollbar?

dangayle:
Maybe the solution isn't to fix the scrolling. Perhaps a sub menu squeezed in to the right of the sidebar can handle the smaller items. Or you remove them entirely and add a fully functioning sub menu to each of the main pages. This does require users to drill down more, but experience has shown that this is often an acceptable compromise.

I also worry about the accessibility of all of this. If a user without a scroll wheel cannot access it all, then that could be an accessibility issue.

PR #4412, opened 2021-01-09
customize Link for mdx

chenxsan:
Improve website performance.

## Why?

Let's take this page for example https://webpack.js.org/api/node/.

Open devtools and click the link in the screenshot below:

image

Here's the network panel:

image

Even with service worker, we still have a lot of requests.

Now let's open the preview of this pull request https://webpack-js-org-git-fork-chenxsan-feature-use-react-route-5aabb7.webpack-docs.vercel.app/api/node/, and click the same link:

image

That's impressive isn't it?

## What's going on

This website is a Single Page React Application with precompiled HTML, but links in body are rendered as `<a>` elements, which would cause browser to load a whole new html page. When we render them as `Link` from `react-router`, it would dynamic load the needed part instead of loading a new html page, and it's possible with `MDX`. Guess this is a good reason to migrate from `.md` to `.mdx`  because we have so many internal links.

### DIFF
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -29,6 +29,7 @@ import PageNotFound from '../PageNotFound/PageNotFound';
 import Vote from '../Vote/Vote';
 import Organization from '../Organization/Organization';
 import Badge from '../Badge/Badge.js';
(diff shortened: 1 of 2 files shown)


montogeek (approved):

montogeek:
Thanks!

PR #4422, opened 2021-01-11
fix(app): Deprecated Sass api

chenxsan:
See the changelog of `sass` here https://github.com/webpack/webpack.js.org/pull/4402:

> Deprecate passing non-% numbers as lightness and saturation to hsl(), hsla(), color.adjust(), and color.change(). This matches the CSS specification, which also requires % for all lightness and saturation parameters. See [the Sass website][color-units] for more details.


snitin315 (approved):

montogeek (approved):

montogeek:
Thanks!

PR #4435, opened 2021-01-14
Don't scroll to top if link was opened in new tab

codeclown:
If user cmd+clicks (or ctrl+clicks) a link in the sidebar, the linked page opens in a new page, and the current tab should not scroll to the top.


chenxsan:
@codeclown Make sense, if you don't mind, I'd like to push some end to end tests in this pull request to prevent potential regression in the future.

codeclown:
Go for it! Thanks!

chenxsan (approved):
@codeclown Thanks for your contribution to webpack.

PR #4442, opened 2021-01-16
fix(docs): expose errors details in loader example

JuniorTour:
### errors detail not exposed:
loader-err-2

### errors detail exposed clearly:
loader-err-1


chenxsan (approved):
Makes sense.

Issue #4458, opened 2021-01-20
Webpack documentation completely broken on Firefox


pmrotule:
# Bug report

**What is the current behavior?**

The page layout of the documentation is completely broken on Firefox.

<img width="755" alt="image" src="https://user-images.githubusercontent.com/10983258/105140938-b3364680-5af8-11eb-9006-d07ea7ae0d76.png">

I personally added "flex: 0" to `.gitter` to fix the issue on my end.

<img width="357" alt="image" src="https://user-images.githubusercontent.com/10983258/105141240-250e9000-5af9-11eb-88d5-32f56a27a283.png">

**If the current behavior is a bug, please provide the steps to reproduce.**

Go to https://webpack.js.org/plugins/split-chunks-plugin/ on Firefox.

**What is the expected behavior?**

Page not being broken like it is on Chrome or Safari.

**Other relevant information:**
webpack version:  Documentation of Webpack 5
Node.js version: N/A
Operating System: macOS Big Sur v11.1
Additional tools: Firefox v84.0.2

alexander-akait:
@chenxsan Feel free to close, I think we need focus on changing cache strategy, a lot of problems with current strategy

pmrotule:
@chenxsan It worked, thanks 🙌 
No more gitter indeed. Feel free to close the issue.

chenxsan:
Let's keep it open until Network-first strategy is implemented.

Issue #4340, opened 2020-12-24
Cache First strategy is bad for a documentation site.


chenxsan:
We have workbox to enable offline feature at the moment, unfortunately, the `precaching` function [only supports Cache-First](https://github.com/GoogleChrome/workbox/issues/1767), and I had to introduce a refresh button https://github.com/webpack/webpack.js.org/pull/4287 to work around the problem of stale page. Anyway, I think we should improve it, maybe change the strategy to others.

PR #4459, opened 2021-01-20
Use Network-first strategy for service worker

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4340
closes https://github.com/webpack/webpack.js.org/issues/4458

We will do this in multiple steps:

1. Implement `NetworkFirst` strategy for those webpack generated assets so service worker will always fetch them from network first and only fallback to cache when offline
2. Since `NetworkFirst` strategy is used, we don't really need to activate the new service worker as soon as possible, i.e., the `UPDATE` button at the right bottom can be removed. But we'll do it after maybe two months to make sure most users have updated to the latest service worker.

cc @alexander-akait


montogeek:
Current Cache first strategy is good for which kind of sites?

montogeek (dismissed):

montogeek:
Would be also worth to run a poll to see how many actually use the site offline.

chenxsan:
> Current Cache first strategy is good for which kind of sites?

Actually I have no idea why workbox precaching would default to Cache first without alternative strategies. It's a nightmare for a documentation site as users have to click the UPDATE button in the page all the time when we update the site frequently.

> Would be also worth to run a poll to see how many actually use the site offline.

Yeah, once I wanted to remove the offline function but since it's been introduced a long time ago, I have no idea whether we should do it.

Network first strategy would be the most safe option for a documentation site as users can still check it even when they're offline.

alexander-akait:
> Network first strategy would be the most safe option for a documentation site as users can still check it even when they're offline.

:+1:

chenxsan:
Let's ship it now, we'll see how it works before removing that UPDATE button.

PR #4462, opened 2021-01-21
docs(api): document __webpack_is_included__

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4451


montogeek:
Interesting function.

PR #4474, opened 2021-01-23
Update 5.md

Neurone:
`--dev` is deprecated. npm 6 suggest to use `--also=dev` and npm 7 suggests to use `--include=dev`. I simply suggest to remove that flag because if webpack is already in devDependencies, it will be updated in any case.

Even using `--save-dev` works but it will also move webpack from `dependencies` to `devDependencies`, and maybe it's not always what the developer wants.

### DIFF
--- a/src/content/migrate/5.md
+++ b/src/content/migrate/5.md
@@ -88,7 +88,7 @@ Now let's upgrade webpack to version 5:
 
 - npm: `npm install webpack@latest`
 
-- Yarn: `yarn add webpack@latest -D`
+- Yarn: `yarn add webpack@latest`
 
 ### Clean up configuration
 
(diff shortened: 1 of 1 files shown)


src/content/migrate/5.md
```diff
@@ -86,7 +86,7 @@ T> webpack 5 removes these options from the configuration schema and will always
 
 Now let's upgrade webpack to version 5:
 
-- npm: `npm install webpack@latest --dev`
+- npm: `npm install webpack@latest`
 
 - Yarn: `yarn add webpack@latest -D`
```
chenxsan, line 91:
```suggestion
- Yarn: `yarn add webpack@latest`
```
Let's remove `-D` for `yarn` as well to keep consistency with `npm` version.

chenxsan (approved):

PR #4476, opened 2021-01-24
docs(configuration): document webpackIgnore for commonjs

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4470

### DIFF
--- a/src/content/api/module-methods.mdx
+++ b/src/content/api/module-methods.mdx
@@ -172,7 +172,7 @@ T> Note that `webpackInclude` and `webpackExclude` options do not interfere with
 The goal of CommonJS is to specify an ecosystem for JavaScript outside the browser. The following CommonJS methods are supported by webpack:
 
 
-### `require`
+### require
 
 ```typescript
 require(dependency: String);
@@ -185,7 +185,10 @@ var $ = require('jquery');
 var myModule = require('my-module');
 ```
 
+It's possible to enable magic comments for `require` as well, see [`module.parser.javascript.commonjsMagicComments`](/configuration/module/#moduleparserjavascriptcommonjsmagiccomments) for more.
+
 W> Using it asynchronously may not have the expected effect.
+
 
 
 ### `require.resolve`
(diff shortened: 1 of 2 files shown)


src/content/api/module-methods.mdx
```diff
@@ -172,7 +172,7 @@ T> Note that `webpackInclude` and `webpackExclude` options do not interfere with
 The goal of CommonJS is to specify an ecosystem for JavaScript outside the browser. The following CommonJS methods are supported by webpack:
 
 
-### `require`
+### require
```
montogeek, line 175: Don't see the need to make it MDX.

src/content/configuration/module.mdx
```diff
@@ -107,6 +107,51 @@ module.exports = {
...
+};
+```
+
+It's allowed to configure those options in [`Rule.parser`](/configuration/module/#ruleparser) as well to target specific modules.
+
+#### module.parser.javascript.commonjsMagicComments
+
+Enable [magic comments](/api/module-methods/#magic-comments) support for CommonJS.
+
+- Type: `boolean`
```
montogeek, line 134: We usually put the types below the title, Why this change?

montogeek (approved):

src/content/api/module-methods.mdx
chenxsan: See this pull request https://github.com/webpack/webpack.js.org/pull/4412, with mdx we can customize the link with `Link` from `react-router` which would improve the page performance a lot because users don't have to load the full page again clicking on a link.

montogeek: Right :)

src/content/configuration/module.mdx
chenxsan: It's explicit, and I believe it's much readable. Moreover, with explicit format we could possibly customize those complex types with [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element in the future.

PR #4494, opened 2021-01-28
docs(configuration): document lazyCompilation

chenxsan:
closes https://github.com/webpack/webpack.js.org/issues/4471 and https://github.com/webpack/webpack.js.org/issues/4520


src/content/configuration/experiments.mdx
```diff
@@ -35,6 +35,41 @@ module.exports = {
...
+Compile entrypoints and dynamic `import`s only when they are accessed.
+
+- Type
+  - `boolean`
+  - `object`
+    ```ts
+    {
+      // define a custom backend
+      backend?: ((
+        compiler: webpack.Compiler,
```
montogeek, line 54: Why in TS?

chenxsan: It's almost impossible to document such a complex type without typescript.

PR #4518, opened 2021-02-02
enable prettier

chenxsan:
Supersede https://github.com/webpack/webpack.js.org/pull/3109 as there's no news so far from @montogeek https://github.com/webpack/webpack.js.org/pull/3109#issuecomment-752839686, thus I'll take care of it.

I've made some changes to `.markdownlint.json` as prettier doesn't provide options for markdown formatting.


alexander-akait:
It is awesome

src/components/Configuration/components.js
```diff
@@ -28,17 +28,17 @@ const Card = ({ body }) => {
...
 Card.propTypes = {
-  body: PropTypes.node
+  body: PropTypes.node,
 };
 export class Details extends Component {
   static propTypes = {
     url: PropTypes.string,
-    children: PropTypes.arrayOf(PropTypes.node)
-  }
+    myChilds: PropTypes.arrayOf(PropTypes.node),
```
montogeek, line 36: Why this change?

src/content/api/cli.md
```diff
@@ -293,21 +293,21 @@ You can also pass multiple values:
...
 
-You can merge two or more different webpack configurations with the help of `--merge`: 
+You can merge two or more different webpack configurations with the help of `--merge`:
 
 ```bash
 npx webpack --config ./first.js --config ./second.js --merge
 ```
 
-__Print result of webpack as a JSON__
+**Print result of webpack as a JSON**
```
montogeek, line 304: This was done by prettier?

montogeek: It doesnt change anything visually.

src/components/Configuration/components.js
chenxsan: image

src/content/api/cli.md
chenxsan: It is. And there's no option to force it to use `__`, that's why I've made changes to `.markdownlint.json`.

src/components/Configuration/components.js
montogeek: Is this a new warning?
Or you want to fix it in this PR?

chenxsan: Not a new one, before there's a `// eslint-disable-next-line` to disable the warning, see https://github.com/webpack/webpack.js.org/blob/master/src/components/Configuration/Configuration.jsx#L11

prettier.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  singleQuote: true,
+  semi: true,
```
alexander-akait, line 3: You can omit this, it is `true` by default

prettier.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  singleQuote: true,
```
alexander-akait, line 2: Why? You can use JSX, so using `"` is preferred here

prettier.config.js
chenxsan: Indeed. But let's keep it as it's explicit and can serve as a guide for anyone who want to check. I believe not everyone knows what're the default values of prettier's options.

prettier.config.js
chenxsan: I have no preference here. I think `singleQuote` was used more often in code before, hence I set it `true`.

chenxsan: Updated.

chenxsan: Unfortunately there's too many errors hard to fix with `singleQuote: false`, thus I'm reverting the change.

alexander-akait: We can return to this latest, let's integrate prettier firstly

PR #4524, opened 2021-02-03
optimize Badge style

chenxsan:
The current `Badge` is too large compared to other body texts when we use it a lot:


snitin315 (approved):

montogeek (approved):

montogeek:
Thanks!

PR #4537, opened 2021-02-03
clean up cross-env

chenxsan:
We can use `--node-env` to set `process.env.NODE_ENV` with the latest `webpack-cli`, i.e.,`cross-env` is no longer needed as per https://github.com/webpack/webpack-cli/issues/2362#issuecomment-771776945


snitin315 (approved):
💯

PR #4565, opened 2021-02-08
removed semicolon from webpack config file

GajenderTyagi:
_There were commas after last properties in webpack config files, they are removed now as property at last dont need a comma and it breaks the rule of a valid JSON_

### DIFF
--- a/src/content/guides/getting-started.md
+++ b/src/content/guides/getting-started.md
@@ -244,8 +244,8 @@ module.exports = {
   entry: './src/index.js',
   output: {
     filename: 'main.js',
-    path: path.resolve(__dirname, 'dist'),
-  },
+    path: path.resolve(__dirname, 'dist')
+  }
 };
 ```
 


chenxsan:
`webpack.config.js` is a JavaScript file, so trailing comma is actually valid.

GajenderTyagi:
> `webpack.config.js` is a JavaScript file, so trailing comma is actually valid.

Yes you are correct @chenxsan, I thought it might be easy for newbies to understand the code as most IDE gives warning on such code. I'll cancel this PR if this is not necessary. Let me know.

chenxsan:
Could you share what kind of warnings IDE would give regarding those trailing commas?

GajenderTyagi:
image

I'm Using semistandard in VS code

chenxsan:
I can confirm `standard` and `semistandard` report errors with trailing commas, however others like Airbnb JavaScript Style Guide, https://google.github.io/styleguide/jsguide.html#features-arrays-trailing-comma require those trailing commas. It would be impossible to satisfy all in those cases.

snitin315:
There is no harm with trailing commas.

GajenderTyagi:
Ok thanks for you reviews.

PR #4586, opened 2021-02-14
use NavLink for SidebarItem

chenxsan:
This website is a Single Page React Application with precompiled HTML, but links in sidebar are rendered as [`<a>`](https://github.com/webpack/webpack.js.org/blob/master/src/components/SidebarItem/SidebarItem.jsx#L31) elements at the moment, which would cause browser to load a whole new html page. When we render them as `NavLink`, it would dynamic load the needed part instead of loading a new html page which could load the page faster.

### DIFF
--- a/src/components/SidebarItem/SidebarItem.jsx
+++ b/src/components/SidebarItem/SidebarItem.jsx
@@ -38,12 +38,6 @@ export default class SidebarItem extends Component {
     );
   }
 
-  scrollTop(event) {
-    if (!event.metaKey && !event.ctrlKey) {
-      window.scrollTo(0, 0);
-    }
-  }
-
   render() {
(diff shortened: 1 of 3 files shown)


chenxsan:
We're having a problem here https://github.com/ReactTraining/react-router/issues/394.

src/components/SidebarItem/SidebarItem.jsx
```diff
@@ -74,7 +70,6 @@ export default class SidebarItem extends Component {
           key={this.props.url}
           className={`${block}__title`}
           to={this.props.url}
-          onClick={this.scrollTop}
```
chenxsan, line 77: `useEffect` in `Page.jsx` will handle it, so this is no longer needed.

PR #4593, opened 2021-02-15
docs: add info about the parallelism with multiple configurations

snitin315:
Refers https://github.com/webpack/webpack.js.org/issues/4587

Add information about the `parallelism` option on the configuration array.


snitin315:
fixing lint.

chenxsan (approved):

PR #4595, opened 2021-02-15
docs(configuration): add options.dependencies

snitin315:
add options.dependencies

fixes https://github.com/webpack/webpack.js.org/issues/4587


src/content/configuration/other-options.md
```diff
@@ -297,7 +297,15 @@ module.exports = {
 
 W> `cache.idleTimeoutForInitialStore` is only available when [`cache.store`](#cachestore) is set to `'pack'`
 
-## `ignoreWarnings`
+## dependencies
+
+`[string]`
+
+It is an array of [`name`](#name) of other configurations in the same array of configuration. All dependencies need to be compiled before this configuration starts compiling.
```
chenxsan, line 304:
```suggestion
A list of [`name`](#name) defining all sibling configurations it depends on. Dependent configurations need to be compiled first.
```

src/content/configuration/other-options.md
```diff
@@ -297,7 +297,15 @@ module.exports = {
...
+
+In watch mode dependencies will invalidate the compiler when the dependency has changed and when a dependency is currently compiling or invalid this config will not compile until the dependency is done.
```
chenxsan, line 306:
```suggestion
In watch mode dependencies will invalidate the compiler when:

 1. the dependency has changed
 2. a dependency is currently compiling or invalid
 
Remember that this configuration will not compile until the dependency is done.
```

chenxsan (approved):

PR #4662, opened 2021-02-26
Fix https settings for own certs

vladimirmartsul:
Hi! If you want to use own https certificate for hmr with devServer you should pass cert & key in the devServer.https object not just root devServer object.

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -582,10 +582,11 @@ With the above setting, a self-signed certificate is used, but you can provide y
 module.exports = {
   //...
   devServer: {
-    https: true,
-    key: fs.readFileSync('/path/to/server.key'),
-    cert: fs.readFileSync('/path/to/server.crt'),
-    ca: fs.readFileSync('/path/to/ca.pem'),
+    https: {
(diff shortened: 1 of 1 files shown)


chenxsan:
I believe it's [a change](https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0) in `webpack-dev-server` v4 which is still in beta. The current documentation targets the latest released version, i.e., v3.

vladimirmartsul:
> I believe it's [a change](https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-beta.0) in `webpack-dev-server` v4 which is still in beta. The current documentation targets the latest released version, i.e., v3.

I see the same https object config one paragraph above at the https://webpack.js.org/configuration/dev-server/#devserverhttp2 so let the bring docs consistent?

snitin315:
Lets allow and add a badge for beta version.

chenxsan:
@vladimirmartsul That would make sense. I've tested against `webpack-dev-server` v3, the object version works as well:

image

chenxsan (approved):
Thanks for your contribution!

Issue #4658, opened 2021-02-25
Remove usage of `clean-webpack-plugin` in favor `output.clean`


alexander-akait:
__Feature to document__

Now we can use https://webpack.js.org/configuration/output/#outputclean

https://webpack.js.org/guides/output-management/

__Author(s)__

Any

__Additional information__

Maybe we have `clean-webpack-plugin` plugin in other places, need check

[ ] I am willing to work on this issue and submit a pull request.

snitin315:
On it.

chenxsan:
@snitin315 Do you need help?

snitin315:
@chenxsan thanks for proposing, I have the bandwidth to work on it today. Sorry for the delay 😄.

PR #4670, opened 2021-03-01
docs(guides): remove `clean-webpack-plugin` in favor of `output.clean`

snitin315:
Resolve https://github.com/webpack/webpack.js.org/issues/4658

### DIFF
--- a/src/content/guides/development.md
+++ b/src/content/guides/development.md
@@ -30,7 +30,6 @@ Let's start by setting [`mode` to `'development'`](/configuration/mode/#mode-dev
 ```diff
  const path = require('path');
  const HtmlWebpackPlugin = require('html-webpack-plugin');
- const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 
  module.exports = {
 +  mode: 'development',
@@ -39,7 +38,6 @@ Let's start by setting [`mode` to `'development'`](/configuration/mode/#mode-dev
      print: './src/print.js',
    },
    plugins: [
-     new CleanWebpackPlugin(),
      new HtmlWebpackPlugin({
 -      title: 'Output Management',
 +      title: 'Development',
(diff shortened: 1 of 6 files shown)


src/content/guides/development.md
```diff
@@ -173,14 +170,13 @@ Let's add an npm script that will start webpack's Watch Mode:
  }
 ```
 
-Tell `CleanWebpackPlugin` that we don't want to remove the `index.html` file after the incremental build triggered by watch. We do this with the [`cleanStaleWebpackAssets` option](https://github.com/johnagan/clean-webpack-plugin#options-and-defaults-optional):
+Tell `webpack` that we don't want to remove the `index.html` file after the incremental build triggered by watch. We do this with the [`output.clean`](/configuration/output/#outputclean):
```
chenxsan, line 173: `index.html` is not removed running `webpack --watch` with `clean: true` in my test. Can you double check it?

snitin315: Yeah, `index.html` is not removed. Updated 👍🏻

src/content/guides/output-management.md
```diff
@@ -184,33 +184,28 @@ As you might have noticed over the past guides and code example, our `/dist` fol
 
 In general it's good practice to clean the `/dist` folder before each build, so that only used files will be generated. Let's take care of that.
```
chenxsan, line 185:
```suggestion
In general it's good practice to clean the `/dist` folder before each build, so that only used files will be generated. Let's take care of that with [`output.clean`](/configuration/output/#outputclean) option.
```

src/content/guides/output-management.md
```diff
@@ -184,33 +184,28 @@ As you might have noticed over the past guides and code example, our `/dist` fol
...
 
-A popular plugin to manage this is the [`clean-webpack-plugin`](https://www.npmjs.com/package/clean-webpack-plugin) so let's install and configure it.
-
-```bash
-npm install --save-dev clean-webpack-plugin
-```
+You can use [`output.clean` option](/configuration/output/#outputclean) for the same purpose.
```
chenxsan, line 187:
```suggestion
```

chenxsan (approved):

PR #4681, opened 2021-03-03
hash bundled js

chenxsan:
One thing I noticed that there's no hash included in the output filename of `index.bundle.js` and `vendor.bundle.js`, this pull request adds `[contenthash]` to the filename to better support long term caching.


chenxsan:
More difficult than I expected as `ssg` needs data from `prod`while `prod` needs data from `ssg` as well.

chenxsan:
Could be easily fixed once we incorporate react helmet to fix the seo problem.

chenxsan:
Superseded by https://github.com/webpack/webpack.js.org/pull/4738.

Issue #4687, opened 2021-03-04
improve description of the `output.library` option and guide for libraries


alexander-akait:
__Feature to document__

https://webpack.js.org/configuration/output/#outputlibrary
https://webpack.js.org/guides/author-libraries/

__Author(s)__

Any

__Additional information__

output.library support object notation https://github.com/webpack/webpack/blob/master/schemas/WebpackOptions.json#L1474, we it is very poorly described

`output.library.export` should be preferred over `output.libraryExport`
`output.library.type` should be preferred over `output.LibraryType`

`output.libraryExport` and `output.LibraryType` were keep for compatibility, but I think we will drop them in future, so we need rewrite it 

https://webpack.js.org/guides/author-libraries/ should be updated too using `output.library.type`/`output.library.export`. 

> With webpack 3.5.5, using libraryTarget: { root:'_' } doesn't work properly (as stated in issue 4824). However, you can set libraryTarget: { var: '_' } to expect the library as a global variable.

Solved and should be removed. 

Also I think will be great to add more example with different types and exports and names, because right now it is very hard to understand from docs how to generate libraries.

[ ] I am willing to work on this issue and submit a pull request.

chenxsan:
I guess you were talking about `output.libraryTarget` instead of `output.LibraryType`? `output.LibraryType` is not exposed as far as I see.

chenxsan:
Related pull request https://github.com/webpack/webpack/pull/10448

PR #4691, opened 2021-03-07
docs(configuration): update output.library

chenxsan:
Related issue https://github.com/webpack/webpack.js.org/issues/4687

Preview url https://webpack-js-org-git-fork-chenxsan-feature-output-library-365426.vercel.app/configuration/output/#outputlibrary


src/content/configuration/output.mdx
```diff
@@ -1305,6 +1305,206 @@ W> Note that not setting a `output.library` will cause all properties returned b
...
+  ```javascript
+  module.exports = {
+    //...
+    output: {
+      library: {
+        root: 'MyLibrary',
+        amd: 'my-library',
+        commonjs: 'my-common-library',
+      },
+      libraryTarget: 'umd',
```
alexander-akait: It can be rewritten on `library.name = { root, amd, commonjs }`, we should avoid using `libraryTarget`

alexander-akait: `libraryTarget` only for compatibility, all library options should be in `output.library`

chenxsan: All values of `libraryTarget` will have respective `library.type` item. For example, here's the new one for `umd` type https://github.com/webpack/webpack.js.org/pull/4691/files#diff-24a08fc6495329508b1a44d5f20bdaf472555d1c495157ee021f9405dea9bf5dR1056.

But I think we could still keep documentation for `libraryTarget` for a while so users can migrate. I'm going to inject a warning at the beginning of every `libraryTarget` encouraging users to use the respective `library.type`.

alexander-akait: Yep, just think we should provide all examples using `library`, and provide notes about `libraryTarget` with small example

chenxsan:
Will update the [author library guide](https://webpack.js.org/guides/author-libraries/) in another pull request with links to the new `output.library.type` once this get merged.

alexander-akait:
> Will update the author library guide in another pull request with links to the new output.library.type once this get merged.

Sounds good

alexander-akait (approved):
Good job

PR #4717, opened 2021-03-13
docs(output): remove leading slash

SandroMiguel:
I believe that the leading slash in the `keep` property does not prevent the directory from being deleted. 
I didn't test the *hook* option.

### DIFF
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -1434,7 +1434,7 @@ module.exports = {
   //...
   output: {
     clean: {
-      keep: /\/ignored\/dir\//, // Keep these assets.
+      keep: /ignored\/dir\//, // Keep these assets.
     },
   },
 };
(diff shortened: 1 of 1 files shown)


chenxsan (commented):
You are right, `/\/ignored\/dir\//` would only keep assets under `parent/ignored/dir` not `ignored/dir`.

src/content/configuration/output.mdx
```diff
@@ -1434,7 +1434,7 @@ module.exports = {
   //...
   output: {
     clean: {
-      keep: /\/ignored\/dir\//, // Keep these assets.
+      keep: /ignored\/dir\//, // Keep these assets.
```
chenxsan, line 1437:
```suggestion
      keep: /ignored\/dir\//, // Keep these assets under 'ignored/dir'.
```

chenxsan (approved):

Issue #4722, opened 2021-03-15
Error loading content.


chenxsan:
Just merged a pull request into the `master` branch, and here's the page after deployment finished on github pages:

![image](https://user-images.githubusercontent.com/1091472/111161085-b2d89a00-85d5-11eb-9a34-de9dc691cc85.png)

It might be caused by a cached version of `index.bundle.js` requesting stale js which is no longer available on the server.

Here's the right one:

![image](https://user-images.githubusercontent.com/1091472/111162423-18795600-85d7-11eb-87d9-db81227dcd7f.png)

I think this pull request https://github.com/webpack/webpack.js.org/pull/4681 is likely to fix it.

chenxsan:
After about 3 or more minutes, the page is finally updated.

PR #4738, opened 2021-03-18
use helmet

chenxsan:
There're multiple issues this pull request tries to fix:

1. There's no way to set `<meta name='description' />` for a single page at the moment, all pages share one description:

    ```html
    <meta name="description" content="webpack is a module bundler. Its main purpose is to bundle JavaScript files for usage in a browser, yet it is also capable of transforming, bundling, or packaging just about any resource or asset."/>
    ```

    It doesn't make sense, every page should be able to have its own description. With this pull request, we can set `description` for every single page by adding a new `description` field to its front matter, and it will fall back to the current one when none `description` set in front matter.
2. We generates static assets from both `webpack.prod.js` and `webpack.ssg.js`, which is a bad idea as it caused a circular dependency problem https://github.com/webpack/webpack.js.org/pull/4681#issuecomment-789710155. With this pull request, only assets generated by `webpack.prod.js` are needed, `webpack.ssg.js` will only be used to generate precompiled HTML pages.
3. Closes https://github.com/webpack/webpack.js.org/issues/4722

The reason I fixed them in one pull request is those issues are related and can't be fixed separately.


src/server.jsx
```diff
@@ -104,9 +45,16 @@ export default (locals) => {
...
-      </html>
-    </StaticRouter>
+      </StaticRouter>
+    </HelmetProvider>
   );
 
-  return '<!DOCTYPE html>' + renderedHtml;
+  const { helmet } = helmetContext;
+
+  const css = Object.keys(assets)
```
chenxsan, line 54: We're still using assets data from server side rendering, which is a problem for offline as sw.js would cache assets from compilation of `webpack.prod.js`.

PR #4571, opened 2021-02-09
docs(api): add info about env variables

snitin315:
Add `WEBPACK_SERVE`, `WEBPACK_BUILD`,  and `WEBPACK_WATCH`.

PR #4766, opened 2021-03-24
docs(api): document WEBPACK_BUNDLE

chenxsan:
Just noticed there's a [`WEBPACK_BUNDLE`](https://github.com/webpack/webpack-cli/blob/d20714f9fc47d2f20ecbbb553956fcadd947c316/packages/webpack-cli/lib/webpack-cli.js#L1942).


snitin315:
Actually we did not mentioned it as per https://github.com/webpack/webpack.js.org/pull/4571#discussion_r572753848

chenxsan:
Thanks for heads-up.

PR #4774, opened 2021-03-25
Fix invalid optimization.splitChunks config

tomchen:
`minSize` and `maxSize` in the previous configuration is obviously wrong, when executed it shows:

```
WARNING in SplitChunksPlugin
Fallback cache group
Configured minSize (19.5 KiB) is bigger than maxSize (0 bytes).
This seem to be a invalid optimization.splitChunks configuration.
```

### DIFF
--- a/src/content/plugins/split-chunks-plugin.md
+++ b/src/content/plugins/split-chunks-plugin.md
@@ -61,9 +61,8 @@ module.exports = {
   optimization: {
     splitChunks: {
       chunks: 'async',
-      minSize: 0,
+      minSize: 20000,
       minRemainingSize: 0,
-      maxSize: 20000,
       minChunks: 1,
       maxAsyncRequests: 30,
       maxInitialRequests: 30,


chenxsan:
Actually it represents the default values as you can tell from the paragraph above:

> This configuration object represents the default behavior of the SplitChunksPlugin.

But yes, it needs an update as I found:

image

We can just remove `maxSize`.

tomchen:
`minSize: 20000,`, no `mazSize`. OK I changed my PR.

tomchen:
@chenxsan I knew the doc states that the code represents the default values but how to enable this default behaviour without copying the whole code there? For me the code splitting doesn't seem enabled with `splitChunks: {chunks: 'async'}`, `splitChunks: {chunks: 'initial'}`, `splitChunks: {}`, `splitChunks: {chunks: 'all'}`, or without the `splitChunks` block. It works only after I paste that whole code snippet.

However I didn't have a clean test and it's possible that my configuration outside `optimization` block altered splitChunks' settings, if so, the question becomes: how to enforce this default behaviour? Is there something like `splitChunks: {chunks: 'default'}`?

chenxsan (approved):

PR #4775, opened 2021-03-25
Document compiler.run() needs compiler.close()

slorber:
The API doc is not very clear regarding the need to call compiler.close() after compiler.run() (only the changelog was).

It seems to me important because persistent cache won't be flushed otherwise. 
Also important to show that the callback takes an error that may be thrown if persistent caching fails.

Preview link: https://webpack-js-org-git-fork-slorber-patch-3-webpack-docs.vercel.app/api/node/#run

### DIFF
--- a/src/content/api/node.mdx
+++ b/src/content/api/node.mdx
@@ -87,8 +87,7 @@ The `run` method is then used to kickstart all compilation work. Upon
 completion, the given `callback` function is executed. The final logging of
 stats and errors should be done in this `callback` function.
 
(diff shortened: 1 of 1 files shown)


slorber:
cc @alexander-akait

slorber:
not sure why. the CLA bot seems stuck

image

chenxsan (approved):
lgtm

chenxsan:
Thanks for your contribution.

PR #4783, opened 2021-03-27
docs(api): document new HMR api

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/4704


montogeek:
Where is MDX used here?

PR #4797, opened 2021-03-30
📝 Clarified point about requiring webpack-cli under Installation

Sawtaytoes:
There's a point in the docs about installing Webpack-CLI. I've never installed this nor is it required. It's my understanding this is only required if you're running `webpack` in the command line rather than calling `webpack(config)` directly in Node.js.

As a separate piece, there _is_ a bug where, for every config I pass to `webpack()`, it outputs this message to `stats`:
```sh
-----
CLI for webpack must be installed.
  webpack-cli (https://github.com/webpack/webpack-cli)


-----
-----
We will use "yarn" to install the CLI via "yarn add -D webpack-cli".

-----
-----
Do you want to install 'webpack-cli' (yes/no):
-----
```

That's a separate issue.


chenxsan (approved):
Yes, `webpack-cli` is only needed when you run it from command line. Regarding the bug, I'm not able to reproduce, you might want to post it on `webpack` or `webpack-cli` repository for their confirmation.

PR #4804, opened 2021-04-02
doc(writers-guide): linkfix

simon04:
Rationale: The link is currently rendered including the `)`, see https://webpack.js.org/contribute/writers-guide/#adding-links


chenxsan (approved):
Would be great to find what's wrong under the hood. But this seems to be an acceptable workaround to me.

chenxsan:
Seems to be a bug in `remark-gfm`, see https://github.com/remarkjs/remark-gfm/issues/12.

PR #4807, opened 2021-04-05
docs(configuration): add new cache options

### LINKED ISSUE #4800: Document webpack change: memory usage improvements, add GC support for memory cache, persistent cache only mode

See pull request: https://github.com/webpack/webpack/pull/12990

---

* adds workaround for a v8 memory leak problem in ICs when code becomes optimized
* add GC to the memory cache and make it configurable
  * memory GC happens by default when persistent cache is enabled and in development mode
  * otherwise memory cache is kept forever
* make maxAge for persistent cache configurable
* GC the oldest content file even if it's not touched (to keep cache size small)
* filesystem cache will now no longer cache in memory, but restore items from disk again when it was persisted
* Caches will be cleared on shutdown. That should allow to GC cache items once they are serialized.

bugfix, memory leak, feature

n/a

no

* new option for `cache.type === "memory"`: `cache.maxGenerations`: Number of generations unused cache entries stay in memory cache at minimum (1 = may be removed after unused for a single compilation, ..., Infinity: kept forever).
* new options for `cache.type === "filesystem"`:
  * `cache.maxMemoryGenerations`: Number of generations unused cache entries stay in memory cache at minimum (0 = no memory cache used, 1 = may be removed after unused for a single compilation, ..., Infinity: kept forever). Cache entries will be deserialized from disk when removed from memory cache.
    * `cache.maxMemoryGenerations: 0`: Persistent cache will not use an additional memory cache. It will only cache items in memory until they are serialized to disk. Once serialized the next read will deserialize them from disk again. This mode will minimize memory usage, but introduces a performance cost.
    * `cache.maxMemoryGenerations: 1`: This will clear items from memory cache once they are serialized and unused for at least one compilation. When they are used again they will be deserialized from disk. This mode will minimize the memory usage while still keeping active items in memory cache.
    * `cache.maxMemoryGenerations` small numbers > 0 will have a performance cost for the GC operation. It gets lower as the number increases.
    * `cache.maxMemoryGenerations` default to 10 in development mode and to `Infinity` in production mode.
  * `cache.maxAge`: Time for which unused cache entries stay in the filesystem cache at minimum (in milliseconds).

jamesgeorge007:
closes #4800


src/content/configuration/other-options.md
```diff
@@ -297,6 +298,72 @@ module.exports = {
...
+
+- `cache.maxGenerations: 1`: Cache entries removed after unused for a single compilation.
+
+- `cache.maxGenerations: Infinity`: Cache entries kept forever.
+
+**webpack.config.js**
+
+```javascript
+module.exports = {
+  //..
```
chenxsan, line 359:
```suggestion
  // …
```

src/content/configuration/other-options.md
```diff
@@ -297,6 +298,72 @@ module.exports = {
...
+    maxAge: 5184000000,
+  },
+};
+```
+
+### cache.maxGenerations
+
+`number`
+
+Number of generations unused cache entries stay in memory cache at minimum.
```
chenxsan, line 349: This seems obscure to me. It would be great to rewrite the sentence instead of just copying it from the webpack source code. From the release page here https://github.com/webpack/webpack/releases/tag/v5.30.0, we can find `Generations ` equals to `rebuilds`, I believe the latter is much easier to understand.

src/content/configuration/other-options.md
```diff
@@ -297,6 +298,72 @@ module.exports = {
 
 W> `cache.idleTimeoutForInitialStore` is only available when [`cache.store`](#cachestore) is set to `'pack'`
 
+### cache.maxMemoryGenerations
+
+`number`
```
chenxsan, line 303: Since those are new features introduced in webpack 5.30.0, let's use `<Badge text='5.30.0+' />` to indicate the version when those features are available.

chenxsan (approved):
Nice job! One last thing, `<Badge />` would only work in `.mdx`, but I can make the change later if you don't have time.

jamesgeorge007:
Good catch, I'll handle it.

chenxsan (approved):

PR #4820, opened 2021-04-07
docs(configuration): fix liveReload cli usage

snitin315:
`--no-live-reload` is an invalid option. Only `--liveReload` is allowed.

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -751,6 +751,7 @@ Usage via the CLI
 ```bash
 npx webpack serve --liveReload
 ```
+Notice that there's no way to disable it from CLI.
 
 ## `devServer.mimeTypes` 🔑
 
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -750,7 +750,7 @@ Usage via the CLI
 
 ```bash
 npx webpack serve --liveReload
-```
+
 Notice that there's no way to disable it from CLI.
 
 ## `devServer.mimeTypes` 🔑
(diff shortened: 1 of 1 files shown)


src/content/configuration/dev-server.md
```diff
@@ -749,7 +749,7 @@ module.exports = {
 Usage via the CLI
 
 ```bash
-npx webpack serve --no-live-reload
+npx webpack serve --liveReload
```
chenxsan, line 752: I notice that it's enabled by default, then how to disable it from CLI?

snitin315: We can not. We can only enable it via cli if disabled in the config.

src/content/configuration/dev-server.md
```diff
@@ -749,7 +749,7 @@ module.exports = {
...
 ```
```
chenxsan, line 754:
```suggestion
Notice that there's no way to disable it from CLI.

```
I would suggest adding a note below as the cli example is added right after the example with `liveReload` set to `false`.

snitin315: Makes sense 👍🏻 .

src/content/configuration/dev-server.md
```diff
@@ -749,8 +749,9 @@ module.exports = {
...
+Notice that there's no way to disable it from CLI.
```
snitin315, line 754:
```suggestion

Notice that there's no way to disable it from CLI.
```

src/content/configuration/dev-server.md
alexander-akait: @snitin315 We should support it

snitin315: We support in `webpack-dev-server@beta` but not in v3.11.2

snitin315: I will migrate the docs after rc/major release.

chenxsan (approved):

Issue #4799, opened 2021-04-01
Advise for official document page


loynoir:
# Background
Install with latest webpack, keep up to latest official document, ran into weird bug.
Dig a little into that bug, seems it is because of a documented 3rd party loader outside of webpack-contrib, is not really webpack5-able.

# Expected behavior
Any 3rd party loaders outside of community [webpack-contrib](https://github.com/webpack-contrib), should 
1. Within official document page, with **hightlight**
2. Not within official document page

# Actual behavior
Official document contains, at least one, 3rd party loader **outside** of community **without highlight**. 
Not really want to point out which one.

# Personal Opinion
Within official document [page](https://webpack.js.org/loaders) of webpack loaders, should only contains 
1. official webpack loaders
2. community supported 3rd party webpack loaders.
3. neccesary packages required by above two

Any 3rd party loaders outside of community [webpack-contrib](https://github.com/webpack-contrib), should 
1. Within official document page, with **hightlight**
2. Not within official document page

Besides, I saw keyword securiy when submit issue. 
Emm, it's also might be a security issue too.

alexander-akait:
I think we need remove `extract-loader` from our docs, this loader is broken...

loynoir:
@alexander-akait `babel-loader` is too essential. I dont think babel-loader would be a problem.  😆

chenxsan:
Generally we would avoid including documentations for third-party loaders or plugins that are not from `webpack` or `webpack-contrib`, readme from their own repositories could serve more up-to-date documentation.

PR #4827, opened 2021-04-07
remove extract-loader

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/4799

### DIFF
--- a/src/utilities/fetch-package-repos.js
+++ b/src/utilities/fetch-package-repos.js
@@ -18,7 +18,6 @@ const fetch = {
       hides: excludedLoaders,
     },
     'babel/babel-loader',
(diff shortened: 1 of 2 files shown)


src/utilities/fetch-package-repos.js
```diff
@@ -18,7 +18,6 @@ const fetch = {
       hides: excludedLoaders,
     },
     'babel/babel-loader',
-    'peerigon/extract-loader',
     'Banno/polymer-webpack-loader',
```
alexander-akait, line 21: I think removing `polymer-webpack-loader` will be good too, it is unused many developers and don't updated

Issue #4196, opened 2020-11-22
Upgrade alex


chenxsan:
The latest version of [alex](https://github.com/get-alex/alex/releases) is 9.1.0 while we have [5.1.0](https://github.com/webpack/webpack.js.org/blob/master/package.json#L90) which is quite outdated. We should consider upgrading it and fix any potential issues.

PR #4840, opened 2021-04-10
remove alex

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/4196.

So far `alex` has never caught even one error, and upgrading it to v9 would need a lot of changes not really necessary in my opinion. Hence I'm removing it.


montogeek:
I will upgrade it.

PR #4850, opened 2021-04-12
docs(guides) fix typo in development.md

petertdinh:


chenxsan (commented):
As you can see from the above paragraph, it's an error we created on purpose for debugging. Hence i'm closing this.

PR #4862, opened 2021-04-13
docs(configuration): fix `--inline` usage

snitin315:
Correct usage - `npx webpack serve --inline`

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -687,8 +687,10 @@ module.exports = {
 Usage via the CLI
 
 ```bash
-npx webpack serve --inline false
+npx webpack serve --inline
 ```
+
+Notice that there's no way to disable it from CLI.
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/dev-server.md
```diff
@@ -687,9 +687,11 @@ module.exports = {
 Usage via the CLI
 
 ```bash
-npx webpack serve --inline false
+npx webpack serve --inline
 ```
 
+Notice that there's no way to disable it from CLI.
```
montogeek, line 693: Disable what?

snitin315: `devServer.inline`

```suggestion
Notice that there's no way to disable `devServer.inline` from CLI.
```

montogeek: Ahh, uhmm, I think it is fine :)

montogeek (approved):

chenxsan (approved):

PR #4879, opened 2021-04-16
docs(configuration): add warning for `liveReload`

snitin315:
Add warning that `liveReload` doesn't work with non-web targets. 

For new webpack users who don't have much idea about the configuration -

https://github.com/webpack/webpack-dev-server/issues/2026#issuecomment-502450338

### DIFF
--- a/src/content/configuration/dev-server.md
+++ b/src/content/configuration/dev-server.md
@@ -746,7 +746,7 @@ npx webpack serve --liveReload
 
 Notice that there's no way to disable it from CLI.
 
-W> Live reloading does not work with non-web targets like `node`, `async-node` etc.
+W> Live reloading works only with web related [targets](/configuration/target/#string) like `web`, `webworker`, `electron-renderer` and `node-webkit`.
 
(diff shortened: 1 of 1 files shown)


rishabh3112 (approved):

src/content/configuration/dev-server.md
```diff
@@ -746,6 +746,8 @@ npx webpack serve --liveReload
 
 Notice that there's no way to disable it from CLI.
 
+W> Live reloading does not work with non-web targets like `node`, `async-node` etc.
```
rishabh3112, line 749: Sorry for approving earlier, but after thinking later I think naming all targets for which it works would be better. `etc` adds ambiguity for reader.

chenxsan: Or maybe we can list just web targets which I believe is much easier to list as per https://webpack.js.org/configuration/target/#string

snitin315: yep 👍🏻

chenxsan (approved):

PR #4889, opened 2021-04-19
docs(configuration): add warning for bail and watch

snitin315:
Refers https://github.com/webpack/webpack-cli/issues/1794

### DIFF
--- a/src/content/configuration/other-options.mdx
+++ b/src/content/configuration/other-options.mdx
@@ -16,6 +16,7 @@ contributors:
   - chenxsan
   - u01jmg3
   - jamesgeorge007
+  - snitin315
 related:
   - title: Using Records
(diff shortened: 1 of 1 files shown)


src/content/configuration/other-options.mdx
```diff
@@ -66,6 +67,8 @@ module.exports = {
 
 This will force webpack to exit its bundling process.
 
+W> You can use `bail` option in [`watch`](/configuration/watch) mode, but it will still force webpack to exit as soon as an error is found.
```
chenxsan, line 70:
```suggestion
W> Avoid using `bail` option in [`watch`](/configuration/watch) mode, as it will force webpack to exit as soon as possible when an error is found.
```

chenxsan (approved):

PR #4945, opened 2021-05-06
remove deprecated loaders

jamesgeorge007:
- [raw-loader](https://github.com/webpack-contrib/raw-loader)
- [url-loader](https://github.com/webpack-contrib/url-loader)
- [file-loader](https://github.com/webpack-contrib/file-loader)

### DIFF
--- a/src/content/loaders/index.md
+++ b/src/content/loaders/index.md
@@ -9,6 +9,7 @@ contributors:
   - hemal7735
   - snitin315
   - anshumanv
+  - jamesgeorge007
 ---
 
 webpack enables use of [loaders](/concepts/loaders) to preprocess files. This allows you to bundle any static resource way beyond JavaScript. You can easily write your own loaders using Node.js.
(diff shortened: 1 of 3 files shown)


chenxsan (commented):
I would rather keeping those pages although they're deprecated. Those are used a lot and there would be many pages on the internet linking to them. Removing them would break users' expectation.

Or we can remove them and redirect the removed url to the v4 version.

cc @alexander-akait what do you think here.

alexander-akait:
@chenxsan it makes sense to redirect :+1: it will be better since most of developers will move to assets modules instead using `file-loader`/`url-loader`/`raw-loader`

jamesgeorge007:
> Or we can remove them and redirect the removed url to the v4 version.

How to go about implementing the redirects? I can see a few listed here:-
https://github.com/webpack/webpack.js.org/blob/747cb66122a5c5812ba7daa8fafde3b827dfad4e/webpack.ssg.js#L58

chenxsan:
> > Or we can remove them and redirect the removed url to the v4 version.
> 
> How to go about implementing the redirects? I can see a few listed here:-
> https://github.com/webpack/webpack.js.org/blob/747cb66122a5c5812ba7daa8fafde3b827dfad4e/webpack.ssg.js#L58

Spot on.

chenxsan (approved):

PR #4962, opened 2021-05-10
Removed --save because it is no longer necessary

adriancuadrado:
The '--save' parameter is implicit in recent versions of npm. You don't have to specify it anymore.
See https://stackoverflow.com/a/19578808/5744858


chenxsan:
Thanks but I would prefer to keep it explicit. Besides, we don't really know which npm version users are using.

PR #4979, opened 2021-05-14
fix: entry path in doc

tanaypratap:
`entry` path currently isn't platform-independent. This change will fix it.


chenxsan:
Can you share which platform did you find the syntax wouldn't work?

chenxsan:
I'm closing this as I believe the syntax is platform independent. Please share a reproducible repository if you find it not true.

PR #4988, opened 2021-05-17
refactor notification bar

chenxsan:
There're multiple problems presented in the current notification bar:

1. The close button is too small to click on a mobile phone

    

    This pull request fixes the problem with a larger close button, and it's much easier to click on the mobile phone:

    CleanShot 2021-06-21 at 20 49 28@2x

2. Notification bar at the top causes the whole page to repaint, in addition, it forces us to reposition our pages when it's visible:

    image

    Moving it to the bottom would avoid the problem.

    CleanShot 2021-06-21 at 20 50 22@2x

Note that this pull request will save us some bits from css:

1. before - 75.6 kB
    

2. after - 75.2 kB


alexander-akait:
We you set v4, no paddings between navigation and select

chenxsan:
> We you set v4, no paddings between navigation and select

Do you mean this problem on v4 when navigation bar is visible?

image

alexander-akait:
Yes :+1:

chenxsan:
> Yes 👍

Got it, I'll fix it soon.

rishabh3112 (approved):
Nice job!

PR #4997, opened 2021-05-19
drop node 10

chenxsan:
Drop node 10 support as it's end of life.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "url": "https://github.com/webpack/webpack.js.org/issues"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "url": "https://github.com/webpack/webpack.js.org/issues"
   },
   "engines": {
-    "node": ">=10.13.0"
+    "node": ">=12"
```
alexander-akait, line 24: Better use LTS. i.e. `12.13.0`, but here it doesn't have big sense

PR #5016, opened 2021-05-25
docs(general): Revise "imposter syndrome" language

### LINKED ISSUE #5015: Suggestion: Revise "imposter syndrome" words
(I'll submit a PR for this because I think it's important but understand others don't necessarily have time to prioritize this.)

I saw a conversation online where several folks had agreed that the webpack documentation had made them feel a little bit badly. It's actually quite common that the language we use can unintentionally discourage beginners and newcomers to the ecosystem.

Words like "simply", "just", etc. can discourage readers by making it sound like a complex system should really be easy to understand. When they have a typical reaction to a complex system -- including some hesitance and confusion -- this can be discouraging. Additionally, these words typically add little value to the writing and so they can be removed with little impact on any messaging.

I intend to [Use this article as a reference](https://css-tricks.com/words-avoid-educational-writing/) and submit a PR to replace or remove some of that language in an attempt to make these great docs that much more welcoming for users, especially new ones.

SeanKilleen:
Resolves #5015.

### DIFF
--- a/src/content/concepts/configuration.md
+++ b/src/content/concepts/configuration.md
@@ -30,7 +30,7 @@ T> The most important part to take away from this document is that there are man
 
 The examples below describe how webpack's configuration can be both expressive and configurable because _it is code_:
 
-## Simple Configuration
+## Introductory Configuration
 
 **webpack.config.js**
 
--- a/src/content/guides/author-libraries.md
+++ b/src/content/guides/author-libraries.md
@@ -207,7 +207,7 @@ Let's update the `output.library` option with its `type` set to [`'umd'`](/confi
 
 Now webpack will bundle a library that can work with CommonJS, AMD, and script tag.
 
-T> Note that the `library` setup is tied to the `entry` configuration. For most libraries, specifying a single entry point is sufficient. While [multi-part libraries](https://github.com/webpack/webpack/tree/master/examples/multi-part-library) are possible, it is simpler to expose partial exports through an [index script](https://stackoverflow.com/questions/34072598/es6-exporting-importing-in-index-file) that serves as a single entry point. Using an `array` as an `entry` point for a library is **not recommended**.
+T> Note that the `library` setup is tied to the `entry` configuration. For most libraries, specifying a single entry point is sufficient. While [multi-part libraries](https://github.com/webpack/webpack/tree/master/examples/multi-part-library) are possible, it is more straight-forward to expose partial exports through an [index script](https://stackoverflow.com/questions/34072598/es6-exporting-importing-in-index-file) that serves as a single entry point. Using an `array` as an `entry` point for a library is **not recommended**.
 
 ## Externalize Lodash
 
--- a/src/content/guides/author-libraries.md
+++ b/src/content/guides/author-libraries.md
@@ -117,7 +117,7 @@ module.exports = {
 };
 ```
 
-It should be familiar have you used webpack to bundle your application. Basically, we're telling webpack to bundle `src/index.js` into `dist/webpack-numbers.js`.
+In the above example, we're telling webpack to bundle `src/index.js` into `dist/webpack-numbers.js`.
 
 ## Expose the Library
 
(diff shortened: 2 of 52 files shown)


SeanKilleen:
While there are typically more edits and files than I'd normally include in a PR, I hope the small changes make for a quick review.

Happy to consider any language changes you don't agree with, and welcome PR review comments. 👍

src/content/concepts/configuration.md
```diff
@@ -30,7 +30,7 @@ T> The most important part to take away from this document is that there are man
 
 The examples below describe how webpack's configuration can be both expressive and configurable because _it is code_:
 
-## Simple Configuration
+## Introductory Configuration
```
chenxsan, line 33: I think we need to update the redirects list here https://github.com/webpack/webpack.js.org/blob/master/webpack.ssg.js#L57, otherwise links to this heading would be break.

src/content/guides/author-libraries.md
```diff
@@ -207,7 +207,7 @@ Let's update the `output.library` option with its `type` set to [`'umd'`](/confi
 
 Now webpack will bundle a library that can work with CommonJS, AMD, and script tag.
 
-T> Note that the `library` setup is tied to the `entry` configuration. For most libraries, specifying a single entry point is sufficient. While [multi-part libraries](https://github.com/webpack/webpack/tree/master/examples/multi-part-library) are possible, it is simpler to expose partial exports through an [index script](https://stackoverflow.com/questions/34072598/es6-exporting-importing-in-index-file) that serves as a single entry point. Using an `array` as an `entry` point for a library is **not recommended**.
+T> Note that the `library` setup is tied to the `entry` configuration. For most libraries, specifying a single entry point is sufficient. While [multi-part libraries](https://github.com/webpack/webpack/tree/master/examples/multi-part-library) are possible, it is more straight-forward to expose partial exports through an [index script](https://stackoverflow.com/questions/34072598/es6-exporting-importing-in-index-file) that serves as a single entry point. Using an `array` as an `entry` point for a library is **not recommended**.
```
chenxsan, line 210:
```suggestion
T> Note that the `library` setup is tied to the `entry` configuration. For most libraries, specifying a single entry point is sufficient. While [multi-part libraries](https://github.com/webpack/webpack/tree/master/examples/multi-part-library) are possible, it is more straightforward to expose partial exports through an [index script](https://stackoverflow.com/questions/34072598/es6-exporting-importing-in-index-file) that serves as a single entry point. Using an `array` as an `entry` point for a library is **not recommended**.
```

src/content/plugins/hot-module-replacement-plugin.md
```diff
@@ -18,7 +18,7 @@ W> HMR should **never** be used in production.
 
 ## Basic Usage
 
-Enabling HMR is easy and in most cases no options are necessary.
+Enabling HMR is straight-forward and in most cases no options are necessary.
```
chenxsan, line 21:
```suggestion
Enabling HMR is straightforward and in most cases no options are necessary.
```

chenxsan (commented):
Great job! Please sign the CLA here https://cla.js.foundation/webpack/webpack.js.org?pullRequest=5016.

src/content/concepts/configuration.md
SeanKilleen: @chenxsan I looked at this list of redirects and couldn't find anything related to these headings. The rest of the redirects listed there still appear to be in place. 

Happy to address it if you're able to add more detail. Thanks!

SeanKilleen:
@chenxsan update;

✅ CLA Signed
✅ Suggestions merged in
❔ Added a question to the last piece of review. 

I think once we sync up on that last question, this should be good to go. Thanks for being open to the PR!

src/content/concepts/configuration.md
chenxsan: Here's the link for current `Simple Configuration` heading `/concepts/configuration/#simple-configuration`, and the updated one would be `/concepts/configuration/#introductory-configuration`, hence we need to append an redirect item like:

```json
'/concepts/configuration/#simple-configuration': '/concepts/configuration/#introductory-configuration',
```

SeanKilleen: @chenxsan ahh, I see. I was misunderstanding. Thank you for the clarification! I'll do this.

SeanKilleen:
@chenxsan I've added the redirect as requested. I think at this point all feedback items have been addressed. 👍

chenxsan (approved):
Thanks for your contribution!

PR #5026, opened 2021-05-27
Fix double negative in grammar.

lhoff:
The title is self-explanatory.

### DIFF
--- a/src/content/guides/web-workers.mdx
+++ b/src/content/guides/web-workers.mdx
@@ -50,4 +50,4 @@ import { Worker } from 'worker_threads';
 new Worker(new URL('./worker.js', import.meta.url));
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/guides/web-workers.mdx
```diff
@@ -50,4 +50,4 @@ import { Worker } from 'worker_threads';
 new Worker(new URL('./worker.js', import.meta.url));
 ```
 
-Note that this is only available in ESM. `Worker` in CommonJS syntax is not supported by neither webpack nor Node.js.
+Note that this is only available in ESM. `Worker` in CommonJS syntax is not supported by either webpack nor Node.js.
```
chenxsan, line 53: Should it be?

```suggestion
Note that this is only available in ESM. `Worker` in CommonJS syntax is not supported by either webpack or Node.js.
```

chenxsan (approved):

PR #5040, opened 2021-06-01
docs(concepts): Adding promise syntax example to Module Federation docs

KyleBastien:
Adds a `promise new Promise()` example to the Module Federation docs. This makes it so this API is now documented, and gives an example of how to use it's power.


src/content/concepts/module-federation.md
```diff
@@ -188,6 +189,58 @@ loadComponent('abtests', 'test123');
...
+Generally, remotes are configured using URL's like in this example:
+
+```js
+new ModuleFederationPlugin({
+  name: 'host',
+  remotes: {
+    app1: "app1@http://localhost:3001/remoteEntry.js",
+  },
+  // ...
+}),
```
chenxsan, line 203:
```suggestion
module.exports = {
  plugins: [
    new ModuleFederationPlugin({
      name: 'host',
      remotes: {
        app1: 'app1@http://localhost:3001/remoteEntry.js',
      },
    }),
  ],
};
```

src/content/concepts/module-federation.md
```diff
@@ -188,6 +189,58 @@ loadComponent('abtests', 'test123');
...
+            console.log('remote container already initialized')
+          }
+        }
+      }
+      res(proxy)
+    })
+    `,
+  },
+  // ...
+}),
```
chenxsan, line 239:
```suggestion
module.exports = {
  plugins: [
    new ModuleFederationPlugin({
      name: 'host',
      remotes: {
        app1: `promise new Promsie(res => {
      let remote
      const urlParams = new URLSearchParams(window.location.search)
      const version = urlParams.get('app1VersionParam')
      // This part depends on how you plan on hosting and versioning your federated modules
      const remoteUrlWithVersion = 'https://localhost:3001/' + version + '/remoteEntry.js'
      try {
        remote = require(remoteUrlWithVersion)['app1']
      } catch (e) {
        delete require.cache[remoteUrlWithVersion]
        remote = require(remoteUrlWithVersion)['app1']
      }
      const proxy = {
        get: (request) => remote.get(request),
        init: (arg) => {
          try {
            return remote.init(arg)
          } catch(e) {
            console.log('remote container already initialized')
          }
        }
      }
      res(proxy)
    })
    `,
      },
      // ...
    }),
  ],
};
```

chenxsan:
@XiaofengXie16 Hi there, do you think you can help review this pull request just in case you have any similar experience?

XiaofengXie16:
@chenxsan I have not personally used this approach before but I just tried to apply this method on top of this example code https://github.com/module-federation/module-federation-examples/tree/master/basic-host-remote and I am getting `ReferenceError: require is not defined`. I　assume there are some additional settings that we need to configure in order to run it correctly? If that's the case, it would be great to specify them in the example as well.

chenxsan:
> @chenxsan I have not personally used this approach before but I just tried to apply this method on top of this example code https://github.com/module-federation/module-federation-examples/tree/master/basic-host-remote and I am getting `ReferenceError: require is not defined`. I assume there are some additional settings that we need to configure in order to run the example code in this PR correctly? (I know this error is not really tied to module federation itself but it will make it more friendly to people who just get started with module federation)If that's the case, it would be great to specify them in the example as well.

Thanks for checking!

chenxsan:
@KyleBastien Any suggestion regarding the `require is not defined` error? Seems webpack won't compile the code defined in remotes of `promise new Promise()` type. Since there's no `require` in browser, hence `require is not defined`.

KyleBastien:
@chenxsan I'm taking a closer look. I copied the majority of this example from @ScriptedAlchemy's nextjs-mf repo here: https://github.com/module-federation/nextjs-mf/blob/main/withModuleFederation.js#L14. But something must be different in this case. Going to dig in futher.

chenxsan:
It seems to be used in server side compilation as per https://github.com/module-federation/nextjs-mf/blob/main/withModuleFederation.js#L66 and https://github.com/vercel/next.js/blob/canary/docs/api-reference/next.config.js/custom-webpack-config.md:

 ```md
- `isServer`: `Boolean` - It's `true` for server-side compilation, and `false` for client-side compilation
```

That's why `require` is available over there.

ScriptedAlchemy:
Require is an example of a node based app. 

Client side you could inject a script and onload resolve the interface. 

I use this to do api calls to figure out what remote version to inject into the dom

KyleBastien:
Thanks @chenxsan and @ScriptedAlchemy. I updated my example based on this feedback, and I implemented a basic example here: https://github.com/KyleBastien/module-federation-examples/blob/b43512578fe12c105799682853938fa4f5761bac/basic-host-remote/app1/webpack.config.js#L39

To also show that it works.

chenxsan (approved):
I've tested the code and it worked great.

PR #5046, opened 2021-06-03
docs(guides/getting-started): add stackblitz live preview

d3lm:
This PR adds a first live example for the first step in the "Getting Started" guide. This should serve as a basis to start iterating on and add further live examples to other guides as well. Here's a screenshot:

image

You can test the example on [StackBlitz](https://stackblitz.com/github/d3lm/webpack.js.org/tree/feat/add-stackblitz-live-previews/examples/getting-started?runScript=start).

**Note**: The link to the example has to be updated before merging into master because right now it's pointing to my fork and not the `master` branch because it's not yet merged. I used a link to my fork so we can properly test the workflow.


src/content/guides/getting-started.md
```diff
@@ -32,6 +32,14 @@ W> The minimum supported Node.js version to run webpack 5 is 10.13.0 (LTS)
 
 ## Basic Setup
 
+<aside class="preview">
+  <h6 class="preview__prefix">live preview</h6>
+  <p>Preview this example live on <a href="https://stackblitz.com">StackBlitz</a>.</p>
+  <a href="https://stackblitz.com/github/d3lm/webpack.js.org/tree/feat/add-stackblitz-live-previews/examples/getting-started/basic-setup?runScript=start" target="_blank">
+    <img src="https://developer.stackblitz.com/img/open_in_stackblitz.svg">
+  </a>
+</aside>
```
sokra, line 41: I think this block is too early here.

I think these buttons should represent the source code state that you would have at this position of the guide.

So this example should be at the end of the chapter.

examples/getting-started/basic-setup/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "documentation",
+    "examples"
+  ],
+  "dependencies": {
+    "lodash": "latest"
+  },
+  "devDependencies": {
+    "webpack": "latest",
+    "webpack-cli": "latest",
+    "serve": "latest"
```
sokra, line 21: This is not something you would end with by following the guide. `lastest` should be replaced by `^x.x.x`. It's fine to not commit a package-lock.json here so when running the example it actually uses the lastest versions according to the range.

examples/getting-started/basic-setup/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "getting-started-basic-setup",
+  "version": "1.0.0",
+  "license": "MIT",
+  "private": true,
+  "author": "",
+  "scripts": {
+    "start": "serve ."
```
sokra, line 8: It would be cool if this is running in background and the user will end up with an empty terminal (maybe via new query argument `?terminal=new` or something like that)

Currently they will end up with `serve .` running and are unable to follow the next steps in the guide, which is running webpack. (It's not obvious that one can cancel the process, and that would also break the browser)

sokra (changes requested):

examples/getting-started/basic-setup/package.json
d3lm: Would it be ok to add a `package-lock.json` though? This would make the boot process much faster because it will skip resolution, which is always a little slow.

examples/getting-started/basic-setup/package.json
d3lm: I am not sure I follow. Are you suggesting to remove `serve` and not run anything for the first step? IMO, each step should have some visual outcome, no? That's why I added `serve` so the page is actually served in the browser. The first chapter doesn't tell the user what to do with it. Might be that I missed something.

examples/getting-started/basic-setup/package.json
d3lm: If so, what version of webpack would you use then? Just `^5`?

sokra: > Would it be ok to add a `package-lock.json` though?

That would be fine for me too.

> If so, what version of webpack would you use then? Just `^5`?

I would use something real. So run `npm install webpack` and use that that gives. `^5.38.1`.
It's not unlikely that some users are copying these files/lines and they should not end up with something weird that we did only for the example.

examples/getting-started/basic-setup/package.json
sokra: Running `serve .` is good. I was only asking to open another terminal, so the user can continue with the instructions in the guide.
They will be asked to run `npm install lodash` next and this would be harder if the terminal is already blocked by another process.

sokra: My idea was that the guide can click the button and follow the guides instructions for that point on directly in the browser.

examples/getting-started/basic-setup/package.json
d3lm: Ok, that makes sense. I ll change it 👍

examples/getting-started/basic-setup/package.json
d3lm: Oh I see. I was thinking to basically create one example per chapter in the getting-started and see the result of the chapter. Because it has all the files they were asked to create. Or do you just mean the steps for the terminal then?

sokra: > I was thinking to basically create one example per chapter in the getting-started and see the result of the chapter.

Yep

d3lm: Updated.

examples/getting-started/basic-setup/package.json
d3lm: Updated.

src/content/guides/getting-started.md
d3lm: Updated.

src/components/StackBlitzPreview/StackBlitzPreview.jsx
```diff
@@ -0,0 +1,22 @@
...
+export default function StackBlitzPreview(props = {}) {
+  const { example = '' } = props;
+
+  const url = `https://stackblitz.com/github/d3lm/webpack.js.org/tree/feat/add-stackblitz-live-previews/examples/${example}`;
+
+  return (
+    <aside className="preview">
+      <h6 className="preview__prefix">live preview</h6>
+      <p>Preview this chapter live on StackBlitz.</p>
+      <a href={url} target="_blank" rel="noreferrer">
```
chenxsan, line 17:
```suggestion
      <a href={url} target="_blank" rel="noopener noreferrer">
```

d3lm: Updated.

src/content/guides/getting-started.mdx
```diff
@@ -317,6 +324,10 @@ webpack 5.4.0 compiled successfully in 1940 ms
 
 T> Custom parameters can be passed to webpack by adding two dashes between the `npm run build` command and your parameters, e.g. `npm run build -- --color`.
 
+For the live preview, once the build is finished, you can run `npm run start` to serve the `dist` folder.
```
chenxsan, line 327:
```suggestion
For the live preview, once the build is finished, you can run `npm start` to serve the `dist` folder.
```

d3lm: This is not necessary. Do you still want to add `run` to it?

src/content/guides/getting-started.mdx
```diff
@@ -24,12 +24,15 @@ contributors:
...
   - myshov
   - anshumanv
+  - d3lm
 ---
 
 webpack is used to compile JavaScript modules. Once [installed](/guides/installation), you can interact with webpack either from its [CLI](/api/cli) or [API](/api/node). If you're still new to webpack, please read through the [core concepts](/concepts) and [this comparison](/comparison) to learn why you might use it over the other tools that are out in the community.
 
 W> The minimum supported Node.js version to run webpack 5 is 10.13.0 (LTS)
 
+<StackBlitzPreview example="getting-started/basic-setup?terminal=" />
```
chenxsan, line 34:
```suggestion
<StackBlitzPreview example="getting-started?terminal=" />
```
I think this needs an update.

CleanShot 2021-06-10 at 08 02 53@2x

d3lm: Yep, indeed. Good catch.

d3lm: Yes good catch!

d3lm: Updated.

chenxsan (dismissed):
I'm good with it.

One thing I noticed that we use `MIT` for examples now which is not the default license for `npm init -y`, the default one would be `ISC`, but I think it's not that important.

Let's see if @sokra has any other suggestions.

d3lm:
Ok, let me change the URL then already, so we don't forget and you can then re-approve.

d3lm:
URL updated.

chenxsan (approved):

snitin315 (approved):
Nice addition. Thanks!

PR #5051, opened 2021-06-06
docs(concept): remove OverridablesPlugin

snitin315:
All code related to overrides was removed - 

https://github.com/webpack/webpack/commit/e78a76b2e00a17c62dcc3ce8a8ffa1b131a6431d#diff-fc7055018e1900b8bb0feffbdfe152f916fd99dec480326df98d54c223ca8ded

### DIFF
--- a/src/content/concepts/module-federation.md
+++ b/src/content/concepts/module-federation.md
@@ -7,6 +7,7 @@ contributors:
   - EugeneHlushko
   - jamesgeorge007
   - ScriptedAlchemy
+  - snitin315
 related:
   - title: 'Webpack 5 Module Federation: A game-changer in JavaScript architecture'
     url: https://medium.com/swlh/webpack-5-module-federation-a-game-changer-to-javascript-architecture-bcdd30e02669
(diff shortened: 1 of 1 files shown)


XiaofengXie16:
I think there is opportunity to work on the description in the ContainerPlugin because it currently refers to the `OverridablesPlugin` which is removed in your PR. 

```
It also uses the `OverridablesPlugin` internally and exposes the `override` API to consumer of the container.
```

src/content/concepts/module-federation.md
```diff
@@ -68,35 +50,6 @@ The `packageName` option allows setting a package name to look for a `requiredVe
...
-```js
-__webpack_override__({
-  // here we override test1 module
-  test1: () => 'I will override test1 module under src',
-});
-```
-
 ### `ContainerPlugin` (low level)
 
 This plugin creates an additional container entry with the specified exposed modules. It also uses the `OverridablesPlugin` internally and exposes the `override` API to consumer of the container.
```
snitin315, line 55:
```suggestion
This plugin creates an additional container entry with the specified exposed modules.
```

chenxsan (approved):

PR #5060, opened 2021-06-10
load manifest.json only for production

chenxsan:
`manifest.json` would only be available to production https://github.com/webpack/webpack.js.org/blob/master/webpack.ssg.js#L116.


snitin315 (approved):

snitin315:
CLA bot is broken. Let's ignore 👍🏻

PR #5066, opened 2021-06-10
docs(optimizations): remove sourceMap option from TerserPlugin options

tokkoro:
docs optimizations.mdx: remove sourceMap option from TerserPlugin options

Removes sourceMap option from options of Terser Plugin as Terser Plugin no longer takes the sourceMap option, but it seems to automatically detect if it process files that have source maps available.

> [webpack-cli] ValidationError: Invalid options object. Terser Plugin has been initialized using an options object that does not match the API schema.
>  - options has an unknown property 'sourceMap'. These properties are valid:
>    object { test?, include?, exclude?, terserOptions?, extractComments?, parallel?, minify? }


tokkoro:
When I open the CLA in https://cla.js.foundation/webpack/webpack.js.org I get following error. How do I sign and where do I can ask help for that?

```
Error

There is no CLA to sign for webpack/webpack.js.org

({"message":"Must specify access token via Authorization header. https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param","documentation_url":"https://docs.github.com/v3/#oauth2-token-sent-in-a-header"})

```

src/content/configuration/optimization.mdx
```diff
@@ -56,7 +56,6 @@ module.exports = {
     minimizer: [
       new TerserPlugin({
         parallel: true,
-        sourceMap: true, // Must be set to true if using source-maps in production
```
jaylinski, line 59: I think it still takes that option, but as part of the `terserOptions`-object:

```
new TerserPlugin({
  parallel: true,
  terserOptions: {
    sourceMap: true, // Must be set to true if using source-maps in production
  },
})
```

tokkoro: When I have `devtool: 'source-map'` set in webpack's config and I build with production mode, it will provide sourcepmap input for minify and inside Terser Plugins folder /lib/minify.js after line 240 it will set the terserOptions for you:
```
  if (inputSourceMap) {
    // @ts-ignore
    terserOptions.sourceMap = {
      asObject: true
    };
  }
```
So I think you are right that you can set it by yourself, but still that it is unnecessary to set that option manually and then IMO it is better to leave it out of Webpack documentation?

tokkoro: ps. I'm unable to re-request your @jaylinski review. This is an one of my first PRs for open source projects so please guide me if I'm not following the right procedures :smile:

jaylinski: > but still that it is unnecessary to set that option manually and then IMO it is better to leave it out of Webpack documentation?

Agree :) (I only commented on this because I found the same issue with the documentation. I'm not an official reviewer, this was just a comment.)

chenxsan: @jaylinski I don't think there's any official ones :) Anyone can help reviewing.

@tokkoro The CLA bot is broken at the moment. I've asked the author for help, stay tuned.

tokkoro: CLA site was working and I accepted the license. Now I'm wating to it to show up here :)

chenxsan (approved):
Thank you for your contribution!

benoittgt:
Do you think it could be back-ported to Webpack 4 documentation. I just had the surprise.

chenxsan:
> Do you think it could be back-ported to Webpack 4 documentation? I just had the surprise.

From what I can tell, you needs to use `terser-webpack-plugin` v4 along with webpack v4. And that `sourceMap` option was removed in [`terser-webpack-plugin` v5](https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/CHANGELOG.md#500-2020-10-14). So it make no sense to back port this change to webpack 4 documentation.

PR #5081, opened 2021-06-14
chore: update examples

d3lm:
This PR updates the examples and slightly improves the `README` as well as adds a `package-lock.json`. By adding a lock file the package manager doesn't have to run resolution which makes booting the example almost instant.


chenxsan:
I can see your point of improved performance, but I would suggest avoiding `package-lock.json` since it's impossible to review it.

examples/getting-started/README.md
```diff
@@ -0,0 +1,11 @@
+# Getting Started
+
+## How to use
+
+Run the following commands in the root directory.
+
+```bash
+❯ npm install
+❯ npm run build
+❯ npm run serve
```
chenxsan, line 10:
```suggestion
$ npm install
$ npm run build
$ npm run serve
```

examples/module-federation/README.md
```diff
@@ -1,17 +1,18 @@
...
+## How to use
 
-```
-yarn install
+Run the following commands in the root directory.
 
-yarn start (this should trigger the lerna command)
+```bash
+❯ npm install
+❯ npm start
```
chenxsan, line 9:
```suggestion
$ npm install
$ npm start
```

chenxsan: Another thing I found is this example won't work under npm 6 in my test. Only yarn works.

d3lm: Oh interesting. That's on local then right?

chenxsan: Right, I'm testing on my local computer.

examples/getting-started/README.md
d3lm: Removed the sign completely.

chenxsan (approved):

PR #5090, opened 2021-06-16
docs(examples): fix StackBlitz example for module federation

SamVerschueren:
This PR fixes the link when clicking on the StackBlitz button. It will install the dependencies, run the `npm start` command and open a second blank terminal.

I also removed the `.stackblitzrc` because it's covered by the query parameters and allows more flexibility like opening a second or third terminal and what not. So in this case, the `.stackblitzrc` is not needed.

### DIFF
--- a/examples/module-federation/.stackblitzrc
+++ b/examples/module-federation/.stackblitzrc
@@ -1,3 +1,0 @@
-{
-  "startCommand": "npm start"
-}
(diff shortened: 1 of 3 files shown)


chenxsan:
Can you update https://github.com/webpack/webpack.js.org/blob/master/src/content/plugins/module-federation-plugin.mdx as well?

CleanShot 2021-06-16 at 18 20 25@2x

SamVerschueren:
Good catch!

chenxsan (approved):
Thanks for the contribution!

Issue #5087, opened 2021-06-15
improve description `DefinePlugin`


alexander-akait:
**Feature to document**

https://webpack.js.org/plugins/define-plugin/

**Author(s)**

Any

**Additional information**

> The DefinePlugin allows you to create global constants which can be configured at compile time.

It is not fully true. Here problems https://github.com/webpack/webpack/discussions/13571?converting=1

Thanks

[ ] I am willing to work on this issue and submit a pull request.

PR #5091, opened 2021-06-16
docs(plugins): revise DefinePlugin

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/5087

### DIFF
--- a/src/content/plugins/define-plugin.mdx
+++ b/src/content/plugins/define-plugin.mdx
@@ -7,9 +7,10 @@ contributors:
   - byzyk
   - EugeneHlushko
   - smonusbonus
+  - chenxsan
 ---
 
-The `DefinePlugin` allows you to create global constants which can be configured at **compile** time. This can be useful for allowing different behavior between development builds and production builds. If you perform logging in your development build but not in the production build you might use a global constant to determine whether logging takes place. That's where `DefinePlugin` shines, set it and forget it rules for development and production builds.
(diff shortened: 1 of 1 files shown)


chenxsan:
cc @matthias-ccri Do you think the revised version is appropriate?

matthias-ccri:
Hi @chenxsan thanks for doing this! I see the phrase "arbitrary code" which at least to me is very similar to "arbitrary code execution", which is a bad thing in the area of software vulnerabilities. That's just my perspective, but it might be good for the docs to say "values or expressions" instead of "arbitrary code".

```markdown
The `DefinePlugin` allows you to replace variables in your code with other values or expressions at **compile** time.
```

I see the previous effort to emphasize "at compile time", which I missed previously, but it reads like "this plugin does stuff at compile time" which is true for everything webpack, since webpack is a compiler. It can be removed:

```markdown
The `DefinePlugin` allows you to replace variables in your code with other values or expressions.
```

And it could be even more direct:

```markdown
The `DefinePlugin` replaces variables in your code with other values or expressions.
```

chenxsan:
@matthias-ccri Actually webpack does have runtime code running in the browser, and I believe plugins can contribute to that as well.

What about:

```
The `DefinePlugin` replaces variables in your code with other values or expressions at **compile** time.
```

matthias-ccri:
That looks good to me!

chenxsan:
@matthias-ccri Thank you, I'm grateful for your help.

Issue #5105, opened 2021-06-23
Keep pathname when selecting language dropdown list


cs09g:
**Feature to document**
It's better to replace hostname only when selecting a language on the dropdown list(ENGLISH, 中文, 한국어).

When you search for something on google and enter the document site then change the language. it moves to the main page of the language. Actually, the pathname is the same and the difference is the hostname.

* [x] I am willing to work on this issue and submit a pull request.

chenxsan:
Makes sense. Feel free to send a pull request.

PR #5108, opened 2021-06-23
feat: #5105 add pathname at the end of dropdown language url

cs09g:
Added pathname to keep the consistency of reading while changing language.

Closes https://github.com/webpack/webpack.js.org/issues/5105

### DIFF
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -44,7 +44,7 @@ function Navigation({
   switchTheme,
 }) {
   const themeSwitcher = () => switchTheme(theme === DARK ? LIGHT : DARK);
-  const [locationHash, setLocationHash] = useState('');
+  const [locationHash, setLocationHash] = useState(hash);
 
   useEffect(() => {
     if (isClient) {
--- a/src/components/Navigation/Navigation.jsx
+++ b/src/components/Navigation/Navigation.jsx
@@ -29,13 +29,15 @@ const { DARK, LIGHT } = THEME;
 export default class Navigation extends Component {
   static propTypes = {
     pathname: PropTypes.string,
+    hash: PropTypes.string,
     links: PropTypes.array,
     toggleSidebar: PropTypes.func,
     theme: PropTypes.string,
     switchTheme: PropTypes.func,
   };
   render() {
-    const { pathname, links, toggleSidebar, theme, switchTheme } = this.props;
+    const { pathname, hash, links, toggleSidebar, theme, switchTheme } =
+      this.props;
     const themeSwitcher = () => switchTheme(theme === DARK ? LIGHT : DARK);
 
     return (
@@ -74,17 +76,17 @@ export default class Navigation extends Component {
                 items={[
                   {
                     title: 'English',
-                    url: `https://webpack.js.org${pathname}`,
+                    url: `https://webpack.js.org${pathname}${hash}`,
                   },
                   {
                     lang: 'zh',
                     title: '中文',
-                    url: `https://webpack.docschina.org${pathname}`,
+                    url: `https://webpack.docschina.org${pathname}${hash}`,
                   },
                   {
                     lang: 'ko',
                     title: '한국어',
-                    url: `https://webpack.kr${pathname}`,
+                    url: `https://webpack.kr${pathname}${hash}`,
                   },
                 ]}
               />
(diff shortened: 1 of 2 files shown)


chenxsan:
Really nice feature. However there is one problem, say I'm browsing `/concepts/plugins/#usage` and want to check the respective Chinese version, clicking the button would bring me to `chinese-site/concepts/plugins/` instead of `chinese-site/concepts/plugins/#usage`.

I think we should retain the URI fragment as well, just like how we did when changing the version of webpack documentation:

CleanShot 2021-06-24 at 12 11 58@2x

Here's the code https://github.com/webpack/webpack.js.org/blob/master/src/components/Sidebar/Sidebar.jsx#L25 just in case you want to take a look.

chenxsan:
Btw, can you sign cla here https://cla.js.foundation/webpack/webpack.js.org?pullRequest=5108?

cs09g:
@chenxsan I've signed! 😀

cs09g:
> Really nice feature. However there is one problem, say I'm browsing `/concepts/plugins/#usage` and want to check the respective Chinese version, clicking the language button would bring me to `chinese-site/concepts/plugins/` instead of `chinese-site/concepts/plugins/#usage`.
> 
> I think we should retain the URI fragment as well, just like how we did when changing the version of webpack documentation:
> 
> CleanShot 2021-06-24 at 12 11 58@2x
> 
> Here's the code https://github.com/webpack/webpack.js.org/blob/master/src/components/Sidebar/Sidebar.jsx#L25 just in case you want to take a look.

Good catch! what if I pass `location.hash` as a prop to use it on `Navigation`?

chenxsan:
I just found a bug in the current implementation.

Here's how to reproduce:

Open https://webpack-js-org-git-fork-cs09g-feature-issue-bdc88a-webpack-docs.vercel.app/concepts/entry-points/#multi-page-application, then change the version. You'll see the URI fragment is missing.

cs09g:
@chenxsan Hmm, the hash is missing. Is that based on the last commit on this PR? Or is something different on the build with on localhost?

chenxsan:
We have this preview https://github.com/webpack/webpack.js.org/pull/5108#issuecomment-867109431 synced with the latest commit of current pull request, so the preview should be up to date.

The build is actually a server-side rendered first, then hydrated on client-side app. So, yes, it's not the same as the dev one on localhost.

cs09g:
@chenxsan I've pushed one more commit after the preview generated. I guess it's fine with the next deployment. check this out. https://webpack-js-org-1ozc7s86g-webpack-docs.vercel.app/

chenxsan:
Neither does it work. You can open https://webpack-js-org-1ozc7s86g-webpack-docs.vercel.app/concepts/entry-points/#multi-page-application to inspect.

But I did notice the discrepancy here:

CleanShot 2021-06-25 at 09 58 38@2x

The `href` prop in React virtual DOM is up to date, however the real DOM is stale.

So the problem could be `hash` is client side only, and it's not hydrated on client side causing the issue. How about handling it with `useEffect`?

cs09g:
I got it. I could check this out with `yarn build-test' on localhost. By the way, It's a class component. It needs to be refactored.

chenxsan:
@cs09g Seems the CLA is not signed, can you double check?

src/components/Navigation/Navigation.jsx
```diff
@@ -26,93 +26,27 @@ import StackOverflowIcon from '../../styles/icons/stack-overflow.svg';
...
+function Navigation({
+  pathname,
+  hash = '',
+  links,
+  toggleSidebar,
+  theme,
+  switchTheme,
+}) {
+  const themeSwitcher = () => switchTheme(theme === DARK ? LIGHT : DARK);
+  const [locationHash, setLocationHash] = useState('');
```
chenxsan, line 47:
```suggestion
  const [locationHash, setLocationHash] = useState(hash);
```
Can we initialize it with a default value from prop? Any possible flaw?

cs09g:
@chenxsan I did. Some of my commits pushed as different emails with Github. Should I re-commit with the email that I signed in CLA?
image

src/components/Navigation/Navigation.jsx
cs09g: I was afraid of `undefined` as a string. but actually, that's not going to happen. I'll fix it

chenxsan:
> @chenxsan I did. Some of my commits pushed as different emails with Github. Should I re-commit with the email that I signed in CLA?
> image

Yeah, the bot is not that smart to handle it.

chenxsan (approved):
Really great enhancement! Now users can toggle between different languages seamlessly. Thanks for your contribution.

cs09g:
@chenxsan Thanks for the warm review!

PR #5115, opened 2021-06-25
docs(configuration): sort output configurations alphabetically

chenxsan:
Sort configurations alphabetically.

## Before

CleanShot 2021-06-25 at 20 22 19@2x

## After

CleanShot 2021-06-25 at 20 23 42@2x

FYI @QC-L @YukJiSoo as this one might bring a lot of changes to the downstream translation repository. Or maybe I should split this pull request into some small ones?  Let me know what you think here.


chenxsan:
I think I'll do it in some small pull requests so it won't be a pain for downstream repository to merge.

Issue #4441, opened 2021-01-16
Proposal: remove contributors section in the web page


chenxsan:
I can understand benefits to show the `Contributors` section in the web page,

1. it might attract new contributors
2. it gives contributors credit for contributing

However, it becomes a problem when there're so many contributors on one page, for example, https://webpack.js.org/guides/code-splitting/

<img src='https://user-images.githubusercontent.com/1091472/104813385-dd131300-5843-11eb-925d-1daff1b54f08.png' width='200' />

If you're browsing that page on mobile phone:

<img src='https://user-images.githubusercontent.com/1091472/104813593-139d5d80-5845-11eb-93d2-23d67c628a86.png' width='200' />

This is sort of inconvenient if users want to scroll to the end just to make sure they don't miss something important after the `Contributors` section.

And I don't think webpack users would care about this section, it's a waste of their times and bandwidths to load contributors' avatar even though we are lazy loading them.

In addition, it will be a problem for contributors as well since they have to scroll multiple pages to see the body text:

<img src='https://user-images.githubusercontent.com/1091472/104845127-c5ed2780-590e-11eb-8106-c40c5e09e754.png' width='200' />

Thus I'm proposing to remove the whole `Contributors` section from the web page and the `contributors` field from the frontmatter of markdown files.

I don't think removing those would stop contributors from contributing, if we really want to show those contributors, we can add a link to pages like https://github.com/webpack/webpack.js.org/contributors-list/master/src/content/guides/code-splitting.md which list all contributors who had contributed.

snitin315:
But It makes me happy to see my name there.

snitin315:
Maybe we can limit it to recent 10-15 contributors name and display the rest with the help of a button labeled as `See more contributors`

chenxsan:
> But It makes me happy to see my name there. 

Let's talk from the perspective of common webpack users, is it beneficial to most common webpack users?

I can understand your frustration to remove the section as I am one of the contributors listed in those pages.

And regarding the 'See More' suggestion, sure it would lessen the pain here, but it would only hide the problem as one day in the future there may be 200 contributors listed on the page, which then would become a real problem to contributors as they have to scroll down the markdown page to see the body part.

alexander-akait:
I agree about it is unnecessary on every page, I think we can create separate page and move all contributors to there, also we can get them from github to avoid manually adding

snitin315:
Single separate page sounds good 👌

montogeek:
[...]

PR #5134, opened 2021-07-01
hide 404 contributors from page

chenxsan:
Some contributors are 404 on github now, it would be better to hide them from the webpage,

CleanShot 2021-07-01 at 08 10 07@2x

Note that we're still keeping them in the source file.

### DIFF
--- a/src/components/Contributors/404.js
+++ b/src/components/Contributors/404.js
@@ -1,0 +1,7 @@
+// contributors 404
+export const contributorsNotFound = [
+  'bartushek',
+  'rynclark',
+  'henriquea',
+  'makuzaverite',
+];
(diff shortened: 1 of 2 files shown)


src/components/Contributors/404.js
```diff
@@ -0,0 +1,7 @@
+// contributors 404
+export const contributorsNotFound = [
+  'bartushek',
+  'rynclark',
+  'henriquea',
+  'makuzaverite',
```
snitin315, line 6: Hardcoding them is not a good idea. Is there any alternative?

chenxsan: What's your suggestion other than maintaining a list manually? I can think of reading the list from github API, however there're thousands of contributors on this repository, it just doesn't make sense to read it every time we build the site because most of the time the list should stay the same.

snitin315: Hmm, reading API everytime doesn't makes sense. Perhaps another reason for removing this section in favor of a separate contributors page. https://github.com/webpack/webpack.js.org/issues/4441

snitin315 (approved):

PR #5146, opened 2021-07-04
docs(configuration): move cache to a separate page

snitin315:
_describe your changes..._

`cache` is an important option with large documentation hence moving to a separate page


chenxsan (changes requested):
Makes sense. However, make sure to update all the redirections here https://github.com/webpack/webpack.js.org/blob/c7f547e6ff7f4c6ee5ba6b8c776b9eb6ebb1008c/webpack.ssg.js#L57 otherwise many links on external websites to the old cache options would be broken.

chenxsan:
Btw, use `.mdx` instead. `<Badge />` component used in the page won't work in `.md`.

snitin315:
@chenxsan updated 👍🏻. Please take a look.

webpack.ssg.js
```diff
@@ -94,6 +94,41 @@ module.exports = (env) =>
           'loaders/url-loader': 'https://v4.webpack.js.org/loaders/url-loader',
           'loaders/file-loader':
             'https://v4.webpack.js.org/loaders/file-loader',
+          'configuration/other-options/#cache': 'configuration/cache',
```
chenxsan, line 97:
```suggestion
          'configuration/other-options/#cache': '/configuration/cache',
```
Need a prefix `/`. Otherwise it won't redirect, e.g., https://webpack-js-org-git-cache-page-webpack-docs.vercel.app/configuration/other-options/#cache will stay on other-options page.

snitin315: Still not working 😞

chenxsan: I just noticed it too. Seems the [redirect plugin](https://github.com/bebraw/redirect-webpack-plugin/blob/master/index.js#L15) won't work against those fragments' redirection. I think we can do it on client side. For example, include another `useEffect(() => {})` in `Site.jsx` to achieve the redirections of fragment type.

```js
useEffect(() => {
// if the current pathname is /configuration/other-options/#cache
// redirect it to /configuration/cache
// same could be applied to others
}, []);
```

snitin315: Thanks, working now 🚀

src/components/Site/clientSideRedirections.js
```diff
@@ -0,0 +1,13 @@
+// RedirectWebpackPlugin won't work for URI fragments
+// hence we redirect them with client side script
+export default function (location) {
```
chenxsan, line 3: I'm adding a function for future client slide redirections. Otherwise we might bloat `Site.jsx`.

snitin315: Awesome, Thanks!

chenxsan (approved):

PR #5147, opened 2021-07-05
docs(Getting Started): update diff

donggua-c:
actually we only added `"private": true,` without deleted `"main": index.js,`


chenxsan:
That‘s not true, see `as well as removing the main entry` in above paragraph.

CleanShot 2021-07-05 at 12 53 24@2x

PR #5160, opened 2021-07-06
New way to invoke onListening

jonathanrevell:
You need to call server.server.address() since server.listeningApp is no longer available in webpack-dev-server 4.0.0+


chenxsan:
The current doc targets the latest webpack-dev-server, i.e. v3. v4 is still in beta as far as I know.

chenxsan:
I'm closing it as it's already addressed per https://webpack.js.org/configuration/dev-server/#devserveronlistening. Thanks anyway.

PR #5161, opened 2021-07-07
docs(configuration): add cache.compression

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/5140

### DIFF
--- a/src/content/configuration/cache.mdx
+++ b/src/content/configuration/cache.mdx
@@ -113,7 +113,11 @@ module.exports = {
 
 `false | 'gzip' | 'brotli'`
 
-Compression type used for the cache files. By default it is `false` for `development` mode and `gzip` for `production` mode.
+<Badge text="5.42.0+" />
+
+Compression type used for the cache files. By default it is `false` for `development` mode and `'gzip'` for `production` mode.
+
+`cache.compression` option is only available when [`cache.type`](#cachetype) is set to `'filesystem'`.
 
 **webpack.config.js**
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/cache.mdx
```diff
@@ -109,6 +109,24 @@ module.exports = {
 };
 ```
 
+### cache.compression
+
+`false | 'gzip' | 'brotli'`
+
+Compression type used for the cache files. By default it is `false` for `development` mode and `gzip` for `production` mode.
```
snitin315, line 116: defaults- https://github.com/webpack/webpack/blob/b53a06437a3e8b2756c550c1be6ccb33fdba03d7/lib/config/defaults.js#L301

src/content/configuration/cache.mdx
```diff
@@ -109,6 +109,24 @@ module.exports = {
 };
 ```
 
+### cache.compression
+
+`false | 'gzip' | 'brotli'`
+
```
chenxsan, line 115:
```suggestion
`false | 'gzip' | 'brotli'`

<Badge text="5.42.0+" />

```

src/content/configuration/cache.mdx
chenxsan, line 116:
```suggestion
Compression type used for the cache files. By default it is `false` for `development` mode and `'gzip'` for `production` mode.

`cache.compression` option is only available when [`cache.type`](#cachetype) is set to `'filesystem'`.
```

chenxsan (approved):

PR #5164, opened 2021-07-08
improve page styling

chenxsan:
1. Move the `Edit document` button to the end.

    Generally, I don't think users would use the button frequently, in addition, when we have a long title, it would becomes a little messy:

    CleanShot 2021-07-08 at 18 58 18@2x

    **Before**:

    CleanShot 2021-07-08 at 19 13 04@2x

    **After**:
    CleanShot 2021-07-08 at 19 12 07@2x


src/components/Page/AdjacentPages.jsx
```diff
@@ -15,7 +15,7 @@ AdjacentPages.propTypes = {
 export default function AdjacentPages({ previous, next }) {
   if (!previous && !next) return null;
   return (
-    <div className="adjacent-links">
+    <div className="adjacent-links print:hidden">
```
chenxsan, line 18: No need to show those links for printed page.

snitin315: Now they are easy to miss, I think they should have better visibility. Maybe we can just move them `10px` up?

chenxsan: Could you can share a screenshot to show why do you think they are easy to miss?

snitin315: Because they are not at bottom of the page, users might have to scroll long if  docs are large for example - 
https://webpack-js-org-git-fork-chenxsan-feature-im-b01bac-webpack-docs.vercel.app/configuration/dev-server/

But I see gatsby is also using the same place for these links -
https://www.gatsbyjs.com/docs/how-to/querying-data/use-static-query/

So I think it will be okay for us too. I don't have a strong opiniom here.

chenxsan: Users care about the content most, not those links. They are for contributors, not that important thing on the page as long as we have it presented somewhere.

snitin315 (approved):

PR #5193, opened 2021-07-18
fix syntax error in environment-variables.md

Oscar-21:
Found a small syntax error where this arrow function is using function body braces instead of adding parentheses so it can return an object literal expression. :)


chenxsan:
Thanks but this doesn't seems right. You can't use `console.log` inside `({})`, and we already have a `return {}` inside the function body.

PR #5196, opened 2021-07-19
add webpack-options page

chenxsan:


src/components/Site/clientSideRedirections.js
```diff
@@ -12,8 +12,9 @@ export default function (location) {
 
   if (location.pathname === '/webpack-options/') {
```
alexander-akait, line 13: Maybe `/webpack/options/` so in future we will add `/webpack-dev-server/options/`

chenxsan: How about `https://webpack.js.org/webpack/?option=output/filename` `https://webpack.js.org/webpack-dev-server/?option=xxx`? It's more terse.

Or `https://webpack.js.org/redirect/webpack/?option=output/filename` which is more future proof in case we will be using `/webpack/` path on this site in future.

chenxsan:
I'm afraid it won't work as the documents don't share a common pattern.

E.g.,

1. `/configuration/module/#modulegenerator`
2. `/configuration/module/#rulelayer`

This pull request would work only for the first pattern. And there could be other patterns besides the aforementioned two.

avivkeller:
Stale

PR #5215, opened 2021-07-22
docs(api): add tip for status handler in HMR API

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/5213

### DIFF
--- a/src/content/api/hot-module-replacement.mdx
+++ b/src/content/api/hot-module-replacement.mdx
@@ -357,6 +357,8 @@ import.meta.webpackHot.addStatusHandler((status) => {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


src/content/api/hot-module-replacement.mdx
```diff
@@ -357,6 +357,8 @@ import.meta.webpackHot.addStatusHandler((status) => {
 });
 ```
 
+T> If a status handler returns a `Promise` then it will wait for the `Promise` to resolve before continuing.
```
chenxsan, line 360:
```suggestion
Bear in mind that when the status handler returns a `Promise`, the HMR system will wait for the `Promise` to resolve before continuing.
```

chenxsan (approved):

PR #5216, opened 2021-07-22
docs(configuration): sort other-options alphabetically

snitin315:
sort `other-options` alphabetically


chenxsan:
It would be painful for the downstream repositories (Chinese translation and korean translation) to sync large changes. That's why I sorted just two or three options in one pull request. So if you are going to do it, I would suggest doing it with small changes maybe every other week because the downstream won't sync that quickly.

snitin315:
I just moved `infrastructureLogging` and `name` up but it seems Github is showing like I moved other options down. 

>So if you are going to do it, I would suggest doing it with small changes maybe every other week because the downstream won't sync that quickly.

I get your point here I will keep this in mind in the future, do I need to update this PR with fewer changes?

chenxsan:
I think we can move just one option in one commit, like this one https://github.com/webpack/webpack.js.org/pull/5187/commits. Also prefer to use semantic commit messages so the downstream can follow it easily.

alexander-akait:
Sometimes I think moving `infrastructureLogging` on own page with example will be great idea, because logging is main for many developers

snitin315:
> Sometimes I think moving `infrastructureLogging` on own page with example will be great idea, because logging is main for many developers

Makes sense to me 👍🏻

chenxsan:
So maybe sort these three as well considering the whole changes are already a lot?

chenxsan (approved):

PR #5217, opened 2021-07-22
docs(configuration): use `--stats-colors`

snitin315:
`--colors` is no longer valid. `--stats-colors` is the latest option.

### DIFF
--- a/src/content/configuration/stats.mdx
+++ b/src/content/configuration/stats.mdx
@@ -452,7 +452,7 @@ module.exports = {
 It is also available as a CLI flag:
 
 ```bash
-webpack-cli --stats-colors
+npx webpack --stats-colors
 ```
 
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
We need fix in out CLI :smile:

snitin315:
> We need fix in out CLI 😄

I think `--stats-colors` & `--no-stats-colors` works as expected,


Screenshot 2021-07-22 at 6 33 09 PM


but CLI logger does not respect `stats.colors.red` which it should but I am afraid `stats.colors.red` will not be accessible everywhere we use `logger`.

alexander-akait:
> but CLI logger does not respect stats.colors.red which it should but I am afraid stats.colors.red will not be accessible everywhere we use logger.

ideally only `--stats-colors` should supports colors, other logger can ignore these, we need to add test

chenxsan (approved):

rishabh3112 (approved):

PR #5224, opened 2021-07-23
upgrade docsearch

chenxsan:
Upgrade docsearch to fix an issue here https://github.com/algolia/docsearch/pull/1021. Thus the `width` is no longer needed.


chenxsan:
Closing as it doesn't really solve the problem :)

PR #5230, opened 2021-07-24
upgrade actions/setup-node@v2

chenxsan:
1. upgrade to `actions/setup-node@v2`
2. use its built-in `cache` option which should fulfill our demand without the complexity of `actions/cache@v2`.


.github/workflows/testing.yml
```diff
@@ -111,26 +81,7 @@ jobs:
         uses: actions/setup-node@v1
```
snitin315, line 81:
```suggestion
        uses: actions/setup-node@v2
```

PR #5246, opened 2021-07-31
fix ci failure

chenxsan:
We don't use any api key for opencollective at the moment, which means it's easy to hit the api limit defined here https://github.com/opencollective/opencollective-api/blob/main/server/routes.js#L75:

> // 10 requests / minute / ip for anonymous requests

And it's failing the CI now https://github.com/webpack/webpack.js.org/actions/runs/1083925129.

This pull request would query its api less often to avoid hitting the api limit.

### DIFF
--- a/src/utilities/fetch-supporters.js
+++ b/src/utilities/fetch-supporters.js
@@ -17,14 +17,10 @@ const absoluteFilename = path.resolve(
   filename
 );
 
-let graphqlEndpoint = 'https://api.opencollective.com/graphql/v2';
+const graphqlEndpoint = 'https://api.opencollective.com/graphql/v2';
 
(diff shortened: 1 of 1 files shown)


src/utilities/fetch-supporters.js
```diff
@@ -42,7 +42,7 @@ const membersGraphqlQuery = `query account($limit: Int, $offset: Int) {
 }`;
 
 // only query transactions in last year
-const transactionsGraphqlQuery = `query transactions($dateFrom: ISODateTime, $limit: Int, $offset: Int) {
+const transactionsGraphqlQuery = `query transactions($dateFrom: DateTime, $limit: Int, $offset: Int) {
```
znarf, line 45: Good job on fixing this.

This was indeed a recent breaking change. We thought it was only used internally and decided to proceed. We should have better monitored usage.

PR #5285, opened 2021-08-10
remove skips

chenxsan:
A follow-up of https://github.com/webpack/webpack.js.org/pull/5284 as that solution is not complete as I tested.


chenxsan:
It just took too long to detect external links. I don't think it's acceptable.

alexander-akait:
Maybe run this task by cron?

chenxsan:
Most of the time we don't need to do this because we'll ensure links are not broken when reviewing, the awesome-webpack page is an exception with a lot of links. But I can do it with some online tools since this seems to be one-off case.

PR #5293, opened 2021-08-13
Fix optimization.moduleIds setting in 5.mdx

akccakcctw:
**Upgrade webpack to 5** is the next step, and we can only use `optimization.moduleIds: 'hashed'` in v4


chenxsan (approved):
You are right, thanks for contribution.

PR #5312, opened 2021-08-20
Fix related link format

atesgoral:
The **Further Reading** section on https://webpack.js.org/guides/code-splitting/ has a dangling `&gt;`:

image

Fixing this by converting the HTML entities to the actual angle brackets:

image

**Note:** The last item in this list completely omits the `<link>` and just mentions `rel="preload"`, so it could also be a good idea to standardize either using a `<link>` or not.

### DIFF
--- a/src/content/guides/code-splitting.mdx
+++ b/src/content/guides/code-splitting.mdx
@@ -38,7 +38,7 @@ related:
     url: https://medium.com/webpack/link-rel-prefetch-preload-in-webpack-51a52358f84c
   - title: Preload, Prefetch And Priorities in Chrome
     url: https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf
-  - title: Preloading content with rel="preload"
+  - title: Preloading content with <link rel="preload" />
     url: https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content
 ---
 
--- a/src/content/guides/code-splitting.mdx
+++ b/src/content/guides/code-splitting.mdx
@@ -32,8 +32,9 @@ contributors:
   - smelukov
   - chenxsan
   - Adarah
+  - atesgoral
 related:
-  - title: &lt;link rel="prefetch/preload"&gt; in webpack
+  - title: <link rel="prefetch/preload"> in webpack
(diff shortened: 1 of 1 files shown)


src/content/guides/code-splitting.mdx
```diff
@@ -32,8 +32,9 @@ contributors:
   - smelukov
   - chenxsan
   - Adarah
+  - atesgoral
 related:
-  - title: &lt;link rel="prefetch/preload"&gt; in webpack
+  - title: <link rel="prefetch/preload"> in webpack
```
chenxsan, line 37:
```suggestion
  - title: <link rel="prefetch/preload" /> in webpack
```
It's a mdx file, seems we have to close the tag.

chenxsan (approved):
Thank you for your contribution!

atesgoral:
@chenxsan I was going to dive deeper into this linting error, but thanks for jumping in and tweaking it!

My hesitancy was due to two things:
1. Why is the markdown linter linting the frontmatter? It's supposed to be metadata whose syntax should be treated differently than the actual MDX content. I feel like this is a problem with the linter or its configuration.
2. Since `<link>` elements are meant to be non-self-closing in the HTML context, and since webpack is a generic bundling tool that is not married to React/JSX, should webpack documentation really bend to the MDX linter's will?

I think there's going to be a better tweak (perhaps at linter configuration level) to this collaborative quick patch we shipped :) I might take a stab at it later, just to sate my curiosity...

chenxsan:
@atesgoral Regarding your second concern, here's an explanation for why we would migrate to MDX https://github.com/webpack/webpack.js.org/pull/4412#issue-552059006 considering this site is actually an application built with React :) Generally, MDX improves the performance a lot because we can take advantage of MDX's [MDXProvider](https://mdxjs.com/advanced/components#mdxprovider).

And yes, it seems to be a bug in [eslint-mdx](https://github.com/mdx-js/eslint-mdx). Let me know if you come to any conclusion. Or I can file an issue over their repository, the maintainer is very kind to help as far as I know.

atesgoral:
@chenxsan I couldn't figure out how to tweak the ESLint configuration to make the problem go away.

chenxsan:
@astonizer I've filed an issue over the eslint-mdx repository, we'll see if it's a bug in it.

PR #5233, opened 2021-07-24
migrate md to mdx

chenxsan:
As explained in https://github.com/webpack/webpack.js.org/pull/4412#issue-552059006, we'll migrate all `.md` files to `.mdx` for improving performance. And this is the pull request to finish a part of the job.

## TODO

Some pages are not compatible with `mdx` so far. Will review them at the end.

- [x] `src/content/branding.md`
- [x] `src/content/index.md`

PR #5316, opened 2021-08-20
drop md configurations

chenxsan:
We've migrated all md files to mdx in https://github.com/webpack/webpack.js.org/pull/5233 and https://github.com/webpack/webpack.js.org/pull/5298, it's time to drop all those md related configurations.


package.json
```diff
@@ -123,8 +121,7 @@
     "remark-extract-anchors": "1.1.1",
     "remark-frontmatter": "^3.0.0",
     "remark-gfm": "^1.0.0",
-    "remark-html": "^13.0.1",
-    "remark-loader": "^4.0.0",
+    "remark-html": "^13.0.0",
     "remark-refractor": "git://github.com/montogeek/remark-refractor.git",
```
JounQin, line 125: `git` protocol is incorrect for other contributors.

```suggestion
    "remark-refractor": "montogeek/remark-refractor",
```

package.json
```diff
@@ -123,8 +121,7 @@
     "remark-extract-anchors": "1.1.1",
     "remark-frontmatter": "^3.0.0",
     "remark-gfm": "^1.0.0",
-    "remark-html": "^13.0.1",
-    "remark-loader": "^4.0.0",
+    "remark-html": "^13.0.0",
```
JounQin, line 124: Any issue after removing it?

chenxsan: Nope. We have some remark plugins using it for testing, see https://github.com/webpack/webpack.js.org/blob/0f40d4ed63077ec53426bd349f432fe60225a6a9/src/remark-plugins/remark-custom-asides/index.test.js#L10. That's why I added it back.

package.json
chenxsan: Thanks, I didn't know it works in this way.

PR #5364, opened 2021-08-24
docs(api): add `serve` command

snitin315:
add `serve` command usage.

### DIFF
--- a/src/content/api/cli.mdx
+++ b/src/content/api/cli.mdx
@@ -37,18 +37,18 @@ Read the [installation guide](/guides/installation) if you don't already have we
 
 webpack-cli offers a variety of commands to make working with webpack easier. By default webpack ships with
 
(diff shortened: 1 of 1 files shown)


src/content/api/cli.mdx
```diff
@@ -190,6 +190,22 @@ npx webpack configtest ./webpack.config.js
...
+Run the webpack dev server.
+
+```bash
+npx webpack serve [options]
+```
+
+**example**
+
+```bash
+npx webpack serve --hot --live-reload
```
alexander-akait, line 204: I think better to use `--static` here, because it is more popular

snitin315: lets `--static` and `--open`: 

```suggestion
npx webpack serve --static --open
```

alexander-akait (approved):

PR #5396, opened 2021-09-01
Use ESM

chenxsan:
The whole remark ecosystem has migrated to ESM now. Packages like `node-fetch` did it as well. Thus I think we should commence the migration, otherwise we might get stuck in the CommonJS world with a lot of outdated packages.

- [x] Closes https://github.com/webpack/webpack.js.org/pull/5337.
- [x] Closes https://github.com/webpack/webpack.js.org/pull/5389
- [x] Closes https://github.com/webpack/webpack.js.org/pull/5350
- [x] Closes https://github.com/webpack/webpack.js.org/pull/5272
- [ ] closes https://github.com/webpack/webpack.js.org/pull/5280
- [x] closes https://github.com/webpack/webpack.js.org/pull/5328
- [x] closes https://github.com/webpack/webpack.js.org/pull/5278
- [x] closes https://github.com/webpack/webpack.js.org/pull/5243

## some problems

1. `remark-gfm 2.0.0` is not compatible with `@mdx-js/loader`. While replacing `@mdx-js/loader` with [`xdm`](https://github.com/wooorm/xdm) would fix the problem, migrating to ESM-only `xdm` is much difficult than I expected when we have [server side rendering](https://github.com/webpack/webpack.js.org/blob/master/src/server.jsx#L33) enabled, so I might do it in another pull request if I can figure out how.


package.json
```diff
@@ -28,15 +28,15 @@
     "clean-printable": "rimraf src/content/**/printable.mdx",
     "preclean": "run-s clean-dist clean-printable",
     "clean": "rimraf src/content/**/_*.mdx src/**/_*.json repositories/*.json",
-    "start": "npm run clean-dist && webpack serve --config webpack.dev.js --env dev --progress --node-env development",
+    "start": "npm run clean-dist && WEBPACK_CLI_FORCE_LOAD_ESM_CONFIG=true webpack serve --config webpack.dev.mjs --env dev --progress --node-env development",
```
alexander-akait, line 31: Should work without this

chenxsan: Cool.

snitin315 (approved):
Nice!

PR #5420, opened 2021-09-07
docs(module-federation): Fix typo

Alevale:
_Small fix to correct a typo on the URL for the module federation "public-path" explanation_


snitin315 (approved):
Thanks!

PR #5475, opened 2021-09-19
docs(guides): remove `buble.js` from the guide

alexander-kilyushin:
Remove `buble.js` from the guide because the project has not been maintained for over a year and the website http://buble.surge.sh/ is no longer available.


chenxsan (approved):
Actually I am able to access that address :) But yeah, buble seems no longer maintained, so makes sense to remove.

PR #5493, opened 2021-09-25
docs(configuration): add tip for `xxhash64` algorithm support for hashing

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/5491

### DIFF
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -697,6 +697,8 @@ module.exports = {
 
 Make sure that the hashing function will have `update` and `digest` methods available.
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.mdx
```diff
@@ -697,6 +697,8 @@ module.exports = {
 
 Make sure that the hashing function will have `update` and `digest` methods available.
 
+T> Since `webpack v5.54.0+`, `hashFunction` supports `xxhash64` as a fast algorithm. It will be used as default when [`experiments.futureDefaults`](/configuration/experiments/#experimentsfuturedefaults) is enabled.
```
chenxsan, line 700:
```suggestion
T> Since `webpack v5.54.0+`, `hashFunction` supports `xxhash64` as a faster algorithm, which will be used as default when [`experiments.futureDefaults`](/configuration/experiments/#experimentsfuturedefaults) is enabled.
```

chenxsan (approved):

PR #5495, opened 2021-09-25
docs(configuration): document experiments.cacheUnaffected

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/5492

### DIFF
--- a/src/content/configuration/cache.mdx
+++ b/src/content/configuration/cache.mdx
@@ -3,6 +3,7 @@ title: Cache
 sort: 12
 contributors:
   - snitin315
+  - chenxsan
 ---
 
 ## cache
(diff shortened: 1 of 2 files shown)


src/content/configuration/cache.mdx
```diff
@@ -109,6 +110,25 @@ module.exports = {
 };
 ```
 
+### cache.cacheUnaffected
+
+Cache computation of modules which are unchanged and reference only unchanged modules. It can only be used along with [`cache.type`](#cachetype) of `'memory'`, besides, [`experiments.cacheUnaffected`](/configuration/experiments/#experimentscacheunaffected) must be enabled to use it.
+
+- Type: `boolean`
+- v5.54.0+
```
snitin315, line 118: Not using `<Badge text="v5.34.0+" />` ?

chenxsan: Too distracting when they are many badges on one page. Body text serves the purpose well.

snitin315 (approved):

PR #5496, opened 2021-09-26
fix: show `Contributors`/`Contributor` correctly

snitin315:
## Before

Screenshot 2021-09-26 at 7 00 36 AM

## After

Screenshot 2021-09-26 at 6 59 46 AM


chenxsan (approved):
Looks good.

Issue #5538, opened 2021-10-05
> There is no such placeholder


alexander-akait:
> There is no such placeholder

<img width="1027" alt="截屏2021-06-28 下午11 09 37" src="https://user-images.githubusercontent.com/11417330/123660740-7ff5a180-d866-11eb-8f85-cfa29146355d.png">

Putting [ext] in Output related configuration items in the document can easily mislead developers. As of 2021-06, people are still misled by this document, which is enough to show that the document needs to be revised.

> 文档里面把 [ext] 放在 Output 相关配置项中，非常容易误导开发者。时至 2021-06 仍有人被这份文档误导，足以说明文档有需要修改的地方。

_Originally posted by @minikinl in https://github.com/webpack/webpack/issues/6753#issuecomment-869773269_

alexander-akait:
We should improve docs and say you can't use it for `output.filename`

PR #5539, opened 2021-10-06
docs(configuration): add warning about `[ext]` and `output.filename`

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/5538

### DIFF
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -580,6 +580,8 @@ Substitutions available on File-level:
 | [name]     | Only filename without extension or path            |
 | [ext]      | Extension with leading `.`                         |
 
+W> `[ext]` is not supported for [output.filename](#outputfilename).
+
 Substitutions available on URL-level:
 
 | Template | Description |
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.mdx
```diff
@@ -580,6 +580,8 @@ Substitutions available on File-level:
 | [name]     | Only filename without extension or path            |
 | [ext]      | Extension with leading `.`                         |
```
chenxsan, line 581:
```suggestion
| [ext]      | Extension with leading `.` (not available for [output.filename](#outputfilename))                        |
```

I would suggest mentioning it in description than another warning block.

src/content/configuration/output.mdx
```diff
@@ -580,6 +580,8 @@ Substitutions available on File-level:
...
 
+W> `[ext]` is not supported for [output.filename](#outputfilename).
+
```
chenxsan, line 584:
```suggestion
```

chenxsan (approved):

PR #5616, opened 2021-10-24
[Docs]: Changes to diff blocks in Code Splitting Section

g4rry420:
Signed-off-by: Gurkiran Singh <gurkiransinghk@gmail.com>

Added the diff blocks code changes between entry dependency and SplitChunksPlugin. Previously, the code changes weren't matching, so I couldn't figure which line to remove or add while I am following the documentation guide.


chenxsan:
They're two parallel options:

CleanShot 2021-10-25 at 08 07 10@2x

`SplitChunksPlugin` code wasn't based on the other.

PR #5646, opened 2021-11-01
Fix `resolveData` casing in `beforeResolve` hook docs.

henryqdineen:
I noticed the callback parameter for other hooks lowercase `resolveData` so I updated this for consistency. Thanks.


chenxsan (approved):
Thanks.

PR #5692, opened 2021-11-15
feat: Take default theme as system theme

harish-sethuraman:
Why this change?
- When the site loads for the first time we default the theme with light. But there are chances were the users might prefer site to take system's preference.

What have I done? 
- Default theme will be same as user's system theme preference.
- When the user changes the system's preference the site's changes the theme too.


chenxsan:
Actually we've thought about it when we were implementing the dark mode. You can never know users prefer which to which unless you make a poll for all users which is not possible, and that dark mode is an afterthought, so as long as it works well, we're not going to change it. But thank you anyway for taking your time.

PR #5711, opened 2021-11-22
docs(configuration): add `chunkLoading: false`

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/5677

### DIFF
--- a/src/content/configuration/entry-context.mdx
+++ b/src/content/configuration/entry-context.mdx
@@ -86,6 +86,8 @@ module.exports = {
 };
 ```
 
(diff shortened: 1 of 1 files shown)


chenxsan:
The `chunkLoading` is a part of the entry context, please add it to the entry context like other peers instead of being presented as a tip.

CleanShot 2021-11-24 at 08 12 53@2x

snitin315:
Actually, it is already there as `chunkLoading: 'jsonp'`

https://webpack.js.org/configuration/entry-context/#entry-descriptor

snitin315:
But I think we can still add it there for another entry point.

chenxsan (approved):

PR #5744, opened 2021-11-30
docs(configuration): Added dependencies docs

dangayle:
Added some docs from the other-options page directly to the multiple configuration docs where it makes more sense.

I had this use case where I wanted one webpack config rely upon the output of a previous run, but I was told it wasn't possible and I couldn't see it in the multiple configs docs, so I assumed it was true.

### DIFF
--- a/src/content/configuration/configuration-types.mdx
+++ b/src/content/configuration/configuration-types.mdx
@@ -90,6 +90,27 @@ module.exports = [
 
 T> If you pass a name to [`--config-name`](/api/cli/#config-name) flag, webpack will only build that specific configuration.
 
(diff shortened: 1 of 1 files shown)


dangayle:
BTW, the CLA assistant page is broken, so I can't fill out that doc.

snitin315:
> BTW, the CLA assistant page linked to in the example PR text is broken, so I can't fill out that doc.

The link is outdated, I will update it soon.

snitin315 (approved):
Makes sense to me 👍🏻

chenxsan (approved):

PR #5757, opened 2021-12-06
update Via the Node.js API dev-server.js

uikoo9:
Now the file dev-server.js runs with an error, including an error in the config part. Modify it to the correct example.


snitin315 (approved):
LGTM, thanks.

PR #5778, opened 2021-12-13
Update docs for entry-context

spicylemonhaha:
I think this current directory has a ambiguity. Does the current directory refer to the root directory of the project? Or is it the directory where webpack.config.js is located?

After I tested it, I found neither of them, the default should be the directory of the script that starts webpack. For example, I ran the script `webpack --config relativebase` in a non-project root directory. Then, the value of the context at this time is the directory where the script is located.

In order to better illustrate my ideas, I wrote a demo code, but I don’t know how to give the sample code, so I created a warehouse to store the sample code. https://github.com/spicylemonhaha/webpack-demo

### DIFF
--- a/src/content/configuration/entry-context.mdx
+++ b/src/content/configuration/entry-context.mdx
@@ -30,7 +30,8 @@ module.exports = {
 };
 ```
 
-By default, the current directory is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
+By default, use the directory where the script that started webpack is loacted. It's recommended to pass a value in your configuration. This makes your configuration 
+independent from CWD (current working directory).
 
 ---
 
--- a/src/content/configuration/entry-context.mdx
+++ b/src/content/configuration/entry-context.mdx
@@ -30,7 +30,7 @@ module.exports = {
 };
 ```
 
-By default, the current working directory of the Node.js is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
+By default, the current working directory of Node.js is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
 
 ---
 
--- a/src/content/configuration/entry-context.mdx
+++ b/src/content/configuration/entry-context.mdx
@@ -30,8 +30,7 @@ module.exports = {
 };
 ```
 
-By default, the current working directory of the Node.js process is used, but it's recommended to pass a value in your configuration. This makes your configuration 
-independent from CWD (current working directory).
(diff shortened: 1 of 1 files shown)


src/content/configuration/entry-context.mdx
```diff
@@ -30,7 +30,8 @@ module.exports = {
 };
 ```
 
-By default, the current directory is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
+By default, use the directory where the script that started webpack is loacted. It's recommended to pass a value in your configuration. This makes your configuration
```
chenxsan, line 33:
```suggestion
By default, use the directory where the script that started webpack is located. It's recommended to pass a value in your configuration. This makes your configuration 
```

chenxsan:
If you check the source code of webpack, that `context` defaults to `process.cwd()`,

CleanShot 2021-12-14 at 19 53 55@2x

So maybe just update the `the current directory` to `the current working directory`?

spicylemonhaha:
First of all, thank you for replying to my pr. After reading your reply and reviewing the relevant code, I think `the current working directory` is already very appropriate, but I think it may be easier for readers to understand it if it is changed to `the current working directory of the Node.js process`. If you still think the former is better, then I will change to the former. Thank you again for replying to my pr, you are the first person to reply to my pr.

src/content/configuration/entry-context.mdx
```diff
@@ -30,7 +30,8 @@ module.exports = {
 };
 ```
 
-By default, the current directory is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
+By default, the current working directory of the Node.js process is used, but it's recommended to pass a value in your configuration. This makes your configuration
```
chenxsan, line 33:
```suggestion
By default, the current working directory of the Node.js is used, but it's recommended to pass a value in your configuration. This makes your configuration 
```

Let's remove that `process`, seems terrifying for novice.

spicylemonhaha: Okay I've removed

src/content/configuration/entry-context.mdx
```diff
@@ -30,7 +30,7 @@ module.exports = {
 };
 ```
 
-By default, the current directory is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
+By default, the current working directory of the Node.js is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
```
chenxsan, line 33:
```suggestion
By default, the current working directory of Node.js is used, but it's recommended to pass a value in your configuration. This makes your configuration independent from CWD (current working directory).
```

chenxsan (approved):
Thanks for your contribution!

PR #5796, opened 2021-12-17
Add more explanation of the webpack 5 web worker bundling functionality

cmdcolin:
I was a little confused when reading the web workers document and have returned to it a couple times and only when i tried it out mysef to see that bundling happens with the "new Worker" usage was I aware that webpack does bundling with this, so I thought I'd add some notes about this to the docs!

Let me know if this makes sense

### DIFF
--- a/src/content/guides/web-workers.mdx
+++ b/src/content/guides/web-workers.mdx
@@ -13,7 +13,7 @@ As of webpack 5, you can use [Web Workers](https://developer.mozilla.org/en-US/d
 new Worker(new URL('./worker.js', import.meta.url));
 ```
 
-The syntax was chosen to allow running code without bundler since using `new Worker`) is available in native ECMAScript modules in the browser (https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker) but in webpack, worker.js and the files that worker.js imports can be bundled.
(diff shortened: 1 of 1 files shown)


chenxsan:
Webpack is a bundler, so it's likely to bundle any files you have unless you tell it not to :) Hence I think it's not necessary to explain it? Otherwise we're going to explain a lot of native ECMAScript features. What do you think?

cmdcolin:
thanks for considering. my basic thought is it is valuable to have extra context.

my thought process is sort of like this:
"this seems like they are just using native js, i don't think it is not doing any bundling on the worker resources because it's just importing a js file by name, that must mean deep-thought.js has to go in the public directory in create-react-app or something like that"

I guess perhaps import.meta.url maybe helps resolve this placement issue in native js, but import meta's are new to me and this also not necessarily explained by the docs, and again, just not clear that bundling happens on this file, or that it could be typescript, etc.

snitin315:
Thanks for the PR. I agree with https://github.com/webpack/webpack.js.org/pull/5796#issuecomment-999637044, I believe the current docs are fine.

cmdcolin:
i'm surprised that you don't think this small change is not helpful. similarly, mentioning that you can import e.g. typescript files, or extensionless files which works with just plain node.js type module resolution, that would make clearer to me.

PR #5819, opened 2021-12-29
Adds the missing 'run' command to server script.

Cambuchi:
Adds the "run" text to the CLI command. Without this you get an output of "unknown command: start"

### DIFF
--- a/src/content/guides/development.mdx
+++ b/src/content/guides/development.mdx
@@ -265,7 +265,7 @@ Let's add a script to easily run the dev server as well:
  }
 ```
 
-Now we can run `npm start` from the command line and we will see our browser automatically loading up our page. If you now change any of the source files and save them, the web server will automatically reload after the code has been compiled. Give it a try!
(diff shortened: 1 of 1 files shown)


chenxsan:
Could you set up a reproducible repo, maybe on https://stackblitz.com/? As far as I know, `npm start` should just work as you can see from https://github.com/webpack/webpack.js.org/blob/master/package.json#L31.

Cambuchi:
Using the config from the tutorial, with my script named "server". On a blank node.js stackblitz.
You can see `npm server` does not work but `npm run server` does.

image

chenxsan:
> 

You are right, `npm server` won't work. But `start` is an exception, you can just run `npm start` or `yarn start` without that `run`. See https://stackoverflow.com/questions/51358235/difference-between-npm-start-and-npm-run-start/51358329 for some answers.

Cambuchi:
Ah, I didn't realize that there were special script keywords, I thought the behavior was supposed to be like "build" in the earlier script example. Thanks for the information. Consider this closed.

PR #5856, opened 2022-01-11
Add `webpack-typescript-directory-compile-plugin` plugin

ssigwart:
Adds `webpack-typescript-directory-compile-plugin` plugin to plugin list.


snitin315 (approved):
Thanks!

PR #5862, opened 2022-01-12
Fix path to correct header

piecioshka:
Hi folks! 👋 

I fix a anchor for `CLI Environment Variables` so used `cli-environment-variables`.


snitin315:
Hi @piecioshka, Thanks for the PR. Can you please accept the CLA?

PR #5846, opened 2022-01-06
update link to medium article

saranshkataria:
the hackernoon website version cannot be found by searching, so changed it to the medium article by the person

- [x] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [x] Make sure your PR complies with the [writer's guide][2].
- [x] Review the diff carefully as sometimes this can reveal issues.
- [x] Do not abandon your Pull Request: [Stale Pull Requests][3].
- **Remove these instructions from your PR as they are for your eyes only.**

[1]: https://github.com/openjs-foundation/EasyCLA#openjs-foundation-cla
[2]: https://webpack.js.org/contribute/writers-guide/
[3]: https://webpack.js.org/contribute/#pull-requests

PR #5876, opened 2022-01-16
docs(configuration): add new `commonjs-static` output type

### LINKED ISSUE #5855: Document webpack change: Enhancement: Improve CJS → ESM support
*A pull request by @JakobJingleheimer was merged and maintainers requested a documentation change.*

See pull request: https://github.com/webpack/webpack/pull/14998

---

In order for Node.js to detect named exports in CJS, they must be statically available. However, without this change, webpack does not produce exports that way—webpack will respect source code that is originally written in such a way, but it won't itself produce it (ex if the source code is ESM).

The outputted code will now look like:

```js
// …

function test() {} // user code

module.exports.test = test // this is the change
```

And when imported into an ESM file:

```js
import * as cjs from './output.cjs';
import { test } from './output.cjs';

console.log(cjs);
/*
[Module: null prototype] {
  __esModule: true,
  default: { test: [Function: test] },
  test: [Function: test]
}
*/

console.log(test)l
/*
[Function: test]
*/
```

Enhancement

Will do once approach etc are confirmed.

No: It requires the user to opt-in via configuration flag. Also, Node.js still automatically adds the named exports to `default` anyway (in addition to exposing them as named exports).

New `output.library.type` option `commonjs-static` which is similar to `commonjs-module` but emits code in a statically analyse-able way, so Node.js is able to construct named exports in it's commonjs esm compat layer.
Exports must be statically known to webpack for this to work. `optimization.providedExports` must not be disabled.

JakobJingleheimer:
Resolves #5855

### DIFF
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -1148,6 +1148,8 @@ T> Wondering the difference between CommonJS and CommonJS2 is? While they are si
 
 ##### type: 'commonjs-static'
 
+<Badge text='5.66.0+' />
+
 ```js
 module.exports = {
   // …
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -27,6 +27,7 @@ contributors:
   - QC-L
   - anshumanv
   - mrzalyaul
+  - JakobJingleheimer
 ---
 
 The top-level `output` key contains a set of options instructing webpack on how and where it should output your bundles, assets, and anything else you bundle or load with webpack.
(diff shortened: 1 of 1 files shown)


src/content/configuration/output.mdx
```diff
@@ -1145,6 +1146,52 @@ If we specify `output.library.name` with `type: commmonjs2`, the return value of
...
+const { doSomething } = require('./output.cjs'); // doSomething => [Function: doSomething]
+```
+
+Consumption (ESM):
+
+```javascript
+import { doSomething } from './output.cjs'; // doSomething => [Function: doSomething]
+```
+
+T> This is useful when source code is written in ESM and the output should be compatible with both CJS and ESM. For further details, please [read this issue](https://github.com/webpack/webpack/issues/14998) or [this article](https://dev.to/jakobjingleheimer/configuring-commonjs-es-modules-for-nodejs-12ed) (specifically, [this section](https://dev.to/jakobjingleheimer/configuring-commonjs-es-modules-for-nodejs-12ed#publish-only-a-cjs-distribution-with-property-exports)).
```
JakobJingleheimer, line 1193: The original issue doesn't provide full insight into why and how this option might be useful, so I included a link to an article I wrote. I'm not sure if that is appropriate (I saw in a recently merged PR here that you do link to external articles, such as #5846). I shall remove if not appropriate.

JakobJingleheimer:
@snitin315 @sokra

src/content/configuration/output.mdx
snitin315: It's good to have a reference here.

src/content/configuration/output.mdx
```diff
@@ -1145,6 +1146,52 @@ If we specify `output.library.name` with `type: commmonjs2`, the return value of
 
 T> Wondering the difference between CommonJS and CommonJS2 is? While they are similar, there are some subtle differences between them that are not usually relevant in the context of webpack. (For further details, please [read this issue](https://github.com/webpack/webpack/issues/1114).)
 
+##### type: 'commonjs-static'
```
snitin315, line 1149: We also mention versions in which this feature will be available via a badge.

```suggestion
##### type: 'commonjs-static' <Badge text='5.66.0+' />
```

chenxsan: Avoid it in heading. We can insert it right after the heading.

JakobJingleheimer: done

chenxsan (approved):
Looks good. Thanks.

PR #5901, opened 2022-01-21
fix: update WebStorm "safe write" to "Back up files before saving"

q0rban:
See https://www.jetbrains.com/help/webstorm/system-settings.html

> **Back up files before saving**: Before saving the file, create a backup. If the save operation is successful, WebStorm deletes the backup. If not, it restores the contents of the original file from backup. This behavior is known as "safe write", and it prevents losing your file in case of a faulty save operation.


chenxsan (approved):
Thanks for your contribution.

PR #5919, opened 2022-01-31
added yarn to installation way

behnammodi:
_describe your changes..._

### DIFF
--- a/src/content/concepts/index.mdx
+++ b/src/content/concepts/index.mdx
@@ -132,7 +132,7 @@ In order to use a plugin, you need to `require()` it and add it to the `plugins`
 **webpack.config.js**
 
 ```javascript
-const HtmlWebpackPlugin = require('html-webpack-plugin'); //installed via npm
+const HtmlWebpackPlugin = require('html-webpack-plugin');
 const webpack = require('webpack'); //to access built-in plugins
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


chenxsan:
Thank you but I think it's not necessary. We have so many package managers like pnpm, etc. other than npm/yarn. Not gonna list them all.

chenxsan:
> Thank you, but I think we can have something like this:
> 
> ```js
> const HtmlWebpackPlugin = require('html-webpack-plugin'); //installed via npm or any package managers
> ```
> 
> Why? because at the first time I read the document I was thinking we have to Install that just by `npm`

Or maybe we can just remove the comment to avoid misunderstanding.

behnammodi:
> > Thank you, but I think we can have something like this:
> > ```js
> > const HtmlWebpackPlugin = require('html-webpack-plugin'); //installed via npm or any package managers
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Why? because at the first time I read the document I was thinking we have to Install that just by `npm`
> 
> Or maybe we can just remove the comment to avoid misunderstanding.

Good idea, I changed

snitin315 (approved):
Thank you.

chenxsan (approved):

PR #5957, opened 2022-02-10
fix svg color for dark mode

chenxsan:
Should be white:

CleanShot 2022-02-10 at 20 00 47@2x

CleanShot 2022-02-10 at 20 01 19@2x


snitin315 (approved):
LGTM

PR #5959, opened 2022-02-11
use parcel css

chenxsan:
It's faster, and the build is smaller.


rishabh3112 (dismissed):

webpack.ssg.mjs
```diff
@@ -44,7 +44,11 @@ export default (env) =>
       libraryTarget: 'umd',
     },
     optimization: {
-      minimizer: [new OptimizeCSSAssetsPlugin({})],
+      minimizer: [
+        new OptimizeCSSAssetsPlugin({
+          minify: OptimizeCSSAssetsPlugin.parcelCssMinify,
+        }),
+      ],
```
snitin315, line 51: [From the `OptimizeCSSAssetsPlugin` docs](https://github.com/NMFR/optimize-css-assets-webpack-plugin#optimize-css-assets-webpack-plugin)

> ⚠️ For webpack v5 or above please use [css-minimizer-webpack-plugin](https://github.com/webpack-contrib/css-minimizer-webpack-plugin) instead.


## 

[And the `css-minimizer-webpack-plugin` recently added `@parcel/css` support ](https://github.com/webpack-contrib/css-minimizer-webpack-plugin#using-custom-minifier-parcelcss)


```js
module.exports = {
  optimization: {
    minimize: true,
    minimizer: [
      new CssMinimizerPlugin({
        minify: CssMinimizerPlugin.parcelCssMinify,
      }),
    ],
  },
};
```

chenxsan: Actually we were already using [`css-minimizer-webpack-plugin`](https://github.com/webpack/webpack.js.org/blob/7d0872295ac1c0d797f5928ef458306f1ae71325/webpack.prod.mjs#L3) if you check the code, just old naming not removed. Doesn't mean we're still using `OptimizeCSSAssetsPlugin`. Let me update the naming to avoid confusion.

snitin315: Perfect, thanks!

snitin315 (approved):

PR #6014, opened 2022-03-04
Fixed a mistake in the package.json diff

tomjdv:
The test script was removed and then added again in the diff, I think it was just meant to be replaced by the build command. As you don't _have_ to remove it in order to add a new script, I just left it in to avoid confusion.

### DIFF
--- a/src/content/guides/getting-started.mdx
+++ b/src/content/guides/getting-started.mdx
@@ -26,6 +26,7 @@ contributors:
   - anshumanv
   - d3lm
   - snitin315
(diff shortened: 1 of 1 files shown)


snitin315 (changes requested):
The diff is correct, please notice the comma (`,`) at the end of the `test` script.

tomjdv:
> The diff is correct, please notice the comma (`,`) at the end of the `test` script.

Ah, I'm sorry, I hadn't noticed that. My bad!

PR #6049, opened 2022-03-20
add disclaimer for third-party packages

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6030

CleanShot 2022-03-20 at 18 40 30@2x

### DIFF
--- a/src/components/Page/Page.jsx
+++ b/src/components/Page/Page.jsx
@@ -91,7 +91,7 @@ export default function Page(props) {
   return (
     <section className="page">
       <Markdown>
-        <h1>{title}</h1>
+        <h1 data-thirdParty={rest.thirdParty}>{title}</h1>
 
         {contentRender}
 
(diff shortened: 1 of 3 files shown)


src/components/Page/Page.jsx
```diff
@@ -93,6 +93,15 @@ export default function Page(props) {
...
         <h1>{title}</h1>
 
+        {rest.thirdParty ? (
+          <div className="italic my-[20px]">
+            <strong className="font-bold">Disclaimer:</strong> {title} is a
+            third-party package maintained by community members, it potentially
+            does not have the same support, security policy or license as
+            webpack, and it is not maintained by webpack.
+          </div>
+        ) : null}
```
snitin315, line 103: Not a required change but just for consistency with other sections.

```suggestion
        {rest.thirdParty && (
          <div className="italic my-[20px]">
            <strong className="font-bold">Disclaimer:</strong> {title} is a
            third-party package maintained by community members, it potentially
            does not have the same support, security policy or license as
            webpack, and it is not maintained by webpack.
          </div>
        )}
```

chenxsan: Not necessary.

snitin315 (approved):

PR #6098, opened 2022-04-15
docs(development): Mention `runtimeChunk: single` for multi-endpoint dev server

f3ndot:
This section of the guide introduces the concept of a dev server for the first time, but has not introduced the concept that multiple endpoints on a single html file served by the webserver is problematic without `optimization.runtimeChunk: 'single'`.

This commit attempts to give new developers a softer landing, and references the Code Splitting section to more deeply understand as well as link to the same pitfall external resource as the Code Splitting section does.

See https://github.com/webpack/webpack/discussions/14873#discussioncomment-2449779 for rationale

### DIFF
--- a/src/content/guides/development.mdx
+++ b/src/content/guides/development.mdx
@@ -225,10 +225,15 @@ Change your configuration file to tell the dev server where to look for files:
      path: path.resolve(__dirname, 'dist'),
      clean: true,
    },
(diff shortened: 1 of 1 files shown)


f3ndot:
Couldn't decide if this should be another Tip blockquote, or just a regular paragraph

PR #6107, opened 2022-04-25
docs(getting started) : Remove unnecessary diffs

k4a-dev:
There is no diff on the line marked as diff.
So remove the diff and change it to a normal line.

### DIFF
--- a/src/content/guides/getting-started.mdx
+++ b/src/content/guides/getting-started.mdx
@@ -287,8 +287,7 @@ Given it's not particularly fun to run a local copy of webpack from the CLI, we 
    "description": "",
    "private": true,
    "scripts": {
--    "test": "echo \"Error: no test specified\" && exit 1"
-+    "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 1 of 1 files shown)


src/content/guides/getting-started.mdx
```diff
@@ -287,8 +287,7 @@ Given it's not particularly fun to run a local copy of webpack from the CLI, we
    "description": "",
    "private": true,
    "scripts": {
--    "test": "echo \"Error: no test specified\" && exit 1"
-+    "test": "echo \"Error: no test specified\" && exit 1",
+     "test": "echo \"Error: no test specified\" && exit 1"
```
chenxsan, line 290: There's a `,` in the end of `"test"` line which can't be ignored.

chenxsan:
There's no need to make those changes, the original one is right if you add that `,`.

k4a-dev:
> There's no changes to the code now if you add that `,`, the original one is actually nothing wrong.

Oh, that's right. I am so sorry.
Close this Pull Request.

chenxsan:
> 

No worries, thanks anyway.

PR #6125, opened 2022-04-30
fix scroll bug

chenxsan:
Say you are on page https://webpack.js.org/configuration/module/#modulegenerator, then you want to go another page's section, e.g.,

Clicking the side bar would load the new page, but it won't scroll to the section.

Not sure since when it's broken.


chenxsan:
I think I've found the culprit https://github.com/webpack/webpack.js.org/pull/5809. The preview site works just fine before this pull request.

snitin315:
Interesting. Not sure how dev-server is related here since it is a dev dependency.

chenxsan:
This is more like an issue existing for a long time according to https://github.com/remix-run/react-router/issues/394, obviously we were not handling the dynamic loaded part.

PR #6163, opened 2022-05-21
fix(example/html): make HTML tags into the correct format

dev-itsheng:
When I jump to [StackBlitz](https://stackblitz.com/github/webpack/webpack.js.org/tree/master/examples/module-federation?file=app2%2Fpublic%2Findex.html&terminal=start&terminal=) in the [Module Federation chapter](https://webpack.js.org/concepts/module-federation/) of the Webpack documentation , I realized that its HTML template file had some problems:

image

According to the HTML specification, `div` tags should not be self-closing.

Although on modern browsers it correctly identifies the offending HTML tag (since it happens to be the only tag in the HTML body) and thus happens to render the correct result. But I think it should keep the correct format anyway, so I made a change.

### DIFF
--- a/examples/module-federation/app1/public/index.html
+++ b/examples/module-federation/app1/public/index.html
@@ -5,6 +5,6 @@
         <title>App 1</title>
     </head>
     <body>
-        <div id="root"/>
+        <div id="root"></div>
     </body>
 </html>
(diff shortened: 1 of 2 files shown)


snitin315 (approved):
LGTM. Thank you.

PR #6191, opened 2022-06-02
Fix typo

oscar-b:
Fix simple typo in the docs.

### DIFF
--- a/src/content/plugins/module-federation-plugin.mdx
+++ b/src/content/plugins/module-federation-plugin.mdx
@@ -34,7 +34,7 @@ module.exports = {
 
 Create a new runtime chunk with the specified name.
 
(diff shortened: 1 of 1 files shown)


src/content/plugins/module-federation-plugin.mdx
```diff
@@ -34,7 +34,7 @@ module.exports = {
 
 Create a new runtime chunk with the specified name.
 
-**webpack.js.org**
+**webpack.config.json**
```
snitin315, line 37:
```suggestion
**webpack.config.js**
```

oscar-b: Trying to do 100 things at the same time ;)

PR #6196, opened 2022-06-03
upgrade react 18

chenxsan:


chenxsan:
Seems this site is too old to upgrade to react 18.

PR #6240, opened 2022-06-25
fix highlighting for current subnav

chenxsan:
CleanShot 2022-06-25 at 11 45 53@2x

The subnav item should be highlighted when we are visiting its child sections.


snitin315 (approved):
LGTM

PR #6269, opened 2022-07-13
ci: try webpack cache

snitin315:
Enable webpack persistent cache on CI.

Comparision - https://github.com/webpack/webpack.js.org/pull/6269#issuecomment-1185138320

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,6 +23,16 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: yarn
 
+      - name: Enable webpack persistent caching
+        uses: actions/cache@v3
+        id: build-webpack-persistent-cache
+        with:
+          path: node_modules/.cache
+          key: build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |-
+             build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
+             build-${{ hashFiles('**/webpack.*.js') }}-
+
       - run: yarn --frozen-lockfile
       - name: Build site
         run: yarn build
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,10 +28,10 @@ jobs:
         id: build-webpack-persistent-cache
         with:
           path: node_modules/.cache
-          key: build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
+          key: build-${{ hashFiles('**/webpack.*.mjs') }}-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |-
-             build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
-             build-${{ hashFiles('**/webpack.*.js') }}-
+             build-${{ hashFiles('**/webpack.*.mjs') }}-${{ hashFiles('**/yarn.lock') }}
+             build-${{ hashFiles('**/webpack.*.mjs') }}-
 
       - run: yarn --frozen-lockfile
       - name: Build site
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -89,7 +89,7 @@ jobs:
         id: build-webpack-persistent-cache
         with:
           path: node_modules/.cache
-          key: build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
+          key: build-${{ hashFiles('**/webpack.*.mjs') }}-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |-
              build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
              build-${{ hashFiles('**/webpack.*.js') }}-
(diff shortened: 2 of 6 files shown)


.github/workflows/testing.yml
```diff
@@ -57,6 +57,16 @@ jobs:
           cache: yarn
 
       - run: yarn
+      
+      - name: Enable webpack persistent caching
+        uses: actions/cache@v3
+        id: build-webpack-persistent-cache
+        with:
+          path: node_modules/.cache/webpack
```
snitin315, line 65:
```suggestion
          path: node_modules/.cache
```

chenxsan:
Any specific benefit to enable it? I can see the time spent on actions similar to before.

snitin315:
I was also checking its benefits. It has actually reduced the webpack build time if you notice -

- `8420ms` [After Cache ](https://github.com/webpack/webpack.js.org/runs/7315338868?check_suite_focus=true#step:6:304)
- `92104ms` [Without Cache](https://github.com/webpack/webpack.js.org/runs/7346389520?check_suite_focus=true#step:5:494)

So it has reduced the time for the build site step.

## Before

Screenshot 2022-07-15 at 8 59 43 AM

## After

Screenshot 2022-07-15 at 9 00 01 AM

chenxsan:
> I was also checking its benefits. It has actually reduced the webpack build time if you notice -
> 
> * `8420ms` [After Cache ](https://github.com/webpack/webpack.js.org/runs/7315338868?check_suite_focus=true#step:6:304)
> * `92104ms` [Without Cache](https://github.com/webpack/webpack.js.org/runs/7346389520?check_suite_focus=true#step:5:494)
> 
> So it has reduced the time for the build site step.
> 
> ## Before
> Screenshot 2022-07-15 at 8 59 43 AM
> 
> ## After
> Screenshot 2022-07-15 at 9 00 01 AM

Doesn't seem consistent, e.g., https://github.com/webpack/webpack.js.org/runs/7260630314?check_suite_focus=true without cache:

CleanShot 2022-07-15 at 11 51 03@2x

While the latest action in this pull request with cache enabled:

CleanShot 2022-07-15 at 11 52 01@2x

snitin315:
Yeah, this inconsistency this actually because of the other scripts we run before building (`src/utilities/fetch-package-repos.mjs`). If you notice the webpack build time only, it is reduced.

Screenshot 2022-07-15 at 4 00 43 PM
Screenshot 2022-07-15 at 4 00 56 PM

snitin315:
@chenxsan Can we merge this?

package.json
```diff
@@ -56,7 +56,8 @@
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
     "prettier": "prettier --write '**/*.{js,json,jsx,css,scss,md,mdx}'",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "rimraf ./node_modules/.cache/webpack"
```
chenxsan, line 60: Any specific reason to clear the cache here?

.github/workflows/testing.yml
```diff
@@ -74,6 +84,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Enable webpack persistent caching
+        uses: actions/cache@v3
+        id: build-webpack-persistent-cache
+        with:
+          path: node_modules/.cache
+          key: build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
```
chenxsan, line 92: I just noticed we don't have any `webpack.*.js` files. All we have are `webpack.*.mjs` files, does it matter?

package.json
snitin315: It's [officially recommended](https://webpack.js.org/guides/build-performance/#persistent-cache).

.github/workflows/testing.yml
snitin315: Yes, it should be `.mjs` otherwise key won't change when the webpack config is changed.
```suggestion
          key: build-${{ hashFiles('**/webpack.*.mjs') }}-${{ hashFiles('**/yarn.lock') }}
```

chenxsan: I think there're many other `*.js`.

snitin315: In this repo?

chenxsan: This pull request, e.g.:

CleanShot 2022-07-19 at 19 28 55@2x

snitin315: Oh yes. Sorry! I got confused.

.github/workflows/deploy.yml
```diff
@@ -23,6 +23,16 @@ jobs:
...
 
+      - name: Enable webpack persistent caching
+        uses: actions/cache@v3
+        id: build-webpack-persistent-cache
+        with:
+          path: node_modules/.cache
+          key: build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |-
+             build-${{ hashFiles('**/webpack.*.js') }}-${{ hashFiles('**/yarn.lock') }}
+             build-${{ hashFiles('**/webpack.*.js') }}-
```
chenxsan, line 34: What about those `*.js`?

snitin315: This slipped my attention.

chenxsan:
Interesting, it's no longer faster than before now.

Both are from this pull request's action logs:

CleanShot 2022-07-20 at 19 57 23@2x

CleanShot 2022-07-20 at 20 00 15@2x

PR #6278, opened 2022-07-15
docs(configuration): add `resolve.extensionAlias`

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/6254


chenxsan (commented):
Might move it above `resolve.extensions` since we are manually sort the headings alphabetically.

chenxsan (approved):

PR #6279, opened 2022-07-16
docs(guides): refine code-splitting code snippets

Matin297:
remove extra call to createElement method in Dynamic Imports section of the code splitting guide.


chenxsan:
I think there's another line needed to be removed right below:

CleanShot 2022-07-17 at 21 02 11@2x

Matin297:
> I think there's another line needed to be removed right below:
> 
> CleanShot 2022-07-17 at 21 02 11@2x

Good catch! I believe due to the changes I made, this line needs a "+" sign right before it, since the second call has a "-" in this code snippet.

chenxsan (approved):
Thanks.

Issue #6286, opened 2022-07-18
configuration.stats has an unknown property 'maxModules'


AmalJossy:
**Feature to document**

In the [Debugging Optimization Bailouts](https://webpack.js.org/plugins/module-concatenation-plugin/#debugging-optimization-bailouts) doc, the snippet provided is
```
module.exports = {
  //...
  stats: {
    // Examine all modules
    maxModules: Infinity,
    // Display bailout reasons
    optimizationBailout: true,
  },
};
```

However, `maxModules` is not a valid parameter anymore, webpack will fail with the following message if the parameter is present in config.
`Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration.stats has an unknown property 'maxModules'`
 
 An explanation of how bailout reasons for all modules can be displayed can also be helpful here

**Author(s)**

@skipjack @TheLarkInn @byzyk 

**Additional information**

N/A

[ ] I am willing to work on this issue and submit a pull request.

PR #6292, opened 2022-07-22
docs(plugins): fix stale content

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6286


snitin315 (approved):

AmalJossy:
@chenxsan I see the reference to `maxModules` has been removed but the correct way to get "Examine all modules" should still be mentioned IMO. Cant we use `modulesSpace` ?

PR #6316, opened 2022-08-05
Explain the externals string syntax specifying externalsType

### LINKED ISSUE #4148: Misleading documentation on externals type or potential bug
The [documentation](https://webpack.js.org/configuration/externals/#string) makes us believe that this example:
```
module.exports = {
  //...
  externals: {
    jquery: 'jQuery'
  }
};
```

will result in `jquery` imports to be replaced with `jQuery` variable:
```
In other words, when a string is provided it will be treated as root (defined above and below).
```
```
root: The library should be available as a global variable (e.g. via a script tag).
```

However it's not exactly true.
The default type of externals is defined by `externalsType` option. Which is by default defined as follows:

https://github.com/webpack/webpack/blob/master/lib/config/defaults.js#L189-L198

So for example in case of commonjs library target `import $ from 'jquery';` will be transformed to `require("jQuery")`.

### LINKED ISSUE #5928: Documentation contains no example for externals: string
https://webpack.js.org/configuration/externals/#string is supposed to document type `string` of externals, but it does not.  It provides examples for type object.

There shall be an example, stating that
```json
{
  "externals": "A"
}
```
and
```json
{
  "externals": {"A": "A"}
}
```
do both the same.

tanhauhau:
Explain the externals string syntax specifying externalsType

Fixes #4148
Fixes #5928

### DIFF
--- a/src/content/configuration/externals.mdx
+++ b/src/content/configuration/externals.mdx
@@ -404,7 +404,7 @@ Note that there will be a `require()` in the output bundle.
 
 ### externalsType.global
 
-Specify the default type of externals as `'global'`. Webpack will read the external as a global variable on the `global` object.
+Specify the default type of externals as `'global'`. Webpack will read the external as a global variable on the [`globalObject`](/configuration/output/#outputglobalobject).
 
 #### Example
 
@@ -421,6 +421,9 @@ module.exports = {
   externalsType: 'global',
   externals: {
     jquery: '$',
+  },
+  output: {
+    globalObject: 'global',
   },
 };
 ```
--- a/src/content/configuration/externals.mdx
+++ b/src/content/configuration/externals.mdx
@@ -59,13 +59,9 @@ import $ from 'jquery';
 $('.my-element').animate(/* ... */);
 ```
 
-The property name `jquery` indicates that the module `jquery` in `import $ from 'jquery'` should be excluded. In order to replace this module, the value `jQuery` will be used to retrieve a global `jQuery` variable, as the default external library type is `var`, see [externalsType](#externalstype).
+The property name `jquery` specified under `externals` in the above `webpack.config.js` indicates that the module `jquery` in `import $ from 'jquery'` should be excluded from bundling. In order to replace this module, the value `jQuery` will be used to retrieve a global `jQuery` variable, as the default external library type is `var`, see [externalsType](#externalstype).
 
-The bundle with external dependencies can be used in various module contexts, such as [CommonJS, AMD, global and ES2015 modules](/concepts/modules).
-
-The external library may be available in any of these forms: global variable, CommonJS, AMD, ES2015 Module, see more in [externalsType](#externalstype).
-
-The following syntaxes are accepted...
+While we showed an example consuming external global variable above, the external can actually be available in any of these forms: global variable, CommonJS, AMD, ES2015 Module, see more in [externalsType](#externalstype).
 
 ### string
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/externals.mdx
```diff
@@ -370,13 +457,36 @@ module.exports = {
...
 Specify the default type of externals as `'node-commonjs'`. Webpack will import [`createRequire`](https://nodejs.org/api/module.html#module_module_createrequire_filename) from `'module'` to construct a require function for loading externals used in a module.
 
+#### Example
+
+```javascript
+import jq from 'jquery';
+jq('.my-element').animate(/* ... */);
+```
+
+**webpack.config.js**
```
chenxsan, line 488: I think the `externals: 'jquery'` is missing in this configuration.

src/content/configuration/externals.mdx
```diff
@@ -386,9 +496,80 @@ module.export = {
...
+Will generate into something like
+
+```javascript
+import { createRequire } from 'module';
+
+const jq = createRequire(import.meta.url)('$');
+jq('.my-element').animate(/* ... */);
+```
+
+Note that there will be a `import` statement in the output bundle.
```
chenxsan, line 508:
```suggestion
Note that there will be an `import` statement in the output bundle.
```

src/content/configuration/externals.mdx
```diff
@@ -59,42 +59,49 @@ import $ from 'jquery';
...
+You can specify the [external library type](#externalstype) to the external using the `${externalsType} ${path}` syntax. This will override the default external library type specified in the [externalsType](#externalstype) option.
+
+For example, if the external library is a [CommonJS module](#externalstypecommonjs), you can specify
 
 ```javascript
-const fs = require('fs-extra');
+module.exports = {
+  //...
+  externals: {
+    jquery: 'commonjs jQuery',
```
chenxsan, line 102:
```suggestion
    jquery: 'commonjs jquery',
```
Should it be `jquery`?

src/content/configuration/externals.mdx
```diff
@@ -370,13 +457,36 @@ module.exports = {
     outputModule: true,
   },
   externalsType: 'module',
+  externals: {
+    jquery: '$',
+  },
```
chenxsan, line 462: Maybe just use `externals: 'jquery'` in those cases? I can understand such usage under the `externalsType.global` section, but here could introduce confusion to users because we'll `import` from `$` now which doesn't really exist.

chenxsan (commented):
Thank u for those examples! I've made some feedback.

src/content/configuration/externals.mdx
```diff
@@ -329,23 +345,23 @@ Supported types:
 
 - `'amd'`
 - `'amd-require'`
-- `'assign'`
-- `'commonjs'`
+- `'assign'` - same as `'var'`
+- [`'commonjs'`](#externalstypecommonjs)
 - `'commonjs-module'`
-- `'global'`
+- [`'global'`](#externalstypeglobal)
```
chenxsan, line 351: This `global` works same as `self` in my test, could u double check it?

src/content/configuration/externals.mdx
```diff
@@ -370,25 +453,122 @@ module.exports = {
...
+  externals: {
+    jquery: 'jquery',
+  },
 };
 ```
 
+Will generate into something like
+
+```javascript
+import * as __WEBPACK_EXTERNAL_MODULE_$_ from 'jquery';
```
chenxsan, line 465:
```suggestion
import * as __WEBPACK_EXTERNAL_MODULE_jquery__ from 'jquery';
```

src/content/configuration/externals.mdx
```diff
@@ -370,25 +453,122 @@ module.exports = {
...
+
+const jq = __WEBPACK_EXTERNAL_MODULE_$_['default'];
```
chenxsan, line 467:
```suggestion
const jq = __WEBPACK_EXTERNAL_MODULE_jquery__['default'];
```

src/content/configuration/externals.mdx
tanhauhau: oh it actually depends on [`output.globalObject`](https://webpack.js.org/configuration/output/#outputglobalobject)

tanhauhau: i've updated the relevant change

chenxsan (approved):

PR #6329, opened 2022-08-12
Autoprefixer-core was deprecated

maazakn:
According to autoprefixer-core: https://www.npmjs.com/package/autoprefixer-core

Autoprefixer-core was deprecated, use autoprefixer instead
Packages autoprefixer and autoprefixer-core was merged in 6.0 release. So please replace autoprefixer-core to autoprefixer in your package.json and code.

### DIFF
--- a/src/content/migrate/3.mdx
+++ b/src/content/migrate/3.mdx
@@ -304,6 +304,26 @@ plugins: [
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


chenxsan:
Thanks but I don't think it belongs to webpack migration.

PR #6330, opened 2022-08-14
explain more on the condition names

tanhauhau:


src/content/configuration/resolve.mdx
```diff
@@ -193,6 +193,52 @@ module.exports = {
 };
 ```
 
+Webpack will match export conditions that is listed within the `resolve.conditionNames` array.
+
+The key order in the `exports` field is significant. During condition matching, earlier entries have higher priority and take precedence over later entries.
```
chenxsan, line 198: Do you mean `conditionNames` field here? `exports` is an object which I believe has no order in JavaScript.

tanhauhau: i thought so too in the beginning, but no, i meant the `exports` object.

see: https://nodejs.org/api/packages.html#conditional-exports

chenxsan (approved):

PR #6337, opened 2022-08-16
docs(loaders): add link to esbuild-loader

MasterOdin:
PR adds `esbuild-loader` as a possible option for a transpiling loader as it's gaining a lot of popularity vs using babel-loader.


chenxsan (approved):
Thanks.

PR #6352, opened 2022-08-23
fix(docs): correct typo in the setupMiddlewares example

TPXP:
The injected middleware is the first in the array, it makes more sense to call it like that. Also, I'm not sure we should condone violence through fists against JS arrays.


chenxsan (approved):
Looks good, thanks.

PR #6353, opened 2022-08-24
fix(docs): syntax fix in "Inline matchResource"

vinicius-caputo:
syntax fix in Inline matchResource example:
"const match = STYLES_REGEXP.match(source);"  is in the wrong order -> "const match = source.match(STYLES_REGEXP);"
"require.resolve('./getStyle');" is missing the "s" to match with the file getStyles.js


chenxsan (approved):
Looks good, thanks for the contribution.

Issue #6356, opened 2022-08-26
Proposal: Update writer's guide for "Configuration defaults and types" to support listing different defaults for production/development


caseyjhol:
The current writer's guide suggests writing defaults as follows:

> **configuration.example.option**
> 
>  `string = 'none': 'none' | 'development' | 'production'`
> 
> Where `= 'none'` means that the default value is `'none'` for the given option.

However, many options have 3 possible defaults depending on whether `mode` is set to none, production, or development.

Proposal: use a "defaults" table instead.

## optimization.moduleIds

`boolean: false` `string: 'natural' | 'named' | 'deterministic' | 'size'`

| Mode          | Default         |
| ------------- | --------------- |
| `'production'`  | `'deterministic'` |
| `'development'` | `'named'`         |
| `'none'`        | `'natural'`       |

Example:
![image](https://user-images.githubusercontent.com/2874325/186944311-d6397cb4-4347-4707-9334-da081665c536.png)

If an option has the same default for all modes, we could simplify it like so:

## optimization.mangleWasmImports

`boolean`

| Default |
| ------- |
| `false` |

Example:
![image](https://user-images.githubusercontent.com/2874325/186945293-0b99a69f-d3da-46e0-b24b-b443b3407ef9.png)

---
This makes it easy to see at a glance what the defaults are for every option.

Related:
#2080, #2302

TusharThakur04:
@alexander-akait i would like to send PRs

alexander-akait:
Please no need to ping me every time, you can pick any PR, and also no need to ping me in PRs too, I see your PRs in unread notifications and I will look at them when will have time to do it, thanks

iblawal:
Hi, I'm a GSoC 2026 applicant interested in the New Documentation Website project. I'd like to work on this issue - I'll update the defaults formatting to use tables as proposed. I'll submit a PR shortly.

abhaymauryadev:
Hi, I’m interested in working on this issue. I'm also Gsoc 2026 applicant, I saw previous discussions— is this still open for contribution, or should I pick another issue?

iblawal:
Hi, thanks for reaching out! Actually, I believe this specific update has already been addressed and merged in my previous PR (referencing [Link your PR number here]). You might want to check the latest version of the Writer's Guide or look for other open issues labeled 'good first issue' or 'documentation' to avoid duplicating work. Good luck with your application

abhaymauryadev:
[...]

PR #6367, opened 2022-08-31
Use table for default values

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6356

### DIFF
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -34,7 +34,11 @@ The top-level `output` key contains a set of options instructing webpack on how 
 
 ## output.assetModuleFilename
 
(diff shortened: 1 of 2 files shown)


caseyjhol:
Is it the goal of this PR to update all the defaults in the docs as well (i.e. using values from https://github.com/webpack/webpack/blob/main/lib/config/defaults.js)?

chenxsan:
> 



> Is it the goal of this PR to update all the defaults in the docs as well (i.e. using values from https://github.com/webpack/webpack/blob/main/lib/config/defaults.js)?

Yes, will likely update all the defaults under the `Configuration` section unless the case doesn't fit well with a table.

PR #6437, opened 2022-10-09
CI: cancel previous runs for pull requests

chenxsan:


.github/workflows/cancel.yml
```diff
@@ -0,0 +1,13 @@
+name: Cancel
+on:
+  workflow_run:
+    workflows: ["Ensure PR"]
+    types:
+      - requested
```
snitin315, line 6:
```suggestion
on:
  pull_request:
```

.github/workflows/cancel.yml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  cancel:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: styfle/cancel-workflow-action
+      with:
+        workflow_id: ${{ github.event.workflow.id }}
```
snitin315, line 13:
```suggestion
        workflow_id: testing.yml
        access_token: ${{ github.token }}
```

snitin315 (commented):
You can run on PR event

.github/workflows/cancel.yml
chenxsan: Nah, it won't run against those pull requests from forks https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks. E.g, I'm submitting this pull request from my forked repository.

snitin315: [Try this](https://github.com/webpack-contrib/css-loader/blob/master/.github/workflows/cancel.yml#L2), we have similar in webpack-contrib and it runs on PR form forks.

chenxsan: Thanks. Any reason to avoid `workflow_run`? Seems fine to me according to the documentation.

chenxsan: Maybe we can only test it after getting it merged, not in this one.

snitin315: I'm good with it if it works 👍

PR #6439, opened 2022-10-10
test workflow

chenxsan:


chenxsan:
@snitin315 FYI, it works fine from what I can tell. See https://github.com/webpack/webpack.js.org/actions/runs/3218817669/jobs/5263438132

PR #6441, opened 2022-10-10
docs(configuration): update configuration page

chenxsan:
1. the `js-with-links-details` code block is hard to maintain, actually we haven't touched it for a long time now, and some configurations inside are outdated, besides, it looks terrible on mobile. However, I can see the benefit of giving an overview of the whole webpack configuration, we might figure out another way to present it in future.
2. the `init` section is outdated as well, so I've updated it with a new one.


src/content/configuration/index.mdx
```diff
@@ -45,1128 +46,46 @@ If for some reason you want to use a different configuration file depending on c
 
 ## Options
```
snitin315, line 47: We may want to change this heading now since we don't show options now. Maybe something like `Scaffolding a webpack project`

src/content/configuration/index.mdx
```diff
@@ -43,1130 +44,48 @@ If for some reason you want to use a different configuration file depending on c
...
+## Set up a new webpack project
 
 W> webpack applies configuration defaults after [plugins defaults](/contribute/writing-a-plugin/#configuration-defaults) are applied.
 
-Use [webpack-cli's `init` command](/api/cli/#init) to rapidly generate webpack configuration files for your project requirements, it will ask you a couple of questions before creating a configuration file.
+Webpack has a huge set of options which might be overwhelming to you, please take advantage of [webpack-cli's `init` command](/api/cli/#init) which could rapidly generate webpack configuration files for your project requirements, it will ask you a couple of questions before creating a configuration file.
 
 ```bash
-npx webpack-cli init
+$ npx webpack init
```
harish-sethuraman, line 54:
```suggestion
npx webpack init
```
I feel like removing the $ would be better. It'll be a little hard to copy command?

chenxsan: I don't think it matters if you're going to select the texts manually. But I do consider introducing a copy button in near future for those code blocks.

snitin315 (approved):

src/content/configuration/index.mdx
harish-sethuraman: In case we consider to include copy button it is better to remove these $ prefixes. Considering no other documentation site have them ex: https://beta.reactjs.org/learn/start-a-new-react-project, eslint.org

chenxsan: It should be easy to remove `$` in event handler if we want to. But I do agree users might copy `$` without knowing its intention here. And in the case of multiple-line commands, `$` will cause more trouble. Thanks for your input.

IdkMan2:
I don't understand why have you removed documentation regarding an overview of all available options. The only way to do this now is to dig in the commits history in this repo. These options overview is pretty useful though.

chenxsan:
> I don't understand why have you removed documentation regarding an overview of all available options. The only way to do this now is to dig in the commits history in this repo. These options overview is pretty useful though.

Hi, the current implementation is hard to maintain or update causing stale content, if you know any better ways to maintain, please share with us so we can add it back.

IdkMan2:
You are still maintaining [typescript types declarations](https://github.com/webpack/webpack/blob/main/types.d.ts), perhaps you could reuse them to automatically generate an overview of these options based on the [`Configuration` interface](https://github.com/webpack/webpack/blob/c18203c89447cd6728aa1f9d77b03aff7ae6e03f/types.d.ts#L2077)?

There are already projects which offer similar functionalities, like [TypeDoc](https://typedoc.org/guides/overview/) or [DocDash](https://github.com/clenemt/docdash), I don't know if they would be helpful for you.

At least it is some idea that can be explored.

PR #6446, opened 2022-10-11
enable markdownlint for mdx

chenxsan:
So we can catch errors like https://github.com/webpack/webpack.js.org/pull/6441#discussion_r991824393

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -3,13 +3,17 @@
   "MD002": false,
   "MD004": { "style": "dash" },
   "MD007": { "indent": 2 },
-  "MD013": { "line_length": 600 },
+  "MD013": { "line_length": 600, "code_blocks": false },
+  "MD024": {
+    "siblings_only": true
+  },
   "MD026": false,
   "MD029": { "style": "ordered" },
   "MD033": false,
   "MD034": false,
   "MD036": false,
   "MD041": false,
+  "MD053": false,
   "no-hard-tabs": false,
   "whitespace": false
 }
--- a/.markdownlintignore
+++ b/.markdownlintignore
@@ -1,3 +1,1 @@
-.vale/**/*.md
-.github/**/*.md
 node_modules
--- a/.markdownlintignore
+++ b/.markdownlintignore
@@ -1,0 +1,3 @@
+.vale/**/*.md
+.github/**/*.md
+node_modules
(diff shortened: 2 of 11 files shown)


.markdownlint.json
```diff
@@ -3,13 +3,17 @@
...
+  "MD024": {
+    "siblings_only": true
+  },
   "MD026": false,
   "MD029": { "style": "ordered" },
   "MD033": false,
   "MD034": false,
   "MD036": false,
   "MD041": false,
+  "MD053": false,
```
chenxsan, line 16: This one is disabled for temporary because of https://github.com/DavidAnson/markdownlint/issues/571

snitin315 (dismissed):

.markdownlintignore
```diff
@@ -0,0 +1,3 @@
+.vale/**/*.md
+.github/**/*.md
```
harish-sethuraman, line 2:
```suggestion
.vale
.github
```
doesn't markdownlint lint only markdown files? we might not actually need *.md here? we can just add both directories alone?

chenxsan: yeah, it's sort of superfluous. Actually I notice there's no `.md` files under `.vale` now, and we should apply rules to `.github` directory as well.

package.json
```diff
@@ -44,8 +44,8 @@
     "lint": "run-s lint:*",
     "lint:js": "npm run lint-js .",
     "lint-js": "eslint --cache --cache-location .cache/.eslintcache",
-    "lint:markdown": "npm run lint-markdown *.md",
-    "lint-markdown": "markdownlint --config ./.markdownlint.json --ignore '.vale/**/*.md' --ignore '.github/**/*.md'",
+    "lint:markdown": "npm run lint-markdown '**/*.{md,mdx}'",
```
harish-sethuraman, line 47:
```suggestion
    "lint:markdown": "npm run lint-markdown '*.{md,mdx}'",
```

previously it was `*.md` and now it is `**/*` this will miss the readme.md in the main directory? and run only for sub directories? correct me if i'm wrong?

chenxsan: I believe the one before was wrong since `*.md` won't match markdown files recursively. `markdownlint-cli` uses `glob` for those patterns, as you can see from https://github.com/isaacs/node-glob/blob/main/README.md:

> ** If a "globstar" is alone in a path portion, then it matches zero or more directories and subdirectories searching for matches. It does not crawl symlinked directories.

So `readme.md` will be included as well.

harish-sethuraman: ✅ LGTM

harish-sethuraman (approved):

PR #6458, opened 2022-10-16
docs(Configuration): fix lang for code block

chenxsan:
There's no `console` language for code block https://github.com/wooorm/refractor#data

### DIFF
--- a/src/content/configuration/dev-server.mdx
+++ b/src/content/configuration/dev-server.mdx
@@ -1654,7 +1654,7 @@ module.exports = {
 
 Usage via CLI:
 
-```console
+```bash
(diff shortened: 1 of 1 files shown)


snitin315 (approved):
LGTM. Can we throw lint errors for such errors?

chenxsan:
> LGTM. Can we throw lint errors for such errors?

Really good idea. There's a rule https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md040 requiring us to specify a language for code block, unfortunately we can't specify which languages to be included. But I think it could be feasible with a custom rule. I'll try it later.

PR #6459, opened 2022-10-18
docs: clean up .js-with-links

chenxsan:
We have this `.js-with-links` from [remark-refractor](https://github.com/montogeek/remark-refractor) which would generates links inside code block.

Unfortunately it's totally broken at the moment:

I have no idea when and how we broke it, and I have no interest to fix it for two reasons:

1. it's hard to maintain.
2. Editors like VSCode won't apply syntax highlighting to such a code block, which means we are on our own when editing the code block.
    
    

Hence I would like to take this opportunity to clean up them.


src/content/api/node.mdx
```diff
@@ -42,14 +42,12 @@ import webpack from 'webpack';
...
+```js
 const webpack = require('webpack');
 
-webpack({
-  // [Configuration Object](/configuration/)
-}, (err, stats) => { // [Stats Object](#stats-object)
+webpack({}, (err, stats) => {
   if (err || stats.hasErrors()) {
-    // [Handle errors here](#error-handling)
+    //
```
harish-sethuraman, line 50:
```suggestion
```

chenxsan: It's a placeholder, guess it makes more sense to use `// ...`.

snitin315 (dismissed):
Makes sense to move away from this 👍🏻

Issue #6480, opened 2022-11-04
Unreadable `<code>` nested inside `link` in a `tip`


mukeshgurpude:
Page: https://webpack.js.org/concepts/plugins/
Theme: dark

![image](https://user-images.githubusercontent.com/55982424/199867577-9d791d9a-bbb8-4524-9f30-1b58af536311.png)

First link in the above image(webpack-source) is unreadable in dark mode. Somehow the styles have messed up the visibility of the element.

snitin315:
@chenxsan Could this be because of https://github.com/webpack/webpack.js.org/pull/6438? Maybe we need to improve Tip styles for the dark theme too.

chenxsan:
Filed a pull request here https://github.com/webpack/webpack.js.org/pull/6481, let me know what you think.

PR #6481, opened 2022-11-04
optimize blockquote styles for dark mode

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6480

CleanShot 2022-11-04 at 12 59 55@2x

CleanShot 2022-11-04 at 13 00 09@2x

### DIFF
--- a/src/styles/dark.scss
+++ b/src/styles/dark.scss
@@ -64,6 +64,34 @@ body {
       color: #82b7f6;
     }
   }
+
+  aside.tip,
+  aside.warning,
+  aside.preview,
+  aside.todo {
+    background-color: #222222;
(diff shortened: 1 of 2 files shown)


snitin315 (approved):
Looks great

src/styles/dark.scss
```diff
@@ -64,6 +64,34 @@ body {
       color: #82b7f6;
     }
   }
+
+  aside.tip,
```
harish-sethuraman, line 68: just a NIT. You could have actually extracted the `aside` separately and nested `&.tip`. `&.warning` that way we could have moved all aside code to one place.

PR #6487, opened 2022-11-05
Add globalObject when type is umd

tsq:
https://webpack.js.org/configuration/output/#outputglobalobject

When targeting a library, especially when libraryTarget is 'umd', this option indicates what global object will be used to mount the library. To make UMD build available on both browsers and Node.js, set `output.globalObject` option to 'this'. Defaults to self for Web-like targets.


chenxsan (approved):
Thanks.

PR #6495, opened 2022-11-08
fix naming in sidebar version select

Simek:
I have spotted that almost everywhere on the docs webpack is referred to as `webpack` (lowercase), when in the sidebar version select the capitalized name has been used, so I have corrected it for the naming consistency.


chenxsan (approved):
Thanks.

PR #6517, opened 2022-11-19
docs(api): webpack-cli v5 documentation updates

snitin315:
Refers [webpack-cli v5 release](https://github.com/webpack/webpack-cli/blob/master/CHANGELOG.md#500-2022-11-17)


src/content/api/cli.mdx
```diff
@@ -596,15 +600,15 @@ module.exports = (env, argv) => {
 };
 ```
 
-### node-env
+### define-process-env-node-env
```
chenxsan, line 603: I think it's better to keep both versions. Most users might not realize it's webpack-cli v5+ only, and it will take a long time for users to upgrade to webpack-cli v5.

E.g.,

```
### node-env

for `webpack-cli v4`

### define-process-env-node-env

for `webpack-cli v5+`
```

snitin315: Makes sense 👍

chenxsan (approved):

PR #6519, opened 2022-11-21
docs: add section on how to specify the package name for module federation

### LINKED ISSUE #5476: With module federation, submodules of shared modules are not actually shared if they are imported individually
# Bug report

Importing a submodule of a shared module yields a different instance of the submodule in each `ModuleFederationPlugin`, instead of a single shared instance.

I created a repository to showcase the issue: [module-federation-with-shared-submodules](https://github.com/antoinetissier/module-federation-with-shared-submodules).

It is based on the [shared context](https://github.com/module-federation/module-federation-examples/tree/master/shared-context) module federation example, with the difference that on the `main` branch the different module federation plugins import the shared context from the shared library through a submodule import rather than a root import:

```diff
- import { NameContext } from "@shared-context/shared-library";
+ import { NameContext } from "@shared-context/shared-library/src/context/NameContext";
```

The problem is that when doing this, the context is not actually shared, and each module federation plugin gets its own instance: `app1` provides "Billy" as context value above the `Welcome` component taken from `app2`, but `Welcome` uses a different context instance in its call to `useContext` - the one bundled with `app2`.
=> This leads to seeing "No name provided" in `app1` instead of "Billy".

Not being able to rely on submodules imports is a problem, because importing the whole shared library might make the bundle size much bigger in real life cases.
For example, importing just the `<Button />` component from [ant-design](https://github.com/ant-design/ant-design) instead of the whole library makes a huge difference:
```diff
- import { Button } from "antd";
+ import Button from "antd/lib/button";
```

I would expect the submodules to be shared between the different module federation plugins.

**Possible workaround**
Sharing the submodule itself in the webpack configs of the module federation plugins seems to work as expected:
```diff
      shared: [
        "react",
        "react-dom",
        {
-          "@shared-context/shared-library": {
-            import: "@shared-context/shared-library",
+          "@shared-context/shared-library/src/context/NameContext": {
+            import: "@shared-context/shared-library/src/context/NameContext",
            requiredVersion: require("../shared-library/package.json").version,
         },
        },
      ]
```

You can check that on the `shared-submodule` branch of the repo. After rebuilding, "Billy" is rendered as expected in `app1`:
```
git checkout shared-submodule
yarn clean-install && yarn clean && yarn build && yarn start
```

However, having to do this for every submodule of a shared library would be painful, and not scale well.
Please let me know if there is a better to way to approach this issue.

webpack version:  5.52.1
Node.js version: 14.17.1
Operating System: Windows 10

Christian24:
Hello,

I added a little section on how to specify the package name when using `shared` with module federation plugin. This should fix #5476.

### DIFF
--- a/src/content/plugins/module-federation-plugin.mdx
+++ b/src/content/plugins/module-federation-plugin.mdx
@@ -89,7 +89,7 @@ module.exports = {
 };
 ```
 
-The `/` syntax allows you to access all files of a package.
+The `/` syntax allows you to access all files of a package. However, it should be used only where necessary, because it has an impact on performance especially in `development` mode.
 
 #### Specify package versions
 
--- a/src/content/plugins/module-federation-plugin.mdx
+++ b/src/content/plugins/module-federation-plugin.mdx
@@ -5,6 +5,7 @@ contributors:
   - XiaofengXie16
   - chenxsan
   - burhanuday
+  - christian24
 related:
   - title: Module Federation
     url: /concepts/module-federation/
@@ -48,11 +49,15 @@ module.exports = {
 };
 ```
 
-### Specify package versions
+### Sharing libraries
+
+With the `shared` key in the configuration you can define libraries that are shared between your federated modules. The package name is the same as the one found in the `dependencies` section of your package.json. However, by default webpack will only share the root level of a library.
+
+#### Specify package versions
 
 There are three ways to specify the versions of shared libraries.
 
-#### Array syntax
+##### Array syntax
 
 This syntax allows you to share libraries with package name only. This approach is good for prototyping, but it will not allow you to scale to large production environment given that libraries like `react` and `react-dom` will require additional requirements.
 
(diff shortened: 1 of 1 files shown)


evenstensberg (commented):
CC @ScriptedAlchemy

ScriptedAlchemy (approved):
Excellent information

evenstensberg:
@Christian24 mind rebasing?

Christian24:
> @Christian24 mind rebasing?

@evenstensberg give me a couple hours. :)

evenstensberg (approved):

Christian24:
@evenstensberg is there something else I need to do here?

snitin315 (approved):

snitin315:
I think we can merge. Deploy previews from forks need spproval on vercel, not sure who has the access.

evenstensberg:
I can too but need to figure out what is wrong with the deployment phase. For reference, I can approve the deploy but it doesn't update the pipeline or comments the deployment url...

PR #6550, opened 2022-12-08
docs(guides): fix broken links in typescript guide

qeleb:
Fixed 3 relative links which must begin from /src/ to work


chenxsan:
Thanks, but those links are set for production, not development. You can preview it on https://webpack-js-org-git-fork-qeleb-main-webpack-docs.vercel.app/guides/typescript/ to see that the current change would break the links on production now. We're going to have link like `https://webpack-js-org-git-fork-qeleb-main-webpack-docs.vercel.app/src/content/concepts/configuration.mdx` which doesn't exist.

PR #6555, opened 2022-12-13
docs(guides): improve document

chenxsan:
Improve document.

### DIFF
--- a/src/content/guides/production.mdx
+++ b/src/content/guides/production.mdx
@@ -152,7 +152,7 @@ Feel free to run those scripts and see how the output changes as we continue add
 
 ## Specify the Mode
 
-Many libraries will key off the `process.env.NODE_ENV` variable to determine what should be included in the library. For example, when `process.env.NODE_ENV` is not set to `'production'` some libraries may add additional logging and testing to make debugging easier. However, with `process.env.NODE_ENV` set to `'production'` they might drop or add significant portions of code to optimize how things run for your actual users. Since webpack v4, specifying [`mode`](/configuration/mode/) automatically configures [`DefinePlugin`](/plugins/define-plugin) for you:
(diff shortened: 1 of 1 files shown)


src/content/guides/production.mdx
```diff
@@ -165,7 +165,7 @@ Many libraries will key off the `process.env.NODE_ENV` variable to determine wha
   });
 ```
 
-T> Technically, `NODE_ENV` is a system environment variable that Node.js exposes into running scripts. It is used by convention to determine dev-vs-prod behavior by server tools, build scripts, and client-side libraries. Contrary to expectations, `process.env.NODE_ENV` is not set to `'production'` **within** the build script `webpack.config.js`, see [#2537](https://github.com/webpack/webpack/issues/2537). Thus, conditionals like `process.env.NODE_ENV === 'production' ? '[name].[contenthash].bundle.js' : '[name].bundle.js'` within webpack configurations do not work as expected.
+T> Technically, `NODE_ENV` is a system environment variable that Node.js exposes into running scripts. It is used by convention to determine dev-vs-prod behavior by server tools, build scripts, and client-side libraries.
```
snitin315, line 168: This guide doesn't mention the usage of `--node-env`.

chenxsan: Indeed, seems I was wrong to assume it's related to `--node-env`, description updated. But still it's stale content as there's no `-p` anymore [from what I can tell](https://github.com/webpack/webpack-cli/pull/1693)? Hence it doesn't make sense to link to it.

chenxsan: Thanks for heads-up, absolutely my oversight. I've went over that thread, and think we can still make some improvements here. Let me know what you think.

snitin315 (approved):

PR #6571, opened 2022-12-23
Fix some grammar

klu22:


snitin315:
@klu22 Can you accept CLA?

klu22:
Yes, signed just now

chenxsan:
Thank you for taking your time. But I don't see those changes necessary.

E.g., what's wrong with `could rapidly`? What's the benefit changing it to `can quickly`?

And why would use 2 or more sentences when you can make them in one?

klu22:
> E.g., what's wrong with `could rapidly`? What's the benefit changing it to `can quickly`?

The latter is shorter, easier to read, and more natural to native English speakers.

> And why would use 2 or more sentences when you can make them in one?

The sentences I revised were [run-on sentences](https://www.wordnik.com/words/run-on%20sentence), which are grammatically incorrect.

chenxsan:
> > E.g., what's wrong with `could rapidly`? What's the benefit changing it to `can quickly`?
> 
> The latter is shorter, easier to read, and more natural to native English speakers.
> 
> > And why would use 2 or more sentences when you can make them in one?
> 
> The sentences I revised were [run-on sentences](https://www.wordnik.com/words/run-on%20sentence), which are grammatically incorrect.

Webpack doesn't serve only native English speaker, it has users all around the world who can read English, and I don't think we're going to change all `rapidly` to `quickly` or even forbid the use of `rapidly` on this site, that doesn't make sense to me.

Regarding the grammar errors, since I'm not really a good English speaker, so let me get some help from chatgpt:

klu22:
> Webpack doesn't serve only native English speaker, it has users all around the world who can read English

Proper English is better than broken English.

>I don't think we're going to change all rapidly to quickly or even forbid the use of rapidly on this site, that doesn't make sense to me.

I never said we should do that.

> Regarding the grammar errors, since I'm not really a good English speaker, so let me get some help from chatgpt:

ChatGPT isn't an authoritative source, because its output is often incorrect. Consider this link:
https://www.wordnik.com/words/run-on%20sentence

chenxsan:
> > Webpack doesn't serve only native English speaker, it has users all around the world who can read English
> 
> Proper English is better than broken English.
> 
> > I don't think we're going to change all rapidly to quickly or even forbid the use of rapidly on this site, that doesn't make sense to me.
> 
> I never said we should do that.
> 
> > Regarding the grammar errors, since I'm not really a good English speaker, so let me get some help from chatgpt:
> 
> ChatGPT isn't an authoritative source, because its output is often incorrect. Consider this link: https://www.wordnik.com/words/run-on%20sentence

I reckon t's all bikeshedding here. Anyway, I'll leave it open just in case any English experts want to step in and help review it.

klu22:
> I reckon it's all bikeshedding here. 

Since you argued against my commit, it's natural that I defend it and explain why I think it's good.

>Anyway, I'll leave it open just in case any English experts want to step in and help review it.

Fair enough. I agree that we could use a second opinion.

snitin315 (changes requested):
I don't feel we should accept this change as a standalone commit. Grammarly doesn't show these as grammar mistakes.

klu22:
The sentences I edited are all run-on sentences that use comma splices:

http://writing.umn.edu/sws/quickhelp/grammar/runons.html

As you know, online grammar checking software like Grammarly or ChatGPT are not always reliable. This very case is an example. Please read about run-on sentences and comma splices in the link above and see how they apply, quite straightforwardly, to the sentences I edited.

PR #6583, opened 2023-01-05
docs(plugins): clarify `SourceMapDevToolPlugin` usage

dgautsch:
While implementing the `SourceMapDevToolPlugin` in my project I was following the documentation and noticed it caused an error in the build. I copied this from the documentation:

```javascript
new webpack.SourceMapDevToolPlugin({
  append: '\n//# sourceMappingURL=https://example.com/sourcemap/[url]',
  filename: '[name].map',
});
```

Ultimately I encountered this error: `Multiple assets emit different content to the same filename`

After some research I noticed it should be set to `[file].map[query]` as specified in this issue https://github.com/webpack/webpack/issues/9732#issuecomment-555461786

I figured this was worth fixing so others don't encounter the same issue I did.


chenxsan (approved):
Thanks.

PR #6600, opened 2023-01-17
fix api rate limit

chenxsan:
Investigating api rate limit problem which was rampant recently.

E.g., https://github.com/webpack/webpack.js.org/actions/runs/3947765127/jobs/6756970622#step:6:28:


src/utilities/fetch-package-repos.mjs
```diff
@@ -33,17 +32,6 @@ const fetch = {
 };
 
 async function main() {
-  let api;
-  if (process.env.CI && process.env.CI === true) {
```
chenxsan, line 37: `process.env.CI` is of type `string`, not `boolean`, so the original code was actually wrong which means the api wasn't authenticated at all before.

src/utilities/fetch-package-readmes.mjs
```diff
@@ -84,8 +85,13 @@ async function main() {
         headmatter = yamlHeadmatter(basic);
       }
 
-      const response = await fetch(url);
-      const content = await response.text();
+      const { data: content } = await api.repos.getReadme({
```
chenxsan, line 88: `fetch` was not authenticated before.

PR #6622, opened 2023-01-28
Use native concurrency

chenxsan:
According to https://github.com/styfle/cancel-workflow-action, `cancel-workflow-action` is no longer recommended, and we can apply the native `concurrency` feature on workflow level per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency.


.github/workflows/testing.yml
```diff
@@ -3,6 +3,10 @@ name: Ensure PR
 on:
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
snitin315, line 9: Should we define it at the job level?

chenxsan: I think it makes more sense to run at workflow level since we need to rerun all jobs against the latest commit and cancel any running jobs, besides, it's much simpler and easier to maintain.

snitin315 (approved):

snitin315 (approved):
LGTM

PR #6636, opened 2023-02-12
docs(guides): clarify bundled files served directory

weidehai:


src/content/guides/development.mdx
```diff
@@ -236,7 +236,7 @@ This tells `webpack-dev-server` to serve the files from the `dist` directory on
 
 T> The `optimization.runtimeChunk: 'single'` was added because in this example we have more than one entrypoint on a single HTML page. Without this, we could get into trouble described [here](https://bundlers.tooling.report/code-splitting/multi-entry/). Read the [Code Splitting](/guides/code-splitting/) chapter for more details.
 
-T> `webpack-dev-server` serves bundled files from the directory defined in [`output.path`](/configuration/output/#outputpath), i.e., files will be available under `http://[devServer.host]:[devServer.port]/[output.publicPath]/[output.filename]`.
+T> `webpack-dev-server` serves bundled files from the directory defined in [`output.publicPath`](/configuration/output/#outputpublicpath), i.e., files will be available under `http://[devServer.host]:[devServer.port]/[output.publicPath]/[output.filename]`.
```
chenxsan, line 239: Nah, this doesn't seem right. `output.path` defines where webpack stores its bundled files, and wds would serve those files from that directory, even though `output.publicPath` would rewrite the served path, it doesn't mean that wds would serve files from `output.publicPath`.

weidehai: may be should provide more detail explanation here？my friend tell me he is feel puzzled when reading here，and he is beginner for web pack

chenxsan: Just ignore those options and use the default values until you really need them, only then those options would make sense :)

PR #6643, opened 2023-02-15
!perf: change webpack hash function to improve speed of hash calculation

WarrenLee19:
change: webpack.common.mjs
change webpack hash function to improve speed of hash calculation
set output.hashFunction to 'xxhash64'


chenxsan (approved):
Thanks.

PR #6647, opened 2023-02-16
Update hot-module-replacement.mdx

kertof:
Updates the documentation regarding HMR support in Angular. 

HMR support is now built-in the Angular CLI, and does not require an external package anymore, and does not require any changes in the codebase.


chenxsan (approved):
Thanks.

Issue #6646, opened 2023-01-28
Documentation incorrect or incomplete for `/plugins/split-chunks-plugin/#splitchunksname`


benjaminblack:
The documentation reads:

> If you choose to specify a function, you may find the `chunk.name` and `chunk.hash` properties (where chunk is an element of the chunks array) particularly useful in choosing a name for your chunk.

The `chunk.hash` property seems to not exist. [Minimal example demonstrating the problem](https://stackblitz.com/edit/webpack-webpack-js-org-m6dszs?file=webpack.config.js).

I.e., in `chunks.forEach(({name, hash}) => ...)`, `name` is `"main"` but `hash` is `undefined`.

alexander-akait:
@chenxsan Feel free to close if you already fixed it

PR #6650, opened 2023-02-18
docs(plugins): remove chunk.hash

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6646

Removing `chunk.hash` to avoid confusion due to potential `undefined` `chunk.hash`.


snitin315 (approved):
Are these lockfile changes related?

chenxsan:
> Are these lockfile changes related?

Nah, not related. I've pulled the upstream code and run `yarn` which results in the lockfile changes.

PR #6651, opened 2023-02-19
Make .deleted changes non-selectable.

DerDemystifier:
Simple change, but would make guides easier to follow. This would make people easily copy the entire file's content instead of manually making line changes, especially if there are many changes to make in multiple files.


chenxsan (approved):
Great idea!

PR #6656, opened 2023-02-21
docs(configuration): remove [ext] for filename

weidehai:
it seem there no support [ext] placeholder in entery filename

See https://webpack.js.org/configuration/output/#template-strings

CleanShot 2023-02-24 at 15 52 18@2x


chenxsan (approved):
Thanks.

PR #6659, opened 2023-02-25
upgrade react-spring

chenxsan:
Fix https://github.com/webpack/webpack.js.org/pull/6658.

There was a bug in `react-spring` https://github.com/pmndrs/react-spring/issues/2097 when used in ESM environment, hopefully it will get fixed in near future so we can remove the `fullySpecified ` option.


snitin315 (approved):

joshuaellis:
This might be a good opportunity to instead just install `@react-spring/web` instead of `react-spring` by the way :)

PR #6663, opened 2023-02-28
upgrade react-spring

chenxsan:
Follow-up of https://github.com/webpack/webpack.js.org/pull/6659 as the bug in react-spring has been fixed, hence we can remove the `fullySpecified` rule in `webpack.config.js`.

Closes https://github.com/webpack/webpack.js.org/pull/6662

### DIFF
--- a/src/utilities/githubAPI.mjs
+++ b/src/utilities/githubAPI.mjs
@@ -2,7 +2,7 @@ import { Octokit as GithubAPI } from '@octokit/rest';
 import { createActionAuth } from '@octokit/auth-action';
 /** @type import('@octokit/rest').Octokit */
 let api;
-console.log('process.env.CI', JSON.stringify(process.env.CI));
+console.log('process.env.CI', JSON.stringify(process.env.CI), process.env.CI === '1');
 if (process.env.CI && process.env.CI === 'true') {
   const auth = createActionAuth();
   const authentication = await auth();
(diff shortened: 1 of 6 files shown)


chenxsan:
Weird, something changes the `process.env.CI` value from `true` to `1`:

CleanShot 2023-02-28 at 10 35 15@2x

src/utilities/githubAPI.mjs
```diff
@@ -2,7 +2,11 @@ import { Octokit as GithubAPI } from '@octokit/rest';
 import { createActionAuth } from '@octokit/auth-action';
 /** @type import('@octokit/rest').Octokit */
 let api;
-if (process.env.CI && process.env.CI === 'true') {
+if (
+  process.env.CI &&
+  process.env.GITHUB_ACTION &&
```
chenxsan, line 7: Detect `process.env.GITHUB_ACTION` as well since vercel would set `process.env.CI` too, see https://vercel.com/guides/how-do-i-resolve-a-process-env-ci-true-error

snitin315: Did you figure out which package is setting `CI` to `1`?

chenxsan: See https://github.com/cypress-io/github-action/blob/9674a20f82e9e45ec75aa66038310b00e2f24657/index.js#L223

PR #6699, opened 2023-03-18
Remove excess change in getting-started.mdx

AlexMisiulia:


snitin315 (changes requested):
It's not an extra change. There is a comma added at the end of the script.

AlexMisiulia:
> It's not an extra change. There is a comma added at the end of the script.

Ah, you're right. Closing it, thanks for the feedback

PR #6712, opened 2023-03-24
docs(Configuration): add output.workerPublicPath

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6711

### DIFF
--- a/src/content/configuration/output.mdx
+++ b/src/content/configuration/output.mdx
@@ -2311,7 +2311,7 @@ module.exports = {
 
 `string`
 
-Set a public path for Worker, defaults to value of [output.publicPath](#outputpublicpath). Avoid this option unless your worker scripts are located at a different path from your other script files.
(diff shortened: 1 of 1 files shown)


chenxsan:
DocSearch no longer working 
CleanShot 2023-03-24 at 12 23 03@2x

src/content/configuration/output.mdx
```diff
@@ -2306,3 +2306,20 @@ module.exports = {
   },
 };
 ```
+
+## output.workerPublicPath
+
+`string`
+
+Set a public path for Worker, defaults to value of [output.publicPath](#outputpublicpath). Avoid this option unless your worker scripts are located at a different path from your other script files.
```
chenxsan, line 2314:
```suggestion
Set a public path for Worker, defaults to value of [output.publicPath](#outputpublicpath). Only use this option if your worker scripts are located in a different path from your other scripts.
```

PR #6727, opened 2023-03-31
docs: append `webpack-stats-viewer` in code-splitting

moonrailgun:
I make a tools for show more detail which existed in webpack. At my work for improve my project performance, I found i need more and more detail content which can help me. so i build it. And i wanna make more people like me known it.

Its copy from my repo:

> Many same webpack analyzer tools can show summary, but as a developer, when i try go deep into, and think why those file bundled into those chunk, those tools make me powerless, because of i cannot get more detail. So i develop this tool, no magic, just plain data. I think i need it.
> This is for in-depth study of webpack chunks composition and for deep website performance optimization

---------


chenxsan (approved):
Looks good.

PR #6736, opened 2023-04-04
fix bug in readme processing

chenxsan:
Noticed a bug in https://github.com/webpack/webpack.js.org/actions/runs/4600783997/jobs/8127825773?pr=6734#step:7:9 caused by the readme of https://github.com/webpack-contrib/postcss-loader:

```md
### `postcssOptions`

See the file [`./src/config.d.ts`](./src/config.d.ts).

Type:
```

Both matched relative links should be replaced with new one.


snitin315 (approved):
Good catch.

PR #6740, opened 2023-04-04
Update output-management.mdx

Hydrock:
In Preparation section we see creating two modules:

1) src/index.js
2) src/print.js

Inside webpack config we made two enrty points:

```
entry: {
    index: './src/index.js',
    print: './src/print.js',
},
```

We, also, made changes inside html:

```
<!DOCTYPE html>
 <html>
   <head>
     <meta charset="utf-8" />
    <title>Output Management</title>
    <script src="./print.bundle.js"></script>
   </head>
   <body>
    <script src="./index.bundle.js"></script>
   </body>
 </html>
```

We expose print.bundle.js and index.bundle.js module, bucause after building we will have two separated modules

But, there is anoter odd code. We import print.js module inside index.js module. I not understand for what.

We will have separated modules, connected on index.html, but index.js already include print module inside itself. It's weird, or i am stupit.

Please explain me, why we should include print module indide index module?

_describe your changes..._

- [ ] Read and sign the [CLA][1]. PRs that haven't signed it won't be accepted.
- [ ] Make sure your PR complies with the [writer's guide][2].
- [ ] Review the diff carefully as sometimes this can reveal issues.
- [ ] Do not abandon your Pull Request: [Stale Pull Requests][3].

### DIFF
--- a/src/content/guides/output-management.mdx
+++ b/src/content/guides/output-management.mdx
@@ -46,6 +46,7 @@ export default function printMe() {
 
 And use that function in our `src/index.js` file:
 
+
 **src/index.js**
 
 ```diff


chenxsan:
Yeah, it's sort of confusing at the first sight. But how are you gonna use `printMe` function defined in `print.js` if you don't import it? And note that it's not exposed as a global variable in our example so you can't just call `printMe`.

Here's how I would reason about it,

1. first, you write code in ESM, there's nothing special here even if you're going to use webpack to bundle your code
2. you decide to output two bundles. In webpack, one entry would only output one bundle, that's why we define two entries. The contrived example is for education purpose, so it might not be easy to reason about.

Hope it helps.

Hydrock:
> Yeah, it's sort of confusing at the first sight. But how are you gonna use `printMe` function defined in `print.js` if you don't import it? And note that it's not exposed as a global variable in our example so you can't just call `printMe`.
> 
> Here's how I would reason about it,
> 
> 1. first, you write code in ESM, there's nothing special here even if you're going to use webpack to bundle your code
> 2. you decide to output two bundles. In webpack, one entry would only output one bundle, that's why we define two entries. The contrived example is for education purpose, so it might not be easy to reason about.
> 
> Hope it helps.

First of all, thank you for the answer - it's very valuable.

But. This example is very strange. It is designed for beginners, but only misleads them.

Giving an example of the use of technology, an example is given in order to show the practical benefits of technology. Specifically, in this example, two bundles are created.

But it doesn't make any sense. The print module is already included in the index module.

What is the meaning of the fact that we connected the script tag?
```
<script src="./print.bundle.js"></script>
```

it doesn't do anything and only confuses. I think you should agree that the example needs to be redone.

If you wanted to show how the assembly of individual modules works and their application on an html page as separate modules, you need to assemble two separate bundles - for example, in one bundle the hello function, and in the other the print function. Then, connect the modules to the web page using the script tag. And use these functions (modules) on the page itself

What was the point of connecting print at all.a js module on html if the print function which is not even global? It turns out that we already have 2 print functions, while one is not global and there is no access to it.

This is the same if we were explaining to children how gasoline works in a car, but instead of refueling the car, we poured gasoline on the road. It doesn't make any sense.

I am already quite an experienced developer, but even this example has brought me into a stupor, not to mention beginners. Examples should be objective and taken from life tasks, otherwise why else make such wonderful tools as webpack.

I hope for your understanding, I'm sorry if I misunderstood something.

chenxsan:
> > Yeah, it's sort of confusing at the first sight. But how are you gonna use `printMe` function defined in `print.js` if you don't import it? And note that it's not exposed as a global variable in our example so you can't just call `printMe`.
> > Here's how I would reason about it,
> > 
> > 1. first, you write code in ESM, there's nothing special here even if you're going to use webpack to bundle your code
> > 2. you decide to output two bundles. In webpack, one entry would only output one bundle, that's why we define two entries. The contrived example is for education purpose, so it might not be easy to reason about.
> > 
> > Hope it helps.
> 
> First of all, thank you for the answer - it's very valuable.
> 
> But. This example is very strange. It is designed for beginners, but only misleads them.
> 
> Giving an example of the use of technology, an example is given in order to show the practical benefits of technology. Specifically, in this example, two bundles are created.
> 
> But it doesn't make any sense. The print module is already included in the index module.
> 
> What is the meaning of the fact that we connected the script tag?
> 
> ```
> <script src="./print.bundle.js"></script>
> ```
> 
> it doesn't do anything and only confuses. I think you should agree that the example needs to be redone.
> 
> If you wanted to show how the assembly of individual modules works and their application on an html page as separate modules, you need to assemble two separate bundles - for example, in one bundle the hello function, and in the other the print function. Then, connect the modules to the web page using the script tag. And use these functions (modules) on the page itself
> 
> What was the point of connecting print at all.a js module on html if the print function which is not even global? It turns out that we already have 2 print functions, while one is not global and there is no access to it.
> 
> This is the same if we were explaining to children how gasoline works in a car, but instead of refueling the car, we poured gasoline on the road. It doesn't make any sense.
> 
> I am already quite an experienced developer, but even this example has brought me into a stupor, not to mention beginners. Examples should be objective and taken from life tasks, otherwise why else make such wonderful tools as webpack.
> 
> I hope for your understanding, I'm sorry if I misunderstood something.

Thank you for the explanation, now I see the problem after running the code. I agree the contrived example is bad, will see how we can improve it.

PR #6818, opened 2023-05-11
fix api limit of opencollective

chenxsan:
We kept hitting the api limit of opencollective recently.

Let's see if adding a personal token would help a little.

Note that this only targets the `deploy.yml` action at the moment as there're [some limitation of environment secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) stops us from applying to others:

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.


snitin315 (approved):
LGTM

PR #6821, opened 2023-05-12
handle api limit automatically

chenxsan:
This is a much better version to handle the opencollective api limit.

The code will automatically wait until api limit is lifted then continue, and it supports running fetch simultaneously in different tasks, no more manually restart the failed tasks!


snitin315 (approved):
LGTM

Issue #6838, opened 2023-05-20
Scrollbar not visible in sidebar.


Riyan911:
# Bug report

**What is the current behavior?**
Scrollbar is not visible in microsoft edge linux. 
**If the current behavior is a bug, please provide the steps to reproduce.**

**What is the expected behavior?**

**Other relevant information:**
webpack version:  
Node.js version:
Operating System:
Additional tools:

chenxsan:
I believe it was designed this way https://github.com/webpack/webpack.js.org/commit/c7d70be3f00c7642c8dd4392b34315d13c35c94e. And in your screencast, I can see you scrolled sidebar successfully while reaching the end fast as `Under The Hood` is the last item in sidebar menu.

Riyan911:
I meant to say that that the scrollbar isn't visible on the left hand side

chenxsan:
> I meant to say that that the scrollbar isn't visible on the left hand side

Then it's designed this way. But I agree it's sacrificing accessibility for good looks.

PR #6839, opened 2023-05-21
show scrollbar for sidebar

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6838

### DIFF
--- a/src/components/Sidebar/Sidebar.scss
+++ b/src/components/Sidebar/Sidebar.scss
@@ -19,12 +19,6 @@
   top: -1px;
   overflow: scroll;
   max-height: 100vh;
-  // To hide scrollbar in Firefox
-  scrollbar-width: none;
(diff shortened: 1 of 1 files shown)


src/components/Sidebar/Sidebar.scss
```diff
@@ -19,12 +19,6 @@
   top: -1px;
   overflow: scroll;
```
chenxsan, line 20:
```suggestion
  overflow-y: scroll;
```

src/components/Sidebar/Sidebar.scss
```diff
@@ -17,14 +17,8 @@
   padding: 1.5em;
   position: sticky;
   top: -1px;
-  overflow: scroll;
+  overflow-y: scroll;
```
chenxsan, line 20:
```suggestion
  overflow-y: auto;
```

PR #6888, opened 2023-06-08
docs(plugins): document regexp for splitChunks.chunks

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/6882

### DIFF
--- a/src/content/plugins/split-chunks-plugin.mdx
+++ b/src/content/plugins/split-chunks-plugin.mdx
@@ -98,7 +98,7 @@ By default webpack will generate names using origin and name of the chunk (e.g. 
 
 ### splitChunks.chunks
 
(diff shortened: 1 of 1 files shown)


src/content/plugins/split-chunks-plugin.mdx
```diff
@@ -134,6 +134,19 @@ module.exports = {
 };
 ```
 
+If you are using webpack version 5.8.6 or later, you can also pass a regular expression:
```
snitin315, line 137:
```suggestion
If you are using webpack version 5.86.0 or later, you can also pass a regular expression:
```

snitin315 (changes requested):

snitin315 (approved):

PR #6919, opened 2023-06-17
docs: add docs for the css experiment

burhanuday:
_describe your changes..._

Closes https://github.com/webpack/webpack.js.org/issues/6909

Add documentation for `experiments.css`

### DIFF
--- a/src/content/configuration/experiments.mdx
+++ b/src/content/configuration/experiments.mdx
@@ -202,8 +202,7 @@ Experimental features:
 - <Badge text="5.87.0+" /> Style-specific fields resolution in `package.json` files:
   webpack will look for `style` field in `package.json` files and use that if it
   exists for imports inside CSS files.
-- Content hash for CSS files: webpack will generate a content hash for CSS files
-  and use it in the filename. This is useful for long-term caching.
+- Content hash for CSS files: webpack will generate a content hash for CSS files and use it in the filename. This is useful for long-term caching.
 - CSS extraction: Webpack will extract CSS into a separate file. This functionality replaces the need for `mini-css-extract-plugin` and `css-loader`, as it provides native support.
 - CSS imports: webpack will inline CSS imports into the generated CSS file.
 - Hot Module Reload (HMR): webpack supports HMR for CSS files. This means that changes made to CSS files will be reflected in the browser without a full page reload.
--- a/src/content/configuration/experiments.mdx
+++ b/src/content/configuration/experiments.mdx
@@ -184,6 +184,28 @@ Detect changes to remote resources and upgrade them automatically.
 
 Enable native CSS support. Note that it's an experimental feature still under development and will be enabled by default in webpack v6, however you can track the progress on [GitHub](https://github.com/webpack/webpack/issues/14893).
 
+- Type:
+
+  - `boolean`
+  - `CssExperimentOptions`
+
+    ```ts
+    {
+      exportsOnly?: boolean;
+    }
+    ```
+
+Experimental features:
+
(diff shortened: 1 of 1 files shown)


src/content/configuration/experiments.mdx
```diff
@@ -184,6 +184,28 @@ Detect changes to remote resources and upgrade them automatically.
...
+
+    ```ts
+    {
+      exportsOnly?: boolean;
+    }
+    ```
+
+Experimental features:
+
+- Inbuilt support for CSS and CSS Modules. Use the `.module.css` extension for CSS Modules.
```
alexander-akait, line 200: We need to clarify what you shoudl set `css/auto` to enable this, I think in future we should use `css/auto` by default for any CSS files

burhanuday: based on the filename, we automatically use css modules rule
https://github.com/webpack/webpack/blob/dc4d97429b2013f7770ba785721b65cf82c2ef04/lib/config/defaults.js#L728-L729
dont need to set css/auto explicitly

alexander-akait: Oh, yeah, missing that

src/content/configuration/experiments.mdx
```diff
@@ -184,6 +185,33 @@ Detect changes to remote resources and upgrade them automatically.
...
+      exportsOnly?: boolean;
+    }
+    ```
+
+Experimental features:
+
+- CSS Modules support: webpack will generate a unique name for each CSS class. Use the `.module.css` extension for CSS Modules.
+- <Badge text="5.87.0+" /> Style-specific fields resolution in `package.json` files:
+  webpack will look for `style` field in `package.json` files and use that if it
+  exists for imports inside CSS files.
```
chenxsan, line 204: Maybe include an example for this list item?

src/content/configuration/experiments.mdx
```diff
@@ -184,6 +185,33 @@ Detect changes to remote resources and upgrade them automatically.
...
+- Content hash for CSS files: webpack will generate a content hash for CSS files
+  and use it in the filename. This is useful for long-term caching.
+- CSS extraction: Webpack will extract CSS into a separate file. This functionality replaces the need for `mini-css-extract-plugin` and `css-loader`, as it provides native support.
```
chenxsan, line 207:
```suggestion
- CSS extraction: webpack will extract CSS into a separate file. This functionality replaces the need for `mini-css-extract-plugin` and `css-loader`, as it provides native support.
```
Let's keep consistency.

src/content/configuration/experiments.mdx
burhanuday: > Btw, did you insert many soft line breaks?

Strange. The line break gets added by prettier. Was able to remove it from one point but not the other

burhanuday: > Maybe include an example for this list item?

It is a little difficult to explain that point with an example since it has to do with module resolution inside of `node_modules` folder. Will give it a try

burhanuday: > Maybe include an example for this list item?

Added an example with using bootstrap since they do have that field in their package.json file https://github.com/twbs/bootstrap/blob/main/package.json#L40

burhanuday: > Btw, did you insert many soft line breaks?

This is happening due to MDX component Badge. VSCode considers the file as markdown and thats just how it formats it for me 😅

chenxsan (approved):
Looks good to me, thanks.

PR #6935, opened 2023-06-28
docs(debugging): remove `node-nightly` in favor of `node`

styfle:
Since `node --inspect` has been available for 7 years, I think its time to remove `node-nightly` from the docs.


chenxsan (approved):
Thanks.

PR #6981, opened 2023-07-28
docs(typescript): update typescript guide on using third party libs

ross-u:
The TypeSearch feature is no longer available, as explained [here](https://www.typescriptlang.org/dt/search). Therefore the reference link to the TypeSearch page and related explanation can be removed.


chenxsan (approved):
Thanks.

PR #6987, opened 2023-08-02
docs(configuration): correct Condition reference

undergroundquizscene:
Based on the name in the sidebar, I believe this should be called “Condition”, not “Condition.test”.


chenxsan (approved):
Thanks, you're right, we don't really have condition.test item.

PR #7002, opened 2023-08-23
docs(contribute): remove broken link to the documentation team

dkdk225:
I was unable to find a new documentation team link. As a solution i suggest removing it completely

### DIFF
--- a/src/content/contribute/index.mdx
+++ b/src/content/contribute/index.mdx
@@ -8,6 +8,7 @@ contributors:
   - dhedgecock
   - tbroadley
   - EugeneHlushko
+  - dkdk225
 ---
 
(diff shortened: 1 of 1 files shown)


snitin315 (commented):
The link is actually valid, but it's accessible only for webpack org members on GitHub

dkdk225:
Oh i see. Do you have a suggestion for improvement (like making the link avalible to non-members) or should we leave it as it is

dkdk225:
Oh i see. Then should i change commit/PR name or do you suggest another solution ?

evenstensberg (approved):
lgtm, an author can just cc a maintainer by looking at github contributors page

snitin315 (approved):

PR #7068, opened 2023-11-23
Added info about folder separator on windows

WhyINeedToFillUsername:
see https://stackoverflow.com/a/56084864/4255158 for more context. There should be info about this in the docs!


chenxsan (approved):
thanks.

PR #7075, opened 2023-12-08
docs: add notice for deprecated compilation hooks

burhanuday:
Add a deprecation notice for deprecated hooks

These hooks are deprecated and deprecation notice exists in code but not in the docs - https://github.com/webpack/webpack/pull/10876


burhanuday:
@chenxsan @alexander-akait bumping this

chenxsan (approved):

Issue #7084, opened 2023-12-28
Bug: Broken Link to CommonJS Specification


wheezyturbo:
### Bug

[CommonJS Broken Link](https://webpack.js.org/concepts/modules/#what-is-a-webpack-module)

### Additional information:
![Screenshot from 2023-12-28 10-42-21](https://github.com/webpack/webpack.js.org/assets/77047619/63e35878-5599-4c55-b56c-02e44fcc2530)
![Screenshot from 2023-12-28 10-42-44](https://github.com/webpack/webpack.js.org/assets/77047619/45c39151-4bb0-42ae-851d-61c3bd9cc7e0)

The link to the CommonJS specification on the webpack.js.org documentation is currently broken. When users click on the CommonJS link, it directs them to a "Not Found" page with the following error:

Location in Documentation: [Link location - https://webpack.js.org/concepts/modules/#what-is-a-webpack-module]
Current URL: [Broken URL - https://wiki.commonjs.org/specs/modules/1.0/]

### Steps to Reproduce:

Go to [link location](https://webpack.js.org/concepts/modules/#what-is-a-webpack-module) in the webpack.js.org documentation.
Click on the "CommonJS" link.
Additional Context:
The CommonJS specification is crucial for users to understand different module systems. A working link will improve user experience and ensure accurate information.

alexander-akait:
Hello, do you want to send a PR?

wheezyturbo:
> Hello, do you want to send a PR?

Yes, I would like to

wheezyturbo:
#7086

PR #7086, opened 2023-12-31
chore(docs): Fix broken CommonJS link in webpack.js.org documentation

wheezyturbo:
This PR addresses a broken link issue in the webpack.js.org documentation. The previous link for the CommonJS specification was leading to a "Not Found" page. I've updated the link to the correct URL (https://wiki.commonjs.org/wiki/Modules/1.1) to ensure users can access the accurate information about CommonJS modules.
Resolves https://github.com/webpack/webpack.js.org/issues/7084


snitin315 (approved):
LGTM, thanks!

PR #7091, opened 2024-01-02
docs: add note about how publicPath interacts with asset modules

Aghassi:
This updates the documentation to better define how `publicPath` interacts with `asset` module types. Specifically, it is poorly documented on how to dynamically set the `publicPath` on the fly for these module types. This is important and useful in an enterprise environment where there may be different upstream CDNs or other locations that the app hosts assets on that aren't the server or a static site.

### DIFF
--- a/src/content/guides/asset-modules.mdx
+++ b/src/content/guides/asset-modules.mdx
@@ -78,6 +78,30 @@ module.exports = {
 }
 ```
 
+### Public Path
+
+By default, the `asset` type does `__webpack_public_path__ + import.meta` under the hood. This means that setting the `output.publicPath` in your config will allow you to override the url that the `asset` loads from.
+
+#### On The Fly Override
+
+If you set the `__webpack_public_path__` in code, the way you need to achive it so as not to break the `asset` loading logic is to make sure you run it as the first code in your app and not use a function to do so. An example of this would be having a file called `publicPath.js` with contents
+
+```javascript
+__webpack_public_path = https://cdn.url.com
+```
+
+And then in your `webpack.config.js` updating your `entry` field to look like
+
+```javascript
+entry: ['./publicPath.js', './App.js']
+```
+
+Alternatively, you can do the following in your `App.js` without modifying your webpack config. The only downside is you have to enforce ordering here and that can collide with some linting tools.
+
+```javascript
+import './publicPath.js'
+```
+
 ## Resource assets
 
 **webpack.config.js**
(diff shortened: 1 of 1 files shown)


alexander-akait (dismissed):

snitin315:
@Aghassi Lint is failing on CI, can you fix?

Aghassi:
Yup let me do that

Aghassi:
Is there a good way to understand the Typeerror from CI that Cypress is throwing? The stack trace is pretty sparse.

alexander-akait:
Weird, we need to check it

@snitin315 Can you look why our tests are failed?

src/content/guides/asset-modules.mdx
```diff
@@ -78,6 +78,32 @@ module.exports = {
 }
 ```
 
+### Public Path
+
+By default, the `asset` type does `__webpack_public_path__ + import.meta` under the hood. This means that setting the `output.publicPath` in your config will allow you to override the url that the `asset` loads from.
+
+#### On The Fly Override
+
+If you set the `__webpack_public_path__` in code, the way you need to achive it so as not to break the `asset` loading logic is to make sure you run it as the first code in your app and not use a function to do so. An example of this would be having a file called `publicPath.js` with contents
```
Aghassi, line 87:
```suggestion
If you set the `__webpack_public_path__` in code, the way you need to achieve it so as not to break the `asset` loading logic is to make sure you run it as the first code in your app and not use a function to do so. An example of this would be having a file called `publicPath.js` with contents
```

src/content/guides/asset-modules.mdx
```diff
@@ -78,6 +78,32 @@ module.exports = {
 }
 ```
 
+### Public Path
+
+By default, the `asset` type does `__webpack_public_path__ + import.meta` under the hood. This means that setting the `output.publicPath` in your config will allow you to override the url that the `asset` loads from.
```
snitin315, line 83:
```suggestion
By default, under the hood, the `asset` type does `__webpack_public_path__ + import.meta`. This means that setting the `output.publicPath` in your config will allow you to override the URL from which the `asset` loads.
```

src/content/guides/asset-modules.mdx
```diff
@@ -78,6 +78,32 @@ module.exports = {
 }
 ```
 
+### Public Path
```
snitin315, line 81:
```suggestion
## Public Path
```

src/content/guides/asset-modules.mdx
```diff
@@ -78,6 +78,32 @@ module.exports = {
...
+
+By default, under the hood, the `asset` type does `__webpack_public_path__ + import.meta`. This means that setting the `output.publicPath` in your config will allow you to override the URL from which the `asset` loads.
+
+#### On The Fly Override
```
snitin315, line 85:
```suggestion
### On The Fly Override
```

snitin315:
Fixed, it was because of the incorrect heading levels.

snitin315 (approved):
LGTM, thanks for contributing.

PR #7139, opened 2024-02-14
Fix typo in __dirname and __filename options

monkbroc:
node.__filename and node.__dirname accept 'node-module' as an option, not 'node_module'

Using 'node-module' works, whereas using 'node_module' gives the following error:
```
[webpack-cli] Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration.node.__filename should be one of these:
   false | true | "warn-mock" | "mock" | "node-module" | "eval-only"
   -> Include a polyfill for the '__filename' variable.
```

### DIFF
--- a/src/content/configuration/node.mdx
+++ b/src/content/configuration/node.mdx
@@ -54,7 +54,7 @@ Options:
 
 ## node.\_\_filename
 
-`boolean` `'mock' | 'warn-mock' | 'node_module' | 'eval-only'`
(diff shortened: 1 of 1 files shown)


alexander-akait (approved):

alexander-akait:
Thank you

monkbroc:
You're welcome

PR #7180, opened 2024-03-13
replace outdated references to libraryTarget with library.type

tanyabouman:
The `libraryTarget` option has a warning that support might be dropped and it's better to use `library.type`.  I left `libraryTarget` in other sections that are marked as maybe dropping support later because that seems consistent for those options.


alexander-akait (approved):

alexander-akait:
Thank you

PR #7251, opened 2024-05-07
docs: add use Rspack to build performance guide

chenjiahan:
This PR adds the "use Rspack" guide to the build performance document.

Using Rspack can be a pragmatic option for Webpack users.  If the user has already tried to optimize the build performance of the webpack application and has not achieved the expected results, then using Rspack would be a good choice.

Feel free to close this if this is inappropriate for webpack documentation ❤️

### DIFF
--- a/src/content/guides/build-performance.mdx
+++ b/src/content/guides/build-performance.mdx
@@ -26,6 +26,14 @@ Use the latest webpack version. We are always making performance improvements. T
 
 Staying up-to-date with **Node.js** can also help with performance. On top of this, keeping your package manager (e.g. `npm` or `yarn`) up-to-date can also help. Newer versions create more efficient module trees and increase resolving speed.
 
+### Use Rspack
+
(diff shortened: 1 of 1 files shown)


chenjiahan:
@alexander-akait Can you review this PR? Thanks! ❤️

chenxsan:
Not sure it's a good idea to add it to build performance guide, it's more like a replacement of webpack from what I can tell. Maybe https://webpack.js.org/comparison/ if you want to update the table.

snitin315:
Yes, comparison table should be the place for Rspack.

chenjiahan:
Thanks for the suggestion!  Indeed, Rspack seems more like a replacement for webpack. Unlike other replacements, Rspack aims to be compatible with and accelerate the webpack ecosystem, rather than break it.

The comparison table looks a bit outdated, and adding Rspack to it might not catch the users' attention. I will close this PR first, then see if we can add anything to the webpack documentation. ❤️

PR #7258, opened 2024-05-18
ushers user into https page

gabrielfalcao:


snitin315:
This change is not required.

PR #7298, opened 2024-07-03
upgrade react to v18

chenxsan:


chenxsan:
Seems impossible, lol.

PR #7300, opened 2024-07-06
add a comma

Nala60:
I added a comma in "Awesome, isn't it? [...]" because I felt it would fit.


chenxsan (approved):
Not really necessary, but I'm ok with it.

Issue #6748, opened 2023-04-07
Lets better document `resolve.unsafeCache` 


TheLarkInn:
**Feature to document**

**Author(s)**
@TheLarkInn @alexander-akait We need to reference the code for this. 

**Additional information**

[ ] I am willing to work on this issue and submit a pull request.

PR #7303, opened 2024-07-09
docs(configuration): update resolve.unsafeCache

chenxsan:
Close https://github.com/webpack/webpack.js.org/issues/6748. Related https://github.com/webpack/webpack/issues/7976#issuecomment-820706796

Screenshot-mQRW2Msw@2x

### DIFF
--- a/src/content/configuration/resolve.mdx
+++ b/src/content/configuration/resolve.mdx
@@ -716,7 +716,7 @@ module.exports = {
 
 ### resolve.unsafeCache
 
-`RegExp` `[RegExp]` `boolean: true`
+`object` `boolean: true`
 
 Enable aggressive, but **unsafe**, caching of modules. Passing `true` will cache everything.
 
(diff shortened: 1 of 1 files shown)


src/content/configuration/resolve.mdx
```diff
@@ -731,15 +731,34 @@ module.exports = {
 };
 ```
 
-A regular expression, or an array of regular expressions, can be used to test file paths and only cache certain modules. For example, to only cache utilities:
+When an object is provided, webpack will use it as cache.
+
+For example, you can supply a [Proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) instead of regular one:
```
snitin315, line 736:
```suggestion
For example, you can supply a [Proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) instead of a regular one:
```

src/content/configuration/resolve.mdx
```diff
@@ -716,7 +716,7 @@ module.exports = {
 
 ### resolve.unsafeCache
 
-`RegExp` `[RegExp]` `boolean: true`
+`object` `boolean: true`
```
snitin315, line 719: We use `=` to indicate [the default value](https://webpack.js.org/contribute/writers-guide/#configuration-defaults-and-types).
```suggestion
`object` `boolean = true`
```

PR #7319, opened 2024-07-19
upgrade husky

chenxsan:


chenxsan:
Hmm, seems something's wrong with github action.

Issue #7323, opened 2024-07-23
Imporve https://webpack.js.org/guides/package-exports/#alternatives


alexander-akait:
**Feature to document**

Need improve our docs - https://github.com/webpack/enhanced-resolve/pull/429, only first valid thing thing will be used

**Author(s)**

No
**Additional information**

No

[ ] I am willing to work on this issue and submit a pull request.

PR #7326, opened 2024-07-24
docs(guides): update alternatives

chenxsan:
Closes https://github.com/webpack/webpack.js.org/issues/7323

### DIFF
--- a/src/content/guides/package-exports.mdx
+++ b/src/content/guides/package-exports.mdx
@@ -67,6 +67,20 @@ Example:
 
 Here `package/things/apple` might be found in `.../package/good-things/apple` or in `.../package/bad-things/apple`.
 
+W> As of version 5.93.1, webpack's behavior has been updated to align with Node's behavior. It now selects the first valid path without attempting further resolutions and throws an error if the path cannot be resolved.
+
+For example, given the following configuration:
+
+```json
+{
+  "exports": {
+    ".": ["-bad-specifier-", "./non-existent.js", "./existent.js"]
+  }
+}
+```
+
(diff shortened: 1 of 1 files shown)


alexander-akait (dismissed):

src/content/guides/package-exports.mdx
```diff
@@ -67,6 +67,20 @@ Example:
 
 Here `package/things/apple` might be found in `.../package/good-things/apple` or in `.../package/bad-things/apple`.
 
+W> As of version 5.93.1, webpack's behavior has been updated to align with Node's behavior. It now selects the first valid path without attempting further resolutions and throws an error if the path cannot be resolved.
```
chenxsan, line 70:
```suggestion
W> As of version 5.94.0, webpack's behavior has been updated to align with Node's behavior. It now selects the first valid path without attempting further resolutions and throws an error if the path cannot be resolved.
```

src/content/guides/package-exports.mdx
```diff
@@ -67,6 +67,20 @@ Example:
...
+
+```json
+{
+  "exports": {
+    ".": ["-bad-specifier-", "./non-existent.js", "./existent.js"]
+  }
+}
+```
+
+Webpack 5.93.1+ will now throw an error since `non-existent.js` is not found while the previous behavior would have resolved to `existent.js`.
```
chenxsan, line 82:
```suggestion
Webpack 5.94.0+ will now throw an error since `non-existent.js` is not found while the previous behavior would have resolved to `existent.js`.
```

PR #7353, opened 2024-08-13
docs: Replace overkill regex with a much simpler one

askoufis:
The example given in the docs for `NormalModuleReplacementPlugin` uses quite an expensive regex for what it's trying to achieve. This exact regex was the source of a sizeable performance hit in one of our applications. I've updated it to be simpler and more performant.


chenxsan (approved):
lgtm. Thanks.

PR #7360, opened 2024-08-16
Fix `webpackExports` section header

matthias-ccri:
On the https://webpack.js.org/api/module-methods/ page, the `webpackExports` section was not the right format. It wasn't an actual heading, and as a consequence it didn't have an anchor link.


chenxsan (approved):
lgtm, thanks.

PR #7364, opened 2024-08-21
Mention end of  support for subpath mapping exports by Node > 17

fbessou:
Subpath mapping exports are no longer handled by Node.js. I removed it from the list of supported platforms.

See : https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V17.md#2021-10-19-version-1700-current-bethgriggs


chenxsan (commented):
Screenshot-yEeA9TPF@2x

Maybe format it like this?

chenxsan (approved):
lgtm, thanks :)

PR #7367, opened 2024-08-24
docs: add `Rule.assert`

snitin315:
Refers https://github.com/webpack/webpack.js.org/issues/7280

https://github.com/webpack/webpack/blob/09543e7d8e0e7dd1703207193bcc3c3252874636/schemas/WebpackOptions.json#L4400-L4406

### DIFF
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -722,6 +722,35 @@ The order of evaluation is as follows:
 2. [`rules`](#rulerules)
 3. [`oneOf`](#ruleoneof)
 
+## Rule.assert
+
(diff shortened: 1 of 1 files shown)


src/content/configuration/module.mdx
```diff
@@ -722,6 +722,35 @@ The order of evaluation is as follows:
 2. [`rules`](#rulerules)
 3. [`oneOf`](#ruleoneof)
 
+## Rule.assert
+
+A [`Condition`](#condition) taht allows you to match the import assertion of a dependency and apply specific rules based on the assertion type.
```
snitin315, line 727:
```suggestion
A [`Condition`](#condition) that allows you to match the import assertion of a dependency and apply specific rules based on the assertion type.
```

PR #7395, opened 2024-09-15
docs(plugins): replace a typo with a convenient word

Ahmed-Hakeem:
I believe that `caramelized` was just a typo


chenxsan (approved):
Thanks.

PR #7397, opened 2024-09-16
Update entry-context.mdx

cengizilhan:
multiple js config entry added

the technical document in which you wanted to determine the input and output names of different files was not included. That's why I added this.

### DIFF
--- a/src/content/configuration/entry-context.mdx
+++ b/src/content/configuration/entry-context.mdx
@@ -175,6 +175,19 @@ module.exports = {
 
 For example: you can use dynamic entries to get the actual entries from an external source (remote server, file system content or database):
 
+### Multiple Entry Points in Different Folders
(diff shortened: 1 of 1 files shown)


src/content/configuration/entry-context.mdx
```diff
@@ -175,6 +175,19 @@ module.exports = {
 
 For example: you can use dynamic entries to get the actual entries from an external source (remote server, file system content or database):
 
+### Multiple Entry Points in Different Folders
+In some scenarios, you might want to organize your entry points into different folders for better structure and maintainability. Here's how you can define multiple entry points located in different folders:
+```
+module.exports = {
+  //...
+  entry: {
+    './js/one': './js/one.js/,
```
evenstensberg, line 184: key should not be a fp.

cengizilhan: You are right, entry keys are usually not file paths.  
Here it was intentional, just to make the custom location clearer in the example.

PR #7452, opened 2024-11-03
ci: use Node.js LTS version

snitin315:
`latest` resolves to v23.1.0 which has a regression https://github.com/nodejs/node/issues/55609
Failing our test & deploy actions.
Screenshot 2024-11-03 at 10 41 13 AM


chenxsan:
No wonder I can't figure out what's going on.

PR #7461, opened 2024-11-03
docs(plugins): update EnvironmentPlugin warning

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/7448

### DIFF
--- a/src/content/plugins/environment-plugin.mdx
+++ b/src/content/plugins/environment-plugin.mdx
@@ -27,7 +27,7 @@ new webpack.DefinePlugin({
 });
 ```
 
-T> Not specifying the environment variable raises an "`EnvironmentPlugin` - `${key}` environment variable is undefined" error.
+W> `EnvironmentPlugin` checks `process.env` for the specified variable. If it’s missing, it searches for a default value provided in the configuration. If neither the environment variable nor the default value is defined, an error will be thrown: **"`EnvironmentPlugin` - `${key}` environment variable is `undefined`"**.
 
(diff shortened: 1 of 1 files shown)


fasiha:
I was hoping that this PR would include a note about the issue raised in https://github.com/webpack/webpack.js.org/issues/7448 where, if you use `process.env.SOMETHING` and that EnvironmentPlugin doesn't know about it, Webpack will happily ignore `process.env.SOMETHING` and include it in the output.

snitin315:
>`EnvironmentPlugin` checks `process.env` for the specified variable. 

@fasiha This is mentioned in the warning EnvironmentPlugin checks only for the variables specified within this plugin. Otherwise, no errors.

fasiha:
@snitin315 thanks for the note! Do you mean this warning?:

> W> `EnvironmentPlugin` checks `process.env` for the specified variable. If it’s missing, it searches for a default value provided in the configuration. If neither the environment variable nor the default value is defined, an error will be thrown: **"`EnvironmentPlugin` - `${key}` environment variable is `undefined`"**.

Re-reading this, I don't think it makes clear that the plugin ignores usages of `process.env` that it doesn't know about? So perhaps we could add a line to this paragraph, something like what you said:

> `EnvironmentPlugin` checks only for variables specified by this plugin. Using `process.env` with unspecified variables results in those usages being ignored: for example, `process.env.UNKNOWN` will be left intact.

snitin315:
@fasiha sure, feel free to send a PR

PR #7490, opened 2024-12-08
docs(module): add `module.parser.css.import` option

snitin315:
Refers https://github.com/webpack/webpack.js.org/issues/7449

### DIFF
--- a/src/content/configuration/module.mdx
+++ b/src/content/configuration/module.mdx
@@ -262,6 +262,9 @@ module.exports = {
       css: {
         // Parser options for css modules
 
+        // Enable/disable `@import` at-rules handling, available since webpack 5.97.0
+        // type: boolean
+        import: true,
         // Use ES modules named export for css exports, available since webpack 5.90.0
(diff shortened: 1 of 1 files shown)


src/content/configuration/module.mdx
```diff
@@ -281,9 +284,65 @@ module.exports = {
...
+  ```css
+  /* styles.css */
+  @import './reset-styles.css';
+
+  body {
+    background-color: red;
+  }
+  ```
+
+T> To filter specific imports, you can use Webpack's built-in [IgnorePlugin](plugins/ignore-plugin/). The [`filter` option](/loaders/css-loader/#filter), as available in `css-loader`, is not supported.
```
snitin315, line 341:
```suggestion
T> To filter specific imports, you can use Webpack's built-in [IgnorePlugin](/plugins/ignore-plugin/). The [`filter` option](/loaders/css-loader/#filter), as available in `css-loader`, is not supported.
```

PR #7516, opened 2025-01-11
docs(config): added `type :  "module"` to the package.json in getting started section

RajeevPullat:
added `type :  "module"` to the package.json in getting started section

The getting started doc's package.json was missing `type:"module"` setting. Because of that the users following the docs get error at the import statements.


snitin315 (changes requested):
The [getting started example](https://stackblitz.com/github/webpack/webpack.js.org/tree/main/examples/getting-started?file=webpack.config.js&terminal=) seems to be working fine even without `type: "module"`

PR #7517, opened 2025-01-11
chore(scripts): updated npm scripts that use rimraf and glob patterns to use `--glob` option to fix `Illegal characters in path` error

RajeevPullat:
updated npm scripts that use rimraf and glob patterns to use `--glob` option to fix `Illegal characters in path` error

If we are using rimraf from the command line and using glob patterns, we have to use `--glob` command line option. Otherwise, we will get a `Error: Illegal characters in path.` error. So, this is fixed with adding `--glob` option to npm scripts that use rimraf and glob patterns.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,9 @@
   },
   "scripts": {
     "clean-dist": "rimraf ./dist",
-    "clean-printable": "rimraf src/content/**/printable.mdx",
+    "clean-printable": "rimraf --glob src/content/**/printable.mdx",
     "preclean": "run-s clean-dist clean-printable",
-    "clean": "rimraf src/content/**/_*.mdx src/**/_*.json repositories/*.json",
+    "clean": "rimraf --glob src/content/**/_*.mdx src/**/_*.json repositories/*.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -25,9 +25,9 @@
   },
   "scripts": {
     "clean-dist": "rimraf ./dist",
-    "clean-printable": "rimraf src/content/**/printable.mdx",
+    "clean-printable": "rimraf --glob src/content/**/printable.mdx",
     "preclean": "run-s clean-dist clean-printable",
-    "clean": "rimraf src/content/**/_*.mdx src/**/_*.json repositories/*.json",
+    "clean": "rimraf --glob src/content/**/_*.mdx src/**/_*.json repositories/*.json",
```
snitin315, line 30: It works without glob as well.
Screenshot 2025-01-24 at 6 01 47 AM

snitin315:
Thanks for the PR, but I don't think this is required.

RajeevPullat:
image
There is issue when using with npm as above. Thats why I tried to create a PR. I am using Node Version 20.17.0 and NPM version 11.0.0

PR #7556, opened 2025-02-24
Update getting-started.mdx

prasadcm:
The latest version of npm init sets type to commonjs by default. This will fail when we run webpack


snitin315:
We need to improve the getting started guide in general given that we now have scaffolding too.

PR #7612, opened 2025-04-19
fix alignment of navigation icons

chenxsan:
Seems the tooltip introduced have caused a alignment problem:

Screenshot-ddJ3pZL7@2x


snitin315 (approved):
Thanks!

Issue #7620, opened 2025-04-29
Document `errorCause` and `errorErrors`


alexander-akait:
**Feature to document**

Ref: https://github.com/webpack/webpack/pull/19453

**Author(s)**

Any

**Additional information**

`errorCause` outputs `cause` of any `Error` in stats
`errorErrors` outputs `errors` of `AggregateError` in stats

By default `true` for all

Possible values: `"auto"`/`true`/`false`

[ ] I am willing to work on this issue and submit a pull request.

PR #7621, opened 2025-04-30
docs(configuration): add `stats.errorCause`

snitin315:
Fix https://github.com/webpack/webpack.js.org/issues/7620


alexander-akait (commented):
We need to add `errorErrors` too, type and default value are same

it will print `errors` for https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError

alexander-akait (approved):

PR #7652, opened 2025-08-31
docs(contributing): add instructions to generate github token

brunocroh:
While running the `build` and/or `fetch*` scripts, many requests are made to GitHub. If a user does not have `GITHUB_TOKEN` in their environment variables, these requests will fail due to GitHub API rate limits. Instructions have been added to reduce friction for new contributors.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -13,16 +13,20 @@ Start by ensuring that you have Node.js installed and forking the repository:
 - Fork the **webpack.js.org** repo from [the main repository][2].
 - `git clone <your-clone-url> && cd webpack.js.org`
 
-Some operations in the build process require a GitHub token. To create and export one:
-
-- Go to [GitHub Personal Access Tokens](https://github.com/settings/personal-access-tokens) and generate a new token with the repo and workflow scopes.
-- Export it in your terminal before running build or fetch commands:
+> To run the project, you must run scripts that download the website content from GitHub.
+> To avoid hitting rate limits, you'll need to use your GitHub token.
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTING.md
```diff
@@ -13,10 +13,15 @@ Start by ensuring that you have Node.js installed and forking the repository:
 - Fork the **webpack.js.org** repo from [the main repository][2].
 - `git clone <your-clone-url> && cd webpack.js.org`
 
+Some operations in the build process require a GitHub token. To create and export one:
```
evenstensberg: Mention the fetch scripts here. Also, I think we can do this instead:

`GITHUB_TOKEN=123 node fetch-path.js`

package.json
```diff
@@ -35,6 +35,7 @@
     "fetch": "run-p fetch:*",
```
evenstensberg, line 35: `fetch` already fetches all

brunocroh: It doesn’t run `fetch-repos`, and if you don’t run `fetch-repos` first, `fetch` fails.

I think this is the reason why the `build` runs `fetch-repos` first.

https://github.com/webpack/webpack.js.org/blob/d803b5444a27fb66cbd7a5eeba501b9c43a5b98b/package.json#L39

.github/CONTRIBUTING.md
brunocroh: I’m not sure if I understood it correctly. I’ve updated it, but please let me know if I misunderstood something.

package.json
evenstensberg: Hmm weird, but okay!

PR #7663, opened 2025-10-05
docs: update example <script> loading strategy

zowiebeha:
In the code examples, a `<script>` tag is
placed just before the ending `</body>` tag.
This causes the script's external resource to
download and execute only after the HTML has been
parsed up to that point.

For organization and faster script evaluation,
we can consolidate scripts in the `<head>`
and enable their defer attribute, which will
download the external resource in parallel with
HTML parsing. Once the HTML is parsed, the
deferred script will execute.

### DIFF
--- a/src/content/guides/getting-started.mdx
+++ b/src/content/guides/getting-started.mdx
@@ -201,7 +201,7 @@ Now, since we'll be bundling our scripts, we have to update our `index.html` fil
  </html>
 ```
 
-T> Deferred loading is an alternative to the above, where instead scripts are consolidated into the `<head>` and are given the `defer` attribute. This strategy downloads the external script resource in parallel with document parsing, and will execute the script after parsing has finished. This is in contrast to the above, in which the parser pauses to download and then execute the external resource.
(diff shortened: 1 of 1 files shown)


alexander-akait (commented):
defer is alternative loading, so let's add another one alternative example with defer, don't change currents

src/content/guides/getting-started.mdx
```diff
@@ -200,6 +201,8 @@ Now, since we'll be bundling our scripts, we have to update our `index.html` fil
  </html>
 ```
 
+T> Deferred loading is an alternative to the above, where instead scripts are consolidated into the `<head>` and are given the `defer` attribute. This strategy downloads the external script resource in parallel with document parsing, and will execute the script after parsing has finished. This is in contrast to the above, in which the parser pauses to download and then execute the external resource.
```
evenstensberg: great! Maybe you can add a link to a place that this is covered more in detail?

https://bitsofco.de/async-vs-defer/

evenstensberg:
Thank you for the contribution! 💙

PR #7679, opened 2025-10-09
chore: use correct link for mini-css-extract

evenstensberg:


evenstensberg:
wrong namespace, closing.

Issue #7737, opened 2026-01-05
Notification regarding version of documentation cannot be dismissed


rjgotten:
The notification alerting the end-user they're looking at the Webpack v5 documentation rather than the v4 documentation cannot be dismissed. There's no explicit button to dismiss it. It keeps reappearing on every page.
It actually **blocks** essential parts of the pages.

[EDIT]
A close button is rendered on Chrome.
NOT on Firefox.

(Also `<div role="button">` is not an actual button.)

pranjalisr:
Hi @rjgotten can you assign this to me. I would like to work on it by hiding it with custom CSS

alexander-akait:
Let's remove notifier at all, webpack@4 outdated

pranjalisr:
@alexander-akait Can you please assign this to me? I’d really like to work on its implementation.

evenstensberg:
Send a PR to remove the banner, no need to assign anyone here. First serve, first merge

alexander-akait:
Fixed by https://github.com/webpack/webpack.js.org/pull/7746

PR #7746, opened 2026-01-15
docs: remove non-dismissible documentation notification bar

pranjalisr:
**Summary**

This PR removes the global documentation notification bar that informs users they are viewing a different documentation version. Fixes issue #7737

The notification cannot be dismissed in Firefox, blocks essential parts of the page, and relies on non-semantic interaction patterns (div role="button"). While a close button is rendered in Chrome, it does not appear in Firefox, causing the banner to reappear on every page load and negatively impact usability and accessibility.

Given that the notification is rendered globally and provides no actionable value, the component has been turned into a no-op to fully remove the problematic UI without touching layout wiring or introducing browser-specific logic.

Fix / Docs

No.
This change removes a purely presentational UI component from the documentation site. There is no business logic or functional behavior to test, and the repository does not currently include tests covering documentation UI banners.

No.
The change only affects the documentation UI and does not impact Webpack functionality, APIs, or consumer applications.

No additional documentation is required. The notification has been removed entirely rather than replaced or modified.

### DIFF
--- a/src/components/NotificationBar/NotificationBar.jsx
+++ b/src/components/NotificationBar/NotificationBar.jsx
@@ -1,0 +1,36 @@
+import { useState, useEffect, lazy, Suspense } from 'react';
+import testLocalStorage from '../../utilities/test-local-storage';
+
+const MessageBar = lazy(() => import('./MessageBar'));
+
+export const version = '3';
+export const localStorageIsEnabled = testLocalStorage() !== false;
+
+const barDismissed = () => {
+  if (localStorageIsEnabled) {
+    return localStorage.getItem('notification-dismissed') === version;
+  }
+  return false;
+};
+
+export default function NotificationBar() {
+  // hide the bar in the beginning
+  const [dismissed, setDismissed] = useState(true);
+  const onClose = () => {
+    setDismissed(true);
+  };
+  useEffect(() => {
+    // update dismissed value when component mounted
+    setDismissed(() => barDismissed());
+  }, []);
+
+  return (
+    <>
+      {dismissed === false ? (
+        <Suspense fallback={<div />}>
+          <MessageBar onClose={onClose} />
+        </Suspense>
+      ) : undefined}
+    </>
+  );
+}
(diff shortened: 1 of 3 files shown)


src/content/guides/web-workers.mdx
```diff
@@ -87,4 +87,22 @@ import { Worker } from 'worker_threads';
...
+
+```js
+// worker.js
+
+// Define the public path dynamically before loading other modules
+/* eslint-disable camelcase */
+__webpack_public_path__ = self.location.origin + '/assets/';
+
+// Then import or execute the rest of your worker logic
+importScripts('deep-thought.js');
```
alexander-akait: Avoid these changes

src/components/NotificationBar/NotificationBar.jsx
```diff
@@ -1,36 +1,3 @@
...
-    <>
-      {dismissed === false ? (
-        <Suspense fallback={<div />}>
-          <MessageBar onClose={onClose} />
-        </Suspense>
-      ) : undefined}
-    </>
-  );
+  return null;
 }
```
alexander-akait: Why we need this component at all?

src/content/guides/web-workers.mdx
pranjalisr: Thanks for the review!
These lines were carried over from my earlier PR. I’ve removed them from this one. Please let me know if further adjustments are needed.

src/components/NotificationBar/NotificationBar.jsx
pranjalisr: Agreed — it’s no longer needed. I’ve removed the NotificationBar completely along with all related files and references.  The notifier is no longer part of the codebase at all. Please take another look and let me know if any other change is required.

src/components/Site/Site.jsx
```diff
@@ -294,7 +293,7 @@ function Site(props) {
         </Route>
       </Routes>
       <Footer />
-      <NotificationBar />
```
alexander-akait, line 297: let's keep `NotificationBar` component for future notification (like webpack@6), just remove `<NotificationBar />`, like you done

pranjalisr: Thanks! I've restored the NotificationBar component so it remains available for future notifications (e.g., webpack@6), and only removed the <NotificationBar /> usage from Site.jsx as mentioned.
Let me know if you want any other adjustments!

PR #7747, opened 2026-01-16
fix: replace notification dismiss div with semantic button (#7737)

Kritank07:
Problem
The notification alerting users about the documentation version cannot be dismissed in certain browsers, specifically Firefox. This is caused by using a non-semantic <div> with role="button" for the dismissal logic. This implementation also creates accessibility barriers, as it does not follow standard HTML practices for interactive elements.

Solution
Updated the dismissal element in the NotificationBar component from a <div> to a semantic <button type="button">.

Maintained existing CSS classes and the onClick handler to ensure visual and functional consistency.

Ensured the CloseIcon remains correctly nested within the new button element for proper click targeting.

### DIFF
--- a/src/components/NotificationBar/MessageBar.jsx
+++ b/src/components/NotificationBar/MessageBar.jsx
@@ -35,6 +35,7 @@ export default function MessageBar(props) {
             <button
               type="button"
               className="px-20 self-stretch inline-flex items-center cursor-pointer"
+              style={{ background: 'none', border: 'none'}}
               onClick={close}
             >
               <CloseIcon


evenstensberg:
This doesnt look right.

# Expected
Screenshot 2026-01-18 at 03 45 07

# Recieved
Screenshot 2026-01-18 at 03 44 31

Kritank07:
I have updated the dismiss element to a semantic <button> to fix the Firefox issue. I also added a CSS reset (background: none, border: none) to ensure the visual appearance matches the original design exactly. Please check the latest commit."

evenstensberg:
I've already approved it 😄

Kritank07:
Thank you for the approval! I've confirmed that all checks have passed and the unrelated changes have been separated into their own branch. Is there anything else needed before this can be merged?

PR #7755, opened 2026-01-20
Fix service worker installation failure on Firefox

### LINKED ISSUE #7738: Service worker fails to install/register on Firefox
Service worker fails to install/register on Firefox:

Image

pranjalisr:
**Summary**
Fixes #7738 
Fix SW install failure in Firefox by replacing cache.addAll() with Promise.allSettled() to prevent installation from failing due to individual precache errors. Also deduplicates manifest URLs before caching.

---

Code change

---

No.  

---

No. 

---

No documentation changes are required.

### DIFF
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -262,7 +262,6 @@ function Site(props) {
         />
         <meta name="mobile-web-app-capable" content="yes" />
         <link rel="icon" sizes="192x192" href="/icon_192x192.png" />
-        <link rel="icon" sizes="512x512" href="/icon_512x512.png" />
         <meta name="apple-mobile-web-app-capable" content="yes" />
         <meta name="apple-mobile-web-app-status-bar-style" content="black" />
         <meta name="apple-mobile-web-app-title" content="webpack" />
--- a/src/components/Site/Site.jsx
+++ b/src/components/Site/Site.jsx
@@ -262,6 +262,7 @@ function Site(props) {
         />
         <meta name="mobile-web-app-capable" content="yes" />
         <link rel="icon" sizes="192x192" href="/icon_192x192.png" />
+        <link rel="icon" sizes="512x512" href="/icon_512x512.png" />
         <meta name="apple-mobile-web-app-capable" content="yes" />
         <meta name="apple-mobile-web-app-status-bar-style" content="black" />
         <meta name="apple-mobile-web-app-title" content="webpack" />
--- a/src/sw.js
+++ b/src/sw.js
@@ -58,7 +58,6 @@ self.addEventListener("activate", (event) => {
     ),
   );
 });
-
 registerRoute(
   ({ url }) => manifestURLs.includes(url.href),
   new NetworkFirst({
--- a/src/sw.js
+++ b/src/sw.js
@@ -41,7 +41,6 @@ self.addEventListener("install", (event) => {
     })(),
   );
 });
-
 self.addEventListener("activate", (event) => {
   // - [x] clean up outdated runtime cache
   event.waitUntil(
--- a/src/sw.js
+++ b/src/sw.js
@@ -88,10 +88,7 @@ setDefaultHandler(new NetworkOnly());
 
 // fallback to app-shell for document request
 setCatchHandler(({ event }) => {
-  switch (event.request.destination) {
-    case "document":
-      return caches.match("/app-shell/index.html");
-    default:
-      return fetch(event.request);
+  if (event.request.destination === "document") {
+    return caches.match("/app-shell/index.html");
   }
 });
--- a/src/sw.js
+++ b/src/sw.js
@@ -88,7 +88,10 @@ setDefaultHandler(new NetworkOnly());
 
 // fallback to app-shell for document request
 setCatchHandler(({ event }) => {
-  if (event.request.destination === "document") {
-    return caches.match("/app-shell/index.html");
+  switch (event.request.destination) {
+    case "document":
+      return caches.match("/app-shell/index.html");
+    default:
+      return Response.error();
   }
 });
--- a/src/sw.js
+++ b/src/sw.js
@@ -2,7 +2,6 @@
 import { CacheableResponsePlugin } from "workbox-cacheable-response";
 import { cacheNames } from "workbox-core";
 import { ExpirationPlugin } from "workbox-expiration";
-
 import {
   registerRoute,
   setCatchHandler,
--- a/src/sw.js
+++ b/src/sw.js
@@ -44,7 +44,6 @@ self.addEventListener("install", (event) => {
 self.addEventListener("activate", (event) => {
   // - [x] clean up outdated runtime cache
   event.waitUntil(
-    // clean up those who are not listed in manifestURLs
     caches.open(cacheName).then((cache) =>
       // clean up those who are not listed in manifestURLs
       cache
@@ -93,6 +92,6 @@ setCatchHandler(({ event }) => {
     case "document":
       return caches.match("/app-shell/index.html");
     default:
-      return Response.error();
+      return fetch(event.request);
   }
 });
--- a/src/sw.js
+++ b/src/sw.js
@@ -44,6 +44,7 @@ self.addEventListener("install", (event) => {
 self.addEventListener("activate", (event) => {
   // - [x] clean up outdated runtime cache
   event.waitUntil(
+     // clean up those who are not listed in manifestURLs
     caches.open(cacheName).then((cache) =>
       // clean up those who are not listed in manifestURLs
       cache
--- a/src/sw.js
+++ b/src/sw.js
@@ -2,6 +2,7 @@
 import { CacheableResponsePlugin } from "workbox-cacheable-response";
 import { cacheNames } from "workbox-core";
 import { ExpirationPlugin } from "workbox-expiration";
+
 import {
   registerRoute,
   setCatchHandler,
@@ -12,7 +13,6 @@ import {
   NetworkOnly,
   StaleWhileRevalidate,
 } from "workbox-strategies";
-/* eslint-enable import/no-extraneous-dependencies */
 
 const cacheName = cacheNames.runtime;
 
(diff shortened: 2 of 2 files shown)


pranjalisr:
@alexander-akait Just following up to kindly request a review of this PR when you have availability.
Thank you.

alexander-akait:
@evenstensberg Can you look at this and test, I think we can merge it

src/sw.js
```diff
@@ -31,26 +31,37 @@ const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
...
   );
 });
-
 self.addEventListener("activate", (event) => {
-  // - [x] clean up outdated runtime cache
   event.waitUntil(
-    caches.open(cacheName).then((cache) => {
-      // clean up those who are not listed in manifestURLs
+    caches.open(cacheName).then((cache) =>
       cache.keys().then((keys) => {
```
evenstensberg: dont remove the comments please

pranjalisr: I have added back the comments.

pranjalisr: Re-adding the comments causes the lint/test checks to fail. Is there a recommended way to retain the comments without breaking the rules?

pranjalisr: Fixed

src/sw.js
```diff
@@ -31,26 +31,38 @@ const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
...
+    })(),
   );
 });
-
 self.addEventListener("activate", (event) => {
-  // - [x] clean up outdated runtime cache
   event.waitUntil(
-    caches.open(cacheName).then((cache) => {
-      // clean up those who are not listed in manifestURLs
+    // clean up those who are not listed in manifestURLs
```
evenstensberg: comments should be where they were earlier so the diff is good.

src/sw.js
```diff
@@ -31,26 +31,38 @@ const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
...
+            const response = await fetch(url, { cache: "no-store" });
+            if (response && response.ok) {
+              await cache.put(url, response);
+            }
+          } catch (err) {
+            console.warn("[sw] Failed to precache:", url, err);
+          }
+        }),
+      );
+    })(),
```
alexander-akait, line 41: On each manifest item we generate fetch, don't look like a good idea, it kills the idea of cache, looks like AI generate without understanding

pranjalisr: My intention was specifically to prevent Firefox from aborting the install when a single precache request fails. I agree the manual fetch approach isn’t ideal for precaching. I am happy to adjust it.
What pattern would you recommend here?

evenstensberg: We need to minimize amount of fetches in the service worker script (this one), so maybe we can re-fetch the failed script instead of all? Maybe that works?

pranjalisr: One approach could be to keep cache.addAll(manifestURLs) as it is & identify the specific URL that fails in Firefox and exclude it from the precache list (or handle it separately), instead of changing the entire install logic.

Does that sound good, or would you suggest a different pattern here?

evenstensberg: sure

pranjalisr:
Tested in Firefox using production build (yarn build + yarn serve-dist). Service worker installs successfully and no longer aborts on install.
Confirmed in Application → Service Workers that sw.js is running.

Screenshot 2026-02-13 at 16 43 13






@evenstensberg Can you please check this? If this sounds Ok then I will push the code changes.

src/sw.js
```diff
@@ -31,26 +31,38 @@ const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
...
+
+      await Promise.all(
+        manifestURLs.map(async (url) => {
+          try {
+            const response = await fetch(url, { cache: "no-store" });
+            if (response && response.ok) {
+              await cache.put(url, response);
+            }
+          } catch (err) {
+            console.warn("[sw] Failed to precache:", url, err);
```
ovflowd: nit: does it even matter console logging it?

src/sw.js
```diff
@@ -31,26 +31,38 @@ const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
...
 self.addEventListener("install", (event) => {
   event.waitUntil(
-    caches.open(cacheName).then((cache) => cache.addAll(manifestURLs)),
+    (async () => {
+      const cache = await caches.open(cacheName);
+
+      await Promise.all(
+        manifestURLs.map(async (url) => {
+          try {
+            const response = await fetch(url, { cache: "no-store" });
```
ovflowd: wwon't this be repeatedly fetching an existing resource? I think that's not really good.

src/sw.js
```diff
@@ -31,26 +31,38 @@ const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
 });
 self.addEventListener("install", (event) => {
   event.waitUntil(
```
ovflowd, line 36: I'd refactor this to be something like this:

```tsx
caches.open(cacheName).then((cache) => 
  Promise.allSettled(manifestURLs.map(cache.add))
```

ovflowd: cache.add per MDN docs https://developer.mozilla.org/en-US/docs/Web/API/Cache/add functionallity does the same. So the whole code can be cleaned.

The try/catch approach is also not ideal.

ovflowd (changes requested):
Appreciate the contribution. Code is not _ideal_, made a few suggestions :)

src/sw.js
```diff
@@ -45,10 +45,12 @@ self.addEventListener("activate", (event) => {
       const keys = await cache.keys();
       await Promise.all(
         keys.map((request) =>
```
ovflowd: This can be refactored:

```tsx
for (request in keys) {
  if (manifestURLs.includes(request.url)) {
    await cache.delete(request);
  }
}
```

This is cleaner, easier to understand + we're not really using the Promise return value. Creating empty Promise.resolve() adds overhehead.

pranjalisr: Ok. I will make changes run locally and the update.

pranjalisr: I’ve refactored the activate cleanup to use a simple for loop and removed the unnecessary Promise.resolve().

src/sw.js
```diff
@@ -25,32 +25,32 @@ const otherManifest = [
...
+const EXCLUDED_PRECACHE_URLS = new Set([
+  new URL("/icon_512x512.png", self.location).href,
+]);
+const manifestURLs = [...manifest, ...otherManifest]
+  .map((entry) => new URL(entry.url, self.location).href)
+  .filter((href) => !EXCLUDED_PRECACHE_URLS.has(href));
+
 self.addEventListener("install", (event) => {
   event.waitUntil(
     caches.open(cacheName).then((cache) => cache.addAll(manifestURLs)),
```
ovflowd: I think you undid your previous changes... That actually fix the issue, with Promise.allSettled ;)

pranjalisr: I'm sorry. I accidentally reverted the install behavior while refactoring.
I’ll restore the fix. I’ll push an update shortly.

pranjalisr:
Made changes and re-ran locally on Firefox. It's running
Please review code if anything needs to be changed will do that accordingly

Screenshot 2026-02-15 at 22 38 02

src/sw.js
```diff
@@ -25,34 +25,50 @@ const otherManifest = [
...
+
+const optionalManifestURLs = [...EXCLUDED_PRECACHE_URLS];
+
 self.addEventListener("install", (event) => {
   event.waitUntil(
-    caches.open(cacheName).then((cache) => cache.addAll(manifestURLs)),
+    (async () => {
+      const cache = await caches.open(cacheName);
+
+      // Fail install only if "required" assets fail
```
ovflowd: this line isn't really needed, and you don't really need a hardcoded EXCLUDED_PRECACHE_URLS. Simply do 

```tsx
await Promise.allSettled(
  allManifestURLs.map((url) => cache.add(url)),
);
```

This will already ignore failures gracefully.

pranjalisr: Made code changes. Checked locally, running fine. Please review

ovflowd (dismissed):
LGTM! Thanks for the contrubution :)

src/sw.js
```diff
@@ -25,34 +25,37 @@ const otherManifest = [
     url: "/app-shell/index.html",
   },
 ];
-const manifestURLs = [...manifest, ...otherManifest].map((entry) => {
-  const url = new URL(entry.url, self.location);
-  return url.href;
-});
+
+const allManifestURLs = [...manifest, ...otherManifest].map(
```
evenstensberg: doubt it is a problem but lets filter out so we dont have duplicate urls before we access the cache

pranjalisr: What do you suggest? Should I make changes in allManifestURLs ?

evenstensberg: yes just filter out dupes, easy to do if you convert it to a set and back to an array.

pranjalisr: Made code changes as suggested. Checked locally, working fine on firefox browser.

alexander-akait (commented):
Please rebase

/cc @evenstensberg please take a look

evenstensberg:
I'll review once rebased

pranjalisr:
I will rebase within 24hrs, travelling today so don't have access to laptop for today

pranjalisr:
@evenstensberg please review. Working fine locally on firefox browser
</br>
Screenshot 2026-03-18 at 19 19 22
</br>
Screenshot 2026-03-18 at 19 21 19

src/sw.js
```diff
@@ -87,6 +92,6 @@ setCatchHandler(({ event }) => {
     case "document":
       return caches.match("/app-shell/index.html");
     default:
-      return Response.error();
+      return fetch(event.request);
```
evenstensberg: I dont think this is smart. Lets say something is wrong with your computer and it doesnt have a good network. You will always re-fetch until it suceeds. will be very expensive an circular-loop.

pranjalisr: removed the fetch(event.request) fallback.
I also removed the non-essential 512x512 favicon that was causing the Firefox error, tested locally and the service worker now installs and runs cleanly.
Please review

src/components/Site/Site.jsx
```diff
@@ -262,7 +262,6 @@ function Site(props) {
         />
         <meta name="mobile-web-app-capable" content="yes" />
         <link rel="icon" sizes="192x192" href="/icon_192x192.png" />
-        <link rel="icon" sizes="512x512" href="/icon_512x512.png" />
```
evenstensberg, line 265: dont remove

pranjalisr: then it's causing error in network tab </br>
Screenshot 2026-03-29 at 00 44 00

evenstensberg: it means that its missing in our dist folder, we need to add it back.

pranjalisr: restored the icon. Running fine locally, please review 
Screenshot 2026-03-29 at 15 05 25

src/sw.js
```diff
@@ -83,10 +88,7 @@ setDefaultHandler(new NetworkOnly());
 
 // fallback to app-shell for document request
 setCatchHandler(({ event }) => {
-  switch (event.request.destination) {
-    case "document":
-      return caches.match("/app-shell/index.html");
-    default:
-      return Response.error();
```
evenstensberg, line 90: dont remove the error handling please

alexander-akait:
Fixed in the main, anyway thanks for the PR

PR #7767, opened 2026-01-28
docs: clarify array entry behavior with examples

### LINKED ISSUE #3788: The behavior of `entry is array` should be described more specifically
## Feature request

The behavior of `entry is array` should be described more specifically

image

**What is motivation or use case for adding/changing the behavior?**

When the user sees the document, it should be clear how to use it and what the usage scenario is.

**How should this be implemented in your opinion?**

I think the document should be like this:

## Array Syntax

```js
module.exports = {
  entry: {
    main: ['react', 'react-dom']
  }
};
```

What happens when you pass an array to entry? Passing an array of file paths to the entry property creates what is known as a "multi-main entry". This is useful when you would like to inject multiple dependent files together and graph their dependencies into one "chunk".

another usage scenario is when you want inject a pollyfill or some special code, you can use array entry instead of import by hand.

example 1:

```js
module.exports = {
  entry: {
    // have not to do `import('@babel/polyfill')` in entry.js file
    main: ['@babel/polyfill', './entry.js']
  }
};
```

example 2:

Common development environment configuration

```js
module.exports = {
  entry: {
    // inject hot reload code
    main: ['webpack-dev-server/client?https://localhost:8080/', 'webpack/hot/dev-server', './entry.js']
  }
};
```

**Are you willing to work on this yourself?**

yes, but i do not good at english very well, i need help😄

alwaysalearner1234:
This PR improves the documentation for array entry usage by explaining
that it creates a single entry chunk and by adding common real-world
examples such as polyfills and development-only scripts.

Fixes #3788

### DIFF
--- a/src/content/concepts/entry-points.mdx
+++ b/src/content/concepts/entry-points.mdx
@@ -73,6 +73,7 @@ Single Entry Syntax is a great choice when you are looking to quickly set up a w
 Usage: `entry: { <entryChunkName> string | [string] } | {}`
 
 **webpack.config.js**
+
 ```javascript
 module.exports = {
   entry: {


src/content/concepts/entry-points.mdx
```diff
@@ -211,4 +227,4 @@ module.exports = {
 
 **Why?** In a multi-page application, the server is going to fetch a new HTML document for you. The page reloads this new document and assets are redownloaded. However, this gives us the unique opportunity to do things like using [`optimization.splitChunks`](/configuration/optimization/#optimizationsplitchunks) to create bundles of shared application code between each page. Multi-page applications that reuse a lot of code/modules between entry points can greatly benefit from these techniques, as the number of entry points increases.
 
-T> As a rule of thumb: Use exactly one entry point for each HTML document. See the issue [described here](https://bundlers.tooling.report/code-splitting/multi-entry/#webpack) for more details.
+
```
alexander-akait: Please return this line

alwaysalearner1234: Thanks for the feedback! I’ve added the missing webpack.config.js code example for the array-based entry to keep the documentation consistent.

alwaysalearner1234: @alexander-akait  
Formatting issues have been fixed and the requested line has been restored.  
Please let me know if anything else needs adjustment.

alexander-akait:
Please fix lint

alwaysalearner1234:
Thank you for the review and feedback. Glad to see this merged!

PR #7777, opened 2026-02-04
docs(blog): add details on `tsconfig.json` alias resolution and support for `import.defer()`

bjohansebas:
**Summary**

These last two features were added before I left home and before the release went out, so they were missing. @webpack/tsc I’m not sure who manages the social media here, but it would be pretty cool to have this shared officially on the webpack accounts.

### DIFF
--- a/src/content/blog/webpack-5-105.mdx
+++ b/src/content/blog/webpack-5-105.mdx
@@ -47,7 +47,7 @@ see the [Webpack Security Policy](https://github.com/webpack/webpack/security/po
 
 Module resolution can be configured to read `compilerOptions.baseUrl` and `compilerOptions.paths` from `tsconfig.json` and apply those aliases when resolving imports. As a result, it is no longer necessary to install and configure the external `tsconfig-paths-webpack-plugin` just to make TypeScript path aliases work during bundling.
 
-In the webpack configuration, the `resolve.resolver.tsconfig` option is used to specify which tsconfig file should be used as the source (which is also useful in monorepos or setups with multiple tsconfig files):
+In the webpack configuration, the `resolve.tsconfig` option is used to specify which tsconfig file should be used as the source (which is also useful in monorepos or setups with multiple tsconfig files):
 
 ```js
 // webpack.config.js


src/content/blog/webpack-5-105.mdx
```diff
@@ -90,6 +130,15 @@ if (React.useId) {
 }
 ```
 
+## Support `import.defer()` for Context Module
+
+Webpack now supports `import.defer()` even when the import is a context module (that is, when the import path is built dynamically and webpack needs to include a set of possible files). In practice, this makes it much easier to defer the loading of a module selected at runtime, even when that module comes from a dynamic expression pointing to a directory. This enables patterns like “choose between a.js or b.js and load it only when needed”, with webpack generating the appropriate context so that import.defer() works correctly.
```
bjohansebas, line 133: @alexander-akait I’m not entirely sure, but that’s how I understood it from the context of the PR.

src/content/blog/webpack-5-105.mdx
```diff
@@ -41,6 +43,44 @@ For more information on reporting security vulnerabilities,
...
+
+In the webpack configuration, the `resolve.resolver.tsconfig` option is used to specify which tsconfig file should be used as the source (which is also useful in monorepos or setups with multiple tsconfig files):
+
+```js
+// webpack.config.js
+module.exports = {
+  resolve: {
+    resolver: {
+      tsconfig: "./tsconfig.json", // Can be `false` (disabled), `true` (use the default tsconfig.json), a path to a `tsconfig.json` file, or an object with configFile and references options.
+    },
```
bjohansebas:
```suggestion    
    tsconfig: "./tsconfig.json", // Can be `false` (disabled), `true` (use the default tsconfig.json), a path to a `tsconfig.json` file, or an object with configFile and references options.
```

src/content/blog/webpack-5-105.mdx
```diff
@@ -41,6 +43,42 @@ For more information on reporting security vulnerabilities,
 see the [Webpack Security Policy](https://github.com/webpack/webpack/security/policy)
 .
 
+## `tsconfig.json` Alias Resolution Without Requiring plugins
+
+Module resolution can be configured to read `compilerOptions.baseUrl` and `compilerOptions.paths` from `tsconfig.json` and apply those aliases when resolving imports. As a result, it is no longer necessary to install and configure the external `tsconfig-paths-webpack-plugin` just to make TypeScript path aliases work during bundling.
+
+In the webpack configuration, the `resolve.resolver.tsconfig` option is used to specify which tsconfig file should be used as the source (which is also useful in monorepos or setups with multiple tsconfig files):
```
bjohansebas:
```suggestion
In the webpack configuration, the `resolve.tsconfig` option is used to specify which tsconfig file should be used as the source (which is also useful in monorepos or setups with multiple tsconfig files):
```

src/content/blog/webpack-5-105.mdx
bjohansebas: Tomorrow I’ll be out all day, so feel free to modify the text. If you need me, I’ll be keeping an eye on Discord.

src/content/blog/webpack-5-105.mdx
```diff
@@ -90,6 +128,19 @@ if (React.useId) {
 }
 ```
 
+## Support `import.defer()` for Context Module
+
+Webpack now supports `import.defer()` even when the import is a context module (that is, when the import path is built dynamically and webpack needs to include a set of possible files). In practice, this makes it much easier to defer the loading of a module selected at runtime, even when that module comes from a dynamic expression pointing to a directory. This enables patterns like “choose between a.js or b.js and load it only when needed”, with webpack generating the appropriate context so that import.defer() works correctly.
+
+{/* eslint-disable */}
+
+```text
```
bjohansebas, line 137: Because ESLint throws an error right away for import.defer(), I mentioned it in the text as a workaround. As far as I know, there isn’t a rule that can be disabled for this

PR #7787, opened 2026-02-07
style: add hover animation to copy button in code block

TusharThakur04:
**Summary**

This PR updates the copy button behavior in code blocks so that it becomes visible only when the user hovers over the block

style

no, this PR only changes the css properties

no

**Preview**

[Screencast from 2026-02-07 19-44-15.webm](https://github.com/user-attachments/assets/78097ab6-50a1-4d5a-b6e8-67adc58aeaf0)

### DIFF
--- a/src/components/CodeBlockWithCopy/CodeBlockWithCopy.scss
+++ b/src/components/CodeBlockWithCopy/CodeBlockWithCopy.scss
@@ -71,6 +71,7 @@
 
   &:focus {
     outline: none;
+    opacity: 1;
   }
 
   &:active {


TusharThakur04:
@evenstensberg can you review this PR

Brennvo (changes requested):
This has accessibility implications. Now, there is no indication of what is receiving focus when these buttons are navigated to via keyboard.

To fix this, a `:focus` pseudo-class needs to be added to the button styles.

alexander-akait:
@evenstensberg Can you review and look at this?

evenstensberg:
Yes

Issue #7789, opened 2026-02-07
[Fix]: 'hastscript' module has no default export


TusharThakur04:
The 'hastscript' module does not provide a default export. 
Importing it as a default import causes an error when running **yarn start**.

### How to reproduce

run command **yarn start** after completing the project setup

<img width="960" height="619" alt="Image" src="https://github.com/user-attachments/assets/d3edeab4-925f-4640-aee5-fa00f0d06d11" />

TusharThakur04:
@alexander-akait  should i make a PR?

alexander-akait:
PR welcome

PR #7793, opened 2026-02-08
fix: handle hastscript no-default-export + unblock fresh install/start

pranjalisr:
Fixing #7789
**Summary**

This PR updates webpack.common.mjs to resolve hastscript using createRequire instead of a named ESM import.
Previously, CI failed because hastscript did not provide a named h export in certain environments. Using require("hastscript") ensures compatibility with both CommonJS and ESM builds of the package.

Fix.

No.

No.

**No documentation changes are required.**
N/A

**Verification:**
Verified locally. Running fine

Screenshot 2026-02-08 at 19 54 57

### DIFF
--- a/webpack.common.mjs
+++ b/webpack.common.mjs
@@ -1,6 +1,8 @@
 import { createRequire } from "node:module";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { h } from "hastscript";
 import MiniCssExtractPlugin from "mini-css-extract-plugin";
 import autolink from "remark-autolink-headings";
 import remarkEmoji from "remark-emoji";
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -156,7 +156,12 @@
     "webpack-dev-server": "^5.2.3",
     "webpack-merge": "^6.0.1",
     "workbox-webpack-plugin": "^7.4.0",
-    "yarn-deduplicate": "^6.0.2"
+    "yarn-deduplicate": "^6.0.2",
+    "workbox-core": "^7.4.0",
+    "workbox-routing": "^7.4.0",
+    "workbox-strategies": "^7.4.0",
+    "workbox-cacheable-response": "^7.4.0",
+    "workbox-expiration": "^7.4.0"
```
alexander-akait: Why you add them?

pranjalisr: These were added while resolving the Workbox import issues locally. After reworking the sw.js imports, they’re not strictly necessary as direct dependencies anymore since workbox-webpack-plugin already brings them in.

I can remove them to avoid unnecessary dependency additions and keep the change focused on the hastscript fix.

pranjalisr: I have removed them successfully.

TusharThakur04: i dont know but seems ai generated

TusharThakur04:
This PR solves these issues have a look #7833

pranjalisr:
> We don't need such solution

Ok. I will make relevant changes and push a new code soon.

pranjalisr:
@alexander-akait I need one advice here please. 

I pushed a simple code change which was working fine locally but is causing CI issues here. So my another approach would be to use createRequire to load hastscript via CommonJS inside the ESM config and normalize its export by checking whether it’s exported directly as a function or under hashscript. If neither is found, it will throw an error

Is this ok or do you suggest something ? I just need 1 hint and I will apply it in clean way.

TusharThakur04:
@pranjalisr this issue is because of yarn berry version, there is **no need to make a fix in code**.
migrate to yarn classic i.e. 1.x versions and pull the latest yarn.lock from upstream,
that would resolve this issue.
you can go through the discussion happened in this PR - #7833

alexander-akait:
Let's close

PR #7798, opened 2026-02-08
docs(guides): add tip for multiple entries in output management

Brennvo:
**Summary**

When I was going through the [Output Management](https://webpack.js.org/guides/output-management/) guide, I started reading the bundle files (specifically `print.bundle.js`) to see what webpack was outputting. Upon inspection, I found a side-effect free IIFE that caught be by surprise, and made me question my understanding of the guide so far.

Hence, I figured it might be helpful to have a new "Tip" section for any readers like me who decided to dig deeper into the `/dist` folder.

However, I understand if this is not appropriate for the documentation, especially given its length. I originally took this as a personal note (to better grasp my own understanding), but I thought it might be worth suggesting it as a documentation entry.

Documentation enhancement.

No.

No.

N/A

### DIFF
--- a/src/content/guides/output-management.mdx
+++ b/src/content/guides/output-management.mdx
@@ -10,6 +10,7 @@ contributors:
   - AnayaDesign
   - chenxsan
   - snitin315
+  - Brennvo
 ---
 
 T> This guide extends on code examples found in the [`Asset Management`](/guides/asset-management) guide.


src/content/guides/output-management.mdx
```diff
@@ -133,6 +133,11 @@ We can see that webpack generates our `print.bundle.js` and `index.bundle.js` fi
...
 
+T> If you inspected `print.bundle.js`, you might have noticed that it did not contain the `printMe` function. Instead, it contained only an [IIFE](/concepts/why-webpack/#iifes---immediately-invoked-function-expressions) with no observable side effects.<br /><br />
+This is because in our current setup, webpack treats each entry point as a standalone program that runs when the bundle is loaded. Because `print.js` only exports a function and never calls `printMe()`, executing it as its own entry does not require any executable runtime code. Thus, [webpack is able to omit that code as an optimization](/guides/tree-shaking/) since it would have no observable effect.<br /><br />
+However, since `index.js` imports and uses `printMe`, webpack includes that function directly in `index.bundle.js`. Remember that webpack is a module _bundler_, so `index.bundle.js` contains all the code it needs to run rather than unresolved import statements.<br /><br />
+Multiple entry points are used here solely to demonstrate output management and the `HtmlWebpackPlugin`. In a real project, you would typically avoid creating a separate entry point for utility modules like this.
+
```
Brennvo, line 141: Here is a user friendly preview of this text (in Github Flavored Markdown):

> [!NOTE]
> If you inspected `print.bundle.js`, you might have noticed that it did not contain the `printMe` function. Instead, it contained only an [IIFE](https://webpack.js.org/concepts/why-webpack/#iifes---immediately-invoked-function-expressions) with no observable side effects.
>
> This is because in our current setup, webpack treats each entry point as a standalone program that runs when the bundle is loaded. Because `print.js` only exports a function and never calls `printMe()`, executing it as its own entry does not require any executable runtime code. Thus, [webpack is able to omit that code as an optimization](https://webpack.js.org/guides/tree-shaking/) since it would have no observable effect.
>
>However, since `index.js` imports and uses `printMe`, webpack includes that function directly in `index.bundle.js`. > Remember that webpack is a module _bundler_, so `index.bundle.js` contains all the code it needs to run rather than unresolved import statements.
>
> Multiple entry points are used here solely to demonstrate output management and the `HtmlWebpackPlugin`. In a real project, you would typically avoid creating a separate entry point for utility modules like this.

alexander-akait: let's also add an example where different entry points depend on each other to show how to solve it if you need it

Brennvo: @alexander-akait wouldn't this be covered in the [Code Splitting](https://webpack.js.org/guides/code-splitting/#prevent-duplication) guide for preventing duplication? If so, I can just mention it in this tip and link to there?

alexander-akait: Let's add a link to this section in this case, to  navigate users in such case :+1:

Brennvo: Sounds good! Just pushed an update.

alexander-akait (approved):

alexander-akait:
Thanks

PR #7825, opened 2026-02-19
fix(a11y): add main landmark to page content

pranjalisr:
**Summary**
Fixes Lighthouse a11y: “Document does not have a main landmark”
**Before**
Screenshot 2026-02-19 at 01 34 36

**After**
Screenshot 2026-02-19 at 01 51 04

minor code change

Not required

No

N/A


pranjalisr:
Most of the diff is indentation/Prettier reflow due to adding a main wrapper, no logic changes neither copy pasted

src/components/Page/Page.jsx
```diff
@@ -100,49 +100,51 @@ export default function Page(props) {
   }
   return (
     <section className="page">
```
alexander-akait: We should not use `section` here, only `main` due spec

pranjalisr: Replaced section with main. checked locally, it's working fine on browser. now only issue is of prettier which is failing CI. I can open another PR for that, after it merges I can rerun CI on this PR and it should pass.

pranjalisr: I have raised PR #7876 that will fix lint error in MessageBar.jsx

PR #7828, opened 2026-02-19
fix(a11y): removed links rely 

pranjalisr:
**Summary**

Fixed Links rely on color to be distinguishable issue in Lighthouse and increased accessibility to 94
Screenshot 2026-02-19 at 15 03 28

Code changes

No

No

N/A


pranjalisr:
@evenstensberg I tried every approach for this audit without using underline text-decoration, but nothing worked, I think this won't pass without underline, but that also won't look good.
Do you suggest any other approach

Issue #7803, opened 2026-02-09
[Fix]: fetch npm script


TusharThakur04:
running `fetch` script is failing 

### How to reproduce
   Run command `GITHUB_TOKEN=<your-token-here> yarn fetch-all`

<img width="860" height="229" alt="Image" src="https://github.com/user-attachments/assets/5fafd1b7-7591-46af-989a-81e5a2c49647" />

TusharThakur04:
@alexander-akait should i make the PR?

alexander-akait:
Pr welcome

PR #7804, opened 2026-02-09
fix: change 'fetch' script by replacing wildcard with explicit fetch …

TusharThakur04:
fixes #7803 

**Summary**
This PR changes `fetch` script by replacing wildcard with explicit fetch scripts

**What kind of change does this PR introduce?**
fix

**Did you add tests for your changes?**
no

**Does this PR introduce a breaking change?**
no

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

PR #7790, opened 2026-02-07
fix: replace default import with named import from 'hastscript' module

TusharThakur04:
**Summary**
This PR resolves the runtime error caused by importing 'hastscript' as a default export. The import has been updated to use the correct named export instead.

**What kind of change does this PR introduce?**
fix

**Did you add tests for your changes?**
no

**Does this PR introduce a breaking change?**
no

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

PR #7833, opened 2026-02-20
docs(contrib): specify to use only 1.x yarn version

TusharThakur04:
Fixes: #7803, #7789 
Closes: #7793, #7790, #7804

**Summary**
This PR specifies that project should be set up using `yarn classic v(1)` as `yarn berry (v2)/(v3)/(v4)` changes the yarn.lock file, don't executes the scripts properly, and produces runtime error.

docs

no

no

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,6 +10,10 @@ Start by ensuring that you have Node.js installed and forking the repository:
 
 - Install [Node.js][1] if you have not already.
 - Install [Yarn][13] if you have not already. Yarn is an alternative to npm, and it is required for building the site as we use `yarn` specific features like `resolutions` which are not available in other package managers like `pnpm`.
+
+  > This project uses **Yarn Classic (v1)**.
+  > Ensure you are using **Yarn 1.x** version.
+
 - Fork the **webpack.js.org** repo from [the main repository][2].
 - `git clone <your-clone-url> && cd webpack.js.org`
 


TusharThakur04:
@evenstensberg please review

.github/CONTRIBUTING.md
```diff
@@ -10,6 +10,10 @@ Start by ensuring that you have Node.js installed and forking the repository:
 
 - Install [Node.js][1] if you have not already.
 - Install [Yarn][13] if you have not already. Yarn is an alternative to npm, and it is required for building the site as we use `yarn` specific features like `resolutions` which are not available in other package managers like `pnpm`.
+
+  > This project uses **Yarn Classic (v1)**.
+  > Ensure you are using **Yarn 1.x** version.
```
ovflowd, line 15: Maybe add an external link to how to install Yarn 1? :)

TusharThakur04: sure

.github/CONTRIBUTING.md
```diff
@@ -10,6 +10,10 @@ Start by ensuring that you have Node.js installed and forking the repository:
 
 - Install [Node.js][1] if you have not already.
 - Install [Yarn][13] if you have not already. Yarn is an alternative to npm, and it is required for building the site as we use `yarn` specific features like `resolutions` which are not available in other package managers like `pnpm`.
```
TusharThakur04, line 12: actually its there, this link makes a user install yarn classic
there might be case someone has pre-installed berry like me :)
and just start contributing without ensuring the version

ovflowd:
> @evenstensberg please review

You don't need to do direct mentions, we get pinged with new PRs :)

divyeshagrawal:
Hi @ovflowd , would it make sense to specify packageManager in package.json to prevent accidental usage of other versions of Yarn?

That might help enforce consistency across contributors’ environments

Happy to open a follow-up PR if this sounds reasonable 🙂

Ref: https://nodejs.org/docs/latest-v20.x/api/packages.html#packagemanager

TusharThakur04:
I also thought of this idea earlier but it not recommended for production environment 
[stability index](https://nodejs.org/api/documentation.html#stability-index)

divyeshagrawal:
> I also thought of this idea earlier but it not recommended for production environment 
> [stability index](https://nodejs.org/api/documentation.html#stability-index)
> 

Thanks for the clarification, that makes sense.

Another option I have in mind is having a small pre-install task that fails when current version of yarn !== v1, purely to surface unsupported setups as soon as possible.

divyeshagrawal:
Just noticed webpack/webpack is already using packageManager. Curious if that makes a difference here, or if the contexts are different.

.github/CONTRIBUTING.md
alexander-akait: No need, we will migrate from yarn soon, so this word is not needed

TusharThakur04: ok, you can close the PRs and issue i have tagged
they were happening because of yarn berry versions

alexander-akait: yeah, let's close, we will migrate soon

PR #7831, opened 2026-02-19
clarify asset/bytes behavior and fix example

GreenUpOu:
**Summary**

The asset/bytes example was incorrect, as it treated the imported value as a
string, while asset/bytes modules actually export raw bytes (Uint8Array).
The example has been fixed to correctly decode the bytes before using them
as text.

Additionally, the documentation has been clarified to explicitly describe
the different output types of asset/source and asset/bytes modules, removing
ambiguity around the "as is" wording.

**What kind of change does this PR introduce?**

docs

**Did you add tests for your changes?**

No, this PR only updates documentation and examples.

**Does this PR introduce a breaking change?**

No.

**If relevant, what needs to be documented once your changes are merged or what have you already documented?**

The corrected asset/bytes example and the clarified descriptions of
asset/source and asset/bytes behavior are already documented in this PR.

PR #7834, opened 2026-02-20
Add GreenUpOu to asset-modules contributors list

GreenUpOu:
GreenUpOu contributor was missed in PR https://github.com/webpack/webpack.js.org/pull/7831


GreenUpOu:
@alexander-akait thank you very much for this opportunity and for your time!

PR #7871, opened 2026-02-24
refact(styles): remove repeated code block styles

TusharThakur04:
**Summary**
This PR removes unnecessary repeated styles for code block as they are already defined under [`Markdown.scss`](https://github.com/webpack/webpack.js.org/blob/main/src/components/Markdown/Markdown.scss)

styles

no

no

### DIFF
--- a/src/components/Markdown/Markdown.scss
+++ b/src/components/Markdown/Markdown.scss
@@ -359,13 +359,15 @@ $topHeightMobileWithBanner: $bannerHeight + $topHeightMobile;
   }
 
   pre {
-    background-color: rgba(238, 238, 238, 0.35);
-    background-color: getColor(elephant);
-    font-size: 13px;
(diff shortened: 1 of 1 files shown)


TusharThakur04:
in `CodeBlockWithCopy.jsx` the pre tag has been given the class name `code-block` so styles applied to that are basically getting applied to `pre` tag :
Screenshot from 2026-02-25 20-32-04

---
But in `Markdown.scss` styles have already been applied to `pre` tag: 
 
Screenshot from 2026-02-25 20-31-46

alexander-akait:
In this case we don't need `code-block` className anymore

alexander-akait:
I found that color and box-shadow are losing if we remove it

TusharThakur04:
@alexander-akait fixed

alexander-akait (approved):

PR #7887, opened 2026-02-27
fix(navbar): improve CSS alignment

Katari-8055:
### What does this PR do?
Fixes CSS alignment issues in the Navbar component.

### Why is this needed?
The current Navbar layout appears misaligned on certain screen sizes.

### Notes
No existing issue was found for this problem, so this PR directly addresses it.

### Before
Before

### After
After


alexander-akait:
Thanks

PR #7888, opened 2026-02-27
Fix table orientation in the comparison page (#2236)

karanrajsurya:
I have fixed the comparison table issue. People were unable to read the table as it was too long so I changed it to more readable form without changing any content. (Reverted the module changes and only pushed the table this time. Please let me know if anything else should be adjusted.)

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -11,4 +11,4 @@ jobs:
       - name: "Checkout Repository"
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: "Dependency Review"
-        uses: actions/dependency-review-action@3c4e3dcb1aa7874d2c16be7d79418e9b7efd6261 # v4.8.2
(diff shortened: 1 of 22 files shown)


.markdownlint.json
```diff
@@ -3,7 +3,7 @@
   "MD002": false,
   "MD004": { "style": "dash" },
   "MD007": { "indent": 2 },
-  "MD013": { "line_length": 600, "code_blocks": false },
+  "MD013": { "line_length": 1600, "code_blocks": false },
```
alexander-akait: Please don't change it

karanrajsurya: Actually I transposed the table so the table becomes more visible. Its fine if you want to retain that part, i will find another way to make the table actually work and readable without changing other parts of the project.

karanrajsurya:
I fixed the table by dividing it into 4 categories and under the line width constraint, making them far better accessible.


Screenshot 2026-02-27 213401

alexander-akait:
This page is outdated, and we need to remove it, don't think we need it in now days

karanrajsurya:
> This page is outdated, and we need to remove it, don't think we need it in now days

Alright sir, its okay if you want to remove the page. I'd be happy if my commit gets merged or to help if this page need some improvement or anything.

PR #7925, opened 2026-03-05
Upgrade react

### LINKED ISSUE #7920: update react to the latest version
We need to update `react` to the latest version, related dependencies should be updated too, like `react-tiny-popover` and maybe other, also need to check all pages on warnings from react side and fix them (but I don't think we will have many of them)

theEquinoxDev:
## Upgrade React to v19

Closes #7920 

### What changed

**Dependencies upgraded**
- `react` and `react-dom`: 17.0.2 -> 19.1.0
- `react-tiny-popover`: 5 -> 8.1.6
- `react-helmet-async`: 2.0.5 -> 3.0.0
- `@react-spring/web`: 9.7.5 -> 10.0.3

**`react-visibility-sensor` removed**
This package uses `ReactDOM.findDOMNode()`, which was removed in React 19. Both
`Contributors.jsx` and `Support.jsx` now use the native `IntersectionObserver` API
to achieve the same lazy-load behavior, with no additional dependencies.

**`SidebarItem.jsx` refactored**
Converted from a class component to a functional component to eliminate the
`UNSAFE_componentWillReceiveProps` lifecycle method, which React 19 no longer supports.

**`index.jsx` updated**
Replaced the legacy `ReactDOM.render` / `ReactDOM.hydrate` calls with `createRoot` and
`hydrateRoot` from `react-dom/client`, as required by React 18+.

**`server.jsx` updated**
`react-helmet-async` v3 with React 19 no longer populates the helmet context during
SSR (React 19 handles head tags natively). Added a null guard to prevent a crash in
the static site generator.

**`Configuration/components.js` updated**
`react-tiny-popover` v8 switched to a named export and renamed the `position` prop to
`positions`. Updated the import and JSX accordingly.

**`package.json` scripts fixed**
`rimraf` v6 requires an explicit `--glob` flag for glob patterns. Two scripts
(`clean` and `clean-printable`) were missing this flag, causing them to silently do
nothing on Node 20+.

### Testing
- `npm start` - dev server runs cleanly, no React warnings in the browser console
- `npm run build` - production and SSG builds complete successfully
- `npm run jest` - all 27 tests pass

Let me know if anything needs changing or if I missed something.


src/components/CodeBlockWithCopy/CodeBlockWithCopy.scss
alexander-akait, line 51: Please avoid these changes, it is only about react, not styles

src/components/Support/Support.jsx
```diff
@@ -177,119 +189,109 @@ export default class Support extends Component {
...
-                          ? supporter.avatar
-                          : SmallIcon
-                      }
-                      alt={
-                        supporter.alt ||
-                        (supporter.name || supporter.slug
-                          ? `${supporter.name || supporter.slug}'s avatar`
-                          : "avatar")
-                      }
-                      onError={this._handleImgError}
```
alexander-akait, line 264: Why this was just removed?

package.json
```diff
@@ -3,28 +3,31 @@
...
+  "engines": {
+    "node": ">= 20.9.0"
+  },
   "scripts": {
     "clean-dist": "rimraf ./dist",
-    "clean-printable": "rimraf src/content/**/printable.mdx",
+    "clean-printable": "rimraf --glob src/content/**/printable.mdx",
     "preclean": "run-s clean-dist clean-printable",
-    "clean": "rimraf src/content/**/_*.mdx src/**/_*.json repositories/*.json",
+    "clean": "rimraf --glob src/content/**/_*.mdx src/**/_*.json repositories/*.json",
```
alexander-akait, line 30: Why they was added?

alexander-akait:
Also please reabase

package.json
ovflowd: AI-generated PR with unrelated changes 🤷

alexander-akait:
Close in favor of https://github.com/webpack/webpack.js.org/pull/7929

package.json
theEquinoxDev: > AI-generated PR with unrelated changes 🤷

was caught up with some things so wasn't much active... here's my justification :)

The --glob flag was added because rimraf does not process glob patterns unless globbing is explicitly enabled.
Without --glob , a command like:
rimraf src/content/**/_*.mdx wont' match the pattern and will effectively do nothing. Adding --glob ensures the intended files are actually removed during the clean step.

Reference: https://www.npmjs.com/package/rimraf

this was not an AI-generated change. While testing the clean step locally I ran into the behavior where the glob pattern wasn't being processed. I did use AI to understand the error I encountered, but the change itself came from verifying the behavior and testing it locally.

PR #7928, opened 2026-03-06
fix(sw): return inner promise in activate cache cleanup

ryzrr:
Summary
In the service worker's activate event handler, the inner cache.keys().then() promise was never returned from the outer .then() callback. this caused event.waitUntil() to resolve immediately with undefined before any stale cache entries were deleted. Old previous cache entries from previous deployment would be accumulated in users browsers with no guaranteed of cleanup. this fix returns the full promise chain and replaces the for loop with Promise.all() so all deletions run in parallel and event.waitUntil() only resolves once all stale entries are gone.

fix

No, the service worker runs in a browser context outside the Jest test environment. 

No

N/A

Use of AI
N/A

### DIFF
--- a/src/sw.js
+++ b/src/sw.js
@@ -38,16 +38,18 @@ self.addEventListener("install", (event) => {
 self.addEventListener("activate", (event) => {
   // - [x] clean up outdated runtime cache
   event.waitUntil(
-    caches.open(cacheName).then((cache) => {
+    caches.open(cacheName).then((cache) =>
       // clean up those who are not listed in manifestURLs
-      cache.keys().then((keys) => {
-        for (const request of keys) {
-          if (!manifestURLs.includes(request.url)) {
-            cache.delete(request);
-          }
-        }
-      });
-    }),
+      cache
+        .keys()
+        .then((keys) =>
+          Promise.all(
+            keys
+              .filter((request) => !manifestURLs.includes(request.url))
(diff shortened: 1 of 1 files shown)


alexander-akait:
@evenstensberg Please take a look

alexander-akait:
@ryzrr Do you have an issue about it?

ryzrr:
@alexander-akait  No existing issue for this one , I noticed it while going through the sw code and will be happy to open an issue first if you prefer that before raising  a PR.

alexander-akait:
@ryzrr No need, just a question

src/sw.js
```diff
@@ -38,16 +38,18 @@ self.addEventListener("install", (event) => {
 self.addEventListener("activate", (event) => {
   // - [x] clean up outdated runtime cache
   event.waitUntil(
-    caches.open(cacheName).then((cache) => {
+    caches.open(cacheName).then((cache) =>
       // clean up those who are not listed in manifestURLs
-      cache.keys().then((keys) => {
```
ovflowd, line 43: To be fair, the inner Promise doesn't _need_ to be returned. But indeed the outer Promise will resolve eagerly when the inner one hasn't finished resolving yet.

ryzrr: Thanks for the clarification ! My intention was to ensure that all cache delete operations are awaited so event wait until resolves only after the cleanup finishes.

PR #7938, opened 2026-03-07
fix(remark-responsive-table): correctly extract header text for data-th

ryzrr:
Summary
the remark-responsive-table plugin reads table header text using th.children[0].value to set the data-th attribute on each cell. This only works for plain text header and when  table header contains complex markdown such as code, bold, or links, children[0] is a structural node that has no .value property, causing data-th to set to "undefined". On the mobile, the CSS uses content : attr(data-th) to show the column label before each cell  ,so any affected table would show blank or "undefined" labels instead of the header text. 

fix

No, the existing test continue to pass. 

No

N/A

Use of AI
N/A

### DIFF
--- a/src/remark-plugins/remark-responsive-table/remark-responsive-table.mjs
+++ b/src/remark-plugins/remark-responsive-table/remark-responsive-table.mjs
@@ -1,4 +1,5 @@
 // this plugin was first based on https://github.com/montogeek/remark-responsive-tables
+import { toString } from "mdast-util-to-string";
 import { visit } from "unist-util-visit";
 
 export default function remarkResponsiveTable() {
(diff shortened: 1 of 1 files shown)


alexander-akait:
Where I can see the fixed result? Can you link a page?

ryzrr:
You can see a responsive table on this page:
https://webpack.js.org/configuration/devtool/ or at  /dev-server
In mobile layout this table uses the `data-th` attribute to show the column label before each cell.
but headers on this page are plain text, so they work fine with the current implementation. Incase, if a table header contains markdown (like inline `code`, bold text, or links), `th.children[0].value` may return `undefined` because the first child isn't always a text node.
and by using `toString(th)` ensures the correct header text is extracted regardless of the markdown structure. 

Example:
| `Option` | Description |  **Quality** |

alexander-akait:
Thanks

PR #2058, opened 2018-04-17
fix: various fixes on concepts for v4 and some rephrasing

jeremenichelli:
Initially this PR solves https://github.com/webpack/webpack.js.org/issues/2035 issue.

I also change some parts and code snippets that are not valid with v4 release.

FInally, I encounter a lot of issues with code snippets and content in each section. There were lots of sentences saying exactly the same thing over and over, probably tangling more the understanding an entry user needs on what each concept means in webpack.

Would your words on my changes, the intention was to summarize and simplify the content to be a little more consumable.

You can check the formatted version of the concepts page [here](https://github.com/webpack/webpack.js.org/blob/762a338f28e361d70ae494d1ff12b330441b4a1b/src/content/concepts/index.md) which might help review the content.

PR #7939, opened 2026-03-07
docs(concepts): clarify zero-config entry default and error message

### LINKED ISSUE #2035: using webpack without config file: can't resolve ./src  message error
Hi,

webpack 4.x.x can be used without configuration file (as said at https://webpack.js.org/concepts/  - "Since v4.0.0 webpack does not require a configuration file"). 
What is not said  explicitly is that the root file must be **index.js**. If you put any other file name at src folder the error message is : 
```
ERROR in Entry module not found: Error: Can't resolve './src' in '/home/leandro/repositorios-git/INE5646-2018/teste_webpack'
```
My sugestion: Alter docs and error message to let explicit the real cause of the problem: file ./src/index.js not found.

  Leandro

saishankar404:
## what this does

adds a tip block to `concepts/index.mdx` and `concepts/entry-points.mdx` that explains:

- when running webpack without a config file, the entry defaults to `./src/index.js`
- if that file is missing, webpack throws `ERROR in Entry module not found: Error: Can't resolve './src'`
- why the error says `./src` rather than `./src/index.js` (webpack resolves the directory first, then looks for `index.js` inside it)
- how to fix it by pointing `entry` to the actual filename in a `webpack.config.js`

## why

the 2018 PR #2058 added a one-liner naming the default but didn't address the confusing mismatch between the error message (`./src`) and the actual missing file (`./src/index.js`). users still hit this confusion as noted in the last comment on the issue. the tip on the entry-points page also helps users who land there directly.

closes #2035

### DIFF
--- a/src/content/concepts/entry-points.mdx
+++ b/src/content/concepts/entry-points.mdx
@@ -12,6 +12,7 @@ contributors:
   - adyjs
   - anshumanv
   - ritikbanger
+  - saishankar404
 ---
 
 As mentioned in [Getting Started](/guides/getting-started/#using-a-configuration), there are multiple ways to define the `entry` property in your webpack configuration. We will show you the ways you **can** configure the `entry` property, in addition to explaining why it may be useful to you.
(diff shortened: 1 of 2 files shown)


saishankar404:
> Looks good, please accept CLA

done! thanks.

PR #7941, opened 2026-03-08
fix: replace reserved import prop with loadContent

Katari-8055:
## Summary
This PR resolves a naming conflict caused by using the reserved JavaScript keyword `import` as a prop. The prop has been renamed to `loadContent` across the relevant components to avoid compatibility issues with modern JavaScript engines.

## Changes
- Replaced `import` prop with `loadContent`
- Updated references in:
  - App.jsx
  - server.jsx
  - Site.jsx

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,5 +1,5 @@
 import Site from "./components/Site/Site.jsx";
 
 export default function App() {
-  return <Site import={(path) => import(`./content/${path}`)} />;
+  return <Site loadContent={(path) => import(`./content/${path}`)} />;
 }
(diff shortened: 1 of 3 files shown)


alexander-akait:
Where is `import` is reserved?

Katari-8055:
> Where is `import` is reserved?

As import is reserved in JavaScript module syntax . While it can technically be used as an object property like props.import, I renamed it to loadContent to avoid confusion with the module import syntax and to improve code readability.

alexander-akait:
@let's merge it after updating react to avoid merge conflicts

PR #7954, opened 2026-03-08
fix(process-readme): fix broken npm URL normalization

ryzrr:
Summary
 a broken npm url normalization in process-readme.mjs , The linkFixer function is used during build time to process all upstream loader and plugin READMEs fetched from GitHub and any npmjs.com link (without www.) in those READMEs was being silently corrupted into https:undefined/package/...    a completely broken URL.

Bug fix. 

The existing tests in process-readme.test.mjs all pass but still a  targeted test for npm URL normalization could be added as a follow-up.

No. 

If relevant, what needs to be documented?

No need.

Use of AI
 Slightly for refining my words .

### DIFF
--- a/src/utilities/process-readme.mjs
+++ b/src/utilities/process-readme.mjs
@@ -30,7 +30,7 @@ function linkFixerFactory(sourceUrl) {
     const oldHref = href;
 
     if (href.includes("//npmjs.com")) {
-      href = href.replace("//www.npmjs.com");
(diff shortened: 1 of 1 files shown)


avivkeller (approved):
It's worth noting this is all scheduled to be re-written at a point in the near future

ryzrr:
Okay Understood  , Thanks for reviewing it .

Issue #7772, opened 2026-02-02
Inconsistent module syntax in tutorial (ESM `import` vs. CommonJS `require`)


Brennvo:
In the "Getting Started" tutorial section, [`webpack.config.js` is initialized using ESM syntax](https://webpack.js.org/guides/getting-started/#using-a-configuration):

> **`webpack.config.js`**
> 
> ```javascript
> import path from "node:path";
> import { fileURLToPath } from "node:url";
> 
> const __filename = fileURLToPath(import.meta.url);
> const __dirname = path.dirname(__filename);
> 
> export default {
>   entry: "./src/index.js",
>   output: {
>     filename: "main.js",
>     path: path.resolve(__dirname, "dist"),
>   },
> };
> ```

However, in the section that immediately follows ("Asset Management"), [CommonJS syntax is used](https://webpack.js.org/guides/getting-started/#using-a-configuration):

> **`webpack.config.js`**
> 
> ```diff
>  const path = require('path');
> 
>  module.exports = {
>    entry: './src/index.js',
>    output: {
> -    filename: 'main.js',
> +    filename: 'bundle.js',
>      path: path.resolve(__dirname, 'dist'),
>    },
>  };
> ```

This can be confusing to somebody who is completing the tutorial sequentially.

alexander-akait:
PR welcome

the-shreyash:
Hi! I’ve submitted a PR that aligns the **Getting Started** guide with a CommonJS `webpack.config.js` example: #7786.

This follows the approach discussed here.  
Happy to adjust, rebase, or close it if it overlaps with other changes. Thanks!

aayushi-tomar:
Hi, I would like to work on aligning the module syntax in the tutorial by using CommonJS consistently across sections. Please let me know if that approach works.

PeterWayne12:
Hi
I’d like to work on this issue if it’s still open.  
I’ve reviewed the tutorial section and noticed the inconsistency between ESM (`import`) and CommonJS (`require`) syntax.
I’d be happy to standardize the examples according to the preferred module format used in the documentation.

Please let me know if I can proceed.

PR #7979, opened 2026-03-10
docs: modernize 'Writing a Loader' guide to ESM syntax

hagemaruwu:
**Summary**

The "Writing a Loader" guide was mixing CommonJS (`require`/`module.exports`) 
and ESM (`import`/`export`) syntax across different code examples, which is 
confusing for readers. This PR makes all examples consistent by using ESM 
throughout. It also fixes a missing `__dirname` definition in the 
`test/compiler.js` example that would cause a runtime error in an ESM context.

part of #7772

docs

No — this is a documentation-only change. No functional code was modified.

No.

No additional documentation needed. This PR is itself a documentation fix.

**Use of AI**

I used an AI assistant to help identify inconsistencies 
in the file and understand the difference between CommonJS and ESM module 
syntax. All code changes were applied manually by me. I reviewed 
the webpack AI policy before submitting.

### DIFF
--- a/src/content/contribute/writing-a-loader.mdx
+++ b/src/content/contribute/writing-a-loader.mdx
@@ -346,7 +346,7 @@ npm install --save-dev webpack memfs
 
 ```js
 import path from "node:path";
-import { fileURLToPath } from "node:url";
+import { Volume, createFsFromVolume } from "memfs";
 import webpack from "webpack";
 
 const __dirname = import.meta.dirname;


src/content/contribute/writing-a-loader.mdx
```diff
@@ -341,9 +346,13 @@ npm install --save-dev webpack memfs
 
 ```js
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { Volume, createFsFromVolume } from "memfs";
 import webpack from "webpack";
 
+const __filename = fileURLToPath(import.meta.url);
```
ovflowd: This is an obsolete way to get the filename and dirname nowadays in Node.js -- you can check the new way on Node.js docs or even ask AI for it.

ovflowd: But the TL;DR is import.meta.dirname and import.meta.filename

hagemaruwu: Thank you for the review! I've updated both examples to use 
'import.meta.dirname` instead of the `fileURLToPath` approach.

src/content/contribute/writing-a-loader.mdx
```diff
@@ -227,18 +230,18 @@ T> If the language only accepts relative urls (e.g. `url(file)` always refers to
 
 ### Common Code
 
-Avoid generating common code in every module the loader processes. Instead, create a runtime file in the loader and generate a `require` to that shared module:
+Avoid generating common code in every module the loader processes. Instead, create a runtime file in the loader and `import` it as a shared module:
```
alexander-akait: We should write here that commonjs syntax supported too, but we recommended to use ECMA modules

hagemaruwu: Thanks for the review @alexander-akait! I have updated the prose to explicitly mention that `require` is still supported while recommending `import`.

hagemaruwu: Also restored the missing 
`memfs` import that was accidentally removed.

PR #7999, opened 2026-03-12
fix: replace invalid anchor with button in PageLinks

### LINKED ISSUE #7998: Accessibility: invalid <a> element without href in PageLinks.jsx
In PageLinks.jsx, the "Print this page" action is rendered as an <a> element without an href.

Using an anchor without an href creates invalid link semantics and can cause accessibility and keyboard navigation issues.

Expected behavior:
Interactive actions that do not navigate should use a <button> element instead of <a>.

Possible fix:
Replace the <a> element with <button type="button"> while preserving the existing styling and click behavior.

mr-baraiya:
Fixes #7998

In `PageLinks.jsx`, the **"Print this page"** action was rendered as an `<a>` element without an `href`, which creates invalid link semantics and may cause accessibility and keyboard navigation issues.

This PR replaces it with a semantic `<button type="button">` while preserving the existing styling and behavior.

### DIFF
--- a/src/components/PageLinks/PageLinks.jsx
+++ b/src/components/PageLinks/PageLinks.jsx
@@ -26,9 +26,9 @@ export default function PageLinks({ page = {} }) {
         Edit this page
       </a>
       <Separator />
-      <a className={classes} onClick={_handlePrintClick}>
+      <button type="button" className={classes} onClick={_handlePrintClick}>
         Print this page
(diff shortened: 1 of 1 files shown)


alexander-akait:
@mr-baraiya Please in future feel free to send a PR straight away if you found a bug, need to create issues because it is more like spam

PR #8002, opened 2026-03-12
Enhancement: Add fallback for missing SVG Rendering in SplashViz.jsx

rahul-kr-rai:
**Summary**

This PR adds null/undefined checks with a fallback message for SVG rendering in the `SplashViz.jsx` component. It ensures that if `HomeSVG.body` is missing, the component will not break and instead display a clear fallback.

**Enhancement:** improves resilience and user feedback in case of missing SVG data.

No

No

**Use of AI**
GitHub Copilot

### DIFF
--- a/src/components/SplashViz/SplashViz.jsx
+++ b/src/components/SplashViz/SplashViz.jsx
@@ -26,7 +26,9 @@ export default class SplashViz extends Component {
         </h1>
         <div
           className="splash-viz__modules"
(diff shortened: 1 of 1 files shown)


rahul-kr-rai:
Fair point! Since it's a static import, the build process would catch any missing assets anyway. I'll revert this to keep the code lean.

alexander-akait:
Let's close because we don't need it, we can revisit this in future if we will have situations when we don't have content

PR #8004, opened 2026-03-13
docs: clarify emit hook usage in writing-a-plugin guide

raj-sapalya:
This PR adds a clarification note to the "Writing a Plugin" guide.
Some examples in the guide use the `emit` hook for simplicity.
However, modern Webpack plugins often use `compilation.hooks.processAssets`
for asset processing.
This note helps readers understand the modern approach without modifying
the existing example code.


raj-sapalya:
> We should fix it and use `processAssets` in guides

Thanks for the review! I'll update the example to use `processAssets` instead of `emit` and push the changes.

PR #8008, opened 2026-03-13
chore(Page): update content PropTypes to reflect supported types

mr-baraiya:
Update `content` PropTypes to reflect the supported content types.

The component accepts content as a string, function, or Promise-like object, but the previous PropTypes only supported a Promise shape.

### DIFF
--- a/src/components/Page/Page.jsx
+++ b/src/components/Page/Page.jsx
@@ -188,8 +188,8 @@ Page.propTypes = {
     PropTypes.func,
     PropTypes.shape({
       // eslint-disable-next-line unicorn/no-thenable
-      then: PropTypes.func,
-      default: PropTypes.func,
+      then: PropTypes.func.isRequired,
+      default: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
     }),
   ]),
 };


mr-baraiya:
`content` can be passed as a string or function through MDX rendering in the docs.
For example, MDX pages may pass `content` as a string or a render function depending on how the page is compiled.
I updated the PropTypes to explicitly reflect these supported types instead of the previous Promise-only shape.

If needed, I can adjust the PropTypes further to remove `any` and make them stricter.

alexander-akait:
> If needed, I can adjust the PropTypes further to remove any and make them stricter.

Yes, let's make it stricter

mr-baraiya:
Updated the PropTypes to remove `any` and make the `content` type stricter.

src/components/Page/Page.jsx
```diff
@@ -184,10 +184,12 @@ Page.propTypes = {
...
   pages: PropTypes.array,
   content: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
     PropTypes.shape({
       // eslint-disable-next-line unicorn/no-thenable
-      then: PropTypes.func.isRequired,
-      default: PropTypes.string,
+      then: PropTypes.func,
+      default: PropTypes.func,
```
alexander-akait: it can be string or function and `then` is required for shape

mr-baraiya: Updated the `PropTypes` so content can be string, function, or Promise-like and made then required in the shape.

PR #8011, opened 2026-03-13
fix(content-tree): apply filtered children during restructure

ryzrr:
Summary
The content tree restructuring pipeline called item.children.filter(filter) but discarded the returned array and this made filtering a no-op and it could silently break future filter rules in navigation/content generation.
So this PR assigns the filtered result back to item.children before sorting, so the pipeline behavior matches its contract.

Bug fix

Yes

No

No additional docs required.

Use of AI
Yes, I used an Ai collaborator to help review my initial logic and to assist in debugging the Jest ESM environment configuration but still, the identification of the no-op filter bug and the final implementation of the unit tests were manually verified and tested locally to ensure the pipeline correctness.

### DIFF
--- a/src/utilities/content-tree-enhancers.mjs
+++ b/src/utilities/content-tree-enhancers.mjs
@@ -144,7 +144,7 @@ export function restructure(item, options) {
   if (item.children) {
     for (const child of item.children) restructure(child, options);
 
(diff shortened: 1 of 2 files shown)


ryzrr:
thanks!

PR #8014, opened 2026-03-14
fix(a11y): improve keyboard accessibility for sidebar and external links

hagemaruwu:
**Summary**

Found a few accessibility issues while using the site with keyboard-only navigation and a screen reader. These changes fix them:

Sidebar toggle (SidebarItem) — The chevron icons (▶) next to sidebar items were raw <svg> elements with onClick. They couldn't be reached with Tab or activated with Enter/Space, which blocks keyboard-only users from expanding sidebar sections. Wrapped them in <button> and added aria-label + aria-expanded.

Mobile close button (SidebarMobile) — The close button was a <span role="button"> instead of a real <button>. Using a <span> with role="button" is an anti-pattern — real <button> elements are keyboard-accessible by default and don't need the ARIA role workaround. Added aria-label.

External links (Link) — All external links use target="_blank" but don't warn screen reader users that the link opens in a new tab. Added a visually hidden "(opens in a new tab)" text per WCAG H83 guidelines.

fix — accessibility improvement, no logic or visual changes.

No new unit tests — these components don't have existing tests. Verified manually: sidebar toggles are keyboard-focusable and operable via Enter/Space.

No. Zero visual difference.

**Use of AI**

AI was used to identify accessibility issues by auditing the component source code against WCAG guidelines, and to suggest the appropriate ARIA attributes (aria-label, aria-expanded). All code changes were written and reviewed manually.

### DIFF
--- a/src/components/Link/Link.jsx
+++ b/src/components/Link/Link.jsx
@@ -8,7 +8,10 @@ const Link = ({ to = "", url, ...props }) => {
 
   if (to.startsWith("http") || to.startsWith("//")) {
     return (
-      <a href={to} target="_blank" rel="noopener noreferrer" {...others} />
+      <a href={to} target="_blank" rel="noopener noreferrer" {...others}>
+        {others.children}
+        <span className="screen-reader-only">(opens in a new tab)</span>
+      </a>
     );
   }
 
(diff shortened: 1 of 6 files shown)


evenstensberg:
I've tested on keyboard, looks good. Can't reproduce this fix on external links, do you have an example you can point me to?

hagemaruwu:
hi @evenstensberg 
sorry for the delay, i was out for my friends birthday

here is the visual proof 

Screenshot 2026-03-18 at 9 31 40 AM

and you can see there is no visual ui change 

Screenshot 2026-03-18 at 9 33 11 AM

hope it clears things up.

PR #8021, opened 2026-03-15
Fix: write empty _supporters.json fallback when API fetch fails in development

Debraj2024:
## Problem

When contributors clone the repo and run `yarn start` for the first time 
without running `yarn fetch` first, the site crashes with:

TypeError: _supporters_json__WEBPACK_IMPORTED_MODULE_5__ is not iterable

This happens because:
- `_supporters.json` is gitignored (`src/**/_*.json`)
- There is no fallback when the API call fails, leaving no file behind
- New contributors may miss the README instruction to run `yarn fetch` first

## Fix

1. **Skip fetching if file already exists** — If `_supporters.json` already 
   exists, skip the API call entirely. Can be overridden with `FORCE_FETCH=true`.

2. **Graceful fallback in development** — If the API fetch fails and 
   `NODE_ENV` is not `production`, write an empty `[]` to `_supporters.json` 
   with a clear warning instead of crashing.

## Testing

1. Clone the repo fresh
2. Run `yarn`
3. Disconnect internet
4. Run `yarn fetch:supporters`
5. ✅ Warning message appears, `_supporters.json` created with `[]`
6. ✅ `yarn start` loads without errors

## Notes

- `sw.js` lint errors in this PR are pre-existing and unrelated to this change
- Running `yarn fetch` before `yarn start` as documented will still work as before
- `FORCE_FETCH=true yarn fetch:supporters` forces a fresh fetch when needed

### DIFF
--- a/src/utilities/fetch-supporters.mjs
+++ b/src/utilities/fetch-supporters.mjs
@@ -21,27 +21,27 @@ const absoluteFilename = path.resolve(
   filename,
 );
 
-let graphqlEndpoint = "https://api.opencollective.com/graphql/v2";
-
-if (process.env.OPENCOLLECTIVE_API_KEY) {
-  // rate limit is 100 requests per minute with personal access token
-  // rate limit is 10 requests per minute without personal access token
+if (fs.existsSync(absoluteFilename) && process.env.FORCE_FETCH !== "true") {
   console.log(
-    "Using personal access token to fetch supporters from OpenCollective",
+    `${filename} already exists, skipping fetch. Use FORCE_FETCH=true to override.`,
   );
-  // by default a personal access token of @chenxsan was used as I don't have access to the webpack one
-  // @doc https://graphql-docs-v2.opencollective.com/access#with-a-personal-token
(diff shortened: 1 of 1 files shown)


TusharThakur04:
theres no issue at all, you just need to run the [`yarn fetch-all`](https://github.com/webpack/webpack.js.org/blob/b0368cf3e05f49070ee09bee5c6ae601d08e78bf/package.json#L36) cmd before `yarn start`.
it will automatically run [`fetch:supporters`](https://github.com/webpack/webpack.js.org/blob/b0368cf3e05f49070ee09bee5c6ae601d08e78bf/package.json#L34) to install required modules.
and dont forget to generate the github token, please refer to [contrib guide](https://github.com/webpack/webpack.js.org/blob/main/.github/CONTRIBUTING.md)

Debraj2024:
@TusharThakur04 Thanks for the feedback! You're right that yarn fetch-all handles this when followed correctly. The issue I ran into was running yarn start directly without yarn fetch-all first — which is easy to miss as a new contributor.

TusharThakur04:
> which is easy to miss as a new contributor.

should always read the get-started/contrib guide thoroughly.

evenstensberg:
pr template not used

PR #8029, opened 2026-03-16
docs: add running locally section and supporters.json setup note

imanchalsingh:
While trying to run the project locally, I ran into a module error
caused by a missing `_supporters.json` file.

I’ve added a small "Run Locally" section to the contributing docs to
document the setup steps that worked for me, in case it helps future
contributors. Happy to adjust if this should be placed elsewhere.


alexander-akait:
I think we can make it better, adding an empty file and update content during the build

anchals22:
> I think we can make it better, adding an empty file and update content during the build

Thanks for the suggestion! That makes sense. I’ll update the implementation to create an empty file first and populate the content during the build process instead. Appreciate the guidance.

anchals22:
> Just add an empty file, no need to make a script for this and check we are adding content in `_supporters.json` on the build step

Actually, the _supporters.json file is ignored by .gitignore, so creating it manually wouldn’t appear for other contributors. That’s why I added a mechanism to auto-generate the file during the build process, ensuring everyone can generate it locally without manual steps.

alexander-akait:
In this case add a `fallback` in `resolve` in our configuration when files doesn't exist

anchals22:
> In this case add a `fallback` in `resolve` in our configuration when files doesn't exist

Thank you for the suggestion! Adding a fallback makes sense, I’ll take care of it.

TusharThakur04:
@alexander-akait i dont think there is any need for the solution
on running `github__token yarn fetch-all`, it automatically runs `fetch:supporters` which in turn fetches all required modules to run the project.
Following the setup guide is all needed.

alexander-akait:
@TusharThakur04 We can simplify such things, but maybe you are right, in our guide you need to run this

PR #8035, opened 2026-03-17
refactor(Dropdown): fix inefficient event handler .bind(this) bindings

rahul-kr-rai:
**Summary**

This PR addresses the inefficient event handler binding within the `Dropdown` component. Previously, `.bind(this)` was being used inside the `render()` method and lifecycle hooks, which causes new function instances to be created on every single render. This has been refactored by converting the relevant class methods to arrow functions and utilizing inline arrow functions for parameterized handlers in the render tree, improving overall component performance.

- Avoids creating `new function` instances on every `render`
- Enables React optimization opportunities
- Improves `performance` in large component trees

No new tests were added.

No

N/A

**Use of AI**
Yes, used Google Gemini to increase productivity and safely refactor component logic.

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -47,8 +47,8 @@ export default class Dropdown extends Component {
       <nav
         className={`dropdown ${className}`}
         ref={(el) => (this.dropdown = el)}
-        onMouseOver={() => this._toggle(true)}
-        onMouseLeave={() => this._toggle(false)}
+        onMouseOver={this._handleMouseOver}
+        onMouseLeave={this._handleMouseLeave}
       >
         <button
           ref={(el) => (this.dropdownButton = el)}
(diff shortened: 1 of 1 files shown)


rahul-kr-rai:
Please Review

src/components/Dropdown/Dropdown.jsx
```diff
@@ -50,15 +47,15 @@ export default class Dropdown extends Component {
       <nav
         className={`dropdown ${className}`}
         ref={(el) => (this.dropdown = el)}
-        onMouseOver={this._toggle.bind(this, true)}
-        onMouseLeave={this._toggle.bind(this, false)}
+        onMouseOver={() => this._toggle(true)}
+        onMouseLeave={() => this._toggle(false)}
```
alexander-akait: Why extra `() => `?

rahul-kr-rai: At first, I replaced `.bind(this)` for the inline arrow function to handle the `this` context, but I realize now that 
`() =>` still creates a brand new function reference on every single render, which defeats the purpose of the PR. I've just pushed a commit to extract these into dedicated class methods (`_handleMouseOver` and `_handleMouseLeave`) so the references remain stable.

PR #8052, opened 2026-03-18
docs(entry-points): add 'See also' section with related concept links

mr-baraiya:
Added a **"See also"** section to `entry-points.mdx` with links to related concepts (Output, Modules, Module Resolution, Loaders, Plugins) to improve navigation and help users better understand how these topics are connected.

### DIFF
--- a/src/components/Link/Link.jsx
+++ b/src/components/Link/Link.jsx
@@ -8,7 +8,10 @@ const Link = ({ to = "", url, ...props }) => {
 
   if (to.startsWith("http") || to.startsWith("//")) {
     return (
-      <a href={to} target="_blank" rel="noopener noreferrer" {...others} />
+      <a href={to} target="_blank" rel="noopener noreferrer" {...others}>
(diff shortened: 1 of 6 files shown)


src/content/concepts/entry-points.mdx
```diff
@@ -179,6 +180,16 @@ export default {
...
 
+---
+
+## See also
+
+- [Output](/concepts/output/)
+- [Modules](/concepts/modules/)
+- [Module Resolution](/concepts/module-resolution/)
+- [Loaders](/concepts/loaders/)
+- [Plugins](/concepts/plugins/)
```
alexander-akait: @evenstensberg Do we need this here? Because we have the same in the sidebar

mr-baraiya: You’re right, since these links are already available in the sidebar, this might be redundant.

I’ll remove the "See also" section to keep the page clean and consistent with the existing structure.

PR #8054, opened 2026-03-18
docs(configuration): clarify usage of multiple configurations with ex…

mr-baraiya:
Improve clarity in `configuration.mdx` by adding a brief explanation and example for using single vs multiple configurations.


mr-baraiya:
> No, I disagree, you should use multi configuration in many cases, I even can't list them

You're right — multi-configuration is useful in many cases. I'll update the wording to avoid suggesting it should generally be avoided and instead focus on clarifying when it can be helpful.

PR #8088, opened 2026-03-21
docs: modernize loader example using css-loader

ryzrr:
Summary
Replaced the deprecated `raw-loader` example with the actively supported `css-loader` in the Concepts documentation.

Updated the "Loaders" section in `src/content/concepts/index.mdx` to reflect modern Webpack conventions. The example previously contained dead code (`raw-loader`) which has natively been replaced by Webpack 5 Asset Modules. I swapped it out for `css-loader` to perfectly maintain the  purpose of the `use` property example without breaking the context of the guide.

No

No

N/A

Use of AI
N/A

### DIFF
--- a/src/content/concepts/index.mdx
+++ b/src/content/concepts/index.mdx
@@ -112,14 +112,14 @@ export default {
     filename: "my-first-webpack.bundle.js",
   },
   module: {
-    rules: [{ test: /\.css$/, use: "css-loader" }],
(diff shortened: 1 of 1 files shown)


ryzrr:
Makes sense regarding experiments.css! So should I swap this to a ts-loader or babel-loader instead?

alexander-akait:
I think `babel-loader` is the good example here

ryzrr:
Okay  , working on it then.

PR #8094, opened 2026-03-22
docs(guides): fix lodash classification from peerDependency to depend…

### LINKED ISSUE #6896: Misguiding document on authoring libraries

jay7-tech:
Fixes #6896.

(Incorrect advice to use peerDependencies for utility libraries).

The Authoring Libraries guide was advising developers to treat utility libraries like lodash as peerDependencies to avoid bundle bloat when externalizing them. However, a in the issue, if a library relies on lodash internally at runtime, it must be listed in dependencies.

peerDependencies are strictly for packages that a consumer must install directly in their own app to satisfy the library's environment (eg React for a UI component library). dependencies guarantees the package manager installs the utility library automatically.

This PR simply correctly updates the text to advise using dependencies instead, while keeping the explanation of how to externalize it in the webpack config to prevent duplication.

Fix / Documentation

NA

No.

N/A

Use of AI

None.

### DIFF
--- a/src/content/guides/author-libraries.mdx
+++ b/src/content/guides/author-libraries.mdx
@@ -252,7 +252,15 @@ T> Note that the `library` setup is tied to the `entry` configuration. For most 
 
 ## Externalize Lodash
 
-Now, if you run `npx webpack`, you will find that a largish bundle is created. If you inspect the file, you'll see that lodash has been bundled along with your code. In this case, we'd prefer to treat `lodash` as a _peer dependency_. Meaning that the consumer should already have `lodash` installed. Hence you would want to give up control of this external library to the consumer of your library.
+Now, if you run `npx webpack`, you will find that a largish bundle is created because `lodash` is bundled directly into your code. 
+
+To prevent duplicate bundles, you can externalize `lodash` in your webpack config. How you classify `lodash` in your `package.json` then depends entirely on your library's architecture:
+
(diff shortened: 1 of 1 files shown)


src/content/guides/author-libraries.mdx
```diff
@@ -252,7 +252,7 @@ T> Note that the `library` setup is tied to the `entry` configuration. For most
 
 ## Externalize Lodash
 
-Now, if you run `npx webpack`, you will find that a largish bundle is created. If you inspect the file, you'll see that lodash has been bundled along with your code. In this case, we'd prefer to treat `lodash` as a _peer dependency_. Meaning that the consumer should already have `lodash` installed. Hence you would want to give up control of this external library to the consumer of your library.
+Now, if you run `npx webpack`, you will find that a largish bundle is created. If you inspect the file, you'll see that lodash has been bundled along with your code. In this case, we'd prefer to treat `lodash` as a dependency. By declaring it as a `dependency` in your `package.json` and externalizing it in your webpack config, you give up control of this external library to the consumer's package manager, preventing duplicate bundles.
```
alexander-akait: It can be `dependency`, dev dependency and peer dependency depending on how you bundling your library

jay7-tech: Yeaahh, 
The ideal dependency classification strictly relies on the consumers bundling architecture.
I've updated the guide to explicitly map out dependencies (npm resolution), devDependencies (pre-bundled dists), and peerDependencies (externalized singletons like React/Lodash) so the scenarios are unambiguous. I Just pushed the patch

PR #8130, opened 2026-03-27
fix(build): prevent rimraf glob crash on Windows by injecting --glob …

jay7-tech:
I hit a terminal crash when trying to run npm run build locally on Windows.

The clean-printable and clean scripts throw an EINVAL (Illegal characters) error because rimraf ^6 stops globbing by default and relies on the Unix shell to expand **. On Windows, cmd passes ** literally, which causes fs.rm to crash instantly.

Adding the explicit --glob flag to these scripts lets rimraf handle the wildcards internally, fixing the build crash completely for Windows contributors.

build / fix

N/A (This restores the ability for Windows users to execute the existing build scripts).

No.

N/A

Use of AI

Debugged locally from a custom Windows development environment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,9 +22,9 @@
   "main": "n/a",
   "scripts": {
     "clean-dist": "rimraf ./dist",
-    "clean-printable": "rimraf --glob src/content/**/printable.mdx",
+    "clean-printable": "rimraf \"src/content/**/printable.mdx\"",
     "preclean": "run-s clean-dist clean-printable",
-    "clean": "rimraf --glob src/content/**/_*.mdx src/**/_*.json repositories/*.json",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -22,9 +22,9 @@
   "main": "n/a",
   "scripts": {
     "clean-dist": "rimraf ./dist",
-    "clean-printable": "rimraf src/content/**/printable.mdx",
+    "clean-printable": "rimraf --glob src/content/**/printable.mdx",
     "preclean": "run-s clean-dist clean-printable",
-    "clean": "rimraf src/content/**/_*.mdx src/**/_*.json repositories/*.json",
+    "clean": "rimraf --glob src/content/**/_*.mdx src/**/_*.json repositories/*.json",
```
alexander-akait: You can use `\"`, instead extra flag

PR #8131, opened 2026-03-27
docs: Correct dependency type and externals explanation in Authoring Libraries guide

### LINKED ISSUE #6896: Misguiding document on authoring libraries

Abhishekgupta1215:
Closes #6896 

Motivation
This PR fixes incorrect guidance in the Authoring Libraries guide around dependency classification for lodash.

The guide previously suggested installing lodash as a dev dependency and described it as a peer dependency when externalized. Since lodash is directly imported by library source code, it should be documented as a regular dependency. The externals configuration should be explained as a bundling strategy, not a dependency-type rule.

What changed

Updated install instructions to:
keep webpack and webpack-cli in devDependencies
install lodash as a regular dependency
Replaced wording that implied lodash should be a peer dependency
Clarified externals behavior:
lodash is not bundled into library output
consumer bundlers can deduplicate lodash
script-tag consumers must provide lodash separately
docs

No. This is a documentation-only change and does not modify runtime behavior.

No.

This PR itself updates the relevant documentation in the Authoring Libraries guide, so no additional follow-up docs are required.

Use of AI
Yes. I used AI assistance to help draft and refine documentation wording and structure. I reviewed all generated text, checked its technical correctness, and made final decisions manually.

### DIFF
--- a/src/content/guides/author-libraries.mdx
+++ b/src/content/guides/author-libraries.mdx
@@ -33,14 +33,15 @@ The basic project structure would look like this:
 +     └── ref.json
 ```
 
-Initialize the project with npm, then install `webpack`, `webpack-cli` and `lodash`:
+Initialize the project with npm, then install `webpack`, `webpack-cli` as dev dependencies, and `lodash` as a regular dependency:
 
 ```bash
 npm init -y
-npm install --save-dev webpack webpack-cli lodash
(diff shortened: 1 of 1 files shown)


alexander-akait:
Duplicate of https://github.com/webpack/webpack.js.org/pull/8094

PR #8173, opened 2026-04-06
refactor(Support): rewrite class component to functional

sujalgoel:
**Summary**

Part of #8161. Converts Support from a class component to a functional component using hooks. Moves the supporters computation into `useMemo` to satisfy the `react-hooks/purity` lint rule. Also adds tests.

Refactor. No behavior or visual changes.

No.

N/A

**Use of AI**

No.

### DIFF
--- a/src/components/Support/Support.jsx
+++ b/src/components/Support/Support.jsx
@@ -108,7 +108,7 @@ export default function Support({ rank, type }) {
   useEffect(() => {
     const observer = new IntersectionObserver(
       ([entry]) => {
-        if (entry.isIntersecting && !inView) {
+        if (entry.isIntersecting) {
           setInView(true);
           observer.disconnect();
         }
@@ -121,7 +121,7 @@ export default function Support({ rank, type }) {
     }
 
     return () => observer.disconnect();
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, []);
 
   const { supporters, minimum, maximum, maxAge, limit, random } =
     useMemo(() => {
--- a/src/components/Support/Support.jsx
+++ b/src/components/Support/Support.jsx
@@ -101,6 +101,25 @@ function handleImgError(event) {
   imgNode.setAttribute("src", SmallIcon);
 }
 
+function filterByAge(result, age) {
+  const now = Date.now();
+  return result.filter(
+    (item) =>
+      item.firstDonation && now - new Date(item.firstDonation).getTime() < age,
+  );
+}
+
+function shuffleSlice(arr, n) {
+  const result = [...arr];
+  for (let i = 0; i < n; i++) {
+    const other = Math.floor(Math.random() * (result.length - i));
+    const temp = result[other];
+    result[other] = result[i];
+    result[i] = temp;
+  }
+  return result.slice(0, n);
+}
+
 export default function Support({ rank, type }) {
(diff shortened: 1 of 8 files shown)


src/components/Support/__snapshots__/Support.test.jsx.snap
```diff
@@ -0,0 +1,7 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`Support matches snapshot for backer rank 1`] = `
+<h2>
+  Backers
+</h2>
+`;
```
alexander-akait, line 7: Can we add more tests and handle more cases, including, image not loaded and other stuff

sujalgoel: working on it!

src/components/Support/Support.jsx
```diff
@@ -95,215 +95,208 @@ const AVATAR_CLASSES = {
...
     }
-  }
 
-  componentWillUnmount() {
-    if (this.observer) {
-      this.observer.disconnect();
-    }
-  }
+    return () => observer.disconnect();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
alexander-akait: We need to rewrite it to do not have any `eslint-disable-line`, they are enabled to prevent errors/problems, here we have it

src/components/Support/Support.jsx
```diff
@@ -95,215 +95,208 @@ const AVATAR_CLASSES = {
...
-    if (typeof maxAge === "number") {
-      const now = Date.now();
-      supporters = supporters.filter(
-        (item) =>
-          item.firstDonation &&
-          now - new Date(item.firstDonation).getTime() < maxAge,
-      );
-    }
+      if (typeof age === "number") {
+        // eslint-disable-next-line react-hooks/purity
```
alexander-akait, line 159: We still have `eslint-disable-next-line react-hooks/purity`, can you look at this, we should avoid any `eslint` problems

alexander-akait:
@sujalgoel Latest Sponsors is broken, no images, please check it and add test cases

sujalgoel:
Looked into this. The Latest Sponsors filter keeps supporters whose `firstDonation` is within the last 14 days, evaluated against `Date.now()` at runtime. The `_supporters.json` file is regenerated by `fetch:supporters` at build time, so once a build sits unrebuilt for a while the rolling 14-day window can shrink.

Reproduced locally. With the JSON that ships in the latest preview build (Apr 22), no supporters fall in the window today. After running `npm run fetch:supporters` against the live OpenCollective API, 3 recent sponsors appear (Apr 16, Apr 17, Apr 26), and the section renders all 3 avatars correctly: SmallIcon shows first, then swaps to the avatar URL once IntersectionObserver fires. `naturalWidth: 128` for every image, no errors.

Same behavior as the previous class component, no regression from the refactor. Triggering a rebuild on the preview should populate the section.

Added test coverage for the latest rank:
- new `Support.latest.test.jsx` with a 12 recent + 1 stale fixture: verifies the limit of 10, exclusion of supporters older than 14 days, SmallIcon initial state, and avatar swap after the observer fires
- two cases added to `Support.test.jsx`: maxAge/limit text in the description, and zero images when the supporter list is empty

alexander-akait (approved):

PR #8180, opened 2026-04-09
fix: eliminate dark mode Flash of Unstyled Content (FOUC)

ryzrr:
Summary 
Fixes the dark mode flash on page load by injecting a very short blocking script into the HTML head.

Bug fix

No, but existing tests were run to ensure nothing broke.

No.

Nothing needs to be documented.

Use of AI?
No

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,6 +1,8 @@
 <!doctype html>
 <html lang="en">
-  <head> </head>
+  <head>
+    <script>(function(){try{var d=document.documentElement;var s=localStorage.getItem('theme');var isDark=false;if(s){isDark=s==='"dark"'||s==='dark'}else{isDark=window.matchMedia('(prefers-color-scheme: dark)').matches;if(isDark){localStorage.setItem('theme','"dark"')}}if(isDark){d.setAttribute('data-theme','dark');d.classList.add('dark')}}catch(e){}})();</script>
+  </head>
   <body>
     <div id="root"></div>
   </body>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -1,6 +1,28 @@
...
+            }
+          }
+          if (isDark) {
+            d.setAttribute("data-theme", "dark");
+            d.classList.add("dark");
+          }
+        } catch (e) {}
+      })();
+    </script>
+  </head>
```
alexander-akait, line 25: Can we fix it without such code?

ryzrr: I think the inline script is the most reliable way to avoid  flash 'cause  it handles both the system theme and the users saved preference. BUT, another option is using CSS with `prefers-color-scheme`, which covers the OS setting without any script.

The trade off is that if  the  user has  chosen a different theme than the OS, there might be a brief flash until React loads.

alexander-akait: > I think the inline script is the most reliable way to avoid flash 'cause it handles both the system theme and the users saved preference. BUT, another option is using CSS with prefers-color-scheme, which covers the OS setting without any script.

I think it is a good choose

ryzrr: I implemented the CSS-only route , but it creates a reversed flash for users who override their OS theme preference.

ryzrr: This ,  CSS-only approach works for most users, but if someone's OS is set to dark but they've manually picked light mode on the site (or vice versa), they still see a brief flash. So we've just moved the problem. Let me know how you'd like me to proceed.

PR #8208, opened 2026-04-25
docs(api): add this.getOptions schema validation example

raj-sapalya:
Documentation improvement — adds a missing example for this.getOptions() with schema validation.

**Summary**

The this.getOptions(schema) API supports passing a JSON Schema for
automatic option validation, but the current docs show only a one-line
description with no example.

Many loader authors still use the separate schema-utils validate()
pattern because there is no example showing that this.getOptions()
handles validation natively when a schema is passed.

This PR adds a minimal working example showing:
- passing a schema to this.getOptions()
- accessing validated options inside a loader
- using additionalProperties: false to catch unknown options at build start

No.

Nothing additional.

### DIFF
--- a/src/content/api/loaders.mdx
+++ b/src/content/api/loaders.mdx
@@ -15,6 +15,7 @@ contributors:
   - chenxsan
   - jamesgeorge007
   - alexeyr
+  - raj-sapalya
 ---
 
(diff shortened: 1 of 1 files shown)


alexander-akait:
Fixed in the main

PR #8212, opened 2026-04-27
docs: correct lodash dependency usage in authoring libraries guide

### LINKED ISSUE #6896: Misguiding document on authoring libraries

gurleen2506:
## Summary
Fixes misleading documentation regarding lodash dependency usage.

## Changes
- Updated installation instructions to install lodash as a runtime dependency
- Clarified that lodash should be listed under `dependencies` since it is used at runtime
- Mentioned that bundling can be avoided using webpack externals

Fixes #6896

### DIFF
--- a/src/content/guides/author-libraries.mdx
+++ b/src/content/guides/author-libraries.mdx
@@ -37,10 +37,11 @@ Initialize the project with npm, then install `webpack`, `webpack-cli` and `loda
 
 ```bash
 npm init -y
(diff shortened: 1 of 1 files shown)


evenstensberg (approved):

alexander-akait:
Fixed

gurleen2506:
Thanks for the review and for resolving the issue.

I’d love to contribute more—please let me know if there are other areas where I can help.

PR #8217, opened 2026-04-29
docs: correct lodash dependency usage in authoring libraries guide

prince-up:
**Summary**

The "Authoring Libraries" documentation incorrectly guides developers to install runtime dependencies like `lodash` as `devDependencies` and to classify them as `peerDependencies` when externalizing them. 

This PR solves this by:
- Updating the installation instructions to place `lodash` in `dependencies` rather than `devDependencies`.
- Rewriting the explanation to clarify that it's a runtime dependency, but we still exclude it from the bundle to prevent bloat.
- Removing the misleading terminology regarding "peer dependency" from the `externals` explanation.

Fixes #7903

- [x] docs

- [x] No, this is a documentation-only change.

- [x] No

This PR itself is a documentation update. No further documentation is needed.

**Use of AI**

No AI was used to generate this documentation change.

### DIFF
--- a/src/content/guides/author-libraries.mdx
+++ b/src/content/guides/author-libraries.mdx
@@ -33,14 +33,15 @@ The basic project structure would look like this:
 +     └── ref.json
 ```
 
-Initialize the project with npm, then install `webpack`, `webpack-cli` and `lodash`:
+Initialize the project with npm, then install `webpack` and `webpack-cli` as development dependencies, and `lodash` as a runtime dependency:
 
 ```bash
 npm init -y
-npm install --save-dev webpack webpack-cli lodash
+npm install --save-dev webpack webpack-cli
+npm install lodash
 ```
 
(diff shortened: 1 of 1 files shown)


alexander-akait (changes requested):
This is wrong, `lodash` should be in `dependencies` only when it is in `externals`, without `externals` we will bundle `lodash` and you don't need to have it in `dependencies`

Priyanshu007k:
looks good to me

prince-up:
> This is wrong, `lodash` should be in `dependencies` only when it is in `externals`, without `externals` we will bundle `lodash` and you don't need to have it in `dependencies`

"Thanks for the clarification @alexander-akait, that makes perfect sense! I have updated the documentation to explain that lodash should initially be installed as a devDependency because it's being bundled, and should only be moved to dependencies once we decide to use externals. Let me know if it looks good now!"

alexander-akait (approved):

PR #8243, opened 2026-05-17
docs(experiments): document experiments.html and its HTML features

bjohansebas:
## Summary

Webpack 5.107 introduces `experiments.html` as the foundation for native HTML entry-point support in webpack 6 (tracked in [#536](https://github.com/webpack/webpack/issues/536)). The flag registers the `html` module type on `NormalModuleFactory` and unlocks several HTML behaviors that landed in the same release:

- **Inline `<style>` tags** routed through the CSS pipeline as virtual `exportType: "text"` modules, so `url()` and `@import` resolve relative to the HTML file.
- **`<script src>` and `<link rel="modulepreload">`** references turned into real webpack entries, with shared runtime via `dependOn` for multiple scripts on the same page, independent chunks for modulepreload, and auto-upgrade to `type="module"` under `output.module`.
- **`webpackIgnore` magic comment for HTML** (cross-references the existing magic comments docs in `api/module-methods`).

Adds the entry to the TOC and a new `### experiments.html` section in `configuration/experiments.mdx` placed after `experiments.futureDefaults`.

Refs:
- https://github.com/webpack/webpack/pull/20902 (experiments.html flag)
- https://github.com/webpack/webpack/pull/20962 (inline styles)
- https://github.com/webpack/webpack/pull/20949 (script src / modulepreload)

## Test plan

- [ ] Visual check of the new section and all subsections
- [ ] Verify the TOC entry links correctly
- [ ] Verify the warning admonition is visible
- [ ] Verify all badges show "5.107.0+"

## Use of AI

Drafted with Claude under human review. The contributor verified each change against the upstream webpack PR before commit.

### DIFF
--- a/src/content/configuration/experiments.mdx
+++ b/src/content/configuration/experiments.mdx
@@ -27,6 +27,7 @@ Available options:
 - [`css`](#experimentscss)
 - [`deferImport`](#experimentsdeferimport)
 - [`futureDefaults`](#experimentsfuturedefaults)
+- [`html`](#experimentshtml)
 - [`lazyCompilation`](#experimentslazycompilation)
 - [`outputModule`](#experimentsoutputmodule)
 - `syncWebAssembly`: Support the old WebAssembly like in webpack 4.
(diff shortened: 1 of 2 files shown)


src/content/configuration/experiments.mdx
```diff
@@ -310,6 +311,83 @@ export default {
 };
 ```
 
+### experiments.html
+
+<Badge text="5.107.0+" />
+
+Enable native HTML entry-point support. This is the foundation for handling `.html` files as first-class webpack modules without `html-webpack-plugin` or `html-loader`. The flag registers the `html` module type on `NormalModuleFactory` and unlocks the HTML behaviors described below.
```
alexander-akait: I want to say it is too early write here about `html-webpack-plugin`, now we support only `html-loader` features, I think we will finish it in 5.108.0

alexander-akait (approved):

PR #8249, opened 2026-05-19
chore: migrate from yarn to npm

alexander-akait:
- Replace yarn.lock with package-lock.json
- Convert `resolutions` to npm `overrides` in package.json
- Drop yarn-deduplicate devDependency and remove it from the `prepare` script
- Flip .npmrc to package-lock=true so npm maintains the lockfile
- Untrack package-lock.json/yarn-error.log in .gitignore
- Switch CI workflows (testing, deploy) and the webpack-persistent-cache action to npm
- Update CONTRIBUTING.md to instruct contributors to use npm
- Update the module-federation example (lerna npmClient, README) to use npm

User-facing webpack documentation in src/content/** still mentions yarn
because webpack itself supports yarn / Yarn PnP for end users.


.npmrc
```diff
@@ -1 +1 @@
-package-lock=false
```
avivkeller, line 1: I think `true` is the default, so this can be removed

alexander-akait: Sometimes AI is stupid :smile:

PR #8291, opened 2026-07-05
style: soften DocSearch keyboard shortcut styling

arechaithanya:
## Summary

This PR reduces the visual prominence of the DocSearch keyboard shortcut keys in the search bar.

The shortcut keys are styled to feel more like passive keyboard hints rather than clickable buttons, allowing the search input to remain the primary focus while preserving readability in both light and dark themes.

### What kind of change does this PR introduce?

- [x] Style/UI improvement

### Did you add tests for your changes?

No. This change only updates the visual styling of the search shortcut.

### Does this PR introduce a breaking change?

No.

### If relevant, what needs to be documented?

No documentation changes are required.

### Use of AI

AI was used to assist with brainstorming and reviewing styling approaches. All code changes were manually implemented, tested, and verified locally.


arechaithanya:
The failing E2E check appears unrelated to this styling change. The failure occurs in src/utilities/fetch-supporters.mjs while fetching supporters from the OpenCollective API (node.account is null), which is the same issue I encountered during local development.

src/styles/tailwind.css
```diff
@@ -33,8 +33,30 @@
 
 /* doc search */
 :root {
-  --docsearch-primary-color: #1d78c1 !important;
+  --docsearch-key-color: rgb(148 163 184);
```
evenstensberg, line 36: could you change this to either rgba or hsla?

PR #8295, opened 2026-07-13
Add atproto-did file with DID entry

avivkeller:


bjohansebas (commented):
what is this?

avivkeller:
It's for a bluesky account, if i can fix deployments. not sure why it's not working?

bjohansebas:
@avivkeller, is this so the domain can be used as the username, or what is the purpose of it?

PR #8338, opened 2026-08-27
docs(blog): add the webpack 5.110 release post

bjohansebas:
Release post for webpack 5.110.0. Stacked on top of #8337 (`docs/webpack-5-110`), so **merge that one first** and review only the last commit here; the base will be retargeted to `main` once #8337 lands.

## What the post covers

- **Performance hints** as the headline: how to turn the suite on, what each group of checks looks at, and `hints: "stats"` for adopting them without turning a build red.
- **Built-in CSS and HTML minification** and the per-asset-type `optimization.minimize` object.
- **Native CSS is the way forward**, which announces that `css-loader`, `style-loader` and `mini-css-extract-plugin` are deprecated in favour of `experiments.css`, with a migration diff and the four caveats that actually bite (module vs global by extension, class name generation moving to the generator options, preprocessors staying as loaders, extracted CSS being a real chunk asset).
- The `nodeModules` externals preset, the `glob` rule condition, tree shaking and scope hoisting work, SSR styles, HTML, output and library options, MultiCompiler, and the usual bug fix and performance section.

## Blocked on two things before this can be published

- [ ] **Benchmarks.** The `### Benchmarks` section is an empty table with a visible placeholder line. It needs real measurements of the same app built both ways (`css-loader` + `mini-css-extract-plugin` against `experiments.css`): cold production build, rebuild with a warm cache, HMR update time, peak memory and emitted CSS bytes.
- [ ] **Banner.** `src/assets/blog/webpack-5-110.png` does not exist yet, so the post has no banner import. It has to be added back the way the previous posts do it.

The date in the filename and the `sort` field are a placeholder too, and should be moved to the actual release date.

## Notes

`prettier`, `eslint` and `markdownlint` pass. The 26 internal links were checked to resolve to a real file and anchor, as were the in-page table of contents anchors.


alexander-akait:
@bjohansebas feel free to merge

PR #8355, opened 2026-09-08
docs: teach the docs native HTML and CSS, drop html-webpack-plugin and the CSS loaders

alexander-akait:
webpack builds HTML and CSS natively now, so the docs should teach that first instead of routing every reader through `html-webpack-plugin`, `mini-css-extract-plugin`, `css-loader` and `style-loader`.

## What changed

**`html-webpack-plugin` is gone from the guides and reference.** The guides that used it — output management, development, HMR, PWA, caching — now make the page itself an entry point (`entry: "./src/index.html"` with `experiments.html`), and webpack emits the page with its scripts injected. `src/content/plugins/html-webpack-plugin.mdx` is deleted, with a redirect to `/guides/native-html/` in `webpack.ssg.mjs`.

**The CSS loaders are gone too.** `css-loader`, `style-loader` and `MiniCssExtractPlugin` no longer appear in the guides; `experiments.css` has defaulted to `'auto'` since 5.109.0, so an imported `.css` file needs no loader and no plugin. Where a reader needs more, the text links to `/guides/native-css/`. Their pages redirect there as well, and the three packages are added to `excludedLoaders`/`excludedPlugins` in `src/utilities/constants.mjs` so the README fetcher stops pulling pages for them.

**The front page shows the new model.** The splash example is built around an HTML entry with hashed output filenames, and shows the emitted `dist/index.html` next to the config. The rotator gained "pages" alongside the existing asset kinds.

**The native HTML guide gained migration help.** The companion-plugin table now covers 13 of the plugins people pair with `html-webpack-plugin`, and a new "Rewriting a plugin that tapped its hooks" section maps the old hooks onto `HtmlModulesPlugin`'s (`injectTags`, `transformTags`, `transformHtml`, `htmlEmitted`) with a worked `NoModulePlugin` example.

## Build listings

The listings show the filenames webpack emits today, verified against webpack/webpack#22008 (merged): a page entry emits no JavaScript of its own, so the scripts it pulls in take the entry's name — `main.bundle.js` for the first, `main1.bundle.js` for the second. Before that fix an HTML entry also emitted a dead JS bundle duplicating the page markup, which is why these listings would have looked different a week ago.

## Checks

`lint:prettier`, `lint:js` and `lint:markdown` pass locally. `lint:prose` needs `vale`, and `lint:links` needs a built `dist`, so both are left to CI here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NuWDbcF2jvq7q1r8aEY5V9


alexander-akait:
`E2E test and link check` is red on `46ef11f`, on 1 assertion out of 48,442:

```
✖ FAIL fragment-check dist/plugins/diagnostics-webpack-plugin/index.html --> /configuration/other-options/#validate
| operator: fragment-check
| expected: id="validate"
|       at: dist/plugins/diagnostics-webpack-plugin/index.html:49:82
```

This is not caused by this PR:

- `plugins/diagnostics-webpack-plugin` is not a checked-in page — `fetch-package-readmes.mjs` generates it at build time from the plugin's README on GitHub, so the link comes from that README, not from this branch. Nothing checked in links to that anchor (`grep -rn "other-options/#validate" src/` finds nothing).
- This branch does not touch `src/content/configuration/other-options.mdx`, and the target page is reached fine — only the fragment is missing.

The underlying gap: webpack has a top-level `validate` option (schema: *"Enable validation of webpack configuration. Defaults to true in development mode. In production mode, defaults to true unless futureDefaults is enabled, then defaults to false."*, and `lib/config/defaults.js` sets `F(options, "validate", () => !(futureDefaults === true && production))`), but `other-options.mdx` has no `validate` section, so the anchor the README points at does not exist. Any build of `main` today produces the same page and the same missing anchor.

Documenting `validate` is the fix, but it belongs in its own PR rather than widening this one. The patch would be a section in `src/content/configuration/other-options.mdx`, alphabetically after `snapshot`:

```md
## validate

`boolean`

Enable validation of the webpack configuration against its schema. Defaults to `true` in development mode. In production mode it also defaults to `true`, unless [`experiments.futureDefaults`](/configuration/experiments/#experimentsfuturedefaults) is enabled, in which case it defaults to `false`.

**webpack.config.js**

\`\`\`js
module.exports = {
  //...
  validate: false,
};
\`\`\`
```

Re-running the job once to rule out a transient README fetch. Everything else on this head is green: Lint, Proselint, Unit Test, Socket ×2, dependency-review, Vercel, and EasyCLA.

---
_Generated by [Claude Code](https://claude.ai/code)_

